Issue #1, opened 2021-02-10
trait constructor calling convention


Shoeboxam:
- switch calling convention from individual functions to impls of constructor traits in trans.rs, meas.rs, core.rs
- refactor trans.rs and meas.rs into folders
- split trans.rs code between mod.rs and dataframe.rs
- add num crate and remove OpenDPInto trait

andrewvyrros:
The PR for this is under #2.

Shoeboxam:
Closed in #2

PR #2, opened 2021-02-10
convert trans.rs to updated calling convention

Shoeboxam:
See #1


rust/opendp/src/trans/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+            // d_out >= d_in * max(|m|, |M|)
+            move |d_in: &u32, d_out: &T| *d_out >= T::from(*d_in).unwrap() * max(num::abs(lower), num::abs(upper)))
+    }
+}
+
+impl<MO, T> MakeTransformation4<SizedDomain<VectorDomain<IntervalDomain<T>>>, AllDomain<T>, SymmetricDistance, MO, usize, T, T, MO> for BoundedSum<SymmetricDistance, T>
+    where T: 'static + Copy + PartialOrd + Sub<Output=T> + NumCast + Mul<Output=T> + Div<Output=T> + Sum<T>,
+          MO: SensitivityMetric<Distance=T>,
+          SymmetricDistance: Metric<Distance=u32>  {
+    fn construct(length: usize, lower: T, upper: T, output_metric: MO) -> Transformation<SizedDomain<VectorDomain<IntervalDomain<T>>>, AllDomain<T>, SymmetricDistance, MO> {
```
andrewvyrros, line 128: It's a bummer that we have to pass in output_metric as a parameter here. Could we not do the same trick as with DatasetMetric, have a new() associated function?

rust/opendp/src/meas/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+    use super::*;
+
+    #[test]
+    fn test_make_base_laplace() {
+        let measurement = LaplaceMechanism::<f64>::construct(1.0);
+        let arg = 0.0;
+        let _ret = measurement.function.eval(&arg);
+        // TODO: Test for base_laplace
+    }
+
```
andrewvyrros, line 121: Can we add a test for VectorLaplaceMechanism too? (I know it's just a sanity check, but it's a place to put a TODO anyway.)

Also would be good to do one for GaussianMechanism.

rust/opendp/src/trans/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+    fn construct(param1: P1, param2: P2) -> crate::core::Transformation<DI, DO, MI, MO>;
+}
+
+pub trait MakeTransformation3<DI: Domain, DO: Domain, MI: Metric, MO: Metric, P1, P2, P3> {
+    fn construct(param1: P1, param2: P2, param3: P3) -> crate::core::Transformation<DI, DO, MI, MO>;
+}
+
+pub trait MakeTransformation4<DI: Domain, DO: Domain, MI: Metric, MO: Metric, P1, P2, P3, P4> {
+    fn construct(param1: P1, param2: P2, param3: P3, param4: P4) -> crate::core::Transformation<DI, DO, MI, MO>;
+}
```
andrewvyrros, line 38: Can we change the associated function name to make()? I think that's more in keeping with the terminology used in the paper.

rust/opendp/src/meas/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+    fn construct(param1: P1, param2: P2) -> crate::core::Measurement<DI, DO, MI, MO>;
+}
+
+pub trait MakeMeasurement3<DI: Domain, DO: Domain, MI: Metric, MO: Measure, P1, P2, P3> {
+    fn construct(param1: P1, param2: P2, param3: P3) -> crate::core::Measurement<DI, DO, MI, MO>;
+}
+
+pub trait MakeMeasurement4<DI: Domain, DO: Domain, MI: Metric, MO: Measure, P1, P2, P3, P4> {
+    fn construct(param1: P1, param2: P2, param3: P3, param4: P4) -> crate::core::Measurement<DI, DO, MI, MO>;
+}
```
andrewvyrros, line 33: Similar to trans, can we change these to make()?

rust/opendp/src/trans/mod.rs
Shoeboxam: I think the reason we didn't add `new()` to SensitivityMetric was because some sensitivities take parameters.

Shoeboxam: The doubt here was that the new() constructor would not have sufficient information to create:
`struct LpSensitivity {p: u32}`
I agree that there is a downside for not defining new for all metrics, because you get inconsistent interfaces-- sometimes you need to pass metrics, sometimes you don't.

andrewvyrros: Ah yes, I forgot about that source of variation.

Is this something that's actually used by some mechanisms? Or in practice, would we only have say L1, L2, L-infinity norm?

Shoeboxam: It's more a question of how general we want to be. I think the vast majority of cases will need L1 or L2 norms. In cases where we need Lp norms, the default constructor can just let p = 1. Unintentionally setting p = 1 would still get caught by runtime checks when chaining. Constructors that emit or accept sensitivity in Lp space can take an extra LpSensitivity argument, or p argument. 

Interesting note, when we start having non-empty metric structs, we can no longer omit the check for metric equality when chaining.

I pushed a commit that adds a `new()` constructor to the `SensitivityMetric` trait.

andrewvyrros: Makes sense, thanks!

andrewvyrros (changes requested):
Looks good, unfortunately unit tests are broken. 😢

rust/opendp/src/lib.rs
```diff
@@ -38,26 +38,31 @@
...
 //!
-//!     // Construct a Measurment to calculate a noisy sum.
-//!     let clamp = trans::make_clamp(bounds.0, bounds.1);
-//!     let bounded_sum = trans::make_bounded_sum_l1(bounds.0, bounds.1);
-//!     let laplace = meas::make_base_laplace(sigma);
-//!     let intermediate = core::make_chain_tt(&bounded_sum, &clamp);
-//!     let noisy_sum = core::make_chain_mt(&laplace, &intermediate);
+//!     // Construct a Measurement to calculate a noisy sum.
+//!     let clamp = trans::Clamp::make(bounds.0, bounds.1);
+//!     let bounded_sum = trans::BoundedSum::make2(bounds.0, bounds.1);
```
Shoeboxam, line 59: Don't miss that this is now `::make2` to disambiguate which trait to dispatch to. `BoundedSum::make` is implemented for both 2 and 3 arguments.

Shoeboxam: If you impl more than one trait with the same function (ex: `MakeTransformation0`, `MakeTransformation1`), rust can't determine which trait to dispatch to based on type or arity.
An example of this is the impls for BoundedSum. The restricted sensitivity impl takes an extra argument for dataset length, causing calling sites to use the fully qualified syntax to disambiguate:
```let bounded_sum = <trans::BoundedSum<_, _, _> as MakeTransformation2<_, _, _, _, _, _>>::make(bounds.0, bounds.1);```
One way we could disambiguate easier is by changing the function from `make` to `makeN`. In most cases, you can still automatically dispatch to the right impl, so the constructor traits can define a default implementation of `make` that points to `makeN`. This way, if you do run into issues with ambiguous trait impls, you can add the arity after `make`, instead of using the fully qualified syntax. This does mean adding a second impl with different arity can be a breaking change to make if both traits happen to be imported.

andrewvyrros: Oh interesting. I remember reading about this issue in the Rust docs, but haven't encountered it in code before now.

The way you split these out into separate functions with explicit arity is a cool solution.

andrewvyrros (approved):

PR #24, opened 2021-02-27
laplace/gaussian/geometric floating-point-safe samplers

Shoeboxam:
Ported, with modifications throughout, from SmartNoise.

- introduces dependencies on openssl and gmp/mpfr
- adds feature flags to disable gmp/mpfr
- restructure meas submodule
- reference new samplers from laplace, geometric and gaussian mechanisms

New samplers:
- standard bernoulli
- bernoulli
- standard uniform (over float types)
- uniform (over float types)
- rademacher (over signed types)
- censored standard geometric (accumulates tail of probability < ~2^-52 into fully-saturated 1022)
- censored geometric
- laplace (mpfr and non-mpfr) (over float types)
- gaussian (mpfr and non-mpfr) (over float types)

### DIFF
--- a/rust/opendp/src/meas/geometric.rs
+++ b/rust/opendp/src/meas/geometric.rs
@@ -13,7 +13,7 @@ pub struct BaseSimpleGeometric<T> {
     data: PhantomData<T>
 }
 
-// gaussian for scalar-valued query
+// geometric for scalar-valued query
 impl<T> MakeMeasurement3<AllDomain<T>, AllDomain<T>, L2Sensitivity<T>, MaxDivergence<f64>, f64, T, T> for BaseSimpleGeometric<T>
     where T: 'static + Clone + SampleGeometric + Sub<Output=T> + Add<Output=T> + DistanceCast {
     fn make3(scale: f64, min: T, max: T) -> Fallible<Measurement<AllDomain<T>, AllDomain<T>, L2Sensitivity<T>, MaxDivergence<f64>>> {
--- a/rust/opendp/src/meas/geometric.rs
+++ b/rust/opendp/src/meas/geometric.rs
@@ -1,0 +1,55 @@
+use std::marker::PhantomData;
+use std::ops::{Sub, Add};
+
+
+use crate::core::{Function, Measurement, PrivacyRelation};
+use crate::dist::{L2Sensitivity, MaxDivergence};
+use crate::dom::AllDomain;
+use crate::error::Fallible;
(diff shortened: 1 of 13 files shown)


Shoeboxam:
I also added a simple scheme to sample laplace noise at 53 bits of precision using mpfr. This sidesteps the obvious inverse transform issue, but it is not yet clear if this leaves other floating vulnerabilities.

andrewvyrros:
Gulp, openssl is a big dependency with a lot of issues. Let's have a discussion about this (and dependencies in general).

Shoeboxam:
For any randomized function, we could take an rng source as a generic-

`fn myfunc<RNG>(rng: RNG, ...) -> ... where RNG: rand::CryptoRng + rand::RngCore`

...and then move the openssl dependency out to the FFI crate. Obviously anyone could impl CryptoRng on a newtype for a non-crypto or static noise generator.

rust/opendp/src/meas/geometric.rs
```diff
@@ -0,0 +1,55 @@
...
+use crate::error::Fallible;
+use crate::meas::{MakeMeasurement3, SampleBernoulli, SampleGeometric, SampleUniform};
+use crate::traits::DistanceCast;
+
+pub struct BaseSimpleGeometric<T> {
+    data: PhantomData<T>
+}
+
+// gaussian for scalar-valued query
+impl<T> MakeMeasurement3<AllDomain<T>, AllDomain<T>, L2Sensitivity<T>, MaxDivergence<f64>, f64, T, T> for BaseSimpleGeometric<T>
```
andrewvyrros, line 17: Why is scale fixed to f64 here, when you've parameterized the others?

Shoeboxam: `SampleBernoulli` has an interface that takes a probability as an f64. This f64 is re-exported in the `SampleGeometric` trait. The `SampleBernoulli` implementation decomposes the f64 and uses the bits in the mantissa to weight the probability.
Here:
https://github.com/opendp/opendp/blob/9ef96dccd4230552dded0b8995db3ea568ade6af/rust/opendp/src/meas/mod.rs#L150-L159

To make the geometric mechanism generic over the output distance, I need to make this snippet generic with respect to the `Ieee754::Significand` associated type. I haven't found a clean way to do this yet.

andrewvyrros: Cool thanks for the explanation!

andrewvyrros (approved):

Issue #33, opened 2021-03-11
Make DistanceCast properly handle f64 -> f32


andrewvyrros:
DistanceCast properly handles rounding for size change and int -> float changes. But in the corner case of f64 -> f32, it's possible that the resulting distance will be smaller.

Issue #30, opened 2021-03-05
Chaining hints


andrewvyrros:
Combinator functions make_chain_mt() and make_chain_tt() need an additional argument for a hint function from the framework paper. This function chooses an intermediate distance so that relations can be chained.

- [x] Specify APIs
- [x] Implement chaining logic
- [x] Add forward/backward map functionality
- [x] Add convenience constructors using stability constant
- [x] Update existing operations to generate relations with maps
- [ ] Write unit tests

andrewvyrros:
Closing this manually.

PR #34, opened 2021-03-11
chaining hints

Shoeboxam:
Linked to #30 

Still need to implement passing hints over FFI.
Suffers from #33.

Stability and privacy relations are oftentimes specified in terms of a constant.  
In these cases, it is possible to bundle a forward and backward map with the stability relation, to construct hints automatically.


andrewvyrros (approved):

andrewvyrros:
Looks great! This was a surprising amount of work, but makes the intention of constructors easier to understand.

PR #51, opened 2021-03-18
ffi distance dispatch

### LINKED ISSUE #41: FFI constructor dispatch for Metrics & Measures
We don't have a way for FFI constructors to dispatch on different metrics. Currently, this is handled in a clumsy way by having separate entry points. (E.g., `opendp_trans__make_bounded_sum_l1()` & `opendp_trans__make_bounded_sum_l2()`.) This should be cleaned up, so that FFI clients can specify the Metrics/Measures they want.

Shoeboxam:
Closes #41. 
Revises the signatures for FFI functions to take input and output distance metrics.

Update 1:
Dispatch in cases where constructor traits are implemented piecemeal (separate impls for Hamming and Symmetric) are not amenable to generic monomorphization functions. This is demonstrated with verbose custom matching code for the BoundedSum ffi. The shorter, non-working version is commented above.

https://github.com/opendp/opendp/blob/55ff41895df5d5f20a01f6474026fc0b18a4f3b2/rust/opendp-ffi/src/trans.rs#L139-L188

There is no impl for making a BoundedSum with an MI: DatasetMetric. There is an impl for making a BoundedSum with SymmetricDistance and an impl for making a BoundedSum with HammingDistance. But the monomorphize function cannot find them because MI is generic with a trait bound.

Update 2 (Mar 23): resolved by defining ad-hoc trait bounds like `trait BoundedSumStability<MI, MO, T>` to build the stability relations, and then adding it as a trait bound on the monomorphize function in the FFI layer. This unifies disparate impls for different combinations of distance metrics behind a common interface that can be dispatched to.


andrewvyrros:
Looks solid except for possible unintended change in opendp-ffi/src/data.rs.

rust/opendp/src/trans/mod.rs
```diff
@@ -104,55 +105,54 @@ pub struct BoundedSum<MI, MO, T> {
...
 
-impl<MO, T> MakeTransformation2<VectorDomain<IntervalDomain<T>>, AllDomain<T>, HammingDistance, MO, T, T> for BoundedSum<HammingDistance, MO, T>
-    where T: 'static + Clone + PartialOrd + Sub<Output=T> + Mul<Output=T> + Div<Output=T> + Sum<T> + DistanceCast,
-          HammingDistance: Metric<Distance=u32>,
-          MO: SensitivityMetric<Distance=T> {
-    fn make2(lower: T, upper: T) -> Fallible<Transformation<VectorDomain<IntervalDomain<T>>, AllDomain<T>, HammingDistance, MO>> {
-        if lower > upper { return fallible!(MakeTransformation, "lower bound may not be greater than upper bound") }
+pub trait BoundedSumStability<MI: Metric, MO: Metric, T> {
+    fn get_stability(lower: MO::Distance, upper: MO::Distance) -> Fallible<StabilityRelation<MI, MO>>;
+}
```
andrewvyrros, line 114: This was a cool trick, and the fancy trait bound made it possible to do our crazy dispatch!

andrewvyrros (commented):
Sorry for any duplicated comments, GitHub web seemed to be confused and this review got messed up.

andrewvyrros (approved):

PR #55, opened 2021-03-24
Restructured Python project:

andrewvyrros:
* Created Python package.
* Reworked library initialization.
* Added unit test skeleton.
* Added pip requirements.


andrewvyrros:
I want to clean up the package stuff a bit so it will plug into GitHub Actions.

andrewvyrros:
I added a simple setuptools config. Next will be to run Python tests as part of the build validation action.

With a little more work, we should be able to drive everything from GitHub Actions, including pushing releases to PyPi & crates.io.

andrewvyrros:
Rebased after the samplers change.

Shoeboxam (approved):
Looks great!

Issue #56, opened 2021-03-24
Integrate proofs on Transformations/Measurements


Shoeboxam:
The exact route we go here is still pretty open-ended.

- Write latex documents on overleaf.com and add links to them from the documentation for each struct.
- Write latex documents and check them in on this repository, link to this repository from the documentation for each struct.
- Other?

Additional questions-
Do we provide attribution for who has written the proof for the relation? 
How do we provide attribution for who has written the proof for the relation?
How do we record vetting on the documents/code?

PR #58, opened 2021-03-24
BaseLaplace proof (v2)

Shoeboxam:
This is an example of one way to integrate proofs into the library for issue #56.

I wrote a latex document for the BaseLaplace measurement on overleaf.com, and included a link to it in the documentation for the BaseLaplace struct in the code. Connor Wagaman shared a WIP document for this with me (and I've made use of it in the linked Overleaf), but it's been a while and I wanted to go ahead and get my version of the proof written up so we can get this discussion started. 

Example proof document links to here: https://www.overleaf.com/read/brvrprjhrhwb

In the SmartNoise library, we checked pdfs and latex into the repository itself. We could go that route here as well. If we do a lot of pair editing on overleaf, it can get tedious to keep the two sources synced.

### DIFF
--- a/rust/opendp-ffi/src/meas.rs
+++ b/rust/opendp-ffi/src/meas.rs
@@ -4,13 +4,13 @@ use crate::core::FfiMeasurement;
 use crate::util;
 use crate::util::TypeArgs;
 use num::NumCast;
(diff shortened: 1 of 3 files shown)


Shoeboxam:
We've come a long way in how we formalize proofs since this PR was opened, and we will be hosting .tex files next to sources.

PR #66, opened 2021-04-02
CountBy, CountByCategories & BaseStability

### LINKED ISSUE #65: Histograms
- Category-based
- Stability-based

Shoeboxam:
Closes #65

- adds a function to FFI to check a relation
- adds a CountByCategories transform (histogram with known categories)
- adds a CountBy transform (histogram with unknown categories, input to BaseStability)
- adds a BaseStability measurement
- FFI functions and tests for each
- some data loaders that can be dropped when we merge the data loader PR
- improved error and internal backtrace formatting on the python side
- python script to find the tightest scale/threshold for stability histograms to release vocabularies from natural language data
- added missing json for make_bounded_sum_n
- dropped any unnecessary type bounds on operation structs that are already represented via associated types
- tighten the constants on the BaseGaussian measurement
- switch the BaseSimpleGeometric sensitivity space to L1Sensitivity
- move tests out of opendp::trans::mod.rs
- move BoundedSum, Count into their own modules


rust/opendp-ffi/src/meas.rs
```diff
@@ -1,6 +1,6 @@
 use std::os::raw::{c_char, c_void};
```
andrewvyrros, line 1: This module is getting pretty bulky, we should probably split it up to mirror the submodules in opendp soon.

For our sanity, it probably makes sense to keep a parallel hierarchy in opendp, opendp-ffi, and the Python packages.

rust/opendp/src/error.rs
```diff
@@ -51,6 +51,9 @@ pub enum ErrorVariant {
     #[error("Failed function execution")]
     FailedFunction,
 
+    #[error("Failed relation")]
+    FailedRelation,
```
andrewvyrros, line 55: What's the distinction between this and a relation that returns false?

rust/opendp/src/meas/stability.rs
```diff
@@ -0,0 +1,113 @@
...
+                }
+                if del >= _n.recip() {
+                    return fallible!(FailedRelation, "cause: del >= n.ln()");
+                }
+                if scale < ideal_scale {
+                    return fallible!(FailedRelation, "cause: scale < d_in / (epsilon * n)")
+                }
+                if threshold < ideal_threshold {
+                    return fallible!(FailedRelation, "cause: threshold < (2. / delta).ln() * d_in / (epsilon * n) + 1. / n");
+                }
```
andrewvyrros, line 91: Shouldn't some of these return false rather than generating an error? Seems like we should reserve the error for programming problems, but if it's just that the relation doesn't hold, we should return false.

python/example/stability_mechanism.py
```diff
@@ -0,0 +1,145 @@
+import opendp
```
andrewvyrros, line 1: This is a great example!

python/example/stability_mechanism.py
```diff
@@ -0,0 +1,145 @@
...
+            return mid
+
+        if passes:
+            end = mid
+        else:
+            start = mid
+
+
+if __name__ == "__main__":
+    main()
```
andrewvyrros, line 145: Pet peeve, there's no newline here, feels weird for a main.

andrewvyrros (commented):
Great stuff! I didn't bother reviewing the transformations moved to new submodules, assume those were just cut & pastes of the existing ones.

Just a few nits. The only thing maybe worth addressing is whether the `PrivacyRelation` in stability.rs should return false instead of Err.

rust/opendp-ffi/src/trans.rs
```diff
@@ -1,22 +1,25 @@
 use std::fmt::Debug;
```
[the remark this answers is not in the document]
Shoeboxam: I'll split it up on a new PR at the end of the PR stack.

rust/opendp/src/meas/stability.rs
Shoeboxam: After writing the initial version of the BaseStability transformation, I couldn't get the relation to pass (it can be pretty picky about distance combinations). I wanted some way for it to explain itself so I could fix it. I was trying to use the FailedRelation variant as a special variant to pass up explanations, not programmer errors. What if, on the python side, we wrap the measurement check in a python function:

```
def check_measurement(d_in, d_out, debug=False):
    if debug:
        return odp.core.measurement_check(d_in, d_out)
    else:
        try:
            return odp.core.measurement_check(d_in, d_out)
        except OdpException as err:
            if err.variant == "FailedRelation":
                return False
            else:
                raise err
```

I am open to dropping this if you disagree. A special purpose error variant that isn't really an error can be confusing.

rust/opendp-ffi/src/trans.rs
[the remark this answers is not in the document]
andrewvyrros: Sounds good!

rust/opendp/src/meas/stability.rs
andrewvyrros: Ah yeah, I could see that. This is our first time exposing relations this way, so not surprising we may need some extra info. Just feels a little funny folding it into the general Error. I'm fine keeping this for now. If it turns out to be useful to have these distinctions, then probably the right thing is to have relations return a dedicated enum type.

andrewvyrros (approved):

PR #71, opened 2021-04-05
BoundedVariance & BoundedCovariance

### LINKED ISSUE #68: (Co)variance
- [x] Implement function
- [ ] Include comments
- [ ] Write tests

Shoeboxam:
Closes #68.

- BoundedVariance transformation
- BoundedCovariance transformation
- Use higher ranked trait bounds to clean up extra clones in aggregators

### DIFF
--- a/rust/opendp/src/trans/mean.rs
+++ b/rust/opendp/src/trans/mean.rs
@@ -87,14 +87,4 @@ mod tests {
         assert_eq!(ret, expected);
         assert!(transformation.stability_relation.eval(&1, &1.).unwrap_test())
     }
-
-    #[test]
-    fn test_make_bounded_mean_int() {
-        let transformation = BoundedMean::<SymmetricDistance, L2Sensitivity<f64>>::make(0., 10., 5).unwrap_test();
-        let arg = vec![1., 2., 3., 4., 5.];
-        let ret = transformation.function.eval(&arg).unwrap_test();
-        let expected = 3.;
-        assert_eq!(ret, expected);
-        assert!(transformation.stability_relation.eval(&1, &1.).unwrap_test())
-    }
 }
--- a/rust/opendp/src/trans/mean.rs
+++ b/rust/opendp/src/trans/mean.rs
@@ -42,7 +42,8 @@ impl<MO: Metric<Distance=T>, T> BoundedMeanConstant<SymmetricDistance, MO> for B
 impl<MI, MO, T> MakeTransformation3<SizedDomain<VectorDomain<IntervalDomain<T>>>, AllDomain<T>, MI, MO, T, T, usize> for BoundedMean<MI, MO>
     where MI: DatasetMetric<Distance=u32>,
           MO: SensitivityMetric<Distance=T>,
-          T: 'static + Clone + PartialOrd + Sub<Output=T> + Mul<Output=T> + Div<Output=T> + Sum<T> + DistanceCast + Float,
+          T: 'static + Clone + PartialOrd + Sub<Output=T> + Mul<Output=T> + Div<Output=T> + DistanceCast + Float,
+          for <'a> T: Sum<&'a T>,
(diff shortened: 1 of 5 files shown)


Shoeboxam:
Covariance doesn't really have a way to call it from FFI at this point, because of a lack of data loaders.

rust/opendp/src/trans/mean.rs
```diff
@@ -53,7 +54,7 @@ impl<MI, MO, T> MakeTransformation3<SizedDomain<VectorDomain<IntervalDomain<T>>>
                 IntervalDomain::new(Bound::Included(lower.clone()), Bound::Included(upper.clone()))),
                              length),
             AllDomain::new(),
-            Function::new(move |arg: &Vec<T>| arg.iter().cloned().sum::<T>() / _length),
```
andrewvyrros, line 56: It's great that we can get rid of these clones, thanks for figuring out how to do it.

rust/opendp/src/trans/variance.rs
```diff
@@ -0,0 +1,175 @@
...
+    where T: Clone + Sub<Output=T> + Div<Output=T> + Add<Output=T> + NumCast + One {
+    fn get_stability_constant(lower: (T, T), upper: (T, T), length: usize, ddof: usize) -> Fallible<T> {
+        let _length = T::from(length).ok_or_else(|| err!(FailedCast))?;
+        let _1 = T::one();
+        let _ddof = T::from(ddof).ok_or_else(|| err!(FailedCast))?;
+        Ok((upper.0 - lower.0) * (upper.1 - lower.1) * _length.clone() / (_length.clone() + _1) / (_length - _ddof))
+    }
+}
+
+type CovarianceDomain<T> = SizedDomain<VectorDomain<IntervalDomain<(T, T)>>>;
```
andrewvyrros, line 102: It's a little sad that we don't have a way to generate data in this shape from Python, but I think this is fine until we settle on a general representation for dataframe-like things (Arrow or whatever).

andrewvyrros (approved):
Looking good!

Shoeboxam:
For posterity- an explanation of HRTB's.

I want to be able to use traits like Sub, Sum, etc in situations where the data is behind a reference. Otherwise we'll have to clone. So lets say I specify a trait bound `Sum<&T>`. But this isn't specific enough- all references have lifetimes, but in most situations the lifetimes are elided.
In a typical situation, if you do this:
```rust
let my_struct = MyStruct;
let my_struct_ref = &my_struct;
```
The compiler can infer my_struct_ref has some anonymous concrete lifetime that captures how long my_struct lasts from the context.

When specifying a trait bound, there is no context to elide the lifetime, so it must be explicitly stated. So I'll update my trait bound-- `Sum<&'static T>`... but this won't work either- when I try to use that bound in the code, I'm using it on a non-static reference, so the compiler complains. It also doesn't make sense to add a new lifetime on the constructor function, because the lifetime we'd want to specify the trait bound for doesn't yet exist at the time we call the constructor.

So what we need is a way to specify a trait bound with a generic lifetime parameter `Sum<&'a T>`, as well as an impl for Sum<&'a T> (any generic lifetime). Indeed, there is such an impl:

```rust
macro_rules! float_sum_product {
    ($($a:ident)*) => ($(
        #[stable(feature = "iter_arith_traits", since = "1.12.0")]
        impl<'a> Sum<&'a $a> for $a {
            fn sum<I: Iterator<Item=&'a Self>>(iter: I) -> Self {
                iter.fold(0.0, Add::add)
            }
        }
    )*)
}
integer_sum_product! { i8 i16 i32 i64 i128 isize u8 u16 u32 u64 u128 usize }
```

Now, we just need a way to specify the trait bound-- we want to say that for any lifetime `'a`, there exists `Sum<&'a T>`. With a HRTB you can say just that: `for<'a> Sum<&'a T>`

PR #73, opened 2021-04-07
FFI data constructors

### LINKED ISSUE #39: FFI data constructors and accessors
When calling OpenDP FFI Measurements/Transformations/Relations, the system requires that values are wrapped as an `FFIObject` (which ensures types compatibility). We need convenience functions to construct and access these from primitive values. These will be used a lot in FFI contexts, so we should think carefully about signatures. Perhaps something like this:

```
pub extern "C" fn opendp_data__new_scalar(type_args: *const c_char, val: *const c_void) -> FfiResult<*mut FfiObject> ...
```

(This would be analogous to the existing `opendp_data__from_string()` & `opendp_data__to_string()` functions, which should be folded into this.)

This should also include convenience wrappers in Python that automatically  Python objects.

andrewvyrros:
Closes #39.

* Refactored Type representation and lookup.
* Added common FFI entry points for data marshaling.
* Wrote glue for easy access from Python.


rust/opendp-ffi/src/util.rs
```diff
@@ -1,61 +1,160 @@
...
+
+lazy_static! {
+    /// The set of registered types. We don't need everything here, just the ones that will be looked up by descriptor
+    /// (i.e., the ones that appear in FFI function generic args).
+    static ref TYPES: Vec<Type> = {
+        let types: Vec<Type> = vec![
+            vec![t!(())],
+            type_vec![bool, char, u8, u16, u32, u64, u128, i8, i16, i32, i64, i128, f32, f64, String],
+            type_vec![(bool, char, u8, u16, u32, u64, u128, i8, i16, i32, i64, i128, f32, f64, String)],
+            type_vec![[bool, char, u8, u16, u32, u64, u128, i8, i16, i32, i64, i128, f32, f64, String]; 1], // Arrays are here just for unit tests, unlikely we'll use them.
```
Shoeboxam, line 138: I was going to say that this seems too niche to be useful. 2-tuples of homogeneous type? But the f32 and f64 variants of tuples are actually used for `(eps, del)` pairs.

Not important, but as I'm thinking through this-- since HammingDistance is zero-sized, can I use object_new to create an FfiObject containing a HammingDistance, by passing a slice of length 1 full of garbage? I don't know of a zero-sized type in C, so on the python side I'd end up constructing a non-empty buffer to pass into rust. But then on the rust side, you cast the pointer and it only reads 0 bytes of data to build the struct.

python/src/opendp/mod.py
```diff
@@ -210,3 +218,84 @@ def make_chain_tt_multi(self, *transformations):
...
+        if type_name == "i32":
+            return ctypes.cast(raw.contents.ptr, ctypes.POINTER(ctypes.c_int32)).contents.value
+        elif type_name == "f64":
+            return ctypes.cast(raw.contents.ptr, ctypes.POINTER(ctypes.c_double)).contents.value
+        elif type_name == "String":
+            return ctypes.cast(raw.contents.ptr, ctypes.c_char_p).value.decode()
+        elif type_name.startswith("Vec<"):
+            if type_name == "Vec<i32>":
+                array = ctypes.cast(raw.contents.ptr, ctypes.POINTER(ctypes.c_int32))
+            elif type_name == "alloc::vec::Vec<f64>":
```
Shoeboxam, line 275: Did you mean to leave `alloc::vec::` on here? This branch is unreachable.

Shoeboxam (approved):

rust/opendp-ffi/src/util.rs
andrewvyrros: Yeah the tuple type generation is a bit lame. I did it this way just to have some examples, and to cover the case of (epsilon, delta) as you mention. (But now I know how to form the product thanks to your last change! 🤣)

I _think_ you could use that trick to construct HammingDistance. You would just pass a NULL for the ptr. We'd need to update the dispatch in `opendp_data__object_new` to include those types. Have we gotten to the point where we need instances of Metrics in client code? We've been getting by with just using the types, but maybe that's breaking down?

python/src/opendp/mod.py
andrewvyrros: Good catch! I'll fix that, and add a test for that case.

rust/opendp-ffi/src/util.rs
Shoeboxam: We don't *really* need metric constructors but it would help partially unjank the identity function. Not particularly in a rush to fix that. We'd also need an accessor for FfiOperation -> FfiObject of domain.

PR #76, opened 2021-04-08
handle unsigned integers, test make_count

pdurbin:
As discussed with @andrewvyrros and @Shoeboxam previously a string was being returned (the fallback).


Shoeboxam:
I think the next step here is dropping the if/else branching and indexing into a `rust_type_descriptor: python_ctype_type` dict. Thanks for patching this!

andrewvyrros:
Thanks Phil!

Yeah, the type lookup is a little cheesy. I did it with branching rather than an index because Vecs need special handling, but I think the upcoming tuple changes rework this anyway.

andrewvyrros (approved):

PR #84, opened 2021-04-23
snapping mechanism

### LINKED ISSUE #83: Snapping mechanism

Shoeboxam:
Closes #83 
Original implementation from @ctcovington in SmartNoise-Core. I've done hardening work, some refactoring and translated it to bit twiddling instead of string manipulation.

This has a strange interface- we need to be able to scale input data to have sensitivity-1, because the protections only apply to sensitivity-1 queries. This means the measure's function needs to know the sensitivity of the input data. To do this, a sensitivity argument is added to the constructor, where the relation checks that this sensitivity is no less than d_in. This implementation assumes that the snapping mechanism for sensitivity-1 queries can also be applied to any query with sensitivity less than one, and still lend the same protections.

### DIFF
--- a/rust/opendp/src/meas/mod.rs
+++ b/rust/opendp/src/meas/mod.rs
@@ -8,6 +8,11 @@ pub mod laplace;
 #[cfg(all(feature="floating-point", feature="contrib"))]
 pub use crate::meas::laplace::*;
 
+#[cfg(feature="contrib")]
+pub mod snapping;
+#[cfg(feature="contrib")]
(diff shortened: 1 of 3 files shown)


Shoeboxam:
We should be able to solve this by writing the snapping mechanism as if it were a combinator, which runs a binary search on the input transformation to find the sensitivity.

Shoeboxam:
Closing this as we intend to use another algorithm (like integer quantization and a faster geometric sampler).

PR #86, opened 2021-04-23
PyTorch SGD

Shoeboxam:
Long-term feature branch for privatizing neural networks in pytorch. 

Prerequisites to run:
1. `pip install torch>=1.9`
1. `cd rust && cargo build --release --no-default-features`
2. `export OPENDP_TEST_RELEASE=1`

Ported from: https://github.com/opendp/smartnoise-core-python/tree/ms-external-sgd

ISSUES:
- make model saving/loading work on rewritten module trees (this works if you attach odometer before restore)
- consider patching `dispatch.rs` to remove the requirement on release binaries, speed up compilation
- validate that passing torch tensors in/out of the opendp lib doesn't leak memory


Shoeboxam:
Split out into a separate repository: https://github.com/opendp/opendp-pytorch/

Issue #88, opened 2021-05-12
Migrate FFI wrappers to use AnyXXX types


andrewvyrros:
Once we're happy with #80, we should migrate the FFI wrappers from the FfiXXX wrappers to the corresponding type-erased AnyXXX types.

For example, `FfiMeasurement` will become `Measurement<AnyDomain, AnyDomain, AnyMetric, AnyMeasure>`.

This should remove a lot of ugly explicit glue code.

PR #89, opened 2021-05-12
Interactive Measurements (and type-erased component types)

### LINKED ISSUE #80: Interactive Measurement Design
High-level design for Interactive Measurements from the framework paper:

- [x] Work out type signatures for queryables and responses
- [ ] Define Rust data structures
- [x] Develop an approach for state management
- [ ] Prototype a few examples

andrewvyrros:
Closes #80.

I kept this as a fairly conservative change. Most of the code is in separate modules (`any` and `interactive`). Aside from a few trait bounds, the existing types and logic stay the same for non-interactive Measurements and Transformations.

If/when we're happy with this, we can migrate the FFI wrappers to use the type-erased (AnyXXX) approach in #88.

* Created type-erased versions of Measurement/Transformation component types.
* Added type-erasing transformers to Measurement/Transformation.
* Defined generic Queryable abstraction.
* Implemented make_adaptive_composition() constructor.
* Added unit tests.
* Created a separate abstraction for polymorphic operations.

### DIFF
--- a/rust/opendp-ffi/src/core.rs
+++ b/rust/opendp-ffi/src/core.rs
@@ -193,7 +193,7 @@ impl Domain for FfiDomain {
     fn member(&self, _val: &Self::Carrier) -> bool { unreachable!() }
 }
 
-#[derive(Clone)]
+#[derive(Clone, PartialEq)]
 pub struct FfiMeasure;
 impl Default for FfiMeasure {
     fn default() -> Self { FfiMeasure }
@@ -202,7 +202,7 @@ impl Measure for FfiMeasure {
     type Distance = ();
 }
 
-#[derive(Clone)]
+#[derive(Clone, PartialEq)]
 pub struct FfiMetric;
 impl Default for FfiMetric {
     fn default() -> Self { FfiMetric }
--- a/rust/opendp/src/any.rs
+++ b/rust/opendp/src/any.rs
@@ -274,14 +274,6 @@ impl Metric for AnyMetric {
 }
 
 impl<DI: Domain, DO: Domain> Function<DI, DO> where DI::Carrier: 'static, DO::Carrier: 'static {
-    pub fn as_any_out(&self) -> Function<DI, AnyDomain> {
-        let function = self.function.clone();
-        let function = move |arg: &DI::Carrier| -> Fallible<Box<dyn Any>> {
-            let res = function(arg);
-            res.map(|o| Box::new(*o) as Box<dyn Any>)
-        };
-        Function::new_fallible(function)
-    }
     pub fn as_any(&self) -> Function<AnyDomain, AnyDomain> {
         let function = self.function.clone();
         let function = move |arg: &Box<dyn Any>| -> Fallible<Box<dyn Any>> {
@@ -344,16 +336,6 @@ impl<DI: 'static + Domain, DO: 'static + Domain, MI: 'static + Metric, MO: 'stat
           DO::Carrier: 'static,
           MI::Distance: 'static + Clone + PartialOrd,
           MO::Distance: 'static + Clone + PartialOrd {
-    pub fn into_any_out(self) -> Measurement<DI, AnyDomain, MI, MO> {
-        Measurement::new(
-            *self.input_domain,
-            AnyDomain::new(*self.output_domain),
-            self.function.as_any_out(),
-            *self.input_metric,
-            *self.output_measure,
-            self.privacy_relation,
-        )
-    }
     pub fn into_any(self) -> Measurement<AnyDomain, AnyDomain, AnyMetric, AnyMeasure> {
         Measurement::new(
             AnyDomain::new(*self.input_domain),
@@ -371,16 +353,6 @@ impl<DI: 'static + Domain, DO: 'static + Domain, MI: 'static + Metric, MO: 'stat
           DO::Carrier: 'static,
           MI::Distance: 'static + Clone + PartialOrd,
           MO::Distance: 'static + Clone + PartialOrd {
-    pub fn into_any_out(self) -> Transformation<DI, AnyDomain, MI, MO> {
-        Transformation::new(
-            *self.input_domain,
-            AnyDomain::new(*self.output_domain),
-            self.function.as_any_out(),
-            *self.input_metric,
-            *self.output_metric,
-            self.stability_relation,
-        )
-    }
     pub fn into_any(self) -> Transformation<AnyDomain, AnyDomain, AnyMetric, AnyMetric> {
         Transformation::new(
             AnyDomain::new(*self.input_domain),
--- a/rust/opendp/src/any.rs
+++ b/rust/opendp/src/any.rs
@@ -1,0 +1,421 @@
+use std::any::Any;
+use std::cmp::Ordering;
+use std::rc::Rc;
+
+use crate::core::{Domain, Function, Measure, Measurement, Metric, PrivacyRelation, StabilityRelation, Transformation};
+use crate::error::*;
+use crate::traits::{FallibleSub, MeasureDistance, MetricDistance};
+
+pub trait Downcast {
+    fn downcast<T: 'static>(self) -> Fallible<T>;
+    fn downcast_ref<T: 'static>(&self) -> Fallible<&T>;
+}
+
+pub struct AnyBox {
+    value: Box<dyn Any>,
+    eq_glue: Rc<dyn Fn(&Self, &Self) -> bool>,
+}
+
+impl AnyBox {
+    pub fn new<T: 'static + PartialEq>(value: T) -> Self {
+        let eq_ = |self_: &Self, other: &Self| {
+            // This downcast of self will always succeed, so equality check is all that's necessary.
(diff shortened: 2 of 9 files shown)


andrewvyrros:
I realized there are some issues with domains in adaptive composition, putting this back to draft.

andrewvyrros:
I separated the concepts of type erasure (AnyXXX) and polymorphism. There's still more cloning than I'd like, but the logic is cleaner anyway.

rust/opendp/src/any.rs
```diff
@@ -0,0 +1,393 @@
...
+        Transformation::new(
+            AnyDomain::new(*self.input_domain),
+            AnyDomain::new(*self.output_domain),
+            self.function.as_any(),
+            AnyMetric::new(*self.input_metric),
+            AnyMetric::new(*self.output_metric),
+            self.stability_relation.as_any(),
+        )
+    }
+}
```
Shoeboxam, line 366: I was surprised that any.rs is a member of the opendp crate, as it seems like an FFI feature. But you "can't define an inherent impl" for a foreign type. If these were split off into standalone functions, then this could be moved outside of core.

```rust
pub fn transformation_into_any<DI: 'static + Domain, DO: 'static + Domain, MI: 'static + Metric, MO: 'static + Metric>(
    trans: Transformation<DI, DO, MI, MO>
) -> Transformation<AnyDomain, AnyDomain, AnyMetric, AnyMetric>
    where DI::Carrier: 'static,
          DO::Carrier: 'static,
          MI::Distance: 'static + Clone + PartialOrd,
          MO::Distance: 'static + Clone + PartialOrd {
    Transformation::new(
        AnyDomain::new(*trans.input_domain),
        AnyDomain::new(*trans.output_domain),
        trans.function.as_any(),
        AnyMetric::new(*trans.input_metric),
        AnyMetric::new(*trans.output_metric),
        trans.stability_relation.as_any(),
    )
}
```

I see it as, any.rs is in core for the .to_any method convenience.

rust/opendp-ffi/src/core.rs
```diff
@@ -193,7 +193,7 @@ impl Domain for FfiDomain {
     fn member(&self, _val: &Self::Carrier) -> bool { unreachable!() }
 }
 
-#[derive(Clone)]
+#[derive(Clone, PartialEq)]
```
Shoeboxam, line 196: Nice thinking ahead for when we have nontrivial types.

rust/opendp/src/any.rs
```diff
@@ -0,0 +1,393 @@
...
+
+impl CloneableAnyBox {
+    pub fn new<T: 'static + PartialEq + Clone>(value: T) -> Self {
+        let eq_ = |self_: &Self, other: &Self| {
+            // This downcast of self will always succeed, so equality check is all that's necessary.
+            self_.value.downcast_ref::<T>() == other.value.downcast_ref::<T>()
+        };
+        let clone_ = |self_: &Self| {
+            Self {
+                value: Box::new(self_.value.downcast_ref::<T>().unwrap().clone()),
```
Shoeboxam, line 61: Minor, but .unwrap_assert would be nice here. I don't consider this blocking.

rust/opendp/src/any.rs
```diff
@@ -0,0 +1,393 @@
...
+impl Downcast for AnyBox {
+    fn downcast<T: 'static>(self) -> Fallible<T> {
+        self.value.downcast().map_or_else(|_| fallible!(FailedCast), |v| *v)
+    }
+    fn downcast_ref<T: 'static>(&self) -> Fallible<&T> {
+        self.value.downcast_ref().ok_or_else(|| err!(FailedCast))
+    }
+}
+
+pub struct CloneableAnyBox {
```
Shoeboxam, line 47: Naming this is impossible, but it's a wrapper for any type that implements Clone and Eq, and provides a way to downcast- not just a wrapper for cloneable types.

rust/opendp/src/any.rs
Shoeboxam: I take it back- while this type isn't explicitly used in interactive.rs, in practice you're likely to want to use it to erase the type from the interactive composition queryable.

rust/opendp/src/poly.rs
```diff
@@ -0,0 +1,117 @@
...
+    pub fn into_poly(self) -> Function<DI, PolyDomain> {
+        let function = move |arg: &DI::Carrier| -> Fallible<Box<dyn Any>> {
+            let res = (self.function)(arg);
+            res.map(|o| Box::new(*o) as Box<dyn Any>)
+        };
+        Function::new_fallible(function)
+    }
+}
+
+impl<DI: Domain> Function<DI, PolyDomain> {
```
Shoeboxam, line 34: This is cool! `eval_poly` with the type arg to downcast to, exists on `Box<Any>` outputs.

rust/opendp/src/interactive.rs
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    /// Processes a query, generating a new Queryable state.
+    fn transition(self, (measurement, d_out_query): &AcQuery<DI, DO, MI, MO>) -> Fallible<(Self, DO::Carrier)>
+        where MO::Distance: Clone + MeasureDistance {
+        self.check_types(measurement)?;
+        self.check_budget(&measurement.privacy_relation, d_out_query)?;
+        let res = measurement.function.eval(&self.data)?;
+        let new = self.update(d_out_query)?;
+        Ok((new, res))
```
Shoeboxam, line 117: Great- function only runs if query is dp and budget sufficient. Budget only consumed if function evaluates.

rust/opendp/src/interactive.rs
```diff
@@ -0,0 +1,274 @@
...
+          MI: 'static + Metric,
+          MI::Distance: 'static + MetricDistance + Clone,
+          MO: 'static + Measure,
+          MO::Distance: 'static + MeasureDistance + Clone {
+    AcMeasurement::new(
+        input_domain.clone(),
+        AllDomain::new(),
+        Function::new(enclose!((input_domain, input_metric, output_measure, d_in_budget, d_out_budget), move |arg: &DI::Carrier| -> AcQueryable<DI, DO, MI, MO> {
+            AcQueryable::new(
+                // TODO: Remove these clones and have the Queryable use refs? (Also remove Clone trait bounds.)
```
Shoeboxam, line 144: You can drop input_domain from enclose!, because you only use input_domain in the closure. Since you only use input_domain in the closure, it is fine for the closure to consume it.

Shoeboxam: To drop these clones I feel like you'd want a lifetime bound on AcState based on the lifetime of AcMeasurement. To do that I think you'd have to somehow pass Function a reference to the measurement it is a part of, to get a reference to AcMeasurement's input_domain (line 140), output_domain (141), input metric (148) and output measure (149). Not going to happen.

Shoeboxam: Put a different way, the desired behavior requires a struct that contains a reference to itself.

Shoeboxam (approved):

Shoeboxam:
Thanks Andy! The type erasure in particular is really impressive. Looking forward to dropping Ffi-specific types in favor of the type-erased versions of the originals.

rust/opendp/src/any.rs
andrewvyrros: That's a great point. I think your initial instinct was right, `any.rs` should be in opendp-ffi. For the more limited form of type erasure we need for composition, I made `poly.rs`. It just didn't occur to me to move the any stuff to opendp-ffi, but I *think* it can live there.

rust/opendp/src/any.rs
```diff
@@ -0,0 +1,393 @@
...
+    value: Box<dyn Any>,
+    eq_glue: Rc<dyn Fn(&Self, &Self) -> bool>,
+    clone_glue: Rc<dyn Fn(&Self) -> Self>,
+}
+
+impl CloneableAnyBox {
+    pub fn new<T: 'static + PartialEq + Clone>(value: T) -> Self {
+        let eq_ = |self_: &Self, other: &Self| {
+            // This downcast of self will always succeed, so equality check is all that's necessary.
+            self_.value.downcast_ref::<T>() == other.value.downcast_ref::<T>()
```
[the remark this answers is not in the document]
andrewvyrros: Haha yeah I was proud of myself for coming up with this.

rust/opendp/src/any.rs
andrewvyrros: Oops good catch, I'll pick that up later.

rust/opendp/src/any.rs
andrewvyrros: Yeah not a great name, I'll try to come up with some alternatives.

rust/opendp/src/poly.rs
andrewvyrros: Yeah I think this should work as our general purpose composition type.

rust/opendp/src/interactive.rs
andrewvyrros: Thanks, I knew there were issues here, this was the cleanest I could work out. I think this is a good section for us to revisit in a pair programming session.

PR #94, opened 2021-05-19
split type_args into individual arguments

### LINKED ISSUE #93: type_args are first class arguments
By convention, functions with generic dispatch take a type_args string as the first argument. We should promote these type_args to first-class arguments. This will simplify bindings generation, make the python api consistent with the rust api, avoid string interpolation on the python side, and avoid string parsing on the rust side.

Shoeboxam:
Closes #93


andrewvyrros:
As expected, the Python tests aren't working. This is only transitional, so I'm overriding and merging.

andrewvyrros (approved):

PR #96, opened 2021-05-24
Migrated FFI wrappers to use type erasure with AnyXXX types:

andrewvyrros:
Closes #88.

Sorry for the huge change!

I made a cleanup pass on the bootstrap.json files, but feel free to overwrite if it conflicts with the new bindings generation.

* Moved any.rs to opendp-ffi.
* Made AnyBox be compile-time polymorphic.
* Implemented AnyObject data wrapper.
* Created a Glue abstraction for type erasure.
* Migrated data loaders.
* Migrated constructors.
* Added some wrapping helpers.
* Implemented translators for FfiResult to Result for easier testing.
* Wrote unit tests for most FFI wrappers.
* Added some traits and utility methods to make constructors and tests cleaner.
* Updated boostrap.json instances.

PR #98, opened 2021-05-27
Fixed tuple issues in opendp_data__object_as_slice()

### LINKED ISSUE #97: opendp_data__object_as_slice() doesn't handle (AnyObject, AnyObject)
In the refactoring of #88, we lost the ability to handle tuples of AnyObject. This is probably OK in the inbound direction (`opendp_data__slice_as_object()`), because it's unlikely for an operation to consume a tuple like that. But that kind of tuple is produced by basic composition, so we should support it in the outbound direction (`opendp_data__object_as_slice()`).

andrewvyrros:
Closes #97.

Note that, snuck into this change, there's a cfg thing in dispatch.rs so that dev and test builds will compile with a limited set of types for faster monomorphization. If this is a problem, I'm happy to back it out and keep it in my own clone.

* Cleaned up tuple logic.
* Added AnyObject support to type dispatch.
* Wrote some unit tests to cover tuple cases.
* Made a cfg with a stripped-down set of types for fast compilation of dev target.
* Picked up suggested tweaks from #96 PR review.

### DIFF
--- a/rust/opendp-ffi/src/data/mod.rs
+++ b/rust/opendp-ffi/src/data/mod.rs
@@ -1,7 +1,6 @@
 use std::collections::HashMap;
 use std::convert::TryFrom;
 use std::ffi::c_void;
-use std::fmt::Debug;
 use std::os::raw::c_char;
 use std::slice;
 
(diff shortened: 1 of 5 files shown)


rust/opendp-ffi/src/data/mod.rs
```diff
@@ -67,13 +66,10 @@ pub extern "C" fn opendp_data__slice_as_object(raw: *const FfiSlice, T: *const c
...
-                // primitively typed tuples
-                dispatch!(raw_to_tuple, [(types[0], @primitives), (types[1], @primitives)], (raw))
-            } else {
-                // boxy tuples
-                dispatch!(raw_to_plain, [(T, @primitives)], (raw))
-            }
+            let types = try_!(element_ids.iter().map(Type::of_id).collect::<Fallible<Vec<_>>>());
+            // In the inbound direction, we can handle tuples of primitives only. This is probably OK,
+            // because the only likely way to get a tuple of AnyObjects is as the output of composition.
+            dispatch!(raw_to_tuple, [(types[0], @primitives), (types[1], @primitives)], (raw))
```
Shoeboxam, line 72: It's nice to get this cleaned up.

rust/opendp-ffi/src/dispatch.rs
```diff
@@ -97,6 +105,41 @@ macro_rules! disp_expand {
     };
 }
 
+#[cfg(debug_assertions)]
```
Shoeboxam, line 108: Looking forward to the quicker debug loop!

Shoeboxam (approved):

PR #151, opened 2021-06-22
add make_count_distinct

### LINKED ISSUE #149: Count-Distinct Transformation -- Implementation

Shoeboxam:
This is just a quick constructor for make_count_distinct.
Closes #149.

### DIFF
--- a/rust/opendp/src/trans/count.rs
+++ b/rust/opendp/src/trans/count.rs
@@ -28,6 +28,24 @@ pub fn make_count<MI, MO, TI>() -> Fallible<Transformation<VectorDomain<AllDomai
         StabilityRelation::new_from_constant(MO::Distance::one())))
 }
 
+
+pub fn make_count_distinct<MI, MO, TI>() -> Fallible<Transformation<VectorDomain<AllDomain<TI>>, AllDomain<MO::Distance>, MI, MO>>
+    where MI: DatasetMetric,
+          MO: SensitivityMetric,
+          MO::Distance: TryFrom<usize> + Bounded + One + DistanceConstant,
+          TI: Eq + Hash {
+    Ok(Transformation::new(
+        VectorDomain::new_all(),
+        AllDomain::new(),
+        Function::new(move |arg: &Vec<TI>| {
+            let len = arg.iter().collect::<HashSet<_>>().len();
+            MO::Distance::try_from(len).unwrap_or(MO::Distance::max_value())
+        }),
+        MI::default(),
+        MO::default(),
+        StabilityRelation::new_from_constant(MO::Distance::one())))
+}
+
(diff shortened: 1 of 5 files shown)


rust/opendp/src/trans/count.rs
```diff
@@ -28,6 +28,24 @@ pub fn make_count<MI, MO, TI>() -> Fallible<Transformation<VectorDomain<AllDomai
         StabilityRelation::new_from_constant(MO::Distance::one())))
 }
 
+
+pub fn make_count_distinct<MI, MO, TI>() -> Fallible<Transformation<VectorDomain<AllDomain<TI>>, AllDomain<MO::Distance>, MI, MO>>
```
andrewvyrros, line 32: Huh, I hadn't noticed this before, is there a reason the output carrier is `MO::Distance`? In other places, for dataset-sized things, don't we fix it to `u32`?

rust/opendp/src/trans/count.rs
```diff
@@ -28,6 +28,24 @@ pub fn make_count<MI, MO, TI>() -> Fallible<Transformation<VectorDomain<AllDomai
...
+    where MI: DatasetMetric,
+          MO: SensitivityMetric,
+          MO::Distance: TryFrom<usize> + Bounded + One + DistanceConstant,
+          TI: Eq + Hash {
+    Ok(Transformation::new(
+        VectorDomain::new_all(),
+        AllDomain::new(),
+        Function::new(move |arg: &Vec<TI>| {
+            let len = arg.iter().collect::<HashSet<_>>().len();
+            MO::Distance::try_from(len).unwrap_or(MO::Distance::max_value())
```
andrewvyrros, line 42: Rather than return `max_value()`, wouldn't it be better to fail?

rust/opendp/src/trans/count.rs
Shoeboxam: The reason the output distance is generic is because the geometric mechanism expects the distance and datatype to share the same type. I could make these independent and add an additional cast to the geometric mechanism. I'm not sure if that's better?

rust/opendp/src/trans/count.rs
Shoeboxam: I could set TO=i8, and if the function fails, I know the data has at least 128 elements. Then repeat for each datatype to get a rough bound on the power-of-2 length of the private data, without being charged for this information.

rust/opendp/src/trans/count.rs
andrewvyrros: Ah thanks, yeah probably not worth it.

rust/opendp/src/trans/count.rs
andrewvyrros: Good point!

andrewvyrros (approved):

PR #157, opened 2021-06-25
Merge dataset metrics; merge ffi constructors

### LINKED ISSUE #147: Merge scalar and vector constructors in FFI wrappers
With #145, we're merging scalar and vector constructors in Rust to be a single constructor, dispatched by input Domain.

We should do the same with the FFI wrappers. To avoid a lot of annoying boilerplate in Python, the default domain could be something like `VectorDomain<AllDomain<_>>`.

### LINKED ISSUE #156: Unify dataset distances as SymmetricDistance
After some discussion, we've decided to unify treatment of dataset distances:

* Keep the name `SymmetricDistance` (do not rename to `SymmetricDifference`).
* Rename `HammingDistance` to `SubstituteDistance`.
* Standardize on `SymmetricDistance` for all appropriate constructors. Remove dispatch for `SubstituteDistance` (but keep the type).
* For applications that desire it, provide a conversion constructor that can cast `Transformation<_, _, SymmetricDistance, _>` to `Transformation<_, _, SubstituteDistance, _>` with a 1/2 stability factor.
* Aside: For things like `make_sum`, we don't try to model the case of (datasets of unknown but same size, neighbors differ by substitutions). This is because it's impossible to compare datasets of different sizes (infinite distance).

Shoeboxam:
Closes #156 
- Remove dataset metric generic from constructors throughout the library in favor of SymmetricDistance
    * from core, ffi layer, json and rust/python tests

Closes #147
- Add Domains to the static ref Types type map
    * adjusted type_vec! macro to take nested generics of the form type_vec!([A B C] <a, b, c>)
- Merged all ffi functions for each constructor. Expose domain type arguments instead
- typing.py adjustments for domains and working with generics
    * add generics argument to parser functions to identify generics in RuntimeTypes
    * add domain types and update RuntimeType parser to detect domains `VectorDomain[AllDomain["T"]]`
    * get_domain_atom, for retrieving the concrete type at the root of a RuntimeType
 - Revise python.rs codegen to
    * allow default arguments to be null (for Option<T> arguments)
    * identify generics when parsing types into RuntimeTypes
    * substitute concrete types into generics in RuntimeTypes

### DIFF
--- a/python/src/opendp/v1/typing.py
+++ b/python/src/opendp/v1/typing.py
@@ -15,11 +15,11 @@ ELEMENTARY_TYPES = {int: 'i32', float: 'f64', str: 'String', bool: 'bool'}
 
 # all ways of providing type information
 RuntimeTypeDescriptor = Union[
-    "RuntimeType",  # as the normalized type -- HammingDistance; RuntimeType.parse("i32")
+    "RuntimeType",  # as the normalized type -- SubstituteDistance; RuntimeType.parse("i32")
     _GenericAlias,  # a python type hint from the std typing module -- List[int]
     str,  # plaintext string in terms of rust types -- "Vec<i32>"
     Type[Union[typing.List, typing.Tuple, int, float, str, bool]],  # using the python type class itself -- int, float
-    tuple,  # shorthand for tuples -- (float, "f64"); (HammingDistance, List[int])
+    tuple,  # shorthand for tuples -- (float, "f64"); (SubstituteDistance, List[int])
 ]
 
 
@@ -102,7 +102,7 @@ class RuntimeType(object):
             # attempt to upgrade strings to the metric/measure instance
             origin = type_name[:start] if 0 < start else type_name
             closeness = {
-                'HammingDistance': HammingDistance,
+                'SubstituteDistance': SubstituteDistance,
                 'SymmetricDistance': SymmetricDistance,
                 'AbsoluteDistance': AbsoluteDistance,
                 'L1Distance': L1Distance,
@@ -236,7 +236,7 @@ class DatasetMetric(RuntimeType):
     pass
 
 
-HammingDistance = DatasetMetric('HammingDistance')
+SubstituteDistance = DatasetMetric('SubstituteDistance')
 SymmetricDistance = DatasetMetric('SymmetricDistance')
 
 
--- a/python/src/opendp/v1/typing.py
+++ b/python/src/opendp/v1/typing.py
@@ -95,6 +95,8 @@ class RuntimeType(object):
         # parse a string-- "Vec<f32>",
         if isinstance(type_name, str):
             type_name = type_name.strip()
+            if type_name.startswith('_'):
+                return GenericType(type_name[1:])
             if type_name.startswith('(') and type_name.endswith(')'):
                 return RuntimeType('Tuple', cls._parse_args(type_name[1:-1]))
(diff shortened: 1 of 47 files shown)


rust/opendp/src/trans/count.rs
```diff
@@ -4,77 +4,56 @@ use std::convert::TryFrom;
...
+        SymmetricDistance::default(),
+        AbsoluteDistance::default(),
+        StabilityRelation::new_from_constant(TO::one())))
 }
 
-impl<QO: FloatConst> CountByConstant<HammingDistance, L2Distance<QO>> for (HammingDistance, L2Distance<QO>) {
-    fn get_stability_constant() -> Fallible<QO> {
-        Ok(QO::SQRT_2())
-    }
+pub trait CountByConstant: SensitivityMetric {
```
andrewvyrros, line 44: It feels weird to make this a sub-trait of `SensitivityMetric`. I get the desire to reduce the number of trait bounds in function signatures, but I think this one actually makes the bounds on `make_count_by_categories()` a little cryptic. I think it'd be more clear with an explicit `where MO: SensitivityMetric + CountByConstant`.

python/src/opendp/v1/typing.py
```diff
@@ -112,11 +117,23 @@ def parse(cls, type_name: RuntimeTypeDescriptor) -> Union["RuntimeType", str]:
...
                 return closeness
 
+            domain = {
+                'AllDomain': AllDomain,
+                'IntervalDomain': IntervalDomain,
+                'VectorDomain': VectorDomain,
+                'OptionNullDomain': OptionNullDomain,
+                'InherentNullDomain': InherentNullDomain,
+                'SizedDomain': SizedDomain,
+                'SmoothedMaxDivergence': SmoothedMaxDivergence
```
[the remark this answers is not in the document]
Shoeboxam: Yep. Removed. Thanks!

rust/opendp-ffi/build/main.rs
```diff
@@ -40,6 +40,10 @@ pub struct Argument {
     // RuntimeType expressed in terms of rust types with generics.
     // Includes various RuntimeType constructors
     rust_type: Option<RuntimeType>,
+    // RuntimeType expressed in terms of rust types with generics.
+    // Includes various RuntimeType constructors
```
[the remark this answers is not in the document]
Shoeboxam: Yes, updated.

rust/opendp/src/trans/count.rs
Shoeboxam: Hmm, the motivation wasn't just to reduce the number of bounds- the stability constant is only well defined with respect to a metric, and it returns a constant of the distance type of that metric. That being said, this still works:

```rust
pub trait CountByConstant<QO> {
    fn get_stability_constant() -> QO;
}
impl<Q: One, const P: usize> CountByConstant<Q> for LpDistance<Q, P> {
    fn get_stability_constant() -> Q { Q::one() }
}
```

With updated trait bound:
`where MO: CountByConstant<MO::Distance> + SensitivityMetric,`

I'm fine with either.

andrewvyrros: Ah OK, makes sense, thanks!

andrewvyrros (approved):
Great stuff! Standardizing on SymmetricDistance and merging the rest of the constructors are both nice wins. It keeps getting better!

PR #162, opened 2021-07-01
Crate and package publish

### LINKED ISSUE #137: Windows build with system libs
We have a workaround to build on Windows (#100), but this disables MPFR and OpenSSL. We need to figure out a solution that includes everything.

### LINKED ISSUE #138: Python PyPI package upload workflow
We need to get the `opendp` Python package into PyPI. It should be a scripted process, rather than manual. GitHub has support for doing this via Actions. Ideally, we should be able to tag a version, and it happens automatically.

We already have a basic `setup.py` going, so hopefully this won't be a ton of work:

- [ ] Make sure all metadata is present and reasonable.
- [ ] Sort out multi-platform issues. (If there isn't a quick solution to #137, then we can disable system libs for now.)
- [ ] Hook into GitHub Actions.

### LINKED ISSUE #139: Rust crate upload workflow
We need to get the `opendp` and `opendp-ffi` Rust crates into `crates.io`. It should be a scripted process, rather than manual. I think GitHub has support for doing this via Actions. Ideally, we should be able to tag a version, and it happens automatically.

We already have a basic setup going, so hopefully this won't be a ton of work:

- [ ] Make sure all metadata is present and reasonable.
- [ ] Sort out multi-platform issues. (If there isn't a quick solution to #137, then we can disable system libs for now.)
- [ ] Hook into GitHub Actions.

### LINKED ISSUE #172: pypi packaging creds

Shoeboxam:
Closes #137
Closes #138
Closes #139 
Closes #172 

Adds a VERSION file to the repo root with the current version.
Adds a check to the CI that ensures all package versions throughout the repository are equivalent.

Adds a windows folder to the repo root with build instructions and patches for windows.
Adds a Github actions release.yml that 
- triggers on github release
- checks that the person that created the release is an admin
- pushes the rust crates to crates.io using the opendp-dev github account
- builds binaries on windows, mac and manylinux
- embeds the binaries into the python package, builds a python wheel, and uploads to pypi


raprasad (changes requested):
Currently the python code is under a `v1` directory which translates to import statements such as:  `from opendp.v1.trans import *`

Consider removing the `v1` from the module namespace unless it's not going to change.

Once people start using the PyPI package, the package itself will be versioned.

andrewvyrros (approved):
Fantastic, will be great to have an automated release process.

PR #164, opened 2021-07-06
Fix "nan" cast; make_is_null transform

### LINKED ISSUE #163: CheckNull trait
This is to cover the specific behavior when the string "nan" is passed when calling make_cast with TO={f64|f32}. 

The output type of make_cast should have a trait CheckNull, to check if the value may be null and is null. If it is, then return None instead. It should be impossible to construct Some(NaN) vector elements, because OptionNullDomain<AllDomain<T>> assumes the AllDomain<T> is non-null. The compiler should be able to inline and drop this check for types that cannot contain inherent nullity.

An alternative is to modify the Cast trait to check for "nan" and error instead of succeeding with a NaN value. This alternative may lead to unexpected behavior when Cast is reused, and is less efficient because it needs to do a bunch of string manipulation.

Shoeboxam:
Closes #163

This PR reuses the CheckNull trait to resolve #163 and to add a make_is_null transform.

### DIFF
--- a/python/test/test_trans.py
+++ b/python/test/test_trans.py
@@ -69,6 +69,30 @@ def test_identity():
     assert ret == arg
 
 
+def test_is_equal():
+    from opendp.v1.trans import make_is_equal
+    tester = make_is_equal(3)
+    assert tester([1, 2, 3]) == [False, False, True]
+
+
+def test_is_null():
+    from opendp.v1.trans import make_split_lines, make_cast_inherent, make_is_null
+    tester = (
+        make_split_lines() >>
+        make_cast_inherent(TI=str, TO=float) >>
+        make_is_null(DIA=InherentNullDomain[AllDomain[float]])
+    )
+    assert tester("nan\n1.\ninf") == [True, False, False]
+
+    from opendp.v1.trans import make_split_lines, make_cast, make_is_null
+    tester = (
+        make_split_lines() >>
+        make_cast(TI=str, TO=float) >>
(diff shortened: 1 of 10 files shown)


python/test/test_trans.py
```diff
@@ -69,6 +69,30 @@ def test_identity():
...
+    assert tester("nan\n1.\ninf") == [True, False, False]
+
+    from opendp.v1.trans import make_split_lines, make_cast, make_is_null
+    tester = (
+        make_split_lines() >>
+        make_cast(TI=str, TO=float) >>
+        make_is_null(DIA=OptionNullDomain[AllDomain[float]])
+    )
+    assert tester("nan\n1.\ninf") == [True, False, False]
+test_is_null()
```
andrewvyrros, line 94: Spurious line?

```suggestion

```

rust/opendp/src/dom.rs
```diff
@@ -17,6 +17,9 @@ use crate::error::Fallible;
 pub struct AllDomain<T> {
     _marker: PhantomData<T>,
 }
+impl<T> Default for AllDomain<T> {
+    fn default() -> Self { AllDomain { _marker: PhantomData } }
```
andrewvyrros, line 21: I think it's a little cleaner if these call `Self::new()`?

Shoeboxam: Agreed, switched.

andrewvyrros (approved):

PR #170, opened 2021-07-07
Put docs in main repo, working API reference

### LINKED ISSUE #148: docs in OpenDP library repo
Currently https://docs.opendp.org is built using Sphinx from content in https://github.com/opendp/opendp-documentation

In order to simplify the release process, we'd like to try having a directory here in the main repo called "docs" that would be the source of truth.

We'll probably set up GitHub Actions to push to GitHub Pages (a new subdomain for now) based on the existing config at https://github.com/opendp/opendp-documentation/blob/latest/.github/workflows/main.yml

I don't want to rename the default branch to "latest" to accommodate the Sphinx extension I'm using so we'll need to revisit the conversation at https://github.com/opendp/opendp-documentation/issues/15#issuecomment-848204550

In terms of content, I created an information architecture document at https://docs.google.com/document/d/1htapY15tco2ErXQaNEvViRNVjTQOAhP4S023w372CM4/edit?usp=sharing

- [ ] IA document
- [ ] add docs directory
- [ ] set up basic Sphinx to publish to (doc-test or similar).opendp.org

pdurbin:
Closes #148 

As part of this pull request, I went ahead and incorporated the Python docs @Shoeboxam added but rather than using his "build docs" script, I tweaked the Sphinx config file. Here's how it looks:

Screen Shot 2021-07-07 at 3 57 37 PM

A few things to note:

- If you'd like to try this branch locally, `make html` should work fine but if you want to try `make versions` you'll have to go into `source/conf.py` and change "main" to "148-docs" in the line that says `smv_branch_whitelist = r'^main$'`
- Right now I have `smv_branch_whitelist` set to "main" but somehow I'd like the docs for "main" to appear at a URL for "latest". To achieve this in the docs repo, we used "latest" rather than "main" as the default branch but I'm hoping we can avoid this.
- I'm including the GitHub Action we use in the docs repo but I changed it (and the redirect page) to "main" instead of "latest". I haven't get set up the gh-pages branch, CNAME, etc. For now the CNAME (in the GitHub Action) is docz.opendp.org until we decide we're happy with this setup of having the docs in the same repo.
- Since I'm not using Mike's "build docs" script, I think we'll have to manually create files like `opendp.v1.meas.rst` when we add a new module. These are basically small index files and if this becomes a problem, we can probably figure out some automation.


pdurbin:
I'm switched this back to a non-draft pull request.

First, you can now preview the live site here: https://docz.opendp.org/en/148-docs/index.html

Yes, one we merge, we'll change "docz" to "docs" and make it the live site.

I still don't have a solution for "latest" vs "main". The new site (this pull request) will use "main", the default branch. I did open an issue upstream today: https://github.com/Holzhaus/sphinx-multiversion/issues/78

docs/source/user/related-projects.rst
```diff
@@ -0,0 +1,22 @@
...
+
+There are a number of projects that are related to OpenDP.
+
+.. contents:: |toctitle|
+	:local:
+
+Google Differential Privacy
+---------------------------
+
+`Google Differential Privacy`_ consistes of libraries to generate ε- and (ε, δ)-differentially private statistics over datasets.
```
Shoeboxam, line 12:
```suggestion
`Google Differential Privacy`_ consists of libraries to generate ε- and (ε, δ)-differentially private statistics over datasets.
```

.github/workflows/docs.yml
```diff
@@ -0,0 +1,62 @@
+name: CI
```
Shoeboxam, line 1:
```suggestion
name: Documentation
```

docs/source/user/related-projects.rst
pdurbin: Good catch!

.github/workflows/docs.yml
```diff
@@ -0,0 +1,62 @@
...
+          git config --global user.name "$GITHUB_ACTOR"
+          git checkout gh-pages
+          rm -rf *
+          # TODO: Update to final hostname.
+          echo docz.opendp.org > CNAME
+          echo "for underscore directories" > .nojekyll
+          cp -r /tmp/build/html/* .
+          git add --all --force
+      - name: Push docs to gh-pages branch
+        if: success() && github.ref == 'refs/heads/main'
```
Shoeboxam, line 56: How often do we want to update the documentation? Per-merge or per-release? 
In this PR https://github.com/opendp/opendp/pull/162, CI will release a new python package and crate whenever there is a push to the `release` branch. If you tie documentation deployment to the release branch, then the documentation will update less frequently, only when a release is cut.

.github/workflows/docs.yml
pdurbin: Good catch. "CI" made more sense in the docs repo but now "Documentation" makes much more sense.

docs/source/conf.py
```diff
@@ -0,0 +1,114 @@
...
+# General information about the project.
+project = u'OpenDP'
+copyright = u'%d' % datetime.now().year
+
+# The version info for the project you're documenting, acts as replacement for
+# |version| and |release|, also used in various other places throughout the
+# built documents.
+#
+# The short X.Y version.
+version = '0.0.1'
```
Shoeboxam, line 50: I was attempting to add some global checks to make sure version numbers across projects are synchronized in the CI release PR. There's a root-level VERSION file and a `.github/assert_version.py` that checks version numbers. I'm not sure if that's how we want to handle things, but if it is, maybe this could read the version from the VERSION file?

.github/workflows/docs.yml
Shoeboxam: I guess it could update the docs each merge into main, where the updated docs includes both the `latest` branch (renamed from `main`) and all of the tags. 

Would checking out from main to a new branch `latest` in the ci script not work?

pdurbin: At least for the time being, while docs are being worked on quite actively, I was thinking the docs site should be updated when pull requests were merged.

In this pull request, the updated docs would appear under "main" (docs.opendp.org/en/main), reflecting the fact that the pull request was merged into the main branch.

Assuming a regular release cycle in the future, I'm open to only updating docs on release. That's what we do for Dataverse. That said, even in the context of Dataverse, when we haven't released for a while, I'm often interested in seeing what the updated docs will look like, so I think there's value in always having "main" or "latest" to be able to preview what's coming in the next release.

> I guess it could update the docs each merge into main, where the updated docs includes both the `latest` branch (renamed from `main`) and all of the tags.

I'm not sure I fully grok this. Is renaming `main` on the table? The extension I'm using does support tags. No problem there. That's my plan for people being able to switch between different versions of the library.

I don't think the `checkout` above will work but in the same vein, we might be able to have the CI whatever is in `main` to a branch called `latest` and then run `make versions`. I haven't tried this yet.

docs/source/conf.py
pdurbin: Yes, I absolutely want us to have a single source of truth for the version. I just threw 0.0.1 in there for discussion. I'm glad you found it. Reading from a file should work, I would imagine.

.github/workflows/docs.yml
Shoeboxam: Right, if we don't rename main, then each time main is pushed to, CI could run something like:
```
git branch -D latest && git checkout -b latest
make versions
```
By "rename from main" I was referring to the checkout. I don't think I'm the guy to decide if we will rename main. I agree that it makes sense to push docs more frequently to a `latest` version.

pdurbin: Oh, I see. Delete the `latest` branch and then create it again from `main`. Yeah, I would hope that would work. I was thinking of a merge from `main` to `latest` but I'm not picky. 😄

Shoeboxam: I'm just brainstorming. `merges` would work too.

pdurbin:
I just made several commits to get the docs previewing again at https://docz.opendp.org/en/148-docs/index.html

This seems necessary if we're going to work on content and various rearranging, as @andrewvyrros has proposed.

Before merging, we'll need to revert a number of commits. I thought 66bd776 would be enough to get previews working again. I don't think 21cfbe4 was actually necessary for me to commit but I kept getting "No matching refs found!". In the end
8c91ef5 got the build working (got that error to go away) which is very strange. It's as if sphinx-multiversion only works when the branch you want to build is specified under "branches" in the GitHub Action. Long term, this is not what we want. We want to be able to do builds of pull requests, not just builds from merges to main.

python/src/opendp/v1/_lib.py
```diff
@@ -10,31 +10,26 @@
...
     }
     if sys.platform not in platform_to_name:
         raise Exception("Platform not supported", sys.platform)
-    return platform_to_name[sys.platform]
-
+    lib_name = platform_to_name[sys.platform]
 
-lib_path = os.path.join(_get_lib_dir(), _get_lib_name())
-lib = ctypes.cdll.LoadLibrary(lib_path)
+    lib = ctypes.cdll.LoadLibrary(os.path.join(lib_dir, lib_name))
```
andrewvyrros, line 32: Something seems funky here. Are these changes intended, or is this a merge issue?

Shoeboxam: This was something I added. When the docs build, it sources the library. If the binaries don't exist, the sourcing fails, so the docs fail to build. So he had to start building the rust binaries from the docs workflow to get the docs to build. This changes it so that if the env var `OPENDP_HEADLESS` is set, then the binaries are not loaded.

pdurbin:
> This seems necessary if we're going to work on content and various rearranging, as @andrewvyrros has proposed.

I just pushed the content reorg @andrewvyrros and I have been talking about: https://docz.opendp.org/en/148-docs/index.html

Please note that many of the pages are blank but the structure is there.

Even thought this pull request is a mix of content and deployment, these topics can be discussed separately.

For content, I'd like to bring up the following comments or questions

- Now that we have a "Resources" section, how should we treat the various links at https://github.com/opendp/opendp-documentation/issues/28 ? Perhaps some of the links can go under the page called "Tutorials". Do we need more pages under "Resources" for these links?
- I had a page called "Making Releases" but didn't see this in the new reorg so I took it out. I still think it's valuable to document your release process but perhaps it's premature since we haven't made a release yet.
- I had a page in that talked about various aspects of contributing with topics like Roadmap, Kanban Board, Issue Tracker, Getting Help, Other OpenDP Projects. Do we want to add some of this back in somewhere?
- DP Creator now has its own page. Is this necessary? It's also listed on the main OpenDP Commons page.
- Related to OpenDP Commons, I renamed index to intro to make the hierarchy work but it seems odd. I could use another pair of eyes here.

I have lots of thoughts on deployment and CI too:

- Should we be pushing the HTML to a different repo? Right now it goes to the gh-pages branch.
- Do we like having "en" in the URLs? Is this overkill? Will we ever have translations?
- Where should we redirect the root to until stable docs exist? "en/latest"?
- sphinx-multiversion has been a mixed blessing. I may want to drop it but we'll need some sort of strategy for versioning, either building our own solution or hosting with a place like Read The Docs.

andrewvyrros:
Thanks Phil, this is coming together!

> - Now that we have a "Resources" section, how should we treat the various links at https://github.com/opendp/opendp-documentation/issues/28 ? Perhaps some of the links can go under the page called "Tutorials". Do we need more pages under "Resources" for these links?

Yeah I think it makes sense to have separate pages under Resources. I think most of the stuff in that issue could fit in a page called References or Educational Material or similar. For Tutorials, I was imagining that being just coding tutorials (like end-to-end instructions focused on a single use case.)

> - I had a page called "Making Releases" but didn't see this in the new reorg so I took it out. I still think it's valuable to document your release process but perhaps it's premature since we haven't made a release yet.

For sure we want to document that. It just felt a little early to have that in our public docs, as we're still inventing it, and don't want to be encouraging people to try to cut their own releases! (@Shoeboxam discovered what seems like an access hole in GH repo permissions around releases.)

> - I had a page in that talked about various aspects of contributing with topics like Roadmap, Kanban Board, Issue Tracker, Getting Help, Other OpenDP Projects. Do we want to add some of this back in somewhere?

The issue tracker and project board stuff is a little tricky, I think that's worth a separate discussion.

For Getting Help, I was thinking that kind of stuff could live under Contact. Alternatively, we could have the top-level module be labeled Help, and have various ways of getting help under there?

- DP Creator now has its own page. Is this necessary? It's also listed on the main OpenDP Commons page.

Yeah I guess it makes sense to have it live under OpenDP Commons.

> - Related to OpenDP Commons, I renamed index to intro to make the hierarchy work but it seems odd. I could use another pair of eyes here.

Yeah something's a little funky there. We probably don't need the second level OpenDP Commons node. Maybe the hierarchy looks like this:

* OpenDP Commons
  * What is the OpenDP Commons?
  * List of OpenDP Commons Tools and Packages
    * OpenDP Library
    * DP Creator
  * Related Projects
    * Google Differential Privacy
    * Additional Projects

> - Should we be pushing the HTML to a different repo? Right now it goes to the gh-pages branch.

I was thinking it'd be cleaner to push the HTML to a separate repo, and not have publishing artifacts in the main repo. But I've never use GH Pages before, so if there's a compelling reason to keep it in a single repo, I'm open to that.

> - Do we like having "en" in the URLs? Is this overkill? Will we ever have translations?

It's probably a bit aspirational, but doesn't seem to harm anything?

> - Where should we redirect the root to until stable docs exist? "en/latest"?

Yeah latest seems reasonable.

- sphinx-multiversion has been a mixed blessing. I may want to drop it but we'll need some sort of strategy for versioning, either building our own solution or hosting with a place like Read The Docs.

Yeah there's always a tradeoff with adopting tools like that. If we have to roll our own solution to get the best docs, then so be it.

andrewvyrros:
I've made it so that the API doc structure is generated dynamically for each release, so we don't have to keep the .rst docs in sync with Python packages. I also came up with a strategy for handling stable and latest branches.

Shoeboxam (approved):
Woohoo! Nice to have in-repo docs! Thanks Phil! Thanks Andy!

PR #173, opened 2021-07-11
Explicit casting traits

### LINKED ISSUE #33: Make DistanceCast properly handle f64 -> f32

Shoeboxam:
Closes #33 

Added traits:
- `TO: RoundCast<TI>` cast from vi: TI to vo: TO with rounding to nearest, ties to even
- `TO: InfCast<TI>` cast where vi: TI <= vo: TO
- `TO: ExactIntCast<TI>` casting for integers where vi: TI == vo: TO, and vo < `TO::MAX_CONSECUTIVE`

Removed `DistanceCast` trait, which still had some edge cases that would round down:
1. When casting from f64 to f32
2. When casting an int greater than 2^p to nearest, ties to even float with precision p

Switch DistanceConstant to use InfCast, instead of DistanceCast.

Remove num::NumCast from most places in the library, in favor of the more explicit cast traits with situationally-appropriate cast behavior.

This allows TO on count queries to be float-valued, so that you can run float-mechanisms on them. We might want to think more about if we want to allow this. If we are unsure, I can add an Integer trait bound on TO.

### DIFF
--- a/rust/opendp/src/traits.rs
+++ b/rust/opendp/src/traits.rs
@@ -1,53 +1,22 @@
+use std::convert::TryFrom;
 use std::ops::{Div, Mul, Sub};
 
-use num::{NumCast, One, ToPrimitive, Zero};
+use num::{NumCast, One, Zero};
 
 use crate::error::Fallible;
-
-pub trait CheckContinuous { fn is_continuous() -> bool; }
-pub trait Ceil: Clone { fn ceil(self) -> Self; }
-macro_rules! impl_is_continuous {
-    ($($ty:ty),+) => {
-        $(
-            impl Ceil for $ty {
-                #[inline]
-                fn ceil(self) -> $ty { self.ceil() }
-            }
-            impl CheckContinuous for $ty {
-                #[inline]
-                fn is_continuous() -> bool {true}
-            }
-        )+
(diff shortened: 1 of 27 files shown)


rust/opendp/src/traits.rs
```diff
@@ -130,95 +84,274 @@ macro_rules! impl_abs_int {
...
+
+
+macro_rules! impl_inf_cast_from {
+    ($ti:ty, $to:ty) => (impl InfCast<$ti> for $to {
+        #[inline]
+        fn inf_cast(v: $ti) -> Fallible<Self> { Ok(From::from(v)) }
+    })
+}
+
+pub trait BoundedInt {
```
andrewvyrros, line 195: I think this name is still a little cryptic, would suggest ExactIntBounds, ExactIntRange, BoundedExactInt or something like that.

andrewvyrros (changes requested):
This is good, makes the different kinds of casting more clear. Nice feedback from the intern work!

andrewvyrros (approved):

PR #174, opened 2021-07-12
close #87, make member fallible

### LINKED ISSUE #87: Domain::member() should be Fallible
For consistency, and to make some Any stuff cleaner, we should make `Domain::member()` return `Fallible<bool>` instead of just `bool`.

Shoeboxam:
Closes #87

This also drops the `any.rs` Bool trait and PhantomData in favor of const generics.


andrewvyrros:
Looks good. Nice use of const generics! I'll pull in the README change from main and merge.

andrewvyrros (approved):

PR #175, opened 2021-07-12
port/debug examples dir

Shoeboxam:
I had to make some adjustments to the AnyMeasurement impls to not wrap Any* members of Measurement structs in an extra Any* layer. The extra wrapper layer breaks PartialEq.

I also had to make _to_string recursively unpack AnyObjects.


andrewvyrros (approved):
Thanks! It's always good to have up to date examples.

PR #176, opened 2021-07-14
N-way basic composition combinator

### LINKED ISSUE #7: Basic Composition Combinator -- Implementation
- [ ] Implement function
- [ ] Include comments
- [ ] Write tests

Shoeboxam:
Closes #7
Incomplete implementation. It might be possible to lift the necessary ::Distance traits up to the Measure trait bounds, to avoid needing to introduce an EpsilonDelta struct.

Andy pushed the design on most of this! Thanks!

- adjusted python to be able to construct a Vec<const AnyMeasurement*>
- add FFI for multi-way composition combinator
- add BasicComposition trait with binary search algorithm for iteratively finding the tightest budget
    * implemented on measures, and captured in the AnyMeasure via a glue function
- adjusted IntoAnyMeasurementOutExt to not double-wrap the relation distances in Any
- add EpsilonDelta struct, and adjust the associated type for SmoothedMaxDivergence to refer to it
    * not sure I like this


Shoeboxam:
Closed in favor of #242.

PR #179, opened 2021-07-15
Properly handle overflows in aggregators

### LINKED ISSUE #178: Overflow handling in Sum
The function in make_bounded_sum and make_bounded_sum_n has no protection against overflow. 

Proposed solution
- make_bounded_sum: Use a saturating sum
- make_bounded_sum_n: Disallow constructing transformations in which n * L < T::MIN or n * U > T::MAX

Shoeboxam:
Closes #178

Adjusts:
- make_bounded_sum (by check)
- make_bounded_sum_n (by saturation)
- make_bounded_variance (by check)
- make_bounded_covariance (by check)
- make_bounded_mean (by check)
- make_count_by_categories (by saturation)
- make_count_by (by saturation)

(by check) means checked arithmetic is performed to compute the maximum/minimum value that may happen at runtime. If the checked arithmetic fails, the computation may fail, so the constructor fails to build the transformation.

(by saturation) means the function itself uses saturating math instead of wrapping or panicking math, which preserves the privacy semantics.

### DIFF
--- a/rust/opendp-ffi/src/trans/count.rs
+++ b/rust/opendp-ffi/src/trans/count.rs
@@ -1,6 +1,5 @@
 use std::convert::TryFrom;
 use std::hash::Hash;
-use std::ops::AddAssign;
 use std::os::raw::{c_char, c_uint};
 
 use num::{Bounded, Integer, One, Zero};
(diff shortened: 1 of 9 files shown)


andrewvyrros (approved):
This is great, another aspect that we're tightening up.

PR #180, opened 2021-07-15
Handle Option<T> constructor arguments as nullable pointers in FFI (2)

### LINKED ISSUE #165: Option<(T, T)> data loader
The geometric mechanism takes an Option<(T, T)> argument, so there needs to be a data loader to support the constant-time geometric mechanism in FFI.

Shoeboxam:
Closes #165
Allow null pointers to be interpreted as None in rust, when the FFI argument is meant to be an Option
This is used in practice for constructing a constant-time base geometric from ffi.

I closed the [old PR](https://github.com/opendp/opendp/pull/167) because I botched up the rebase.


andrewvyrros (approved):
Nice!

PR #184, opened 2021-07-20
Fixed point exploration

Shoeboxam:
Initial exploration of integrating the fixed crate into OpenDP. 

- The bulk of the work to be done here is implementing macros that extend our traits to fixed point types
- Choosing the types to monomorphize in the ffi crate is hairy, as there are two axes of bit depths


Shoeboxam:
Closing this as it has grown stale. One next step for this would be to select a set of monomorphizations and rebase all the trait impls.

PR #188, opened 2021-07-23
Randomized response

### LINKED ISSUE #187: Randomized response -- implementation

Shoeboxam:
Closes #187 

This has three implementations:
1. Randomized response on a bool vector (with central model drawbacks)
2. Randomized response on a bool (local model)
3. Randomized response on a categorical type (with central model drawbacks)

We still need to narrow the design a bit-- 
Why include a local DP algorithm that needs access to the entire dataset?
What should the input metric be for the local implementation?
Can we prune the library size to make the local implementation useful for embedded devices?


Shoeboxam:
Based on feedback from Salil, we should pare this down to only approach 2.

Shoeboxam:
Adjustments made!

PR #195, opened 2021-07-26
Test release fixes

Shoeboxam:
See #194

- updates the metadata on the Cargo.toml and setup.cfg
- removes an extra license file
- adds dummy version number 0.0.0-development
- fixes the release.yml twine upload


andrewvyrros (approved):
Getting there!

PR #219, opened 2021-08-03
Release 0.1.0

### LINKED ISSUE #43: Untrusted mode
Provide a way to activate "untrusted" mode, where privacy guarantees are loosened, and more features are available. This could be used to enable things outside the strict OpenDP constraints:

* newly contributed components that haven't been validated yet
* user-supplied functions (e.g. row_transform)
* permissive floating point calculations

Need to figure out the mechanics of this. Some things we could leverage:

* Rust module(s)
* Rust conditional compilation
* Separate Rust crate
* Python package flags that load different versions of the library

### LINKED ISSUE #200: Update documentation to emphasize -e in python
If a developer omits -e in the pip install, then the python dir is copied into site-packages, and the relative path to find the rust binaries fails. 

We should also link the install docs from the README.md, to make it less likely that people guess their own install command and get this detail wrong!

### LINKED ISSUE #217: PyPi Readme
The setup.cfg should point to a valid readme.

### LINKED ISSUE #220: Restrict Python gen in opendp-ffi build
The opendp-ffi build has a feature, "python", which enables the generation of Python bindings. This is awkward if you want to use opendp-ffi in a Rust project. We should instead gate the code gen to run only when in development mode.

Cargo has an environment variable for this, CARGO_PRIMARY_PACKAGE, which is set only when building a package directly, not when building as a dependency. Unfortunately, this doesn't seem to work.

Instead, we check the value of CARGO_PKG_VERSION, which will be "0.0.0-development" when building in dev mode.

Shoeboxam:
Closes #220
Closes #217 
Closes #200

### DIFF
--- a/.github/build_manylinux_binaries.sh
+++ b/.github/build_manylinux_binaries.sh
@@ -14,4 +14,4 @@ if ! [ -x "$(command -v cargo)" ]; then
 fi
 
 echo ">>> build the binaries";
-cargo +stable build --verbose --release --manifest-path=io/rust/Cargo.toml
+cargo +stable build --verbose --release --manifest-path=io/rust/Cargo.toml --features=untrusted
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -155,6 +155,9 @@ jobs:
           python -m pip install --upgrade pip
           pip install twine wheel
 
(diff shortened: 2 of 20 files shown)


andrewvyrros (approved):
We're in the home stretch!

PR #223, opened 2021-08-09
Example notebooks

ecowan:


Shoeboxam:
1. The data isn't actually in the repository, so this fails to execute. I think adding the 1k pums data sample you refer to would be fine. 
2. It would be better to use the var_names at the top when specifying column names, and keys. 
3. Your `var_names` have an extra pid column.
4. If trivially achievable, we should opt to use types that make this work on debug builds. 
5. For now, explicitly mention that the input data should have no column names. This column metadata should go into an accompanying file in the data dir. Maybe something like public.json? This will make the "Notice the first element was 0" text outdated, as it won't be imputed as zero.
6. Your col_names vary across cells.
7. With the resize transform you can make the chain complete:
```python

chain = (
    # Convert data into Vec<Vec<String>>
    make_split_dataframe(separator=",", col_names=var_names) >>
    # Selects a column of df, Vec<str>
    make_select_column(key="age", T=str) >>
    # Cast the column as Vec<Int>
    make_cast(TI=str, TO=float) >>
    # Impute missing values to 0
    make_impute_constant(0.) >>
    # Clamp age values
    make_clamp(20., 50.) >>
    make_resize_bounded(lower=20., upper=50., length=100, constant=20.) >>
    make_bounded_mean(lower=20., upper=50., n=100, T=float) >>
    make_base_laplace(scale=1.0)
)

res = chain(data)
print(res)
```

Shoeboxam:
I pushed patches for all but 5, 6, 8, 9.

python/src/opendp/mean.py
```diff
@@ -0,0 +1,67 @@
+from opendp.trans import *
+from opendp.meas import *
+
+
+class DPMean(object):
```
Shoeboxam, line 5: What if instead of making a class, you wrote a function `make_dp_mean` that returns a measurement that obeys the given epsilon? Then you could still chain it with combinators (amplifiers, compositors) down the line. I feel like you'd also be better off if you pulled make_split_dataframe outside, because you probably won't want to parse and split the dataframe for every statistic.

```python
splitter = make_split_dataframe(col_names=col_names)
dp_mean = make_dp_mean(index, constant, lower, upper, n, epsilon)
(splitter >> dp_mean)(data)

# down the line when you have multiple measurements:
dp_sum = make_dp_sum(index, constant, lower, upper, n, epsilon2)
comp = make_sequential_composition_static_distances(1, [(dp_mean, epsilon), (dp_sum, epsilon2)])
mean, sum = (splitter >> comp)(data)
```

I also recommend exposing the input distance in your interface as well.

python/src/opendp/_convert.py
```diff
@@ -167,6 +170,9 @@ def _py_to_slice(value: Any, type_name: str) -> FfiSlicePtr:
     if type_name.startswith("Vec<") and type_name.endswith('>'):
         return _vector_to_slice(value, type_name)
 
+    if type_name.startswith("HashMap<") and type_name.endswith('>'):
```
Shoeboxam, line 173: Did you cherry-pick these commits in? They're in the midst of your commit history. I usually have better luck when I rebase. 
`git rebase --onto [branch you want to sit on top of] [commit hash before where your commits start] [your last commit hash]`

I struggle with git management when my commits are not contiguous.

[See also](https://tworavens.slack.com/archives/C6M0ZSM1N/p1615501753004500)

python/src/opendp/mean.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def __init__(self, data, impute_constant=0.):
+        self.data = data
+        self.impute_constant = impute_constant
+
+    def _binary_search(self, predicate, start, end):
```
Shoeboxam, line 11: See the new `binary_search_chain` utility function in `opendp.mod`.

python/test/integration/test_mean.py
```diff
@@ -0,0 +1,32 @@
+import unittest
```
Shoeboxam, line 1: We're already using pytest in other parts of the library, is there any reason not to use pytest across the board?
You can mark integration tests separately:
https://docs.pytest.org/en/latest/example/markers.html

python/test/integration/test_mean.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def test_dp_mean_compute(self):
+        """
+        A test to ensure that the computation chain for DP Mean completes and returns a value
+        :return:
+        """
+        epsilon = 1.
+        col_names = ["A", "B", "C", "D", "E"]
+        column = "E"
+        dp_mean = DPMean(self.data)
```
Shoeboxam, line 26: This seems backwards to me. Typically computation parameters come first, and the data may or may not be used.

PSI had a class interface similar to what you're writing. The class init would accept computation parameters, and then you could use different methods on the class instance to compute accuracies or make a release.

If you do want to keep the class interface, you could make a base class that implements the shift chainer and __call__, and then inherit from that in DPMean. In DPMean's __init__, conduct the binary search and save the discovered chain as a class attribute. Then it would behave similarly to other measurements and transformations, but you could add an additional method for accuracy estimation.

Shoeboxam: The only gain I see at this point from the class interface is that you can glue additional helper methods onto it like accuracy. Not sure if that's worth adopting classes over just plain simple functions that wrap multiple constructors up in one.

python/test/integration/test_mean.py
ecowan: Thanks, I'll change this, I always defaulted to unittest and when it didn't appear I installed it without thinking about other options.

python/src/opendp/_convert.py
ecowan: I didn't think I did, but I'm finding a lot of strange outcomes (repeated code in the notebook, for example) and I'm not sure what happened

python/src/opendp/mean.py
ecowan: This is a great idea, thanks

ecowan:
>     1. The data isn't actually in the repository, so this fails to execute. I think adding the 1k pums data sample you refer to would be fine.
> 
>     2. It would be better to use the var_names at the top when specifying column names, and keys.
> 
>     3. Your `var_names` have an extra pid column.
> 
>     4. If trivially achievable, we should opt to use types that work on debug builds. Changing column names to var_names achieves this.
> 
>     5. For now, explicitly mention that the input data should have no column names. This column metadata should go into an accompanying file in the data dir. Maybe something like public.json? This will make the "Notice the first element was 0" text outdated, as it won't be imputed as zero.
> 
>     6. Your col_names vary across cells. There is a situation where an analyst only cares about the first k columns, so specifying a list of length k avoids loading up the rest of the columns after column k. Not sure if this is something you want in a basic notebook though, it seems super situational.
> 
>     7. With the resize transform you can make the chain complete:
> 
> 
> ```python
> chain = (
>     # Convert data into Vec<Vec<String>>
>     make_split_dataframe(separator=",", col_names=var_names) >>
>     # Selects a column of df, Vec<str>
>     make_select_column(key="age", T=str) >>
>     # Cast the column as Vec<Int>
>     make_cast(TI=str, TO=float) >>
>     # Impute missing values to 0
>     make_impute_constant(0.) >>
>     # Clamp age values
>     make_clamp(20., 50.) >>
>     make_resize_bounded(lower=20., upper=50., length=100, constant=20.) >>
>     make_bounded_mean(lower=20., upper=50., n=100, T=float) >>
>     make_base_laplace(scale=1.0)
> )
> 
> res = chain(data)
> print(res)
> ```
> 
>     1. You're using the laplace mechanism, but not enabling floating point, which is probably why you have every invocation of the laplace commented.
> 
>     2. I'm seeing a lot of repetition at the end of the basic data analysis file.

Thanks for all the comments and edits!

Regarding (1), I meant to ask if adding data to git was okay for the repo, thanks for going ahead with it. On (8), I thought I had `enable_features("floating-point")` at the top. Mentioned above, I must have done something in git that caused the repetitions at the end.

PR #228, opened 2021-08-10
Drop scalar (un)clamp

### LINKED ISSUE #227: Drop scalar (un)clamp
This doesn't have the utility I thought it did. For example, clamping the sum to an extent where the sensitivity is reduced is extremely inefficient. It also isn't really necessary for EG the snapping or geometric mechanisms, as we just pass clamping bounds directly into those mechanisms. It also complicates the implementation.

Shoeboxam:
Closes #227

### DIFF
--- a/python/src/opendp/trans.py
+++ b/python/src/opendp/trans.py
@@ -201,17 +201,14 @@ def make_cast_metric(
 def make_clamp(
     lower,
     upper,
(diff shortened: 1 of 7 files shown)


andrewvyrros (approved):
Sometimes removing code is good!

PR #229, opened 2021-08-10
TotalOrd trait; AllDomain<T: CheckNull>

### LINKED ISSUE #203: Ord vs PartialOrd
`TotalOrd` is equivalent to rust's `Ord` trait.

There's been an ongoing issue about the use of PartialOrd vs TotalOrd in the proof documents. Float data types don't implement Ord, because of nan. Currently in the implementation, the clamp, IntervalDomain, imputers, and a number of other locations have a relaxed PartialOrd trait bound so that we can work with float data types.

The current implementation tracks nullity via the domain, which gives runtime errors when calling constructors. It would be possible to make these errors trigger at compile-time, and replace PartialOrd with Ord, by adding new float types [(N64 and N32) from the noisy-float crate](https://docs.rs/noisy_float/0.2.0/noisy_float/). These newtypes wrap floats in an interface that structurally prevents nans. The noisy-float crate also re-implements many of the traits and functions you would expect of normal floats, as well as adds the all-important `Ord` trait.

The benefits of moving the checks into the type-system are:
- we can upgrade PartialOrd to Ord in many places. This is more in-tune with the proof documents
- we can leave the definition of `AllDomain<T>` as just "the set of all values T may take on"
- compile-time errors

The downsides of moving checks into the type-system are:
- A new library dependency, noisy-float
- All interfaces that work with non-null floats need floats wrapped in the noisy-float newtype

It hasn't been clear to me which option is better, so I've been procrastinating making this change.

Shoeboxam:
Closes #203
Adds the trait requirement CheckNull to T in AllDomain. This formalizes that AllDomain does not carry null values.

### DIFF
--- a/python/test/test_trans.py
+++ b/python/test/test_trans.py
@@ -154,12 +154,11 @@ def test_split_dataframe():
     assert query.check(1, 1)
 
 
-def test_vector_clamp():
+def test_clamp():
     from opendp.trans import make_clamp
     query = make_clamp(lower=-1, upper=1)
     assert query([-10, 0, 10]) == [-1, 0, 1]
     assert query.check(1, 1)
-
 
 
 def test_bounded_mean():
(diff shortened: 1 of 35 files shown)


andrewvyrros (approved):
Thanks for separating this from the other recent changes!

PR #231, opened 2021-08-12
atomic domain in resize transform

### LINKED ISSUE #230: Resize generic atomic domain
make_resize needs to be capable of resizing any VectorDomain<DA>, because it is typically used in a pipeline where DA could be any of many different domains.

Shoeboxam:
Closes #230 

I made the resize transform generic in the library core, but can only provide an interface for a single domain at a time from FFI. A next step would be adding FFI functions for constructing domains themselves. A user could then pass these domains into the resize constructor as an AnyDomain. See https://github.com/opendp/opendp/issues/232

### DIFF
--- a/rust/opendp/src/trans/resize/mod.rs
+++ b/rust/opendp/src/trans/resize/mod.rs
@@ -1,20 +1,23 @@
-use crate::core::{Transformation, Function, StabilityRelation};
+use crate::core::{Transformation, Function, StabilityRelation, Domain};
 use crate::error::Fallible;
 use crate::dist::{SymmetricDistance, IntDistance};
-use crate::dom::{VectorDomain, AllDomain, SizedDomain};
+use crate::dom::{VectorDomain, SizedDomain};
 use std::cmp::Ordering;
 use crate::traits::CheckNull;
 
-pub fn make_resize_constant<TA: 'static + Clone + CheckNull>(
-    constant: TA, length: usize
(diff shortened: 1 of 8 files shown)


rust/opendp/src/trans/resize/mod.rs
```diff
@@ -1,20 +1,23 @@
...
-pub fn make_resize_constant<TA: 'static + Clone + CheckNull>(
-    constant: TA, length: usize
-) -> Fallible<Transformation<VectorDomain<AllDomain<TA>>, SizedDomain<VectorDomain<AllDomain<TA>>>, SymmetricDistance, SymmetricDistance>> {
+pub fn make_resize_constant<DA>(
+    atom_domain: DA,
+    constant: DA::Carrier, length: usize
+) -> Fallible<Transformation<VectorDomain<DA>, SizedDomain<VectorDomain<DA>>, SymmetricDistance, SymmetricDistance>>
+    where DA: 'static + Clone + Domain,
+          DA::Carrier: 'static + Clone + CheckNull {
+    if !atom_domain.member(&constant)? { return fallible!(MakeTransformation, "constant must be a member of DA")}
```
andrewvyrros, line 14: Calling `member()` here feels a little funky. I guess this is a bit of a gap in the framework, and the way we've modeled Domain types. I can't think of a better way to handle this, but I think it's worth a discussion with Salil.

andrewvyrros (approved):
This seems like a reasonable compromise, at least for now. I'm still not 100% sold on creating Domains in app code (#232); it's nice that we've been able to avoid that so far. But maybe it's time to bite the bullet.

Also I'm a little uneasy about the Domain member check on the imputed value. I think that's worth a discussion with Salil.

PR #233, opened 2021-08-12
population amplification

### LINKED ISSUE #159: Amplification -- Implementation
An amplifier for privacy budgets based on simple sampling.

Shoeboxam:
Closes #159.
Joint work with Monika Hu. Thanks! The scope has creeped here to include some architectural changes.

This adds a combinator constructor that accepts a measurement where DI must implement `IsSizedDomain` (either `SizedDomain<VectorDomain<DIA>>` or `AnyDomain`), and a population parameter. The constructor returns a measurement with an adjusted privacy relation.

#247 was opened for this issue. The current implementation does downcasting of AnyDomains in `impl IsSizedDomain for AnyDomain`, which simply retrieves the length. I made a comment below about the different approaches here, and why I chose the approach I did.

In the process, I also:
- implemented ToString for Type, for better errors
- cleaned up the impls on Type
- removed the AnyMeasureDistance and AnyMetricDistance types

The updated AnyObject doesn't have an impl for FallibleSub, but we also don't need it right now. When/if we do need it, we could use approach 2 below. The removal of AnyMeasureDistance and AnyMetricDistance was a side-effect while exploring FFI designs, but I think it is needed for #242.


Shoeboxam:
This PR is blocked because it isn't yet clear how we want to build glue in the FFI layer. It would be ideal if we didn't have multiple different type-erased objects, as it is unclear which type-erased object to construct in bindings languages. We can merge all type-erased objects into the AnyObject by implementing all traits on AnyObject, and failing if possible, otherwise panicking. It is the responsibility of the FFI layer to ensure the proper glue is present in the AnyObject before passing it into the core library.

We also don't want to require all necessary traits to build all glue functions at the time we do type-erasure, because new combinators introduce additional required trait bounds in all other constructors.

I came up with a strategy for adding trait casts to a top-level trait, that allows us to fallibly acquire new traits. This works well for some traits, but still leaks trait bounds on associated types. This is in the branch 159-amplification-trait-cast.

We can still acquire trait bounds that aren't defined on generics via dispatch. There are three locations where this dispatch can be done.
1. At the time of type-erasure. A demo of this is in 159-amplification-eager-glue.
2. Lazily when calling a constructor that needs it. An older commit in this branch does this.
3. On-demand, when you make use of the trait object. No glue is necessary here. This branch currently uses this approach. It only applies in situations where the glue is only needed once, to avoid multiple dispatch lookups.

andrewvyrros (approved):

PR #234, opened 2021-08-12
capture Debug trait in AnyBoxBase

### LINKED ISSUE #201: Capture Debug trait in AnyDomain
Feature suggestion: if Debug were also captured in AnyDomain, then end-users could visually compare to see why domains are mismatched.

Shoeboxam:
Closes #201.
This adds Debug trait bounds... everywhere, which can be pretty annoying.
But now when you attempt to chain incompatible measurements or operations, you can see how the domains or metrics differ. This doesn't capture type information in the output, unfortunately. I've added more comments about this limitation in the linked issue.

`git rebase --onto main 6609ec5 201-anybox-debug`

### DIFF
--- a/python/example/test.py
+++ b/python/example/test.py
@@ -24,7 +24,7 @@ def main():
     # Noisy sum, col 1
     noisy_sum_1 = (
         make_select_column(key="B", T=int) >>
-        # make_clamp(lower=0, upper=10) >>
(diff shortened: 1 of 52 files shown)


Shoeboxam:
Closing, as I have cherry-picked this on top of the library renaming PR instead. See #263

PR #242, opened 2021-08-17
sequential compositor w/ static distances

### LINKED ISSUE #7: Basic Composition Combinator -- Implementation

Shoeboxam:
Closes #7
This PR replaces the basic composition PR based on a suggestion from Salil. The internals are simpler, because the shared behavior captured in the AnyMeasure is just a function to sum distances.

- Removed the EpsilonDelta struct, Midpoint/Tolerance traits, and binary search.
- Replaced `basic_composition` with `sequential_composition_with_static_distances`

### DIFF
--- a/python/src/opendp/_convert.py
+++ b/python/src/opendp/_convert.py
@@ -17,6 +17,8 @@ ATOM_MAP = {
     'i32': ctypes.c_int32,
     'i64': ctypes.c_int64,
     'bool': ctypes.c_bool,
+    'AnyMeasurementPtr': Measurement,
+    'const AnyTransformation *': Transformation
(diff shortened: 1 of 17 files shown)


Shoeboxam:
This is soft-blocked on the architectural changes in #233 amplification. The bindings layer doesn't know to wrap distances in AnyMeasureDistance structs instead of just AnyObjects. If we removed AnyMeasureDistances, then we'd no longer have the ambiguity.

Shoeboxam:
Closing in favor of the approach in #424.

PR #243, opened 2021-08-18
Binary search utilities

### LINKED ISSUE #237: Utilities to simplify app development (especially for DP Creator)
When writing OpenDP apps, it's a little awkward to specify the noise scale of a Measurement. A more typical mindset for an app developer is to provide a desired privacy loss. We need some utilities to bridge this gap.

The canonical use case is constructing the set of Measurements offered in DP Creator.

A simple solution would be single-use functions that map from privacy loss -> sigma for a given mechanism, but this is a bit limiting. More flexible would be a binary search approach.

Since these things would be generally useful to the library, the best solution would be to add a Rust module with Python bindings. If that turns out to be too complicated, a first cut could be simple Python-only helper functions.

Shoeboxam:
Closes #237 
Adds a general binary search algorithm to python to solve for free parameters in computation chains.
Adds distance maps to the resize transformation. The library was finding a very strange loosening.
Allows _free to fail when the interpreter is shutting down at the end of the program.

### DIFF
--- a/python/src/opendp/mod.py
+++ b/python/src/opendp/mod.py
@@ -307,7 +307,6 @@ def binary_search_chain(
     >>> )
     >>> # Find a value in `bounds` that produces a (`d_in`, `d_out`)-chain within `tolerance` of the decision boundary.
     >>> # `make_chain` returns the complete computation chain when given a single numeric parameter.
-    >>> # The function `binary_search_chain` returns the parameterized computation chain.
     >>> chain = binary_search_chain(
     >>>     make_chain=lambda s: pre >> make_base_laplace(scale=s),
     >>>     bounds=(0., 10.), d_in=1, d_out=1.)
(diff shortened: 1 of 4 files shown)


andrewvyrros (approved):

Shoeboxam:
I was wrong about the max steps... It's up to log2(MAX_FINITE_FLOAT / MIN_POSITIVE_FLOAT) steps to uniquely distinguish the minimum float starting at the maximum float. The performance is so poor because floats are distributed evenly on a log scale, so halving the space each time eliminates much less than half of all possible floats. At some point if we need more performance it would make sense to implement a binary search over the bits, or a two-phase binary search over the exponent and then mantissa. Python isn't really suited for that.

There are also problems with catastrophic cancellation in the error check that could be resolved with an additional relative tolerance term.

PR #246, opened 2021-08-23
Reworked User Guide outline

andrewvyrros:
* Simplified getting started section
* Split concepts section
* Added details to application structure


docs/source/user/core-structures.rst
```diff
@@ -0,0 +1,28 @@
+Core Structures
+===============
+
+.. contents:: |toctitle|
+    :local:
+
+Overview
+--------
+
+OpenDP is about creating computations with specific privacy characteristics. These computations are modeled with two core structures in OpenDP: Transformations and Measurements. These structures are in all OpenDP programs, regardless of the underlying algorithm or definition of privacy. By modeling computations in this abstract way, we're able to combine them in flexible arrangements, and reason about the resulting programs.
```
pdurbin, line 10:
```suggestion
OpenDP is focused on creating computations with specific privacy characteristics. These computations are modeled with two core structures in OpenDP: Transformations and Measurements. These structures are in all OpenDP programs, regardless of the underlying algorithm or definition of privacy. By modeling computations in this abstract way, we're able to combine them in flexible arrangements and reason about the resulting programs.
```

docs/source/user/core-structures.rst
```diff
@@ -0,0 +1,28 @@
...
+
+A Transformation is a (deterministic) mapping from datasets to datasets.
+
+Constructors and Functions
+--------------------------
+
+In OpenDP, Measurements and Transformations are created by calling constructor functions. Because Measurements and Transformations are themselves like functions (they can be invoked on an input and return an output), you can think of constructors as higher-order functions: You call them to produce another function, that you will then feed some data.
+
+Examples
+--------
```
pdurbin, line 28: Yes, examples here would be good. It could also be helpful to have an example or a measurement and a transformation above, where they are defined.

The measurement definition itself uses terms that may not be familiar. What kind of outputs? What kind of types?

docs/source/user/programming-framework.rst
```diff
@@ -0,0 +1,7 @@
+Programming Framework
+=====================
+
+.. contents:: |toctitle|
+    :local:
+
+OpenDP is based on a model described in the paper, `A Programming Framework for OpenDP <https://projects.iq.harvard.edu/files/opendp/files/opendp_programming_framework_11may2020_1_01.pdf>`_. This framework lays out a rigorous approach to modeling privacy-preserving computations. OpenDP (the software library) is a faithful implementation of the OpenDP Programming Framework. This allows you to create applications with well defined privacy characteristics.
```
pdurbin, line 7:
```suggestion
OpenDP is based on a model described in the paper `A Programming Framework for OpenDP <https://projects.iq.harvard.edu/files/opendp/files/opendp_programming_framework_11may2020_1_01.pdf>`_. This framework lays out a rigorous approach to modeling privacy-preserving computations. OpenDP (the software library) is a faithful implementation of the OpenDP Programming Framework. This allows you to create applications with well defined privacy characteristics.
```

docs/source/user/getting-started.rst
```diff
@@ -0,0 +1,33 @@
...
+.. code-block:: bash
+
+    % pip install opendp
+
+This will make the OpenDP modules available to your local environment.
+
+Hello, OpenDP!
+--------------
+
+Once you've installed OpenDP, you can write your first program. In the example below, we'll construct an identity ``Transformation``, then invoke it on a string.
```
pdurbin, line 26: Perhaps this really is as simple as it gets but I can imagine people wondering "What is SymmetricDistance and why it it important in hello world?" They might also wonder what the "identity" transformation does in the first place. I suppose for this we could explain what it does here in the text or link to the API docs.

pdurbin (commented):
Overall, I like the direction this pull request is taking.

Some of the new pages (like many of the old pages) have just a placeholder outline. If we have the capacity to put some more content within the outline, it would be great.

I left a few comment inline.

I will say that it's fantastic to be able to pip install OpenDP. Much easier to get up and running!

PR #262, opened 2021-09-02
library API renaming

### LINKED ISSUE #181: Naming (2)
Follow-up on https://github.com/opendp/opendp/issues/141
Our public API would be more accessible if we unified the terms we use. We should rename now, before community meetings and DPCreator integration.

- rename IntervalDomain to BoundedDomain
    * This removes "Interval" terminology from the library
    * `make_bounded_*` constructors expect `BoundedDomain` inputs
- rename `make_bounded_sum_n` to `make_sized_bounded_sum`
    * This corresponds with the input domain `SizedDomain<VectorDomain<BoundedDomain<T>>>`
- rename `make_bounded_mean` to `make_sized_bounded_mean`
    * Should help to avoid confusion like this: https://github.com/opendp/smartnoise-core/issues/355
    * same for covariance and variance
- rename SizedDomain's length attribute to `size`
- rename `length` and `n` arguments on constructors to `size`
- reorder `size` argument to always be before bounds argument(s) for symmetry with constructor names
- change `lower` and `upper` to `bounds`
    * For symmetry with constructor name: `make_sized_bounded_mean(size=100, bounds=(0., 10.))`
    * For consistency with make_base_geometric, which also takes `bounds`
    * Bounds are usually passed together, this is potentially less verbose for users. See [example](https://github.com/opendp/opendp/blob/main/rust/opendp/src/interactive.rs#L266-L267)
    * Downside: adds more AnyObjects to the API
- rename `make_resize_bounded` to `make_bounded_resize`
- be consistent with "Atomic" naming in generics

Shoeboxam:
Closes #181

See #181 for a summary of changes.

Also removed make_parse_column.


rust/opendp/src/trans/dataframe/mod.rs
```diff
@@ -81,49 +80,17 @@ pub fn make_split_dataframe<K>(
...
-    where T: 'static + Debug + Clone + PartialEq + FromStr + Default,
-          K: Eq + Hash + Clone + Debug,
-          T::Err: Debug {
-    let col: &Vec<String> = df.get(key)
-        .ok_or_else(|| err!(FailedFunction, "column does not exist: {:?}", key))?
-        .as_form()?;
-    let col = vec_string_to_str(col);
-    let col = parse_series::<T>(&col, impute)?;
-    replace_col(key, &df, col.into())
-}
```
andrewvyrros, line 101: Awww... I was so proud of myself when I figured out how to implement this. 😂

andrewvyrros (approved):
These new names are an improvement.

rust/opendp/src/trans/dataframe/mod.rs
Shoeboxam: I'm sorry! But the lessons learnt are not forgotten. We'll be using this more in the future.

PR #264, opened 2021-09-03
177 f dp alternative

ngrislain:
# Sarus contribution on f-DP

In the framework of the fellowship program, Sarus has been working on proposing ways to compose Measurements and automate privacy relation composition using f-DP inspired representations of privacy loss (see https://github.com/opendp/opendp/tree/177-f-dp).

This contribution is an alternative attempt at this goal.
- It is closer to the Privacy Loss Distribution (PLD) formalism from (http://proceedings.mlr.press/v108/koskela20b/koskela20b.pdf)
- It does not try to evaluate epsilon,delta pairs using bisection
- It maintains along the Output Measure a function returning a finite valued Privacy Loss Distribution bounding the true privacy loss from below given an input metric distance (similar to 177-f-dp Hints)
- The privacy relation holds a counted reference to the same function
- Composition is evaluated by summing the PLD (hence doing a convolution)
- A finite valued Privacy Loss Distribution can then be simplified by reducing the number of possible values while bounding the true privacy loss from below
- All computations are exact: only `rug::Rational` are used in `opendp/src/sarus/privacy_loss_distribution.rs`
- Rationals are simplified from below during simplification phases.
- Demos with graphs are written. The Graphs are generated with `vega_lite_4`

Both approaches should converge, eventually.


rust/opendp/src/sarus/privacy_loss_distribution.rs
```diff
@@ -0,0 +1,212 @@
...
+                }
+            }
+        }
+        result
+    }
+}
+
+impl Default for PLDistribution {
+    fn default() -> Self {
+        PLDistribution::new([(Rational::from(1),Rational::from(1))])
```
Shoeboxam, line 200: Minor, but this PR won't compile unless I tweak this into an iterator:
```suggestion
        PLDistribution::new(vec![(Rational::from(1), Rational::from(1))].into_iter())
```

PR #275, opened 2021-09-09
accuracy utilities

### LINKED ISSUE #274: Accuracy Utilities

Shoeboxam:
Closes #274 
Adds utilities for converting between accuracy estimates and noise scales for laplacian and gaussian noise.

The laplacian noise estimate can also be used for the geometric distribution.
We might want to think about renaming `make_base_geometric` to `make_base_discrete_laplace` at some point to make this more obvious.
One could make the argument that you should take the ceil of the geometric accuracy, because the support is discrete ...but on the other hand, the probabilities are still approximately continuous, so I don't think this is necessary. Due to this, I have dropped the ceil when porting from Smartnoise-Core.

### DIFF
--- a/rust/opendp-ffi/src/accuracy/bootstrap.json
+++ b/rust/opendp-ffi/src/accuracy/bootstrap.json
@@ -12,7 +12,7 @@
                 "name": "alpha",
                 "c_type": "void *",
                 "rust_type": "T",
-                "description": "Level-alpha, or (1. - alpha)100% confidence. Must be within (0, 1]."
+                "description": "Statistical significance, level-alpha, or (1. - alpha)100% confidence. Must be within (0, 1]."
             },
             {
                 "name": "T",
--- a/rust/opendp-ffi/src/accuracy/bootstrap.json
+++ b/rust/opendp-ffi/src/accuracy/bootstrap.json
@@ -1,0 +1,110 @@
+{
+    "laplacian_scale_to_accuracy": {
+        "description": "Convert a laplacian scale into an accuracy estimate.",
+        "args": [
+            {
+                "name": "scale",
+                "c_type": "void *",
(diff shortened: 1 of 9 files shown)


rust/opendp-ffi/src/accuracy/bootstrap.json
```diff
@@ -0,0 +1,110 @@
...
+                "name": "scale",
+                "c_type": "void *",
+                "rust_type": "T",
+                "description": "Laplacian noise scale."
+            },
+            {
+                "name": "alpha",
+                "c_type": "void *",
+                "rust_type": "T",
+                "description": "Level-alpha, or (1. - alpha)100% confidence. Must be within (0, 1]."
```
andrewvyrros, line 15:
```suggestion
                "description": "Statistical significance, level-alpha, or (1. - alpha)100% confidence. Must be within (0, 1]."
```

rust/opendp-ffi/src/accuracy/bootstrap.json
```diff
@@ -0,0 +1,110 @@
...
+                "name": "scale",
+                "c_type": "void *",
+                "rust_type": "T",
+                "description": "Gaussian noise scale."
+            },
+            {
+                "name": "alpha",
+                "c_type": "void *",
+                "rust_type": "T",
+                "description": "Level-alpha, or (1. - alpha)100% confidence. Must be within (0, 1]."
```
andrewvyrros, line 69:
```suggestion
                "description": "Statistical significance, level-alpha, or (1. - alpha)100% confidence. Must be within (0, 1]."
```

andrewvyrros (changes requested):
Thanks for the additional tests!

andrewvyrros (approved):

Issue #270, opened 2021-09-07
Docs Update -- User Guide Sections by AV


andrewvyrros:
Tracking work on the User Guide done by @andrewvyrros

PR #276, opened 2021-09-09
268 quickstart

### LINKED ISSUE #268: Docs Update -- Quickstart
Fill in the content of docs/source/quickstart.

I think we can largely copy the content from user/getting-started (then add more detail to the latter).

pdurbin:
closes #268

### DIFF
--- a/docs/source/quickstart/index.rst
+++ b/docs/source/quickstart/index.rst
@@ -1,6 +1,30 @@
 Quickstart
 ==========
 
-(In the future, this section will describe how to use the `OpenDP Library`_.)
+.. contents:: |toctitle|
+    :local:
 
-.. _OpenDP Library: https://github.com/opendp/opendp
+Installation
+------------
+
+The easiest way to get started with OpenDP is from Python. Use ``pip`` to install the `opendp <https://pypi.org/project/opendp/>`_ package from PyPI.
+
+.. code-block:: bash
+
+    % pip install opendp
+
+This will make the OpenDP modules available to your local environment.
+
+Hello, OpenDP!
+--------------
+
+Once you've installed OpenDP, you can write your first program. In the example below, we'll construct an identity ``Transformation``, then invoke it on a string.
+
+.. doctest::
+
+    >>> from opendp.trans import make_identity
+    >>> from opendp.typing import SubstituteDistance
+
+    >>> identity = make_identity(M=SubstituteDistance, TA=str)
+    >>> identity("Hello, world!")
+    'Hello, world!'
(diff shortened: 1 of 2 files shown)


docs/source/quickstart/index.rst
```diff
@@ -1,6 +1,30 @@
...
 
-.. _OpenDP Library: https://github.com/opendp/opendp
+Installation
+------------
+
+The easiest way to get started with OpenDP is from Python. Use ``pip`` to install the `opendp <https://pypi.org/project/opendp/>`_ package from PyPI.
+
+.. code-block:: bash
+
+    % pip install opendp
```
Shoeboxam, line 14: This percent reminds me of the `%pip` magic command from notebooks. But the percent is not necessary in a terminal. Why is the percent here?

docs/source/quickstart/index.rst
```diff
@@ -1,6 +1,30 @@
...
+--------------
+
+Once you've installed OpenDP, you can write your first program. In the example below, we'll construct an identity ``Transformation``, then invoke it on a string.
+
+.. doctest::
+
+    >>> from opendp.trans import make_identity
+    >>> from opendp.typing import SubstituteDistance
+
+    >>> identity = make_identity(M=SubstituteDistance, TA=str)
```
Shoeboxam, line 28: The identity transformation seems like a simple transform, but the nature of the function is so unconstrained that it has many free variables. I'd also say we generally work with SymmetricDistance, so using it in the quickstart may be confusing. The identity function also doesn't have much of a use. Perhaps we could switch to clamp?

docs/source/quickstart/index.rst
andrewvyrros: That's an Andy-ism. Now that I know sphinx a little better, I think the right thing to do is this:

```
.. prompt:: bash

     pip install opendp
```

That requires the sphinx-prompt plugin, which I'm adding in #270. As part of that, I can come back and tweak this.

andrewvyrros (approved):
This may need a little TLC after the User Guide intro is updated in #270, but I'd like to get something in ASAP instead of the current placeholder, so merging now.

docs/source/quickstart/index.rst
pdurbin: I had the same thought but let it go (it was copied from the Getting Started page). I typically see `$` instead of `%` for my shell prompt. That said, I like the idea of simply removing it.

Even without the sphinx-prompt plugin (I haven't heard of it), we can still do `.. code-block:: bash`. There are examples on the Logistics page in the Dev Guide.

PR #278, opened 2021-09-09
add Resources section #272

### LINKED ISSUE #272: Docs Update -- Resources
Create a top-level section Resources (contained in `More...` nav-bar item). Collection of various resources for:
* background
* learning about DP
* papers
* whatever

pdurbin:
closes #272

I took a quick swing at at least putting links to papers, etc. onto a Resources page. The formatting is extremely minimal and looks like this right now:

Screen Shot 2021-09-09 at 4 12 17 PM

I'm not sure how fancy we should get. At minimum, we should probably link the titles to the papers like they do at https://differentialprivacy.org/resources/ (they also link the author names, which would be more work):

Screen Shot 2021-09-09 at 4 09 44 PM

We could also try to use full on citations and a Sphinx extension where we keep all these in Bibtex format. Probably overkill, I'd say.

Under the "Papers" section I left comments from @Shoeboxam in parentheses. We can take these out or otherwise rework them.

### DIFF
--- a/docs/source/resources/index.rst
+++ b/docs/source/resources/index.rst
@@ -1,0 +1,92 @@
+Resources
+=========
+
+.. contents:: |toctitle|
+    :local:
+
+Background
+----------
+
+Differential privacy: A primer for a non-technical audience
+
+https://salil.seas.harvard.edu/publications/differential-privacy-primer-non-technical-audience
+
+The Algorithmic Foundations of Differential Privacy
+
+https://www.cis.upenn.edu/~aaroth/Papers/privacybook.pdf
+
+Learning About Differential Privacy
+-----------------------------------
+
+Courses & Educational Materials from the Harvard University Privacy Tools Project
+
+https://privacytools.seas.harvard.edu/courses-educational-materials
+
+Designing Access with Differential Privacy
+
+https://admindatahandbook.mit.edu/book/latest/diffpriv.html
+
+Programming Differential Privacy
+
+https://programming-dp.com/notebooks/cover.html
+
+Privacy and Confidentiality Protection Overview
+
+https://www2.census.gov/cac/nac/meetings/2019-05/garfinkel-privacy-confidentiality-protection.pdf
+
+Papers
+------
+
+The Complexity of Differential Privacy
+
+https://privacytools.seas.harvard.edu/files/privacytools/files/complexityprivacy_1.pdf
+
+Calibrating Noise to Sensitivity in Private Data Analysis
+
+(Laplace mechanism)
+
+https://people.csail.mit.edu/asmith/PS/sensitivity-tcc-final.pdf
+
+On Significance of the Least Significant Bits For Differential Privacy
+
+(Snapping mechanism and how floating-point numbers and the laplace mechanism leak)
+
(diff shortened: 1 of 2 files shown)


docs/source/resources/index.rst
```diff
@@ -0,0 +1,92 @@
+Resources
+=========
+
+.. contents:: |toctitle|
+    :local:
```
Shoeboxam, line 5: The generated website has a :local: index in the right panel. Do we need it in the page contents as well?

docs/source/resources/index.rst
```diff
@@ -0,0 +1,92 @@
...
+
+Background
+----------
+
+Differential privacy: A primer for a non-technical audience
+
+https://salil.seas.harvard.edu/publications/differential-privacy-primer-non-technical-audience
```
Shoeboxam, line 12: I personally think this would work a lot better if the text were links.

docs/source/resources/index.rst
```diff
@@ -0,0 +1,92 @@
...
+(Gradient descent)
+
+https://arxiv.org/pdf/1607.00133.pdf
+
+Rényi Differential Privacy
+
+https://arxiv.org/pdf/1702.07476.pdf
+
+Concentrated Differential Privacy: Simplifications, Extensions, and Lower Bounds
+
```
Shoeboxam, line 85: Maybe we should hold off on these resources until we support them?

docs/source/resources/index.rst
pdurbin: Hmm. I see what you mean. I'm mixed about it. I'm very used to seeing the contents right there at the top of the page. For Dataverse we made a decision to always have it even on pages that don't have many subheadings (for consistency). I started adding these before we switched to the current theme, which is where the sidebar comes from. I could probably train myself to look at the sidebar.

This affects every page in the docs so a dedicated issue is probably best if we'd like to change it.

docs/source/resources/index.rst
pdurbin: Yeah, me too. Maybe just simple links for now. Just link the title, I mean.

docs/source/resources/index.rst
pdurbin: Makes sense. I'm not sure what we support.

docs/source/resources/index.rst
andrewvyrros: Yeah let's figure this out separately, I've opened #283.

docs/source/resources/index.rst
andrewvyrros: Yeah making the titles into links would be good.

docs/source/resources/index.rst
andrewvyrros: I think dropping the entries for Renyi and Concentrated DP should be enough.

andrewvyrros (changes requested):
This is a great start. I think it's worth changing the URLs to links, and removing a couple of papers.

pdurbin:
This is ready for more review. I just did the following:

- Remove Rényi and Concentrated
- Make titles into links
- Move the notes in parens to the side

Here's how it looks:

Screen Shot 2021-09-10 at 1 53 43 PM

andrewvyrros (approved):
This is a good starting point, hopefully to grow in the future.

PR #290, opened 2021-09-15
integration testing

### LINKED ISSUE #289: Integration testing
I wrote integration tests for each of the aggregators and noticed that the types in the FFI layer for `count_by_categories` and `count_by` weren't quite right.

`count_by_categories` is not currently working, as the output sensitivity should be a int, but is a float. Since the Q and T types differ, the aggregator cannot chain with mechanisms.

`count_by` depends on `make_resize`, which only works with numbers. Adjusted `make_resize` to dispatch to primitives instead. This adds support for stability histograms on strings and bools in bindings languages.

Shoeboxam:
Closes #289

### DIFF
--- a/python/src/opendp/accuracy.py
+++ b/python/src/opendp/accuracy.py
@@ -13,7 +13,7 @@ def laplacian_scale_to_accuracy(
     """Convert a laplacian scale into an accuracy estimate.
     
     :param scale: Laplacian noise scale.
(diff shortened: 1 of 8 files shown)


andrewvyrros (dismissed):
Integration tests FTW

andrewvyrros (approved):

Issue #285, opened 2021-09-14
Use Binder for interactive notebooks


pdurbin:
We've talked a bit about Binder in #199 and #261 and I think it should be relatively straightforward to get Binder links working from the notebook examples at https://docs.opendp.org/en/stable/examples/notebooks.html

If you go to https://mybinder.org you can enter the OpenDP GitHub repo URL and a path to a notebook to get a URL like this: https://mybinder.org/v2/gh/opendp/opendp/HEAD?filepath=python%2Fexample%2Fbasic_data_analysis.ipynb

![Screen Shot 2021-09-14 at 11 51 34 AM](https://user-images.githubusercontent.com/21006/133291379-9709d6d5-0e8a-4d17-a143-e409f1c50102.png)

However, you'll get a "no module named opendp" error when you try to run cells:

![133268424-eef2fb02-0499-41df-97c7-d3a7b8431cf4](https://user-images.githubusercontent.com/21006/133289932-fa6ee866-4ee7-40a7-9579-66b007593822.png)

A workaround is to open the terminal and type `pip install opendp` like this.

![Screen Shot 2021-09-14 at 11 33 20 AM](https://user-images.githubusercontent.com/21006/133289759-5a4ec7e8-e710-43c0-be37-24d5aacd6386.png)
![Screen Shot 2021-09-14 at 11 33 48 AM](https://user-images.githubusercontent.com/21006/133289757-50765bfe-250c-4e8e-86d4-dadd5304ae80.png)

The right way to fix this is to give Binder enough information about our dependencies. Really all we need to do is put "opendp" in a `requirements.txt` file. According to https://repo2docker.readthedocs.io/en/latest/usage.html#where-to-put-configuration-files `requirements.txt` can be in the root of the repo, a directory called `binder`, or a directory called `.binder`.

![Screen Shot 2021-09-14 at 11 47 03 AM](https://user-images.githubusercontent.com/21006/133290547-86ca7cf3-1f0e-48c8-a2a3-c517852a0d58.png)

At standup this morning, we discussed how we'd like to keep clutter out of the root of the repo. My take is that a `.binder` directory is the best way to achieve that. At least it'll be hidden from the command line, though it'll still be visible on GitHub. Also, as @andrewvyrros pointed out, we should link to the stable branch when it comes to linking from the notebooks page under examples in the docs.

pdurbin:
[...]

PR #294, opened 2021-09-16
`pip install opendp` for Binder support #285

pdurbin:
If we consider #285 to be "link to Binder" from the docs, this pull request doesn't fully implement this.

Rather, it allows notebooks in the OpenDP GitHub repo to work with Binder.

So as a first step we could merge this and use "main" or "latest" to play around with Binder. Once we cut a release, we could also use "stable". I'd suggest waiting until this is in "stable" (post release) before we follow up with another pull request linking the docs to Binder.


pdurbin:
Oh, as I mentioned in the issue, you can test this with https://mybinder.org/v2/gh/opendp/opendp/285-binder-notebooks?filepath=python%2Fexample%2Fbasic_data_analysis.ipynb

Or you can fill in the form yourself like this:

133307158-742c1072-e5db-4e6c-9053-8b5e2aaefb5b

andrewvyrros (approved):
Looks cool, let's give it a try!

PR #295, opened 2021-09-16
clean up notebook example link, add Binder link #285

pdurbin:
Instead of a raw link to GitHub, I added some surrounding text so it looks nicer.

More importantly, I also linked to Binder so people can try the notebooks from their browser.

For now, the Binder link doesn't have a branch specified so it'll default to "main" (the default) but as discussed in #285 once we cut a release, we can specify "stable" as the branch.

### DIFF
--- a/docs/source/examples/notebooks.rst
+++ b/docs/source/examples/notebooks.rst
@@ -4,4 +4,10 @@ Notebooks
 .. contents:: |toctitle|
 	:local:
 
-For now, see https://github.com/opendp/opendp/tree/main/python/example
+Example notebooks can be found in `python/example`_ in the OpenDP library source tree.
+
+.. _python/example: https://github.com/opendp/opendp/tree/main/python/example
+
+You can open and run these notebooks online in Binder_ (navigate to ``python/example``).
+
+.. _Binder: https://mybinder.org/v2/gh/opendp/opendp/HEAD


Shoeboxam (approved):
Looks much better! I'm going to make one more pass to make sure we don't have empty pages.

RoyRin:
(I know this was merged, so feel free to ignore)
but right now you have: (https://github.com/opendp/opendp/blob/main/docs/source/examples.rst)
```
docs/source/examples.rst
26:You can open and run these notebooks online in Binder_ (navigate to ``python/example``).
28:.. _Binder: https://mybinder.org/v2/gh/opendp/opendp/HEAD
```

 but just so you know you can add an easy tag to the readme.md 
[Binder](https://mybinder.org/v2/gh/opendp/opendp/HEAD)

like here : https://github.com/RoyRin/arxiv_connections

image

pdurbin:
@RoyRin I'm aware of the "launch binder" button and I use it over at https://github.com/pdurbin/dataverse-irc-metrics

No objection from me about adding a "launch binder" button to the README but as I suggest at https://github.com/RoyRin/arxiv_connections/issues/1 I think it either needs to link to a specific notebook or have some explanatory text to indicate where the user should go hunting for them from the root directory (for OpenDP it's `python/example`) once the user is dumped into Jupyter Lab. I've tried and failed to link to a directory of notebooks (you just get sent to the root). Linking to a single notebook works fine.

If you want to create a pull request, please go ahead. I'll leave it to @andrewvyrros and @Shoeboxam to decide if they want it or not. Thanks for the suggestion!

Oh, while I'm thinking of it, I know we closed #285 already but originally we talked about linking to the "stable" branch, which we haven't done yet.

Shoeboxam:
I think we need a README.md in the examples folder anyways, like this:
https://github.com/opendp/smartnoise-samples/tree/master/analysis

Perhaps we could have a note for binder users in there, and link binder to open that README.md?

RoyRin:
Yep, I believe you are right (@pdurbin), I think Binder takes a path to an individual file or to the whole repo, I don't think there's a way to point it only to a subdirectory in a repo (which is unforunate).

@Shoeboxam  I agree with putting a binder link in python/example/README.md makes sense too.
I can make a tiny PR that puts a binder link in python/example/README.md with the words "navigate to /python/examples" next to it.

Shoeboxam:
@RoyRin Just ran a little test and fiddled with urls. Surprised to find that direct links only work for notebook files-- I'm not getting any luck linking directly to a README.md, for example. So it seems like the README.md can't really be used as a landing page for loading binder.

PR #315, opened 2021-09-21
conservative rounding directions in relations

### LINKED ISSUE #150: Checked arithmetic in privacy/stability relations
The relations should use checked_mul, checked_add, etc. operations that fail the relation if the distance is out-of-bounds of the distance type. In rust, compiling in release mode will wrap arithmetic operations that overflow/underflow. A wrapped distance overflow could immensely underestimate budgets.

The relations should also use arithmetic that always rounds towards the direction that make the relations less favorable to passing.

### LINKED ISSUE #314: Underestimated sized bounded sum integer sensitivity
`d_in * (U - L) / 2` truncates down, meaning the sized bounded sum relation underestimates sensitivity by 0.5 when `d_in * (U - L)` is odd and `T` is integral.

Issue raised by Silvia and Connor, thank you!

### LINKED ISSUE #343: `make_base_geometric` float rounding
The geometric mechanism is *mostly* free from floating-point issues. However, the noise scale does get transformed and has the potential for a slight rounding down. These scale transformations should round conservatively. It might be reasonable to ensure this behavior without MPFR.

Shoeboxam:
Closes #150 (main issue) This PR is an initial pass at making the library robust to overflow and rounding in arithmetic.

There were two other issues that were also related to rounding:
Closes #314 
Closes #343

This PR does not add relation relaxation terms for worst-case floating-point rounding in the function (#348).


Shoeboxam:
This became pretty nasty, because I needed to define forward maps with specific rounding modes in division. I decided to "do it right" for floats as well. I think I should give this some time to look for a better approach.

rust/opendp/src/samplers.rs
```diff
@@ -502,17 +513,21 @@ impl<T: CastInternalReal + SampleRademacher + Zero> SampleLaplace for T {
...
+        // (shift / scale + noise) * scale.
+        let mut value = shift.into_internal();
+        // when scaling into the noise coordinate space, round down so that noise is overestimated
+        value.div_assign_round(&scale.clone().into_internal(), Round::Zero);
+        // the noise itself is never scaled. Round away from zero to offset the scaling bias
+        value.add_assign_round(
+            &laplace, if value.is_sign_positive() {Round::Up} else {Round::Down});
+        // postprocessing back to original coordinate space
+        //     remains differentially private via postprocessing
+        Ok(Self::from_internal(value) * scale)
```
andrewvyrros, line 530: Should we factor this out, so that gaussian and Laplace can both use same logic?

rust/opendp/src/trans/mean/mod.rs
```diff
@@ -1,34 +1,36 @@
...
     Ok(Transformation::new(
         SizedDomain::new(VectorDomain::new(
             BoundedDomain::new_closed(bounds)?), size),
         AllDomain::new(),
         Function::new(move |arg: &Vec<T>| arg.iter().sum::<T>() / _size),
         SymmetricDistance::default(),
         AbsoluteDistance::default(),
-        StabilityRelation::new_from_constant((upper - lower) / _size / _2)))
+        StabilityRelation::new_from_forward(
+            move |d_in: &IntDistance| T::inf_cast(d_in / 2)
```
andrewvyrros, line 31: Should be `d_in.inf_div(2)`?

rust/opendp/src/trans/sum/mod.rs
```diff
@@ -22,30 +22,34 @@ pub fn make_bounded_sum<T>(
...
         SizedDomain::new(VectorDomain::new(
             BoundedDomain::new_closed(bounds)?), size),
         AllDomain::new(),
         Function::new(|arg: &Vec<T>| arg.iter().sum()),
         SymmetricDistance::default(),
         AbsoluteDistance::default(),
-        // d_out >= d_in * (M - m) / 2
-        StabilityRelation::new_from_constant((upper - lower) / T::exact_int_cast(2)?)))
+        StabilityRelation::new_from_forward(
+            move |d_in: &IntDistance| T::inf_cast(d_in / 2)
```
andrewvyrros, line 50: Similarly about `inf_div` here?

rust/opendp/src/trans/variance/mod.rs
```diff
@@ -1,31 +1,32 @@
...
-    let _2 = &_1 + &_1;
+    let _2 = T::exact_int_cast(2)?;
 
-    let range = (&upper - &lower) / _2.clone();
-    if range.clone().checked_mul(&range).is_none() {
-        return fallible!(MakeTransformation, "Detected potential for overflow when computing function.")
-    }
+    let range = upper.inf_sub(&lower)?;
+    // check for potential overflow
+    range.inf_div(&_2)?.inf_mul(&range.inf_div(&_2)?)?;
```
andrewvyrros, line 29: A little explanation here would help.

rust/opendp/src/trans/variance/mod.rs
```diff
@@ -38,34 +39,35 @@ pub fn make_sized_bounded_variance<T>(
         SymmetricDistance::default(),
         AbsoluteDistance::default(),
         StabilityRelation::new_from_constant(
-            (upper - lower).powi(2)
-                * _size
-                / (_size + _1)
-                / (_size - _ddof)
-                / _2)))
+            range.inf_mul(&range)?
```
andrewvyrros, line 42: Is this sufficient, such that we don't need the check above on line 29?

rust/opendp/src/traits.rs
```diff
@@ -544,3 +504,328 @@ impl FloatBits for f32 {
...
+}
+impl_alerting_abs_unsigned_int!(u8, u16, u32, u64, u128);
+macro_rules! impl_alerting_abs_float {
+    ($($ty:ty),+) => ($(impl AlertingAbs for $ty {
+        fn alerting_abs(&self) -> Fallible<Self> {
+            Ok(self.abs())
+        }
+    })+)
+}
+impl_alerting_abs_float!(f32, f64);
```
andrewvyrros, line 831: In discussion, we proposed splitting this up, so that all traits are in a single file for clarity, and impls/macros are in a separate file.

andrewvyrros (commented):
This is great stuff!

I agree with your suggestion that we merge the InfXXX and NegInfXXX variations.

rust/opendp/src/samplers.rs
```diff
@@ -398,10 +400,13 @@ impl<T: Clone + SampleGeometric + Sub<Output=T> + Bounded + Zero + One + TotalOr
         // TODO: benchmark execution time on different inputs
         let uniform = f64::sample_standard_uniform(bounds.is_some())?;
         let direction = bool::sample_standard_bernoulli()?;
-        let geometric = T::sample_geometric(shift.clone(), direction,1. - alpha, trials)?;
+        // make prob conservatively larger
+        let geometric = T::sample_geometric(
+            shift.clone(), direction,(1.).inf_sub(&alpha)?, trials)?;
```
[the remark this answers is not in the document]
Shoeboxam: I made another logical pass through this and significantly changed everything. It worked out nicely.

rust/opendp/src/samplers.rs
Shoeboxam: Factored out, added some more comments, and also censored the mpfr flags, FWIW.

rust/opendp/src/trans/mean/mod.rs
Shoeboxam: This one is ok, owing to the newfangled code comments. When odd, we don't consider datasets with an odd number of edits, so we need to compute `(d_in - 1) / 2` anyways.

rust/opendp/src/trans/variance/mod.rs
Shoeboxam: I loosened the check to be equal to the implicit check when computing the constant, but left it duplicated, because the constant calculation would become sensitive to ordering.

rust/opendp/src/traits.rs
Shoeboxam: I did some significant refactoring here.

Shoeboxam:
I did some significant shuffling of the traits, and added a call to censor_flags from mpfr. I also adjusted the tests to use an input distance of 2 when working with sized domains. A surprising side-effect of the new behavior is that on transformations or measurements with an input SizedDomain, `.check(1, 0.)` will always pass. This is because the input dataset distance is divided down to zero. It's technically right, but has the potential to be a footgun. The following will pass:

```python
assert (make_sized_bounded_mean() >> make_base_laplace(0.)).check(1, 0.) 
```

andrewvyrros:
> I also adjusted the tests to use an input distance of 2 when working with sized domains. A surprising side-effect of the new behavior is that on transformations or measurements with an input SizedDomain, `.check(1, 0.)` will always pass. This is because the input dataset distance is divided down to zero. It's technically right, but seems like a footgun. The following will pass:
> 
> ```python
> assert (make_sized_bounded_mean() >> make_base_laplace(0.)).check(1, 0.) 
> ```

Yeah this is a funny quirk, maybe could use a note in the docs?

andrewvyrros (approved):
For such a big, tricky change, this came out nicely! I'm really glad we spent the time to walk through this together, it helped me understand it much better.

Breaking up the traits was a good move, makes things more manageable. Also the cleanup in `samplers.rs` is nice.

It feels like this work puts us in pretty good shape regarding floating point issues.

Shoeboxam:
@andrewvyrros It does! The only floating-point issue I'm currently aware of is #348.

PR #306, opened 2021-09-20
contrib feature flag

Shoeboxam:
See #305

PR #316, opened 2021-09-22
add contrib feature flag to FFI

Shoeboxam:
The FFI layer is expecting the existence of functions that have been cfg'ed! This missed testing because all of our builds used the "contrib" feature implicitly. The "contrib" feature is enabled by default for debug builds via the dev-dependencies.

Finishes #306.


andrewvyrros (approved):
Release builds, so frustrating!

PR #346, opened 2021-09-28
fix continuous noise samplers

### LINKED ISSUE #140: Remove Box from Relation hints and maps
Remove the usage of `Box` in `PrivacyRelation` and `StabilityRelation` hint and map type signatures.

When we did the conversion to AnyXXX types in #88, we were able to remove most usages of `Box`, but overlooked the Relation hints and maps. This clunky and now unnecessary, we should remove `Box` and use plain types.

### LINKED ISSUE #182: NaN in non-mpfr laplacian sampler
The dev sampler for laplacian noise in environments where mpfr is not supported can still produce NaN floats. The sampling is conducted using a u.abs().ln() transform of u ~ Uniform(-.5, .5). The edge case u == 0. triggers a redraw, but the redraw should also happen when u == -0. 

Given that this is a bug in a sampler that is not used in public releases, it is low-priority. There is a fix in the draft external-sgd branch, as that branch is currently using this sampler for performance reasons.

### LINKED ISSUE #345: Noise scaling in continuous distributions
We have a disclaimer that the continuous samplers suffer from floating-point issues.
We do indeed get "exact" noise (non-porous) from MPFR, but the noise is scaled, which introduces gaps and stacks.
The simple fix is to just compute `(v / scale + noise) * scale` instead of `v + noise * scale`.
Once we've added noise, the rescaling back to the input coordinates is postprocessing.

Shoeboxam:
Closes #182. It was possible to sample -0 in the non-MPFR laplace sampler and induce a NaN.
Closes #345. With this change, I think the continuous noise samplers are no longer subject to major floating-point attacks. I am not aware of any significant floating point issues in the continuous noise samplers, beyond some minor rounding in preprocessing. We will of course need to do more vetting.

For Laplace noise we have Von-Neumann's exact standard exponential sampler from MPFR, and add a sign non-disruptively with rademacher noise to get standard laplacian noise. The input is divided by scale before adding the standard laplacian noise. We now have a DP release without scaling the noise. Postprocess the DP release back to the original coordinates by multiplying by scale.

For gaussian noise we have Algorithm N from https://arxiv.org/pdf/1303.6257.pdf, implemented in MPFR, which adds rejection sampling to get standard gaussian noise. I apply the same noise addition approach as above.

Closes #140. I needed to add a backward map to Laplace, for zero-scale noise, and the box got in the way.

Lastly, I adjusted `GeneratorOpenSSL` to store a fallible. If OpenSSL throws an error, it no longer panics, it stores the error in the rng struct. The error is propagated manually after the randomized MPFR computations are done.

### DIFF
--- a/rust/opendp-ffi/src/any.rs
+++ b/rust/opendp-ffi/src/any.rs
@@ -400,16 +400,16 @@ fn make_any_relation<QI: 'static, QO: 'static, AQI: Downcast, AQO: Downcast>(rel
     }
 }
 
-fn make_any_map<QI, QO, AQI>(map: &Option<Rc<dyn Fn(&QI) -> Fallible<Box<QO>>>>) -> Option<impl Fn(&AQI) -> Fallible<Box<AnyMetricDistance>>>
+fn make_any_map<QI, QO, AQI>(map: &Option<Rc<dyn Fn(&QI) -> Fallible<QO>>>) -> Option<impl Fn(&AQI) -> Fallible<AnyMetricDistance>>
     where QI: 'static + PartialOrd,
           QO: 'static + PartialOrd + Clone,
           AQI: Downcast {
     map.as_ref().map(|map| {
         let map = map.clone();
-        move |d_in: &AQI| -> Fallible<Box<AnyMetricDistance>> {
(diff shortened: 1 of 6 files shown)


andrewvyrros (approved):
What a simple and clean solution!

PR #351, opened 2021-09-30
move windows dir

### LINKED ISSUE #342: Move `windows` dir into `rust`
This doesn't need to be a root directory, it's a rust-only patch. This will help keep the repository root tidy.

- [ ] update links in the windows/ directory .sh patch scripts
- [ ] update .github/workflows/release.yaml CI
- [ ] update developer guide documentation (it links to the windows folder)

Shoeboxam:
Closes #342 
The windows patch is specific to rust, so there's no need for it to be cluttering the root.


andrewvyrros (approved):
Nice, I'm proud of the tidy top-level dir we have!

This one's tough to test, but seems sane visually. Rather than go through the hassle of rejiggering things temporarily, this will get tested implicitly when we generate the next release candidate.

PR #356, opened 2021-10-01
fix resize relation

### LINKED ISSUE #355: Resize relation stability constant
There is a bug in the resize relation stability constant. I wrote it with an equivalent number of additions and removals in mind on d_in / 2 different records. But in the worst case, either d_in removals or d_in additions are made. So the correct relation is not `d_out >= d_in + d_in % 2` (rounding up to the nearest even, assuming a nearly even edit distribution); the correct relation is `d_out >= d_in * 2`. If d_in records are removed, then resize could, in the worst case, make commensurately d_in additions of the constant, resulting in d_in * 2 additions/removals.

Thanks @orespo for getting me to look into this!

Shoeboxam:
Closes #355 
This spawns a new issue: for the use-case when N is known, it is impossible to establish SizedDomain without incurring a 2x penalty.


andrewvyrros (approved):
Ouch, this is a bit of a bummer. But good catch!

PR #361, opened 2021-10-04
stability histogram relation relaxation

### LINKED ISSUE #360: Stability mechanism relaxations
- The stability mechanism enforces an upper bound on the privacy parameters, but this bound is unnecessary
    * if a mechanism is eps-DP, it is also (eps+a)-DP for any a >= 0. Similar for delta.
- The count_by stability knows n, and can be tightened to d_in * 2^(1/p - 1). See #158 
- The non-negative checks for epsilon and delta should happen before potential division by zero.
- There is a typo in the error message itself, but regardless I'd like to standardize error handling across mechanisms.

Shoeboxam:
Closes #360


andrewvyrros (approved):
This is a nice little tweak!

PR #370, opened 2021-10-17
stability mechanism hardening

### LINKED ISSUE #369: Stability mechanism hardening
I've been doing some reading on the stability-based histogram measurement. 
https://privacytools.seas.harvard.edu/files/privacytools/files/complexityprivacy_1.pdf#page=24

1. The paper is written such that `q(x) = sum(q(x_i)) / n`, as in, q is not a count, it is a linear query. Therefore when the query is a simple count, the scale and threshold should not have a 1/n factor.
2. From page 13 "Variants of the definition and notation." The stability mechanism expects ordered data. We don't currently have a way to distinguish between ordered and unordered data. I'm hesitant to call this an issue, because the HashMap ordering itself is random (albeit not with secure noise).
3. The thresholding part of the proof assumes Hamming distance d_in = 1. Therefore modifying an input can change two bins. If I'm understanding the proof correctly, I think the magic 2 constant in the thresholding ln should actually be dataset distance. The aggregator would either need to emit a 2-parameter distance consisting of sensitivity and dataset distance, or be fused with the stability measurement.
4. The overall proof is a basic composition of the noise addition and thresholding. This means the calibration of gaussian noise is more nuanced than I accounted for. It should be calibrated similar to the gaussian mechanism, and any delta consumed while noising should eat into the delta budgeted for thresholding.

Shoeboxam:
Closes #369 

Status of each of the points raised in the issue:
1. Fixed
2. This is fine, as the hashmap is unordered
3. Fixed by fusing the aggregator and measurement and [this proof](https://github.com/opendp/opendp/issues/369#issuecomment-948052190)
4. Disabled

What is changed:
- removed `make_base_stability`. We'll need different stability constructors for different kinds of vector-queries anyways.
- added `make_count_by_ptr`, an L1Distance constructor for a propose-test-release stability histogram measurement
- updated docs, tests, examples


andrewvyrros:
Oops, I think some PyTorch models got checked in by mistake!

python/example/network/model_checkpoints/ddp_lstm/model.pt
python/example/network/model_checkpoints/p2p_lstm/model.pt

PR #372, opened 2021-10-25
smooth sensitivity estimators

Shoeboxam:
This PR adds a collection of smooth sensitivity estimators for the median, minimum and maximum, using the the laplace, gaussian and 2-sided geometric. You choose the p-space and the constructor chooses the appropriate noise distribution for your data type.

It mainly comes from the [smooth sensitivity paper](https://cs-people.bu.edu/ads22/pubs/NRS07/NRS07-full-draft-v1.pdf) and work with Monika Hu! 

Remains in draft mode until:
1. more feedback
2. refactor into a quantile estimator
2. test the utility, add more thorough tests
3. check that the geometric is (alpha, beta)-admissible
4. expand alpha-beta impls to more than just i32 and f64
5. is the proof only valid on sized datasets?
5. FFI

### DIFF
--- a/rust/opendp/src/dist.rs
+++ b/rust/opendp/src/dist.rs
@@ -9,8 +9,11 @@ use std::fmt::{Debug, Formatter};
 pub type IntDistance = u32;
 
 /// Measures
-#[derive(Clone)]
 pub struct MaxDivergence<Q>(PhantomData<Q>);
+
+impl<Q> Clone for MaxDivergence<Q> {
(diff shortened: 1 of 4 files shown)


Shoeboxam:
Closing this because smooth sensitivity based estimators haven't shown to be very accurate compared to other estimators. Arguably they do provide accuracy, but you can still work out quantile accuracies from binary tree releases. Will reopen if there as a good argument for adding these.

PR #378, opened 2021-10-27
monomorphize floats in integer queries

### LINKED ISSUE #377: Float-valued integer queries
From this discussion post:
_Originally posted by @joshua-oss in https://github.com/opendp/opendp/discussions/297#discussioncomment-1525790_

Allowing float-valued mechanisms to be used to privatize what could just be integer-valued data has been a bit controversial. We don't want to use floats when we don't have to.

Affected aggregators:
- make_count
- make_count_distinct
- make_count_by_categories

Benefits of allowing float-valued mechanisms to operate on integer-valued queries:
1. The execution time of the geometric mechanism is linear wrt scale. A large scale can be unreasonably slow to execute. The laplace mechanism has a constant runtime, but (via postprocessing) the output distribution becomes coarser as scale increases.
2. We don't have a noising algorithm for L2 space. The discrete gaussian is not ready yet.

Shoeboxam:
Closes #377

See the attached issue for context. I'm also open to waiting on the discrete gaussian and looking into constant time granularity adjustments for the geometric mechanism.

### DIFF
--- a/python/src/opendp/trans.py
+++ b/python/src/opendp/trans.py
@@ -319,7 +319,7 @@ def make_count(
     
     :param TIA: Atomic Input Type. Input data is expected to be of the form Vec<TIA>.
     :type TIA: RuntimeTypeDescriptor
-    :param TO: Output Type. Must be an integer.
+    :param TO: Output Type. Must be numeric.
     :type TO: RuntimeTypeDescriptor
     :return: A count step.
     :rtype: Transformation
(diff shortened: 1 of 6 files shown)


andrewvyrros (commented):
Seems good, pending discussion with Salil on philosophy of this approach.

andrewvyrros (approved):

Shoeboxam:
The general consensus from Salil was that these kinds of best-practices are up to the analyst or third-party. We can indicate a preference towards a specific mechanism, and we do, but the analysis still remains valid in the framework regardless of if you chain an integer query with an integer mechanism or a float mechanism.

It was still unclear if it was better to have a separate cast operation on aggregated values. The current implementation has the capability of counting with floats. If this places unnecessary burden on the proof, we can split them.

PR #382, opened 2021-10-28
Working on accuracy_pitfalls.ipynb

### LINKED ISSUE #332: Accuracy Notebook Port
There is a good resource here, but it's for the old library!
https://github.com/opendp/smartnoise-samples/blob/master/analysis/accuracy_pitfalls.ipynb

We've recently added accuracies to the OpenDP library, so this one is ready to pull over.

ecowan:
Closes #332


python/example/accuracy_pitfalls.ipynb
```diff
@@ -0,0 +1,338 @@
...
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 47,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": []
+  },
```
Shoeboxam, line 11: Dead cell!

python/example/accuracy_pitfalls.ipynb
```diff
@@ -0,0 +1,338 @@
...
+   "source": [
+    "# Accuracy: Pitfalls and Edge Cases\n",
+    "\n",
+    "This notebook describes OpenDP's accuracy calculations, and ways in which an analyst might be tripped up by them.\n",
+    "\n",
+    "### Overview\n",
+    "\n",
+    "#### Accuracy vs. Confidence Intervals\n",
+    "\n",
+    "Each privatizing mechanism (e.g. Laplace, Gaussian) in OpenDP has an associated accuracy that is a function of the requested privacy usage, sensitivity of the function whose values is being privatized, etc. Imagine you have data $D$ and you want, for some function $\\phi$ to return $\\phi(D)$ in a differentially private way -- we will call this value $\\phi_{dp}(D)$. An $\\alpha$-level accuracy guarantee $a$ promises that, over infinite runs of the privatizing mechanism on the data in question,\n",
```
Shoeboxam, line 23: I know this came from before, but now it's just a function of alpha, and either sigma or accuracy.

python/example/accuracy_pitfalls.ipynb
```diff
@@ -0,0 +1,338 @@
...
+    "\n",
+    "In OpenDP (and differentially private data analysis generally), there is an extra layer of uncertainty due to the noise added to $\\phi(D)$ to produce $\\phi_{dp}(D)$. OpenDP's accuracy metrics deal only with the uncertainty of $\\phi_{dp}(D)$ relative to $\\phi(D)$ and not the uncertainty of $\\phi(D)$ relative to $\\phi(P)$.\n",
+    "\n",
+    "#### What is $D$?\n",
+    "\n",
+    "OpenDP allows for analysis of data with an unknown number of rows by resizing the data to ensure consistency with an estimated size (see the [unknown dataset size notebook](https://github.com/opendp/opendp/blob/main/python/example/unknown_dataset_size.ipynb) for more details). Accuracy guarantees are always relative to the resized data $\\tilde{D}$, not the data of unknown size.\n",
+    "\n",
+    "#### Synopsis\n",
+    "\n",
+    "Let's say an analyst releases $\\phi_{dp}(D)$ and gets an accuracy guarantee of $a$ at accuracy-level $\\alpha$. $D$ as a dataset of unknown size drawn from population $P$ and will be resized to $\\tilde{D}$. This suggests that over infinite runs of this procedure,\n",
```
Shoeboxam, line 37: "is a dataset of unknown size..."

python/example/accuracy_pitfalls.ipynb
```diff
@@ -0,0 +1,338 @@
...
+    "    # Resize the dataset to length `count_release`.\n",
+    "    #     If there are fewer than `count_release` rows in the data, fill with a constant of 20.\n",
+    "    #     If there are more than `count_release` rows in the data, only keep `count_release` rows\n",
+    "    make_bounded_resize(size=data_size, bounds=age_bounds, constant=20.) >>\n",
+    "\n",
+    "    # Compute the mean\n",
+    "    make_sized_bounded_mean(size=data_size, bounds=age_bounds)\n",
+    ")\n",
+    "\n",
+    "scale = binary_search(lambda s: check_scale(s, preprocessor, 1, epsilon), bounds=(0.0, 1000.0))\n",
```
Shoeboxam, line 138: This works, but it is a bit more idiosyncratic to use 
`scale = binary_search_param(lambda s: preprocessor >> make_base_laplace(scale), 1, epsilon)`
And drop `def check_scale`

python/example/accuracy_pitfalls.ipynb
```diff
@@ -0,0 +1,338 @@
...
+    "\n",
+    "preprocessor = preprocessor >> make_base_laplace(scale)\n",
```
Shoeboxam, line 140: I think it's best this isn't called preprocessor anymore, it's a measurement now.

python/example/accuracy_pitfalls.ipynb
```diff
@@ -0,0 +1,338 @@
...
+     "name": "#%%\n"
+    }
+   }
+  },
+  {
+   "cell_type": "markdown",
+   "source": [
+    "This performance is as expected. $D$ and $\\tilde{D}$ are actually the exact same data (the maximum age in the raw data is 93, so our clamp to $[0, 100]$ does not change any values, and we know the correct $n$), so our theoretical guarantees on $\\tilde{D}$ map exactly to gaurantees on $D$.\n",
+    "\n",
+    "We now move to a scenario that is still realistic, but where the performance does not translate quite as well. In this case, we imagine that the analyst believes the data to be of size 1050 and uses the default imputation within resize so that the extra 50 elements are drawn uniformly from $\\{ 0, 1, ..., 100 \\}$.\n",
```
Shoeboxam, line 170: Just change this to fill with a constant of 50.

python/example/accuracy_pitfalls.ipynb
```diff
@@ -0,0 +1,338 @@
...
+    "    make_split_dataframe(separator=\",\", col_names=var_names) >>\n",
+    "\n",
+    "    # Selects a column of df, Vec<str>\n",
+    "    make_select_column(key=\"age\", TOA=str) >>\n",
+    "\n",
+    "    # Cast the column as Vec<float>\n",
+    "    make_cast_inherent(TIA=str, TOA=float) >>\n",
+    "\n",
+    "    # TODO: we need to impute missing values sampled from uniform, not a constant (see line 168)\n",
+    "    make_impute_uniform_float(bounds=age_bounds) >>\n",
```
Shoeboxam, line 218: `make_impute_uniform_float` -> `make_impute_constant`

python/example/accuracy_pitfalls.ipynb
```diff
@@ -0,0 +1,338 @@
...
+    "\n",
+    "    # Clamp age values\n",
+    "    make_clamp(bounds=age_bounds) >>\n",
+    "\n",
+    "    # Resize the dataset to length `data_size`.\n",
+    "    #     If there are fewer than `data_size` rows in the data, fill with a constant of 20.\n",
+    "    #     If there are more than `data_size` rows in the data, only keep `count_release` rows\n",
```
Shoeboxam, line 225: `count_release` -> `data_size`

python/example/accuracy_pitfalls.ipynb
```diff
@@ -0,0 +1,338 @@
...
+    "data_size = 1050\n",
+    "\n",
+    "n_sims = 1_000\n",
+    "true_D_tilde_mean_ages = []\n",
+    "dp_D_tilde_mean_ages = []\n",
+    "\n",
+    "D_tilde = np.clip(D.age, age_bounds[0], age_bounds[1])\n",
+    "D_tilde_mean_age = np.mean(D_tilde)\n",
+    "\n",
+    "def check_scale(scale, preprocessor, dataset_distance, epsilon):\n",
```
Shoeboxam, line 204: Same pick as above. Can be removed in favor of a more specialized lib function.

python/example/accuracy_pitfalls.ipynb
```diff
@@ -0,0 +1,338 @@
...
+     "text": [
+      "Accuracy interval (with accuracy value 0.2853) contains the true mean on D_tilde with probability 0.4725\n",
+      "Accuracy interval (with accuracy value 0.2853) contains the true mean on D with probability 0.4725\n"
+     ]
+    }
+   ],
+   "source": [
+    "# establish extra information for this simulation\n",
+    "age_lower_bound = 0.\n",
+    "age_upper_bound = 100.\n",
```
Shoeboxam, line 194: This is dead code.

python/example/accuracy_pitfalls.ipynb
```diff
@@ -0,0 +1,338 @@
...
+    "# establish data information\n",
+    "from opendp.trans import make_split_dataframe, make_cast_default, make_clamp, make_bounded_resize, \\\n",
+    "    make_sized_bounded_mean, make_select_column, make_impute_uniform_float, make_cast, make_cast_inherent\n",
+    "data_path = os.path.join('.', 'data', 'PUMS_california_demographics_1000', 'data.csv')\n",
+    "var_names = [\"age\", \"sex\", \"educ\", \"race\", \"income\", \"married\", \"pid\"]\n",
+    "D = pd.read_csv(data_path, names=var_names)\n",
+    "age = D.age\n",
+    "D_mean_age = np.mean(age)\n",
+    "age_lower_bound = 0.\n",
+    "age_upper_bound = 120.\n"
```
Shoeboxam, line 79: This is dead code.

python/example/accuracy_pitfalls.ipynb
```diff
@@ -0,0 +1,338 @@
...
+     ]
+    }
+   ],
+   "source": [
+    "# establish extra information for this simulation\n",
+    "age_lower_bound = 0.\n",
+    "age_upper_bound = 100.\n",
+    "data_size = 1050\n",
+    "\n",
+    "n_sims = 1_000\n",
```
Shoeboxam, line 197: More dead code!

python/example/accuracy_pitfalls.ipynb
```diff
@@ -0,0 +1,338 @@
...
+    "true_D_tilde_mean_ages = []\n",
+    "dp_D_tilde_mean_ages = []\n",
+    "\n",
+    "D_tilde = np.clip(D.age, age_bounds[0], age_bounds[1])\n",
+    "D_tilde_mean_age = np.mean(D_tilde)\n",
+    "\n",
+    "def check_scale(scale, preprocessor, dataset_distance, epsilon):\n",
+    "    return (preprocessor >> make_base_laplace(scale)).check(dataset_distance, epsilon)\n",
+    "\n",
+    "preprocessor = (\n",
```
Shoeboxam, line 207: I think it would be clearer if this whole chain were a function of one parameter, `data_size`? Maybe `make_mean_with_data_size` (or whatever makes sense to you). Then both chains can be simplified to one function call.

python/example/accuracy_pitfalls.ipynb
```diff
@@ -0,0 +1,338 @@
...
+    "    #     If there are more than `data_size` rows in the data, only keep `count_release` rows\n",
+    "    make_bounded_resize(size=data_size, bounds=age_bounds, constant=20.) >>\n",
+    "\n",
+    "    # Compute the mean\n",
+    "    make_sized_bounded_mean(size=data_size, bounds=age_bounds)\n",
+    ")\n",
+    "\n",
+    "scale = binary_search(lambda s: check_scale(s, preprocessor, 1, epsilon), bounds=(0.0, 1000.0))\n",
+    "\n",
+    "preprocessor = preprocessor >> make_base_laplace(scale)\n",
```
Shoeboxam, line 234: Same argument as before- use the specialized function, rename the preprocessor variable once you've chained with a measurement.

python/example/accuracy_pitfalls.ipynb
```diff
@@ -0,0 +1,338 @@
...
+   "execution_count": 50,
+   "outputs": [],
+   "source": [],
+   "metadata": {
+    "collapsed": false,
+    "pycharm": {
+     "name": "#%%\n"
+    }
+   }
+  }
```
Shoeboxam, line 315: Dead cells.

Shoeboxam (changes requested):

Shoeboxam:
Looking good Ethan, I left a few comments towards cleaning the notebook up a bit more. Thanks!

python/example/accuracy_pitfalls.ipynb
Shoeboxam: This one's up to your judgement ^^

ecowan:
@Shoeboxam @andrewvyrros This is ready for review!

Shoeboxam:
This doesn't seem to have been run. The following error is on the first cell:
`No such file or directory: './data/PUMS_california_demographics_1000/data.csv'`

In the `What is D?` section, could you clarify that accuracy is relative to the preprocessed data-- that is, not just resize preprocessing, but also imputations and clipping are not factored into the accuracy.

ecowan:
> @ecowan This doesn't seem to have been run. The following error is on the first cell: `No such file or directory: './data/PUMS_california_demographics_1000/data.csv'`
> 
> In the `What is D?` section, could you clarify that accuracy is relative to the preprocessed data-- that is, not just resize preprocessing, but also imputations and clipping are not factored into the accuracy.

Thanks, definitely not what I intended to push - in fact I can't even reproduce the error locally. Just reran everything and pushed.

Shoeboxam:
@ecowan: Looking at the output, the notebook doesn't show any difference in the last code block. The last code block should demonstrate how the accuracy is wrt D_tilde, not D, by making D_tilde sufficiently distinct from D that the resulting confidence intervals are underestimated.

And this would be much easier to read if you de-duplicated the first two cells with a shared `make_preprocessor` function of just `data_size`, and removed duplicate variable declarations and dead code. 

Releases is not redeclared in the second code-block, so the second experiment appends to the first experiment. But honestly `releases = [measurement(data) for _ in range(n_sims)]` would be even simpler. 

There's an extra pipe: `|This performance is as expected.` and `guarantees` is misspelled. I would also drop uniform imputation in favor of constant imputation.

Shoeboxam (changes requested):
@ecowan: Looking at the output, the notebook fails to show any difference in the second experiment. It should demonstrate how the accuracy may be overestimated because the accuracy is wrt D_tilde, not D. D_tilde needs to be made sufficiently distinct from D so that the effect is exaggerated and the last confidence interval gives you a greater rejection rate than the requested alpha. 

`releases` is not redeclared in the second code-block, so the second experiment appends to the first experiment. But honestly `releases = [measurement(data) for _ in range(n_sims)]` would be even simpler. This might be why the second experiment fails to show a difference.

I think this would be much easier to read if you de-duplicated the first two cells with a shared `make_preprocessor(data_size)` function and removed duplicate variable declarations and dead code like `true_D_tilde_mean_ages`, `dp_D_tilde_mean_ages`, and the redeclaration of `n_sims`, `D_tilde`, `D_tilde_mean_age`. The deduplicating function would make it much easier to see how the two experiments differ.

There's an extra pipe: `|This performance is as expected.` and `guarantees` is misspelled. So I think more proof-reading might be necessary once fixes and changes are made. I would also drop uniform imputation in favor of constant imputation. Uniform imputation is needlessly complex.

Shoeboxam (approved):
Thanks for the edits, looks pretty good now, and I think I've picked enough on this! Appreciate your patience.

PR #388, opened 2021-10-31
183 sparse histogram alp

### LINKED ISSUE #183: Sparse histogram -- Implementation
Placeholder issue to reserve branch name.

ChristianLebeda:
Closes #183  

Result of my work during the OpenDP Fellows Program 2021.
Adds an implementation of the ALP mechanism for releasing sparse histogram.
Main contributions are the Measurement constructors and a post-processing step.

Note:
The primary constructor takes arbitrary hash functions as a parameter. These take as input a key from the input domain and the output should be an usize below s. I use modulo to ensure that information is not leaked even if poor hash functions are provided. That is, the implementation does not crash if the key is hashed outside the range. However, I do assume that the functions never panic on any input.


rust/opendp/src/meas/histogram_alp/mod.rs
```diff
@@ -0,0 +1,398 @@
...
+pub fn make_alp_histogram_parameterized<K, C, T>(n: usize, alpha: T, scale: T, beta: C, size_factor: u32) 
+        -> Fallible<Measurement<SizedHistogramDomain<K, C>, 
+                                AlpDomain<K, T>, 
+                                L1Distance<C>, MaxDivergence<T>>>
+    where K: 'static + Eq + Hash + Clone + CheckNull,
+          C: 'static + Clone + Integer + CheckNull + DistanceConstant<C> + InfCast<T> + ToPrimitive,
+          T: 'static + num::Float + DistanceConstant<T> + CastInternalReal + InfCast<C>,
+          AlpState<K,T> : CheckNull {
+
+    let beta: f64 = T::inf_cast(beta).unwrap().to_f64()
```
Shoeboxam, line 183: Minor, but you don't have to unwrap here, you can just try with `?`: `T::inf_cast(beta)?.to_f64()`

rust/opendp/src/meas/histogram_alp/mod.rs
```diff
@@ -0,0 +1,398 @@
...
+        -> Fallible<Measurement<SizedHistogramDomain<K, C>, 
+                                AlpDomain<K, T>, 
+                                L1Distance<C>, MaxDivergence<T>>>
+    where K: 'static + Eq + Hash + Clone + CheckNull,
+          C: 'static + Clone + Integer + CheckNull + DistanceConstant<C> + InfCast<T> + ToPrimitive,
+          T: 'static + num::Float + DistanceConstant<T> + CastInternalReal + InfCast<C>,
+          AlpState<K,T> : CheckNull {
+    
+    make_alp_histogram_parameterized(n, T::from(ALPHA_DEFAULT).unwrap(), scale, beta, SIZE_FACTOR_DEFAULT)
+}
```
Shoeboxam, line 205: You could omit this entire constructor if `alpha` and `size_factor` in `make_alp_histogram_parameterized` were adjusted option types:

```rust
let alpha = if let Some(a) = alpha {a} else {T::inf_cast(ALPHA_DEFAULT)?)};
let size_factor = size_factor.unwrap_or(SIZE_FACTOR_DEFAULT);
```

rust/opendp/src/meas/histogram_alp/mod.rs
```diff
@@ -0,0 +1,398 @@
...
+
+const ALPHA_DEFAULT : u32 = 4;
+const SIZE_FACTOR_DEFAULT : u32 = 50;
+
+// Implementation of a mechanism for representing sparse histogram.
+// The mechanism was introduced in the paper:
+// "Differentially Private Sparse Vectors with Low Error, Optimal Space, and Fast Access"
+// Available here: arxiv.org/abs/2106.10068
+
+type SizedHistogramDomain<K, C> = SizedDomain<MapDomain<AllDomain<K>, AllDomain<C>>>;
```
Shoeboxam, line 25: Based on our DMs, I'm thinking we could drop SizedDomain here. The input data doesn't need to be of exactly size n, you just need a reasonable estimate of n to derive the size of the projection that will maximize utility (by properly balancing the odds of hash collisions against the size of the release).

This will make it fail to chain with `make_count_by`, but the SizedDomain emitted from that constructor can be dropped as well. If we do need a `count_by` with sized output domains in the future, we'd name it `make_sized_count_by` anyways.

rust/opendp/src/meas/histogram_alp/mod.rs
Shoeboxam: Transferred from DM's: 
I have some thoughts about naming!
- `make_alp_histogram_simple` could be folded into `make_alp_histogram_parameterized` by making two of the args optional
- I feel like most people will use the parameterized constructor, so what do you think about renaming as follows:
    * `make_alp_histogram` -> `make_alp_histogram_with_hashers`
    * `make_alp_histogram_parameterized` -> `make_alp_histogram`
- Does this implementation work with any integer query? It seems like it would work with, for example, grouped integer sums
    * Do you think `make_base_alp*` would be a better prefix? Or is that too general?

So a typical analyst flow would be `make_count_by(...) >> make_base_alp(...)` or `make_sum_by(...) >> make_base_alp(...)` and if they also wanted to specify hashers, then they could run `make_base_alp_with_hashers(...)`.

Shoeboxam:
Would you mind adding documentation to your functions as well? 

```rust
/// Short description.
///
/// # Arguments
/// * `s` - short description
pub fn make_base_alp_with_hashers...
```

The pub fn's are more important, but at least some descriptions on some of the private functions would be very useful. I'll also reuse this documentation in the FFI layer. Thanks!

Shoeboxam:
The above error is because the `CastInternalReal` trait was moved out of the samplers file and into `traits/cast.rs`. 

I think adjusting the import will fix it:

```rust
use crate::traits::CastInternalReal;
```

Shoeboxam (approved):
This is awesome work! Looking forward to making this available in the language bindings! Thanks for your patience with the process. This will be in 0.4.0, which should release soon.

PR #396, opened 2021-11-17
Merge `opendp` and `opendp-ffi` crates

### LINKED ISSUE #395: Merge `opendp` and `opendp-ffi` crates
We want to automatically generate FFI bindings. One of the necessary steps is to merge the `opendp` and `opendp-ffi` crates.

1. If we generate extern functions via proc-macros, the extern functions live in the opendp crate. Unfortunately, extern functions cannot reliably be re-exported from dependencies on all architectures. 
2. We could try parsing a foreign crate's sources (opendp) from the opendp-ffi build.rs and generate extern functions to the `OUT_DIR`, but crate sources are somewhat isolated. It might be possible to unpack the cargo cache's .crate files for the opendp lib. This seems fragile.

If we merge the crates, and put FFI behind a flag, then the build.rs could read the crate's own sources, and generate extern fn's that don't need to be re-exported.

Shoeboxam:
Closes #395

See attached issue for reasoning.


andrewvyrros:
In addition to the inline comments, there are a few more changes to get this to work for builds and releases:

#### `.github/workflows/release.yml`
* At the top, need `FEATURES: untrusted,ffi`

#### `tools/python_build.sh`
* In `build()`, change the library names to drop `_ffi`.

#### `tools/build_tool.py`
* In `_main()`, change default for `features` to `untrusted,ffi`

#### `tools/publish_tool.py`
* In `rust()`, drop the last two lines (sleep, second publish)

#### `tools/release_tool.py`
* In `version()`, `rust/opendp/...` -> `rust/...`
* In `version()`, remove lines for `opendp-ffi`

andrewvyrros (changes requested):
Huge change!

Shoeboxam:
Thanks for the thoughtful comments Andy!

andrewvyrros:
Looks great! There were some changes to versions of `rug` & `gmp-mpfr-sys` from #405 that were inadvertently dropped from the merged `Cargo.toml`, causing the build to fail on M1. I've restored those.

andrewvyrros (approved):

PR #398, opened 2021-11-17
__del__ destructor for AnyObjects

### LINKED ISSUE #397: Memory leaks in python bindings
There are memory leaks in the python bindings that are consuming memory on some longer-running processes. The memory that needs to be freed explicitly is memory passed to python that was allocated in rust. We are relatively principled about freeing this memory for most data types- FfiSlices, strings, bools, etc. However, we leak the intermediate AnyObjects when converting between python and rust data representations in constructor, check and invocation functions.

Shoeboxam:
Closes #397 

This adds a `__del__` destructor for AnyObjects, so that constructor arguments, invoke arguments/returns, and check arguments are freed. 

It also adds some python refcounting manipulation to prevent the AnyObjects used to build HashMaps from being freed too early, or to be double-freed.

### DIFF
--- a/python/src/opendp/_convert.py
+++ b/python/src/opendp/_convert.py
@@ -276,20 +276,23 @@ def _hashmap_to_slice(val: Dict[Any, Any], type_name: str) -> FfiSlicePtr:
     ffislice = _wrap_in_slice(ctypes.pointer((AnyObjectPtr * 2)(keys, vals)), 2)
 
     # The __del__ destructor on `keys` and `vals` is called and memory freed when their refcounts go to zero.
-    # We can't allow their refcount to go to zero until ffislice is freed.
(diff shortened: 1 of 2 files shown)


andrewvyrros (approved):
This was a tricky one, thanks @Shoeboxam for walking me through it!

It might seem a bit questionable manually setting a `__class__` pointer, but we're just changing it from `AnyObjectPtr` to `ctypes.POINTER(AnyObject)`, which are basically the same thing (modulo a bit of `ctypes` magic).

Note to future selves: If this pattern arises again, we could create an abstraction for accessing an `AnyObject` without taking ownership (something like `SharedAnyObjectPtr`). This could be useful if we want to feed raw C arrays from numpy/pytorch into opendp without unpacking into a list first.

PR #399, opened 2021-11-17
`proc_macro_attribute` extern "C" fn generator

### LINKED ISSUE #131: Automatic generation of FFI wrappers and metadata
Currently, when adding a component, we need to manually create the FFI wrapper function and its associated bootstrap.json metadata. It'd be fantastic if this could be done automatically from the underlying Rust function in a build script. But this is hard!

One of the main stumbling blocks is wrappers requiring complex dispatch on type. The macros in `dispatch.rs` only handle one level of indirection, so this means implementing multiple nesting `monomorphize()` functions.

Shoeboxam:
Closes #131 

This PR focuses only on generating the extern "C" fn's, not the bindings languages.
Early WIP.


Shoeboxam:
Closing in favor of #549

PR #405, opened 2021-11-25
Release scripts

andrewvyrros:
This incorporates a new helper script, `tools/release_tool.py`. It features commands to perform all the steps of a release:

    $ python tools/release_tool.py init -t minor
    $ python tools/release_tool.py changelog
    $ python tools/release_tool.py version -p 1
    $ python tools/release_tool.py preflight
    $ python tools/release_tool.py create
    $ python tools/release_tool.py watch
    $ python tools/release_tool.py postflight
    $ python tools/release_tool.py version
    $ python tools/release_tool.py preflight
    $ python tools/release_tool.py create
    $ python tools/release_tool.py watch
    $ python tools/release_tool.py postflight

In addition, there are meta-commands that run the entire thing for a major/minor/patch release. Typical usage would be something like this:

    $ python tools/release_tool.py minor

This will make a new release branch based on main, bump the minor version number, then create and sanity check RC and regular releases, with all the versioned files updated appropriately.

This change also streamlines the Windows build. It uses the stock Rust toolchain, rather than patching with platform mingw components. It also let's the `gmp-mpfr-sys` crate build its internal copies of `gmp` and `mpfr`, rather than building them manually. This seems to work fine now. The `release.yml` workflow has been refactored to use a separate build script, which will make it possible to create identical builds when preflighting a release. 

    * Added tools directory.
    * Added release_tool.py helper script.
    * Updated CHANGELOG.md to match new format.
    * Added build_tool.py helper script.
    * Broke out common build operations into separate rust_build.sh and python_build.sh scripts.
    * Added publish_tool.py helper script.
    * Refactored release.yml workflow using new scripts.
    * Moved assert_version.py to tools directory.
    * Fixed compiler warnings.
    * Simplified windows build to use vendored builds of gmp and mpfr in gmp-mpfr-sys crate.
    * Removed patch of Rust toolchain for msys2.
    * Cleaned up entire workflow.
    * Implemented preflight test.
    * Restored versions to 0.0.0-development.


Shoeboxam:
It's awesome that Windows builds compile without strange hacks! 
One reason why things could be working now, without tinkering, is because the bundled msys is still similar enough to the system msys. Fingers crossed the compiler is indeed using the system msys libs now.

tools/release_tool.py
```diff
@@ -0,0 +1,355 @@
...
+    conf = read_conf(args)
+    # There are two possibilities for changelog items:
+    # * All unreleased items are going in this release (more common, when the release contains the current main branch).
+    #   In this case, we expect a single heading "Unreleased", containing all unreleased items, which will be promoted
+    #   to a versioned section; for future items, a new "Unreleased" section will be generated.
+    # * A subset of unreleased items are going in this release (less common, when the release omits some changes).
+    #   In this case, we expect a heading "Unreleased (<VERSION>)", containing the subset of unreleased items going in
+    #   this release, which will be promoted to a versioned section, and another "Unreleased" section containing the
+    #   remaining unreleased items, which will be left as is (for future items).
+    full_heading_regex = "^## \\[Unreleased\\](.*)$"
```
Shoeboxam, line 148: How did we both end up writing code to parse markdown at the same time?

tools/release_tool.py
```diff
@@ -0,0 +1,355 @@
...
+    # If this isn't a subset, prepend a new unreleased section.
+    if not subset:
+        replacement = "## [Unreleased](https://github.com/opendp/opendp/compare/stable...HEAD)\\n\\n\\n" + replacement
+    substitution_arg = f"-e 's|{heading_regex}|{replacement}|'"
+    inplace_arg = "-i ''" if platform.system() == "Darwin" else "-i"
+    run_command("Updating CHANGELOG", f"sed {inplace_arg} {substitution_arg} CHANGELOG.md")
+    commit("CHANGELOG", ["CHANGELOG.md"], f"RELEASE_TOOL: Updated CHANGELOG.md for {conf.target_version}.")
+
+
+def version(args):
```
Shoeboxam, line 165: This is unused, but I'm ok with leaving it in .


Could come in handy later down the line if the libs desync and the windows build issues reappear.

Shoeboxam (approved):
This looks awesome! Really promising simplifications to the Windows build!

PR #408, opened 2021-12-08
split `make_count_by_ptr` -> agg + `make_base_ptr`

### LINKED ISSUE #407: Split `make_count_by_ptr`
In order to support pre-aggregates, this measurement should be broken into an aggregator and a privatizer.

- `make_count_by`
- `make_base_ptr`

Shoeboxam:
Closes #407.

This splitting of components takes advantage of the fact that the L1Distance is an upper bound on dataset distance. Will need to re-assess once we add more aggregators. One potential approach is adding a combinator that adjusts a measurement such that the output distance is a tuple that includes a copy of the input distance. This way the input dataset distance can be forwarded to `make_base_ptr`.

This PR also removes the SizedDomain from `make_count_by`, such that it will chain with both `make_base_alp` and `make_base_ptr`.


andrewvyrros (approved):
Nice!

PR #424, opened 2022-02-08
sequential compositor w/ static distances (double dispatch)

### LINKED ISSUE #7: Basic Composition Combinator -- Implementation

Shoeboxam:
Closes #7.
Replaces #242. 

AnyMeasureDistances are gone. An additional dispatch is introduced to prevent the proliferation of composition traits in measurements.

I'm experimenting with a different way of handling the PRs, where the "branch I want to merge into" is the dependency, instead of merging directly into main.


Shoeboxam:
Closing in favor of #482, which simplifies the implementation via privacy maps.

PR #437, opened 2022-03-10
Add new packaging of CLA and associated docs

### LINKED ISSUE #364: CLA web form
CLA file: https://opendp.org/files/opendifferentialprivacy/files/cla_opendp_project_2021.pdf

Currently, contributors submit this manually, and we track the submissions somehow. This is likely an impediment to contributors, and is a hassle for us to manage.

We should make a little web form to handle this automatically. Ideally, this wouldn't be specific to opendp repo, but would live on opendp.org, or at least a separate repo, so that it could be shared across projects.

This is just to generate a list of names that have signed the CLA. If we have that, I think we can check the authors of PRs manually, at least until the list becomes unwieldy. In a perfect world, the PR workflow could do this automatically, but that's not necessary to get started. The main value is in form submission.

andrewvyrros:
Closes #364.

This incorporates the CLA forms and associated instructions into the docs.opendp.org deployment.
It goes hand-in-hand with the new [clabot-config repo](https://github.com/opendp/clabot-config), which has the config files for the `cla-bot` GitHub App which does the CLA validation.

Note that `cla-bot` hasn't been turned on for any repos yet. I'll do that as soon as this PR is merged.


Shoeboxam (approved):
The workflow is pretty slick! Glad to see this doesn't add a lot of traction.

PR #441, opened 2022-03-30
Support aarch64-apple-darwin

### LINKED ISSUE #439: M1 mac binaries
M1 mac machines fail to load the library, because they expect their own binaries.

This seems promising: https://stackoverflow.com/questions/66849112/how-do-i-cross-compile-a-rust-application-from-macos-x86-to-macos-silicon
However, `gmp-mpfr-sys` doesn't support cross-compilation. I think we'll need a self-hosted runner, and rent a cloud service to handle our M1 CI.

andrewvyrros:
Fixes #439. Thanks @Shoeboxam for doing the core of this!

    * Add target option to rust_build.sh
    * Extend release.yml to cross-compile aarch64 and merge into universal binary
    * Fix some Cargo.toml paths missed in #396
    * Update CHANGELOG

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -61,7 +61,7 @@ jobs:
             ~\.cargo\git\db
             ~\AppData\Local\gmp-mpfr-sys
             rust\target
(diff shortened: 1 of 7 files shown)


Shoeboxam (approved):
Looks good! Glad to have this!

PR #442, opened 2022-03-31
Discrete Laplace Mechanism

### LINKED ISSUE #414: Floating-point issue in noise samplers
Dear OpenDP team,

As suggested by @Shoeboxam, I took a look at the approach OpenDP uses to sample noise. My understanding is that it implements three main mitigations against floating-point issues: 
1. using MPFR to generate a hole-free noise distribution centered on zero,
2. computing `(noise + shift/scale)*scale` instead of `noise*scale + shift`, 
3. and rounding in conservative directions at every step.

None of these mitigations address the problem that when summing two floating-point numbers, the result has the precision of the least-precise summand, which creates distinguishing events. In particular, this problem occurs regardless of whether noise is added before or after scaling, so the second mitigation does not work.

One easy way to see this is to take scale=1, which makes Mitigation 2 a no-op. The following proof of concept adds Laplace noise of scale 1 to 0 and to 1, and checks the precision level of the output.

```py
from opendp.trans import *
from opendp.meas import *
from opendp.comb import *
enable_features("contrib")
enable_features("floating-point")

samples = 1000
data = "1,0"
parse = make_split_dataframe(separator=",", col_names=["A", "B"])
noisy_sum = (
    make_cast(TIA=str, TOA=float)
    >> make_impute_constant(0.)
    >> make_clamp(bounds=(0., 1.))
    >> make_bounded_sum(bounds=(0., 1.))
    >> make_base_laplace(scale=1.0)
)
# Noisy sum, col A
sum_a = parse >> make_select_column(key="A", TOA=str) >> noisy_sum
# Noisy sum, col B
sum_b = parse >> make_select_column(key="B", TOA=str) >> noisy_sum

out_a = [sum_a(data) for _ in range(0, samples)]
out_b = [sum_b(data) for _ in range(0, samples)]
mul_a = sum([(o*(2.**53)).is_integer() for o in out_a])
mul_b = sum([(o*(2.**53)).is_integer() for o in out_b])

print(f"{mul_a}/{samples} outputs from 1 are multiples of 2^-53...")
print(f"... but only {mul_b}/{samples} from 0 are.")
````

This prints:

```
1000/1000 outputs from 1 are multiples of 2^-53...
... but only 722/1000 from 0 are.
````

One solution is to determine in advance, based on the parameters of the aggregation, which level of precision you need, and round all outputs to this precision. This is what Google's DP libraries do. Another solution is to use MPRF to generate a hole-free noise distribution *centered on shift, with the right noise scale*, to avoid the floating-point operations entirely.

This issue only happens if you enable the `floating-point` feature, which the documentation warns about. So I'm assuming you do not consider this as a vulnerability, and posting this on GitHub as a regular issue.

Best regards,

Damien

Shoeboxam:
Related to #414


Shoeboxam:
This PR is being subsumed by smaller PRs-- lipschitz mul float, bernoulli sampler adjustments in the rr sampler PR, tm chainer and log-time geometric samplers in separate PRs, etc.

Shoeboxam:
Closing in favor of #493.

PR #444, opened 2022-04-05
Fix docs.rs documentation and historical python documentation

### LINKED ISSUE #432: Python API docs for old releases are same as latest
See example:
https://docs.opendp.org/en/v0.1.0/api/python/opendp.meas.html

In the above example, the documentation includes things like `make_base_ptr`, which don't exist in v0.1.0.

### LINKED ISSUE #433: Rust API docs should be compiled with all features enabled
The rust API docs are by default built only with default features. This means docs.rs doesn't contain documentation for any constructors. Example:
https://docs.rs/opendp/latest/opendp/trans/index.html

Shoeboxam:
Closes #433
Closes #432

This also changes the behavior of `OPENDP_HEADLESS`. The library will always attempt to load `lib`, but if `OPENDP_HEADLESS` is set, failure to load `lib` is not considered an error. This is so that headless may be set in sphinx multiversion via `conf.py`, but the smoke-test doctests still work.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -4,13 +4,13 @@ import sys
 import os
 from datetime import datetime
 
-# We're inside source when this runs.
-sys.path.append(os.path.abspath('../../python/src'))
-# print("*****************************************")
-# [print(p) for p in sys.path]
-# print("*****************************************")
+# docs should be built without needing import the library binary for the specified version
+os.environ["OPENDP_HEADLESS"] = "true"
 
-print("*****************************************")
+# We're inside source when this runs.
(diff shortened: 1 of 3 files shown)


andrewvyrros (commented):
~~It seems like there are still issues with #432. If I do `make versions` and then look at `build/html/en/v0.1.0/api/python/opendp.meas.html`, it contains newer functions (e.g., `make_base_analytic_gaussian`).~~

~~Maybe an issue with munging `sys.path`, the newer versions are somehow getting included?~~

Apologies, was accidentally testing on branch `438-documentation-fixes`!

andrewvyrros (approved):

PR #456, opened 2022-04-13
quantile scorer transformations

### LINKED ISSUE #324: Exponential Mechanism -- Implementation
There's an exponential mechanism we could easily port from previous work on SmartNoise-Core. It would have to sit behind the "floating-point" flag. 
This might be too difficult for a first issue, but having a prior implementation should help:
https://github.com/opendp/smartnoise-core/blob/develop/runtime-rust/src/utilities/mechanisms.rs#L210-L265

An initial design direction could be as follows--
```
input_domain: VectorDomain<AllDomain<TI>> where TI is numeric
output_domain: TO where TO is integral
input_metric: AbsoluteDistance<Q>
output_metric: MaxDivergence<Q>
```

The measurement takes in a vector of scores, and emits the chosen index.

Christina's base-2 exponential would be a more robust solution, but I think that belongs in its own issue.

### LINKED ISSUE #325: Quantile Utility -- Implementation
If we have an exponential mechanism that takes in a vector of scores (#324), we could add an aggregator for quantile utilities. The transformation could behave as follows--

```
input_domain: VectorDomain<AllDomain<T>> where T is numeric
output_domain: VectorDomain<AllDomain<T>>
input_metric: SymmetricDistance
output_metric: AbsoluteDistance<T>
```

The constructor would accept a candidate set, and the function is tasked with converting a vector of numerics into a vector of candidate scores. The exponential mechanism in #324 releases the index into said candidate set.

For a prior implementation, see: https://github.com/opendp/smartnoise-core/blob/develop/runtime-rust/src/components/quantile.rs#L112-L200

In the current state of the library, this output metric and domain will not chain with any other constructor than #324. But VectorDomain is probably not the right output domain, as the aggregated data is no longer a dataset!

A downside with this approach is that multiple quantiles may not be released at once.
Welcoming opinions on other directions.

Shoeboxam:
Closes #325 

* adds quantile scorer transformation (sized and unsized versions)
* adds InfDistance metric, for the distance between adjacent quantile score vectors


Shoeboxam:
* main
  * **PR #456** <a href="https://app.graphite.dev/github/pr/opendp/opendp/456" target="_blank">Graphite</a>  👈
    * **PR #641** <a href="https://app.graphite.dev/github/pr/opendp/opendp/641" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/opendp/opendp/456?utm_source=stack-comment).

Shoeboxam:
Splitting this PR into implementation (#702) and proof (#703).

PR #465, opened 2022-06-09
ordered metrics and metric casters

### LINKED ISSUE #357: Formalize treatment of data ordering
We're typically using a metric, SymmetricDistance, that is well-defined on unordered data. 

On the other hand, we have at least two transformations that are order-dependent:
- bounded sum
- resize

In these transformations the output of the function is dependent on the order of the input data, which isn't accounted for in the privacy calculus. 

For the bounded sum, we could switch the metric to ModularDistance, and use wrapping noise perturbation (thanks Salil). For the resize, we could do randomized sampling without replacement.

I opened this issue in response to https://github.com/opendp/opendp/issues/335#issuecomment-931980395, and based on some prior talks in the library architecture meetings.

Shoeboxam:
Partially resolves #357.

Removes `make_cast_metric` in favor of a more specialized set of metric casters.


andrewvyrros (approved):
Looks good! These types and traits give a lot of nice infrastructure.

Shoeboxam:
Thanks for the helpful code reviews on zoom!

PR #466, opened 2022-06-09
handle ordering in resize

Shoeboxam:
See #357. Doesn't close.

Refactors the resize transformation to shuffle when necessary, in accordance with the ordered-ness of the input and output metric.


andrewvyrros (approved):
Making progress!

Issue #348, opened 2021-09-29
Relation relaxation terms for worst-case floating point rounding


Shoeboxam:
There should be an additional loosening term in aggregator relations to account for the potential worst-case change in output of the function when adding or removing input observations changes floating point rounding directions of following operations.

Shoeboxam:
Closed by #467 #475 #474.

PR #467, opened 2022-06-09
consider finite types in sum

### LINKED ISSUE #458: Support modular metric spaces
We should add constructors for aggregation transformations that use modular arithmetic, as well as noise addition mechanisms that add modular nose.

Our current `sized_bounded_sum` uses a constructor check to ensure that the computation cannot overflow. This can be overly conservative when working with wide bounds.

I'm more in favor of introducing modular constructors than parameterizing the metrics.

Shoeboxam:
This resolves issues with finite data types specifically for the sum.

Adds relaxation term from #348 to the sum. See also #5.


rust/src/trans/sum/int/split/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+                    Ordering::Equal => (neg, pos),
+                }
+            });
+            neg.saturating_add(pos)
+        }
+    })+);
+}
+
+impl_unsigned_int_split_sat_sum! { u8 u16 u32 u64 u128 usize }
+impl___signed_int_split_sat_sum! { i8 i16 i32 i64 i128 isize }
```
andrewvyrros, line 45: Similar to our discussion about the implementation of `MakeBoundedSum` and `MakeSizedBoundedSum` traits, I find it a bit annoying that we have to resort to macros here, as it makes it harder to follow the flow. I understand why, and don't see any cleaner way to do it. It's just frustrating that Rust doesn't have a way to implement a trait for different generic types in different ways.

rust/src/trans/bootstrap.json
```diff
@@ -851,6 +851,72 @@
...
+                "example": {
+                    "function": "get_first",
+                    "params": ["bounds"]
+                }
+            }
+        ],
+        "ret": {
+            "c_type": "FfiResult<AnyTransformation *>"
+        }
+    },
```
andrewvyrros, line 919: Missing entries for the `_int_XXX_sum` variations?

andrewvyrros (commented):
This is amazing work! I think the taxonomy you came up with is great, and having the individual constructors makes it easier to understand what's happening.

I'm having a little trouble reasoning about which versions of the relations are used in the different situations, but I think you can walk me through that.

rust/src/trans/sum/int/modular/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+    let (lower, upper) = bounds.clone();
+    Ok(Transformation::new(
+        VectorDomain::new(BoundedDomain::new_closed(bounds)?),
+        AllDomain::new(),
+        Function::new(|arg: &Vec<T>|
+            arg.iter().fold(T::zero(), |sum, v| sum.wrapping_add(v))
+        ),
+        SymmetricDistance::default(),
+        Modular::default(),
+        StabilityRelation::new_from_constant(lower.alerting_abs()?.total_max(upper)?)
```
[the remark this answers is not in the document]
Shoeboxam: Actually, no! If upper is negative, then |lower| will always be greater than upper. If upper is positive, then the absolute value is a no-op. Therefore the absolute value on upper is unnecessary.

rust/src/trans/sum/mod.rs
```diff
@@ -1,62 +1,106 @@
...
+
+pub trait MakeBoundedSum: Sized + CheckNull + Clone + TotalOrd {
+    fn make_bounded_sum(bounds: (Self, Self)) -> Fallible<BoundedSumTrans<Self>>;
+}
+
+macro_rules! impl_make_bounded_sum_int {
+    ($($ty:ty)+) => ($(impl MakeBoundedSum for $ty {
+        fn make_bounded_sum(bounds: (Self, Self)) -> Fallible<BoundedSumTrans<Self>> {
+            make_bounded_int_monotonic_sum(bounds)
+                .or_else(|_| make_bounded_int_split_sum(bounds))
```
andrewvyrros, line 47: The additional dispatch options here (and in `impl_make_sized_bounded_sum_int`) are nice. But basing things on the constructors returning `Err` feels a little risky, as unexpected failures would be masked. Would it be possible to have explicit choice logic here, rather than relying on errors on the happy path?

Shoeboxam:
Here's a bird's eye view of the algorithms.

First off, we need a solution for overflow. Here are the options:
- checked 
- monotonic saturating sum 
- ordered saturating sum
- split saturating sum
- modular sum (not implemented in this PR, integer only)

Just using an overflow solution is sufficient to make integer sums DP. The ideal approach is to attempt a checked sum, then attempt a monotonic saturating sum, and then fall back to one of the other sums.

Notice that only the ordered saturating sum requires an expensive shuffle or sort (can only sort if metric is unbounded).

In addition, for floats we need to compensate for float rounding. Here are the options for floats:
- truncated sum, add relaxation term to the map, and use one of the overflow solutions
- shuffle, split and rtz with saturation

The second solution is better asymptotically, but for 64 bit floats, it is only better when the datasets are unreasonably large. So I only implemented the first rounding solution, paired with the checked overflow solution and paired with the ordered saturating sum solution.

For floats I opted for the checked solution because the sum is unlikely to be rejected due to the high dynamic range of floats. Should the check fail, I also implemented the ordered sum, because there's a fair chance we'll need to shuffle anyways, when truncating. Truncation only requires ordering if it actually truncates.

Finally, the float relaxation term is dependent on the summation algorithm:
- sequential with O(n^2) sensitivity increase
- pairwise with O(log_2(n) * n) sensitivity increase

Pairwise is not quite as efficient as sequential. There's also utility in having a sequential map around, to calibrate sensitivity for systems that use sequential summation (like MySql, Sqlite).

rust/src/trans/sum/mod.rs
Shoeboxam: I've updated to have explicit logic.

rust/src/trans/sum/int/split/mod.rs
Shoeboxam: Agreed. They don't even permit two blanket impls for generics that are provably disjoint by the type system via associated types.

```rust
struct Floats;
struct Ints;
trait Distingisher {type Assoc;}
impl Distinguisher for f64 {type Assoc = Floats;}
impl Distinguisher for i64 {type Assoc = Ints;}

impl<T: MyTrait<Assoc=Floats>> MakeBoundedSum for T {...}
impl<T: MyTrait<Assoc=Ints>> MakeBoundedSum for T {...}
```

rust/src/trans/sum/float/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+        + SaturatingAdd
+        + CanSumOverflow
+{
+}
+
+// Marker type to represent sequential, or recursive summation
+pub struct Sequential<T>(PhantomData<T>);
+
+// Marker type to represent pairwise, or cascading summation
+pub struct Pairwise<T>(PhantomData<T>);
```
andrewvyrros, line 75: These marker types are really clever. Along with the SumRelaxation/UncheckedSum/SaturatingSum traits, it's a cool way to do dispatch on a single primitive type. I wonder if we could use this technique for other algorithmic variations?

rust/src/trans/sum/float/checked/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+        return fallible!(MakeTransformation, "potential for overflow when computing function")
+    }
+
+    let (lower, upper) = bounds.clone();   
+    let ideal_sensitivity = upper.inf_sub(&lower)?;
+    let relaxation = S::relaxation(size_limit, lower, upper)?;
+
+    Ok(Transformation::new(
+        
+            VectorDomain::new(BoundedDomain::new_closed(bounds)?),
```
andrewvyrros, line 41: Looks like some funky indentation here.

andrewvyrros (commented):
There's a lot going on here, but it really came together cleanly! I think there's a good balance of individual constructors vs dispatching on Domain & Metric. Splitting things up into modules by type and then algorithm is a big win, it helps to keep each scope manageable. The last round of changes really improved the understandability of the code, especially in the top-level convenience constructors. Expressing the choices explicitly (and capturing common abstractions in traits) makes it straightforward to follow the logic, so you know exactly what you're getting. This is really nice.

There are a few very minor issues (some typos and a missing unit test), but they can go into a future change if preferred.

rust/src/trans/sum/mod.rs
```diff
@@ -7,108 +7,218 @@ pub use int::*;
...
+    fn sum_can_overflow(size: usize, bounds: (Self, Self)) -> bool;
+}
+
+impl<T: InfMul + ExactIntCast<usize>> CanSumOverflow for T {
+    fn sum_can_overflow(size: usize, (lower, upper): (Self, Self)) -> bool {
+        T::exact_int_cast(size)
+            .and_then(|size| lower.inf_mul(&size).or(upper.inf_mul(&size)))
+            .is_err()
+    }
+}
```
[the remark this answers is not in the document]
Shoeboxam: Should be *and*, not *or*. Otherwise agreed.

rust/src/trans/sum/int/checked/ffi.rs
```diff
@@ -0,0 +1,38 @@
...
+            + AddIsExact,
+        for<'a> T: Sum<&'a T>,
+        IntDistance: InfCast<T>, {
+        let bounds = try_!(try_as_ref!(bounds).downcast_ref::<(T, T)>()).clone();
+        make_sized_bounded_int_checked_sum::<T>(size, bounds).into_any()
+    }
+    let size = size as usize;
+    let T = try_!(Type::try_from(T));
+    dispatch!(monomorphize, [(T, @integers)], (size, bounds))
+}
```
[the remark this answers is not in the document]
Shoeboxam: Added tests for all mod.rs and ffi.rs. Agreed, this was needed.

rust/src/trans/sum/float/mod.rs
Shoeboxam: I really liked how it doesn't have an explicit argument, but it does have a downside. It ends up essentially becoming an explicit argument:
1. Rust functions don't support defaults for type args, so the type has to be inferred or set explicitly.
2. In this situation, the type can never be inferred from context, so the user has to figure out how to set it when using the rust API (import Pairwise, supply it as a type arg). I am able to set a default type arg in the python API, which I do.

Since it's now an explicit argument, it unfortunately also forces the user to specify the atomic type explicitly, or `_`.

Shoeboxam:
Added tests for all functions and ffi functions, and reformatted all `sum/` files.

andrewvyrros (commented):
Good catch on `sum_can_overflow`. Glad we have a test for this now.

andrewvyrros (approved):
Last round of changes added some nice small improvements!

Let's do this!!!

PR #469, opened 2022-06-12
windows build fix, development environment documentation

### LINKED ISSUE #400: Interactive module should be behind a feature flag
The `interactive` module contains prototype code that isn't ready for prime time. We should put it behind a feature flag, disabled by default, to make this status clear.

Shoeboxam:
* fix the `--no-default-features` cargo flag
    * primarily used for easy windows builds
* smoke-test now checks that the `--no-default-features` flag works
* don't add opendp as a dev-dependency of opendp to enable the untrusted flag by default when testing
    * feature configuration wasn't being passed forward into the secondary opendp crate
    * therefore the secondary opendp crate re-introduces dependencies on openssl and rug
    * add cfg markup to all tests throughout the library that depend on untrusted
* expanded the Development Environment documentation
* fixes #400

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -30,7 +30,10 @@ jobs:
           restore-keys: ${{ runner.os }}-cargo-
 
       - name: Build
-        run: cargo build --verbose --features=untrusted,ffi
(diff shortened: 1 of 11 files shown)


andrewvyrros (approved):
Thanks for figuring this out and rationalizing everything!

PR #472, opened 2022-06-14
feat: rust binary search utilities

### LINKED ISSUE #471: Binary search utilities in Rust
We have binary search utilities in Python, but not in Rust. It would be ideal to have symmetry in our interfaces across languages.

Shoeboxam:
Closes #471

### DIFF
--- a/python/src/opendp/mod.py
+++ b/python/src/opendp/mod.py
@@ -1443,6 +1443,47 @@ def binary_search_param(
 
     return binary_search(lambda param: make_chain(param).check(d_in, d_out), bounds, T)
 
+
+def _call_rust_search(
+    predicate: Callable[[float], bool],
+    T: "RuntimeType | str",
+    search: Callable[[Callable[[float], bool]], Any],
+    bounds: tuple[float, float] | None = None,
+) -> Any:
+    first_exception = None
+    had_success = False
+
+    def wrapped(arg):
+        nonlocal first_exception, had_success
+        try:
+            result = predicate(arg)
+        except Exception as err:
+            if first_exception is None:
+                first_exception = err
+            raise
+        else:
+            had_success = True
+            return result
+
+    try:
+        return search(wrapped)
+    except OpenDPException as err:
+        if first_exception is not None:
+            exc = first_exception
+            if bounds is None and not had_success and hasattr(exc, "add_note"):
+                center = 0.0 if T in {"f32", "f64"} else 0
+                exc.add_note(
+                    "Predicate in binary search always raises an exception. "
+                    f"This exception is raised when the predicate is evaluated at {center}."
+                )
+            raise exc
+        if err.variant == "Search":
+            raise ValueError(err.message)
+        if err.variant == "Type":
+            raise TypeError(err.message)
+        raise err
+
+
 # when return sign is false, only return float
 @overload
 def binary_search(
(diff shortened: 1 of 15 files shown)


Shoeboxam:
* **#472** <a href="https://app.graphite.com/github/pr/opendp/opendp/472?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opendp/opendp/472?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rust/src/utilities/search/mod.rs
```diff
@@ -0,0 +1,356 @@
...
+macro_rules! impl_binary_searchable_int {
+    ($($ty:ty),+ $(,)?) => {
+        $(impl BinarySearchable for $ty {
+            fn midpoint(lower: &Self, upper: &Self) -> Self {
+                if lower < &<$ty>::zero() && upper >= &<$ty>::zero() {
+                    (lower + upper).halve()
+                } else {
+                    lower + (upper - lower).halve()
+                }
+            }
```
mccalluc, line 42: Add a comment to explain why this is necessary.

rust/src/utilities/search/mod.rs
```diff
@@ -0,0 +1,356 @@
...
+    signed_binary_search(predicate, bounds).map(|(value, _sign)| value)
+}
+
+/// Like [`binary_search`], but also returns the direction away from the decision boundary.
+///
+/// A returned sign of `1` means the passing side is above the boundary, and `-1` means it is below.
+pub fn signed_binary_search<T>(
+    predicate: impl Fn(&T) -> bool,
+    bounds: Option<(T, T)>,
+) -> Fallible<(T, i8)>
```
mccalluc, line 78: How much parallelism should I expect between this and the python version? 

In python, there's an extra parameter `return_sign` to indicate whether the result includes this information: I think having a separate function like you have here is a better interface: The variability in the return type in python adds complexity.

Perhaps a follow-up issue to bring that in line with this?

rust/src/utilities/search/mod.rs
```diff
@@ -0,0 +1,356 @@
...
+                if sign > 0 {
+                    if let Some(next) = center.checked_add(1) {
+                        bands.push(next);
+                    }
+                } else if let Some(next) = center.checked_sub(1) {
+                    bands.push(next);
+                }
+
+                for k in 1..=8 {
+                    let offset = match <$ty>::try_from(65_536_i128 * k as i128) {
```
mccalluc: Replace with power of 2, to avoid typos?

rust/src/utilities/search/mod.rs
```diff
@@ -0,0 +1,356 @@
...
+                    return Vec::new();
+                }
+
+                let mut bands = vec![center];
+                if let Some(next) = center.checked_add(1) {
+                    bands.push(next);
+                }
+
+                for k in 1..=8 {
+                    let offset = match <$ty>::try_from(65_536_u128 * k as u128) {
```
mccalluc: ditto.

rust/src/utilities/search/test.rs
```diff
@@ -0,0 +1,72 @@
+use super::*;
+
+#[test]
+fn test_binary_search_matches_python_r_int_behavior() -> Fallible<()> {
```
mccalluc: What's the "r"?

rust/src/utilities/search/test.rs
```diff
@@ -0,0 +1,72 @@
...
+    assert_eq!(binary_search::<i32>(|x| *x <= -5, None)?, -5);
+    assert_eq!(binary_search::<i32>(|x| *x <= 5, None)?, 5);
+    assert_eq!(binary_search::<i32>(|x| *x >= -5, None)?, -5);
+    assert_eq!(binary_search::<i32>(|x| *x >= 5, None)?, 5);
+    assert_eq!(binary_search::<u32>(|x| *x > 5, None)?, 6);
```
mccalluc: This last isn't one of the python tests. Add to python tests, or explain why rust is different?

rust/src/utilities/search/test.rs
```diff
@@ -0,0 +1,72 @@
...
+fn test_exponential_bounds_search_matches_python_r_bands() {
+    assert_eq!(
+        exponential_bounds_search::<i32>(&|x| *x > 5),
+        Some((1, 65_536))
+    );
+    assert_eq!(
+        exponential_bounds_search::<f64>(&|x| *x > 5.0),
+        Some((2.0, 16.0))
+    );
+}
```
mccalluc, line 43: I don't see a corresponding python test?

mccalluc (commented):
Some comments, but they are superficial. Bigger questions:
- Why now? Is something you are planning to use soon?
- Since this will be in the rust core, can you remove the python implementation? (Several of the comments are about python-rust parallels, and if we don't have a duplicate implementation in python, those go away.)
- Can the utilities also be generated for R?
- This feels like a really good target for https://hypothesis.readthedocs.io/ style tests. Probably not in this PR, but it just feels like there is lots of details in the code, and we can't anticipate the edge cases.

python/src/opendp/mod.py
```diff
@@ -1443,6 +1443,47 @@ def binary_search_param(
 
     return binary_search(lambda param: make_chain(param).check(d_in, d_out), bounds, T)
 
+
+def _call_rust_search(
+    predicate: Callable[[float], bool],
+    T: "RuntimeType | str",
+    search: Callable[[Callable[[float], bool]], Any],
+    bounds: tuple[float, float] | None = None,
+) -> Any:
```
mccalluc, line 1452: A little surprised by the two `Any`: consider using an `assert` to constrain the type returned by rust?

python/src/opendp/mod.py
```diff
@@ -1443,6 +1443,47 @@ def binary_search_param(
...
+        nonlocal first_exception, had_success
+        try:
+            result = predicate(arg)
+        except Exception as err:
+            if first_exception is None:
+                first_exception = err
+            raise
+        else:
+            had_success = True
+            return result
```
mccalluc, line 1466:
```suggestion
        had_success = True
        return result
```
Equivalent? but maybe the `else` is more clear?

python/test/test_binary_search.py
```diff
@@ -76,7 +77,36 @@ def predicate(_):
...
+
+    with pytest.raises(ValueError, match="predicate is constant"):
+        _call_rust_search(lambda _: True, "i32", search)
+
+
+def test_binary_search_forwards_untranslated_rust_errors():
+    def search(_):
+        raise OpenDPException("FailedFunction", "callback failed")
+
+    with pytest.raises(OpenDPException, match='FailedFunction\\("callback failed"\\)'):
```
mccalluc, line 101: consider `re.escape` or
```suggestion
    with pytest.raises(OpenDPException, match=r'FailedFunction\("callback failed"\)'):
```

mccalluc (approved):
Thanks. Looks good. A couple questions, but they shouldn't block.

PR #477, opened 2022-06-16
prepare traits/, dom.rs, dist.rs for proofs

Shoeboxam:
This PR adjusts the file structure such that modules are located in singleton folders.
This is to facilitate adapting the [proof definitions](https://github.com/opendp/whitepapers/pull/6) file and [pseudocode definitions](https://github.com/opendp/whitepapers/pull/4) file to the core OpenDP lib.


rust/src/core/mod.rs
```diff
@@ -20,19 +20,27 @@
...
 // DI, DO, MI, MO, TI, TO, QI, QO
 
+mod domains;
+pub use domains::*;
+
+mod metrics;
+pub use metrics::*;
+
+mod measures;
+pub use measures::*;
```
andrewvyrros, line 30: I'm not crazy about having these inside `core`. Feels like that's putting too much in there, especially if they're re-exported.

andrewvyrros (commented):
This is an improved structure, something we should've done from the beginning.

On first impression, I'm not a big fan of putting `domains`, `metrics`, and `measures` inside `core`, especially if they're re-exported. It feels like that makes `core` too much of a grab-bag. I think they're better as separate top-level modules. I'm open to convincing otherwise though!

rust/src/core/mod.rs
Shoeboxam: Moved it back out!

rust/src/trans/sum/float/sequential/mod.rs
```diff
@@ -0,0 +1,121 @@
+use num::{One, Zero};
```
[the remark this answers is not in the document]
Shoeboxam: How did these get in here?

rust/src/core/mod.rs
andrewvyrros: Thanks!

andrewvyrros (approved):
Woot!

PR #478, opened 2022-06-16
split samplers.rs into smaller modules in traits/

Shoeboxam:
This is to facilitate proof-writing for the sampler traits.


andrewvyrros (approved):
More good cleanup!

PR #481, opened 2022-06-23
fix `make_randomized_response_bool`

### LINKED ISSUE #480: `make_randomized_response_bool` fix
For randomized response over booleans we want to calibrate epsilon as such:

```eps = ln(P(X|in=X)/P(!X|in=X)) = ln(p/(1-p)``` for any X in {true, false}.

However, the algorithm is implemented such that `P(X|in=X) = p + (1-p)p`, because the lie could also be X. The correct implementation samples uniformly from all *other* categories, which in the case of the bool is just `{!X}`.

The correct implementation would be:
```Ok(if bool::sample_bernoulli(prob, constant_time)? { *arg } else { !arg })```

With this implementation, `P(X|in=X) = p` and `P(!X|in=X) = 1 - p`.

### PR DESCRIPTION
**Pull request by Shoeboxam**
Closes #480.

### CODE REVIEW

File: rust/src/traits/samplers/bernoulli/mod.rs [hunk 1]
```diff
@@ -56,37 +59,83 @@ pub trait SampleBernoulli<T>: Sized {
...
+            }
+        };
+
+        // Step 2. index into the binary expansion of prob at first_heads_index to get b_i
 
         // number of leading zeros in binary representation of prob
-        //    cast is non-saturating because exponent only uses first 11 bits
-        //    exponent is bounded in [0, EXPONENT_PROB] by check for valid probability and one check
-        let num_leading_zeros = T::EXPONENT_PROB - prob.exponent();
+        //    exponent is bounded in [0, EXPONENT_BIAS - 1] by check for valid probability and one check
```
andrewvyrros, line 120: The language in this comment is a bit tough to parse. Maybe say "exponent is bounded in [0, EXPONENT_BIAS - 1] by check for valid probability, and by returning when prob == 1"

andrewvyrros (commented):
This one grew a bit, but it ended up with a lot of nice improvements. I like the reworking of FloatBits.

Thanks for walking me through the samplers!

andrewvyrros (approved):

PR #484, opened 2022-07-03
fix infinite loop in `make_base_analytic_gaussian`

### LINKED ISSUE #429: Analytic gaussian mechanism suffers from infinite loop
Specifically when epsilon is zero, the algorithm enters an infinite loop.

Minimal reproduction:
```python
from opendp.meas import make_base_analytic_gaussian
enable_features('floating-point', 'contrib')

make_base_analytic_gaussian(1.).check(1., (0., 1e-6))
```

Shoeboxam:
Closes #429

### DIFF
--- a/python/src/opendp/meas.py
+++ b/python/src/opendp/meas.py
@@ -94,15 +94,16 @@ def make_base_gaussian(
 
 
 def make_base_analytic_gaussian(
-    scale,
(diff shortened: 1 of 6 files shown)


andrewvyrros (changes requested):
Looks good, but `test_count` is failing on `u64` for some reason.

andrewvyrros (approved):

Issue #155, opened 2021-06-24
Count By (Categories) Known N -- Implementation


Shoeboxam:
I noticed that the HammingDistance implementation was really a restricted sensitivity proof for known N, and came up with a tighter sensitivity. There needs to be another set of constructors, or an optional arg, for count by and count by categories that take N and operate on sized domains.

andrewvyrros:
Issue for proof: #158.

Shoeboxam:
Initial implementation is in https://github.com/opendp/opendp/tree/155-sized-count-by
- Renamed `make_count_by` to `make_sized_count_by`
- Added a QI parameter to `make_base_geometric`. It defaults to i32.
- When you chain `sized_count_by_categories` with `make_base_geometric`, you have to specify `QI=float`

This has the same sensitivity as the unsized count by categories when distances are in L1 space. Since all other integer aggregators have integer sensitivities, I think I should not add QI to `base_geometric`, and not implement the sized counter for L1 spaces. It also doesn't make sense to open a PR for this until the discrete gaussian is implemented. We could cast the integer counts to floats and chain with the gaussian, but I'd rather not run continuous samplers on discrete data if we don't have to. Therefore this is blocked on #191.

raprasad:
Will be closed by Polars work

Shoeboxam:
Closed by #1540

PR #487, opened 2022-07-04
shorten where clauses with summary traits

### LINKED ISSUE #485: introduce shorthand traits to condense lengthy where clauses
It's become common for interfaces to be littered with many trait bounds, especially due to the Inf* traits. I'd like to cut down on this by introducing some higher level traits, like DistanceConstant did, that inherit multiple traits. This will make the interfaces less verbose, but will require additional unnecessary trait bounds to be satisfied. I don't think these extra trait bounds will cause a problem, because in most cases it is unlikely that another type exists that will meet the minimal set of trait bounds.

```
trait Number: TotalOrd + InfMul + InfDiv + ... {}
impl<T: TotalOrd + InfMul + InfDiv + ...> Number for T {}

trait Integer: Number + ... {}
impl<T: Number + ...> Integer for T {}

trait Float: Number + num::Float + InfLog1P + FloatBits + ... {}
impl<T: Number + num::Float + InfLog1P + FloatBits + ...> Float for T {}
```

A common where clause may look like this:
```
where
    T: DistanceConstant<IntDistance>
        + Zero
        + One
        + ExactIntCast<usize>
        + ExactIntCast<IntDistance>
        + ExactIntCast<T::Bits>
        + InfAdd
        + InfMul
        + InfDiv
        + InfPow
        + FloatBits
        + AlertingAbs
        + SaturatingAdd
        + CheckNull,
```

which would be shortened as such:

```
where T: Float
```

Shoeboxam:
Closes #485.

This also changes the L2 sensitivity of `make_count_by` to 1. The `sqrt(2)/2` sensitivity is only valid when the dataset is sized. [First discovered in this comment/branch](https://github.com/opendp/opendp/issues/155#issuecomment-931617479). We don't actually have any mechanisms that use the adjusted L2 sensitivity at this time.


andrewvyrros (approved):
This really helps make some of those big trait bounds more manageable.

PR #490, opened 2022-07-05
Binary Randomized Response

matchaginseng:
`make_randomized_response_bool` is a function that returns the truthful dataset value `arg` with probability denoted by input variable `prob` and the false value `!arg` with probability `1 - prob`. It has the option to run in `constant_time`. The code can be found in this [Github permalink](https://github.com/opendp/opendp/blob/main/rust/src/meas/randomized_response/mod.rs). 

Files:
* `binary-rr-proof.tex` -- the LaTeX document with the proof.
* `binary-rr-proof.pdf` -- the PDF generated from the TeX document. 

TODO: proof for `sample_bernoulli` function


Shoeboxam:
Two minor nits to get the ball rolling:
1. Could you please remove the .pdf file? I think it would be best to keep binary files out of the repo. We can re-assess further down the line.
2. Could you move the .tex proof file to the `/rust/src/meas/randomized_response` folder?

Exciting! Thanks for opening the PR!

rust/src/meas/randomized_response/mod.tex
```diff
@@ -0,0 +1,250 @@
...
+
+\emph{The use of \texttt{code}-style parameters in the preconditions section below (for example, \texttt{input\_domain}) means that this information should be passed along to the \texttt{Measurements} constructor.}
+
+\subsubsection*{Preconditions}
+To ensure the correctness of the output, we require the following preconditions:
+
+\begin{itemize}
+    \item \textbf{User-specified types:}
+    \begin{itemize}
+        \item Variable \texttt{prob} must be of type \texttt{Q}
```
SalilVadhan, line 119: I was unable to find the definition of type Q in the pseudocode defs doc (though it is used a lot in the metrics section of that doc)

rust/src/meas/randomized_response/mod.tex
```diff
@@ -0,0 +1,250 @@
...
+        \item Variable \texttt{constant\_time} must be of type \texttt{bool}
+        \item Type \texttt{bool} must have trait \texttt{SampleBernoulli<Q>}
```
SalilVadhan, line 121: I didn't find the def of the trait SampleBernoulli<Q> in the pseudocode defs doc (though I can guess what it means)

rust/src/meas/randomized_response/mod.tex
```diff
@@ -0,0 +1,250 @@
...
+    output_domain = AllDomain(bool)
+    input_metric = DiscreteMetric()
+    similarity_metric = MaxDivergence()
+    
+    if (prob < 0.5 or prob >= 1): |\label{line:range}|
+        raise Exception("probability must be in [0.5, 1)")
+
+    c = inf_ln(inf_div(prob, neg_inf_sub(1, prob)))
+    def privacy_map(d_in: u_32) -> u_32: |\label{line:map}|
+        return d_in * c;
```
SalilVadhan, line 145: is there implicit casting here?  Also, I think it would be nicer to do this as "return 0 if d_in = 0, else return c" (see also my suggested edit to def of DiscreteMetric in the pseudocode defs doc)

rust/src/meas/randomized_response/mod.tex
```diff
@@ -0,0 +1,250 @@
...
+
+    def function(arg : bool) -> bool: |\label{line:fn}|
+        if (sample_bernoulli(prob, constant_time)): |\label{line:coinflip}|
+            return arg
+        else:
+            return !arg
+
```
SalilVadhan, line 152: To avoid the control-flow timing channel, might it be nicer to just return "not(arg XOR sample_bernoulli(prob,constant_time))"?

rust/src/meas/randomized_response/mod.tex
```diff
@@ -0,0 +1,250 @@
...
+    %\item \textup{(Appropriate output domain.)} For every element $v$ in \texttt{input\_domain}, \texttt{function}(v) is in \texttt{output\_domain}. 
+    %\item \textup{(Domain-metric compatibility.)} The domain \texttt{input\_domain} matches one of the possible domains listed in the definition of \texttt{input\_metric}, and the domain \texttt{output\_domain} matches one of the possible domains listed in the definition of \texttt{output\_metric}.
+    \item  \textup{(Domain-metric compatibility.)} The domain \texttt{input\_domain} matches one of the possible domains listed in the definition of \texttt{input\_metric}. 
+    
+    \item \textup{(Privacy guarantee.)} Let \texttt{d\_in} be the associated metric on \texttt{input\_domain} and has the associated type for \texttt{input\_metric}, and let $D$ be the similarity measure on probability distributions with the associated type for \texttt{similarity\_metric}. For every pair of elements $v, w$ in \texttt{input\_domain} and every \texttt{d\_in}, if $v, w$ are \texttt{d\_in}-close under \texttt{input\_metric}, then \texttt{function}(v), \texttt{function}(w) are \texttt{privacy\_map(d\_in)}-close with respect to $D$.
```
SalilVadhan, line 173: I suggest the following rephrasing "Let \texttt{d\_in} have the associated type for \texttt{input\_metric}, and let $D$ have associated type for \texttt{similarity\_metric}."

rust/src/meas/randomized_response/mod.tex
```diff
@@ -0,0 +1,250 @@
...
+\begin{note}[Proof relies on correctness of Bernoulli sampler]
+The following proof makes use of the following lemma that asserts the correctness of the Bernoulli sampler function.
+    \begin{lemma}
+    \texttt{sample\_bernoulli(prob, constant\_time)}, the Bernoulli sampler function used in \texttt{make\_randomized\_response\_bool}, returns \texttt{true} with probability (\texttt{prob}) and returns  \texttt{false} with probability (1 - \texttt{prob}).
+    \end{lemma}
+\end{note}
+\end{tcolorbox}
+    \vicki{to do: need to relax the epsilon-delta defns.} 
+    
+    Let $v$ and $w$ be datasets that are \texttt{d\_in}-close with respect to \texttt{input\_metric}. Here, the metric is \texttt{DiscreteMetric} which enforces that $\din = 1$ if $v \ne w$ and $\din = 0$ if $v = w$. The case where $v = w$ is trivial so we only consider $v \ne w$ and assume without loss of generality that $v = \texttt{true}$ and $w = \texttt{false}$. For shorthand, we let $p$ represent \texttt{prob}, the probability that \texttt{sample\_bernoulli} returns \texttt{true}. Observe that $p = [0.5, 1.0)$ otherwise \texttt{make\_randomized\_response\_bool} raises an error.
```
SalilVadhan, line 215: Suggested edit "which enforces that $\din \geq 1$ if $v \ne w$"

SalilVadhan: Also I wouldn't say that the case is $v=w$ - briefly state why this case is fine.  If $v=w$, then the output distributions on $v$ and $w$ are identical, and therefore the max-divergence is 0.

rust/src/meas/randomized_response/mod.tex
```diff
@@ -0,0 +1,250 @@
...
+    \texttt{sample\_bernoulli(prob, constant\_time)}, the Bernoulli sampler function used in \texttt{make\_randomized\_response\_bool}, returns \texttt{true} with probability (\texttt{prob}) and returns  \texttt{false} with probability (1 - \texttt{prob}).
+    \end{lemma}
+\end{note}
+\end{tcolorbox}
+    \vicki{to do: need to relax the epsilon-delta defns.} 
+    
+    Let $v$ and $w$ be datasets that are \texttt{d\_in}-close with respect to \texttt{input\_metric}. Here, the metric is \texttt{DiscreteMetric} which enforces that $\din = 1$ if $v \ne w$ and $\din = 0$ if $v = w$. The case where $v = w$ is trivial so we only consider $v \ne w$ and assume without loss of generality that $v = \texttt{true}$ and $w = \texttt{false}$. For shorthand, we let $p$ represent \texttt{prob}, the probability that \texttt{sample\_bernoulli} returns \texttt{true}. Observe that $p = [0.5, 1.0)$ otherwise \texttt{make\_randomized\_response\_bool} raises an error. 
+    
+    We now consider the max-divergence $D_{\infty}(Y||Z)$ over the random variables $Y = \function(v)$ and $Z = \function(w)$.
+
```
SalilVadhan, line 218: Throughout the sequence of equations below, use \left and \right for readability.

rust/src/meas/randomized_response/mod.tex
```diff
@@ -0,0 +1,250 @@
...
+    \item \textbf{(Domain-metric compatibility.)} For \texttt{binary\_randomized\_response}, this corresponds to showing \texttt{AllDomain(bool)} is compatible with \texttt{DiscreteMetric}. This follows directly from the definition of \texttt{DiscreteMetric}, as stated in the \href{https://github.com/opendp/whitepapers/blob/pseudocode-defns/pseudocode-defns/pseudocode_defns.pdf}{``List of definitions used in the pseudocode"}.
+    
+    \item \textbf{(Privacy guarantee.)} 
+    % We assume proof of correctness of the \texttt{bernoulli\_sampler} as shown in the document ??? (to do). 
+    
+    \begin{tcolorbox}
+\begin{note}[Proof relies on correctness of Bernoulli sampler]
+The following proof makes use of the following lemma that asserts the correctness of the Bernoulli sampler function.
+    \begin{lemma}
+    \texttt{sample\_bernoulli(prob, constant\_time)}, the Bernoulli sampler function used in \texttt{make\_randomized\_response\_bool}, returns \texttt{true} with probability (\texttt{prob}) and returns  \texttt{false} with probability (1 - \texttt{prob}).
```
SalilVadhan, line 209: Can sample_bernoulli fail?  (If so, the lemma would need to incorporate this, and the use of the lemma would have to take this into account.) Also I'd move the reliance on this lemma to the Warning section, since we don't yet have a proof of the lemma.

rust/src/meas/randomized_response/mod.tex
matchaginseng: @Shoeboxam what exactly is the difference between types T and Q? are they both "generic types"?

Shoeboxam: They are both generic types. These are just naming conventions:

```
- D for Domain
- M for Metric or Measure
- T for D::Carrier
- Q for M::Distance
```
The above is ordered by precedence, and we don't introduce redundant types. If there is a constructor where T == Q, then it will only have one generic named T (because T has higher precedence), and a generic Q won't exist in the code. 

If T != Q, then you'll see both generics in use, like in `make_randomized_response`.

Similarly if we have a generic D, then there won't also be a generic T because T is redundant with D::Carrier. So D has higher precedence.

Additionally:
```
- *I for Input. Like TI or MI
- *O for output
- **A for Atom. Like TIA or DA
```

rust/src/meas/randomized_response/mod.tex
Shoeboxam: The function returns errors that are not data-dependent. The primary (and unlikely) cause of this is openssl failing due to lack of system entropy.

matchaginseng: Should we still account for this in the proof? Or is it ok to just mention as a caveat that `sample_bernoulli` could fail due to OpenSSL failing because of lack of system entropy?

Shoeboxam: This was something Salil brought up in the meeting on Thursday. The proof should include failure as a possible output of the mechanism, and first show that releasing a failure state is independent from the dataset. Then analyze the true/false releases as a conditional probability, conditioned on the assumption that failure is not released.

Shoeboxam:
I rebased this and updated it to be consistent with other proofs. This is ready for another round of review.

Shoeboxam:
* main
  * **PR #490** <a href="https://app.graphite.dev/github/pr/opendp/opendp/490" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/opendp/opendp/490?utm_source=stack-comment).

PR #491, opened 2022-07-05
prove Pseudocode defns

matchaginseng:
Files:

* `pseudocode_defns.tex` -- LaTeX document with definitions
* `pseudocode_defns.pdf` -- PDF generated from `pseudocode.defns.tex`

This upload contains definitions of terms that appear in the pseudocode sections of the proofs we have written. As we write more proofs in the future, more terms will be added as they come up. Much of the content is inherited from last year's interns, Sílvia, Grace, and Connor. 

During the definition creation process, some questions arose. They are noted in the tex file using the format "\questionr{...}". The questions also appear in the pdf file generated from the tex file; they are displayed in red text and are prepended by "Question for reviewers:" in blue text.

Feel free to reach out here or on Slack with any questions. Thanks!

Note: let us know where you want this in the repo!


Shoeboxam:
I think the file structure here should get the same treatment as the randomized response proof PR-- but it will need to be split up. There's a section on domains, it can go in `core/domains/mod.tex`. Similarly metrics and measures in `core/metrics/mod.tex` and `core/measures/mod.tex`. Traits and samplers into `/traits/...` Any common code goes into an include file to avoid duplication. 

Looking through this file, a number of the pseudocode definitions duplicate traits, like InfLn and inf_ln. I really want to avoid this kind of duplication. How about the trait definition lists the functions it provides, so that both definitions are in the same place? This should pare the functions section down significantly.

There are still two categories of functions that aren't supplied by traits:
1. impl's on the type itself. An example of this [is `is_some`](https://doc.rust-lang.org/std/option/enum.Option.html#method.is_some).
2. python builtins like `len`, `map`. 

For now, it seems like we'll still have a separate pseudocode definitions document that ties these functions to rust documentation.

proofs/pseudocode-defns/pseudocode_defns.tex
```diff
@@ -0,0 +1,797 @@
...
+\begin{itemize}
+    \item an associated type \texttt{Imputed}. \texttt{VectorDomain::Imputed} is the data type after imputation. 
+    \item an imputation function \texttt{impute\_constant} that replaces a null value with a constant or passes a non-null value through. The pseudocode is \\
+    \texttt{def impute\_constant(x: DA) -> DA::Imputed: \\ return constant if x.is\_null else x}
+    \item  a function \texttt{is\_null} to check if a value is \texttt{null}
+    \item a function \texttt{new} to construct an instance of the domain
+\end{itemize}
+\end{definition}
+\begin{definition}[\texttt{InfCast}]
+A type \texttt{TO} has trait \texttt{InfCast(TI)} $\iffText$ for every value \texttt{val} of type \texttt{TI}, \texttt{inf\_cast(val:TI,TO)} is defined (note in definition \ref{defn:inf-cast} for \texttt{inf\_cast} that it is acceptable for \texttt{inf\_cast(val:TI,TO)} to return an error).
```
SalilVadhan, line 306: this phrasing is not clear.  is returning an error consistent with being "defined" or not?

proofs/pseudocode-defns/pseudocode_defns.tex
```diff
@@ -0,0 +1,797 @@
...
+\textbf{Example:} \texttt{SymmetricDistance} has a domain of \texttt{VectorDomain(AllDomain(T))}, which means that \texttt{SymmetricDistance} can be used to measure the distance between any objects that are vectors of elements of type \texttt{T}. \texttt{SymmetricDistance} has an associated type of \texttt{u32}, which means that a \texttt{u32} value is used to report the distance.
+Please refer to the proof definitions document (see section \ref{sec:versioned-docs}) for definitions of the functions referred to by the metrics below.
+\note{Once the subdomains have been implemented in OpenDP, it will no longer be necessary to list all of the subdomains in the compatible domains section.}
+\begin{definition}[\texttt{AbsoluteDistance(T)}]
+\metricDefn{absolute distance}
+\begin{itemize}
+    \item \textbf{Domain:} \texttt{AllDomain(T)}, where \texttt{T} has the traits \texttt{Sub(Output=T)} (defined in definition \ref{defn:trait-sub}) and \texttt{TotalOrd} (defined in definition \ref{defn:trait-totalord}).
+    \item \textbf{Associated type:} \texttt{Q}.
```
SalilVadhan, line 694: Here Q is not quantified/defined.  I understand that this Q should be a T.  Should also check in the definitions below.

proofs/pseudocode-defns/pseudocode_defns.tex
```diff
@@ -0,0 +1,797 @@
...
+
+\questionr{Are the domain and associated type correct? Are any further traits needed? Is this definition correct?} 
+
+\end{definition}
+
+
+\begin{definition}[DiscreteMetric]
+\metricDefn{discrete metric}
+\begin{itemize}
+    \item \textbf{Domain:} \texttt{AllDomain(T)}
```
SalilVadhan, line 737: Why not allow DiscreteMetric on every domain (not just AllDomains)?

SalilVadhan: Maybe this should be DiscreteMetric<T> or DiscreteMetric<DI> if we allow it to be on every domain?

proofs/pseudocode-defns/pseudocode_defns.tex
```diff
@@ -0,0 +1,797 @@
...
+    \item \textbf{Associated type:} \texttt{Q}.
+    \item \textbf{$d$-close:} For any two elements $u, v$ in \texttt{AllDomain(T)}, where \texttt{T} denotes an arbitrary type, and $d$ of generic type \texttt{u32}, we say that $u, v$ are $d$-close under the discrete metric (abbreviated as $d_{DM}$) when
```
SalilVadhan, line 739: The use of u32 here was what made me think that perhaps Q is a fixed type like u32.  I assume d should be of type Q?

SalilVadhan: I'd say that u,v are d-close if "u\neq v -> d>=1".

Shoeboxam:
I think the scope of this document should be reduced.

* `Types` are redundant with Rust documentation. I think we should not try to duplicate Rust docs 
* `Domains` have been moved into  `domains/mod.rs` and can be removed
* subdomains are implemented in #604. After this PR, we should open PRs to "prove" the `MetricSpace` trait impls as valid pairings.
* keep the mention that Clone, Copy and 'static need not be discussed in the proof
* `Functions in the pseudocode language` are supplied by traits specified in preconditions, or by the type itself
* preserve the notes about assert, list and other pseudocode-specific shorthands
* rename `Classes` to `Core Structures`, drop the pseudocode with preconditions sections, add type hints instead
    * Measurements don't have output_domain anymore
    * Add Odometer
* `Metrics` has been moved into `metrics/mod.rs` and can be removed
* `Measures` has been moved into `measures/mod.rs` and can be removed

Should also mention that arguments must adhere to type hints- they are not just hints, like they are in Python. A value's type must adhere to the type hint.

Shoeboxam:
These definitions have since been systematically integrated into the documentation itself, in proof definition sections. Keeping definiitons closer to the code helps to keep them from drifting. Thanks!

PR #492, opened 2022-07-11
QOL adjustments; python utility to specify default atomic type

### LINKED ISSUE #431: Change the default python bindings' int type from i32 to C long
When you create an integer array in numpy, it defaults to the bit depth of a C long on your system. On most systems this is an i64, meaning the default dtype of numpy integer arrays isn't the same as the default dtype on int transformations and measurements. This is just annoying to work with, because you have to set domains manually throughout the pipeline, or cast data before invocation.

Shoeboxam:
* Adds a utility function to set the default int type, and default float type in the python bindings.
    * From the `set_default_int_type and set_default_float_type` commit.
    * Closes #431
* Renames the rust function `make_resize_constant` to `make_resize`, to be consistent with python
* Adjust the FFI for `make_*resize` such that `TA` is the last type argument
* Adds a check to `RuntimeType.infer` to ensure that the lists are homogeneous
    * Prevents a potential segfault if the input list has mixed types 
* Adjusts the Debug impl for domains, metrics and measures to include the type name

### DIFF
--- a/python/src/opendp/typing.py
+++ b/python/src/opendp/typing.py
@@ -57,7 +57,7 @@ RuntimeTypeDescriptor = Union[
 
 def set_default_int_type(T: RuntimeTypeDescriptor):
     """Set the default integer type throughout the library.
-    This function is particularly when building computation chains with constructors.
+    This function is particularly useful when building computation chains with constructors.
     When you build a computation chain, any unspecified integer types default to this int type.
 
     The default int type is i32.
@@ -72,7 +72,7 @@ def set_default_int_type(T: RuntimeTypeDescriptor):
 
 def set_default_float_type(T: RuntimeTypeDescriptor):
     """Set the default float type throughout the library.
-    This function is particularly when building computation chains with constructors.
+    This function is particularly useful when building computation chains with constructors.
     When you build a computation chain, any unspecified float types default to this float type.
 
     The default float type is f64.
--- a/python/src/opendp/typing.py
+++ b/python/src/opendp/typing.py
@@ -205,10 +205,16 @@ class RuntimeType(object):
             return RuntimeType('Tuple', list(map(cls.infer, public_example)))
 
         if isinstance(public_example, list):
-            return RuntimeType('Vec', [
-                cls.infer(public_example[0]) if public_example else UnknownType(
-                    "cannot infer atomic type of empty list")
-            ])
+            if public_example:
+                inner_type = cls.infer(public_example[0])
+                for inner in public_example[1:]:
+                    other_type = cls.infer(inner)
+                    if other_type != inner_type:
+                        raise TypeError(f"vectors must be homogeneously typed, found {inner_type} and {other_type}")
+            else:
+                inner_type = UnknownType("cannot infer atomic type of empty list")
+
+            return RuntimeType('Vec', [inner_type])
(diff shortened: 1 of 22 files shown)


python/src/opendp/typing.py
```diff
@@ -55,6 +55,37 @@
 ]
 
 
+def set_default_int_type(T: RuntimeTypeDescriptor):
+    """Set the default integer type throughout the library.
+    This function is particularly when building computation chains with constructors.
```
andrewvyrros, line 60: "particularly useful"?

python/src/opendp/typing.py
```diff
@@ -55,6 +55,37 @@
...
+    equivalence_class = ATOM_EQUIVALENCE_CLASSES[ELEMENTARY_TYPES[int]]
+    assert T in equivalence_class, f"T must be one of {equivalence_class}"
+
+    ATOM_EQUIVALENCE_CLASSES[T] = ATOM_EQUIVALENCE_CLASSES.pop(ELEMENTARY_TYPES[int])
+    ELEMENTARY_TYPES[int] = T
+
+
+def set_default_float_type(T: RuntimeTypeDescriptor):
+    """Set the default float type throughout the library.
+    This function is particularly when building computation chains with constructors.
```
andrewvyrros, line 75: "particularly useful"?

python/test/test_comb.py
```diff
@@ -2,7 +2,7 @@
 from opendp.mod import enable_features
 from opendp.meas import *
 from opendp.trans import *
-from opendp.typing import AllDomain, L1Distance, VectorDomain
+from opendp.typing import AllDomain, L1Distance, VectorDomain, set_default_int_type
```
andrewvyrros, line 5: Spurious import of `set_default_int_type`?

python/test/test_trans.py
```diff
@@ -229,6 +229,7 @@ def test_count():
     assert ret == 3
     assert transformation.check(1, 1)
 
+test_count()
```
andrewvyrros, line 232: Hard-wired test?

```suggestion
```

rust/src/ffi/util.rs
```diff
@@ -17,6 +18,13 @@ use crate::domains::{VectorDomain, AllDomain, BoundedDomain, InherentNullDomain,
 
 use super::any::{AnyMeasurement, AnyTransformation};
 
+#[cfg(not(feature="untrusted"))]
+use std::marker::PhantomData;
+#[cfg(not(feature="untrusted"))]
+pub struct Sequential<T>(PhantomData<T>);
+#[cfg(not(feature="untrusted"))]
+pub struct Pairwise<T>(PhantomData<T>);
```
andrewvyrros, line 26: This puzzled me for a minute, until I figured out why the definitions are needed here. Can we move them up to line 12, so it's more clear that this is just a fallback when we don't import the normal definitions?

andrewvyrros (changes requested):
Good stuff!

rust/src/ffi/util.rs
Shoeboxam: Same here. I was surprised to find compile errors after I messed up my feature flags.

andrewvyrros (approved):

PR #493, opened 2022-07-20
log-time geometric variates, discrete laplace sampling

### LINKED ISSUE #457: Geometric variates via binary search
We should explore sampling geometric random variates via binary search so that the sampling runs in logarithmic time instead of linear time. 

The linear time algorithm we currently have runs very slow when the scale parameter is large.  We repeatedly flip a biased coin until finding a heads. The number of failed flips is our geometric sample. 

However, we can repeatedly split the support at the expectation, by making unbiased coin flips to determine if the sample is in the lower or upper portion of the support. Continue to sample standard bernoullis until a falsey, which narrows the support to a band. I'm not sure how to sample from this band, though. One approach is to reuse our current implementation, but use rejection sampling. This reintroduces time complexity issues. 
Another approach is to continue working out midpoints of the band. The geometric distribution is nice in that it is memoryless, so cutting it at any point still produces a geometric distribution, but we are working with a censored geometric distribution. It may be that computing the midpoint of the censored geometric introduces floating-point vulnerabilities.

Shoeboxam:
Fixes #457 
See also #414 (fixes laplacian variates)

WIP

### DIFF
--- a/docs/source/user/application-structure.rst
+++ b/docs/source/user/application-structure.rst
@@ -197,7 +197,7 @@ can be used to convert the earlier discovered noise scale parameter into an accu
     ...  f"the DP estimate differs from the true value by no more than {accuracy} "
     ...  f"at a statistical significance level alpha of {alpha}, "
     ...  f"or with (1 - {alpha})100% = {(1 - alpha) * 100}% confidence.")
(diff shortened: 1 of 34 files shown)


Shoeboxam:
Closing in favor of #502.

PR #495, opened 2022-07-22
prove geometric mechanism

### LINKED ISSUE #114: `make_base_discrete_laplace` -- Proof

matchaginseng:
`make_base_geometric` sets up a function that returns a draw from a censored two-sided geometric distribution. The code can be found in this [Github permalink](https://github.com/opendp/opendp/blob/main/rust/src/meas/geometric/mod.rs).

Files:
* `mod.tex` -- the LaTeX document with the proof.

TODO: proof for sample_two_sided_geometric function

### DIFF
--- a/rust/src/measurements/geometric/mod.tex
+++ b/rust/src/measurements/geometric/mod.tex
@@ -83,7 +83,7 @@
 
 \section{Algorithm Implementation}
 \subsection{Code in Rust}
-The current OpenDP library contains the \texttt{make\_base\_geometric} function implementing geometric measurement for linear elements. This is defined in lines 48-82 of the file \texttt{geometric/mod.rs} in the Git repository (\url{https://github.com/opendp/opendp/blob/main/rust/src/meas/geometric/mod.rs#L48-L82}).
+The current OpenDP library contains the \texttt{make\_base\_geometric} function implementing geometric measurement. This is defined in lines 48-82 of the file \texttt{geometric/mod.rs} in the Git repository (\url{https://github.com/opendp/opendp/blob/main/rust/src/meas/geometric/mod.rs#L48-L82}).
 
 \vicki{is this right?} 
 In \texttt{make\_base\_geometric}, which accepts a parameter \texttt{scale} of type \texttt{QO} and a parameter \texttt{bounds} of type \texttt{Option<(D::Atom, D::Atom)>}, the function takes in a data point \texttt{arg} of type \texttt{T}, \texttt{scale} that is now \texttt{InfCast} to type \texttt{f64}, and \texttt{bounds}. This function will call \texttt{sample\_two\_sided\_geometric}, which provides a sampling from the two-sided geometric distribution centered at \texttt{arg} with scale parameter \texttt{scale} and saturated at the bounds denoted by \texttt{bounds}. The function returns the sampled element given the parameters. 
@@ -176,7 +176,7 @@ For every setting of the input parameters \texttt{constant} to \texttt{make\_bas
 \end{theorem}
 \begin{proof}
 \begin{enumerate}
-    \item \textbf{(Domain-metric compatibility.)} For \texttt{make\_base\_geometric} for linear elements, this corresponds to showing \texttt{AllDomain(T)} is compatible with \texttt{AbsoluteDivergence}. This follows directly from the definition of \texttt{AbsoluteDivergence}, as stated in the \href{https://github.com/opendp/whitepapers/blob/pseudocode-defns/pseudocode-defns/pseudocode_defns.pdf}{``List of definitions used in the pseudocode"}.
+    \item \textbf{(Domain-metric compatibility.)} For \texttt{make\_base\_geometric}, this corresponds to showing \texttt{AllDomain(T)} is compatible with \texttt{AbsoluteDivergence}. This follows directly from the definition of \texttt{AbsoluteDivergence}, as stated in the \href{https://github.com/opendp/whitepapers/blob/pseudocode-defns/pseudocode-defns/pseudocode_defns.pdf}{``List of definitions used in the pseudocode"}.
     \item \textbf{(Privacy guarantee.)} 
     % This proof assumes the correctness of \texttt{sample\_two\_sided\_geometric}, asserted by the following lemma. 
     % \begin{tcolorbox}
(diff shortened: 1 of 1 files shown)


rust/src/meas/geometric/mod.tex
```diff
@@ -0,0 +1,283 @@
...
+        if (scale == 0):
+            return QO_MAX
+        return d_in * inf_div(scale);
+
+    # these are two separate functions in original, but can discuss if we want to combine the two
+    def noise_function(arg : T, scale : QO, bounds : Optional[{T, T}]) -> Function<f64, Optional[{T, T}]>: |\label{line:noise_fn}|
+        return function(inf_cast(scale), bounds)
+        
+    def function(arg : T, scale : f64, bounds: Optional[{T, T}]) -> <D, QO>: |\label{line:fn}|
+        sample_two_sided_geometric(arg, scale, bounds)
```
Shoeboxam, line 148: I don't think you should define this function inside the constructor! It has already been defined outside of the function because of the GeometricDomain trait bound on D. 
Maybe the syntax we should use to refer to it is `D.noise_function`? Then, outside the function, define what `D.noise_function` is when D is `VectorDomain<AllDomain<T>>` and when D is `AllDomain<T>`.

Then you can build the function as follows:
```python
function = D.noise_function(scale, bounds)
```

Happy to adjust naming in the code to make this more clear.

rust/src/meas/geometric/mod.tex
```diff
@@ -0,0 +1,283 @@
...
+    return Measurement(input_domain, output_domain, function, input_metric, similarity_metric, privacy_map)
+
+\end{lstlisting}
+\hanwen{Changed the input metric to L1Distance for vector
+    (was AbsoluteDistance, but I guess AbsoluteDistance could 
+    be viewed as a special case of L1Distance with dimension 1)}
+\hanwen{re. noise function, I guess it's defined separately to account for different input domain (AllDomain and VectorDomain), but everything else looks pretty much the same to me. I would vote for using just one function and call
+sample\_two\_sided\_geometric once inside, plus some brief explanation of the domain?}
+
+\vicki{may be floating point issues with \texttt{is\_sign\_negative}}
```
Shoeboxam, line 159: I don't anticipate floating-point issues here. Checking if the sign is negative essentially amounts to checking if the first bit is set, which is always exact. I'm also ok with rejecting -0. when -0. could technically still be ok.

rust/src/meas/geometric/mod.tex
```diff
@@ -0,0 +1,283 @@
...
+
+\vicki{marked bounds with [SELF::MIN, SELF::MAX] because it's an optional type in the original Rust code, and if not specified, then in SampleTwoSidedGeometric SELF::MIN and SELF::MAX are substituted}
+
+\vicki{the current code does \texttt{d\_out.neg\_inf\_mul(scale)}, but since we're working with maps now, I changed that to \texttt{inf\_div}. worth asking if that's ok though}
```
Shoeboxam, line 163: That's what the current code does now:
https://github.com/opendp/opendp/blob/430392741ead9782304e2eda0d78384530ea1237/rust/src/meas/geometric/mod.rs#L79

It was written the way it was before because it handled zero scales without additional if-control flow. Not possible with maps, so I just added an `if`.

rust/src/meas/geometric/mod.tex
```diff
@@ -0,0 +1,283 @@
...
+ 
+ \textbf{Failure cases.} We still need to account for failure cases within the \texttt{privacy\_map} code. There are two places in which the code raises an exception:
+ 
+ \begin{enumerate}
+     \item \texttt{d\_in} is negative. The code does a check if \texttt{d\_in < 0} in line \ref{line:positivedin} and raises an error and terminates here if the statement is true. 
+     \item \texttt{inf\_div} fails. This step is only reached if \texttt{d\_in} is nonnegative. As defined in the pseudocode definitions doc, \texttt{inf\_div} throws an exception if division overflows from a 32-bit integer. 
+ \end{enumerate}
+ 
+ If none of these cases fail, \texttt{privacy\_map} returns \texttt{d\_in * inf\_div(scale)}.
+\end{enumerate}
```
Shoeboxam, line 270: Do we need to explicitly consider failure cases in the privacy map? Failed evaluation of the privacy map should imply that it is false. The only difference with returning false is that you get richer diagnostic information.

Also, it returns `d_in.inf_div(scale)` (or `inf_div(d_in, scale)` if you prefer)

rust/src/meas/geometric/mod.tex
```diff
@@ -0,0 +1,283 @@
...
+\end{proof}
+% \hanwen{In the original proof def doc, for L1Dist and other related ones, the sum goes from $i=0$ instead of 
+% $i=1$, but I don't think it makes sense to have $0$-dim 
+% vectors? The dim would start with 1, at least...}
+
+% \hanwen{question for reviewer: not sure if it's correct to go from the first line to the second like this, i.e., go from set to point}
+
+\noindent\textbf{Remark.}
```
Shoeboxam, line 278: I think we should omit "linear" throughout this document. From the perspective of this proof, it's an implementation detail of what should be a black box function.

Shoeboxam (changes requested):
This is a great start! Really excited to see this coming together.

rust/src/meas/geometric/mod.tex
```diff
@@ -0,0 +1,283 @@
...
+\date{Summer 2021}
+\begin{document}
+
+
+\maketitle
+\tableofcontents
+
+\section{Algorithm Implementation}
+\subsection{Code in Rust}
+The current OpenDP library contains the \texttt{make\_base\_geometric} function implementing geometric measurement for linear elements. This is defined in lines 48-82 of the file \texttt{geometric/mod.rs} in the Git repository (\url{https://github.com/opendp/opendp/blob/main/rust/src/meas/geometric/mod.rs#L48-L82}).
```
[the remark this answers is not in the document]
matchaginseng: re: linear elements: oops, sorry! that was an oversight, fixing now.

Shoeboxam:
Will need to update this PR until after #685. After that issue is closed, the pseudocode in this PR will simplify.
This PR also needs to be adjusted to use `lib.sty`, and use the standard proof template (Hoare Triple & Proof).

Shoeboxam:
Closing due to inactivity, outdated proof styles, refactored codebase, and overlap with existing new proofs. Thanks!

PR #497, opened 2022-07-22
parallel composition combinators and dataframe partitioner

### LINKED ISSUE #494: Parallel transformations/measurements
We want to express this in terms of product constructions:

```rust
// defined as d(x, x') = sum_i d_{inner}(x_i, x'_i), where x_i is the ith partition
struct ProductMetric<M> { _inner: PhantomData<M> }
impl<M> Metric for ProductMetric<M> {
    type Distance = M::Distance;
    ...
}

struct ProductDomain<D> { inner_domains: Vec<D> }
impl<D> Domain for ProductDomain<D> {
    type Carrier = Vec<D::Carrier>;
}
```

We need the ProductDomain because we still want to allow some variability in the domain of each partition (for instance, in dataset size). However, this design does not allow the dataset in each partition to vary in structure.

We can still allow the dataset/query to vary within each partition by applying different transformations to each partition.

Shoeboxam:
Closes #494

Adds a `ProductDomain<D>`:
```rust
struct ProductDomain<D: Domain> {
    inner_domains: Vec<D>
}
```
Members of this domain are members of the cartesian product of all the domains in inner_domains.

...and adds a `ProductMetric<M>`:
```rust
struct ProductMetric<M> {
    inner_metric: M
}
impl<M: Metric> Metric for ProductMetric<M> {
    type Distance = (M::Distance, IntDistance);
}
```

The distance type is `(k, r)`. `k` is an upper bound on the distance between any corresponding partition in neighboring datasets, and `r` is an upper bound on the number of partitions an individual may influence. The distance k is in terms of the inner metric, which can be any metric (either dataset distances or sensitivities), and could even be `ProductMetric<ProductMetric<M>>`, which expresses distances as `((M::Distance, IntDistance), IntDistance)` for nested partitionings.

It is also valid, and potentially tighter, to let k be a vector of distances. In this case, the distance type is a tuple `(k_i, r)`, where `k_i` is the upper bound on distance in partition i, and r is an upper bound on the number of partitions an individual could influence.
This PR implements the looser, simpler approach, as we don't have a clear use-case for separating distances. This can be changed internally later without affecting external APIs.

These domains and metrics are used in two new combinators, `make_parallel_transformation` and `make_parallel_composition`, that apply a list of transformations or a list of measurements to each partition.

The stability map of  `make_parallel_transformation` is `d_out = (k', r)`, where `k' = max_i t_i.map(k)` and the privacy map of  `make_parallel_composition` is `d_out = r * max_i m_i.map(k)`.
(t_i is the ith transformation, m_i is the ith measurement, and .map is the stability or privacy map)

This PR also adds a `make_partition_by` constructor to partition data. Since we don't currently have user IDs, there is no way to limit the number of partitions a user may contribute to. Thus the stability map does not bound the partition limit, letting it just be `d_in.min(output_partitions)`. This can be tightened in a later PR.

Thanks for all the help on this Shurong/Wanrong!

### DIFF
--- a/python/src/opendp/_convert.py
+++ b/python/src/opendp/_convert.py
@@ -244,7 +244,7 @@ def _vector_to_slice(val: Sequence[Any], type_name: RuntimeType) -> FfiSlicePtr:
             return ctypes.c_char_p(val.encode())
         array = (ctypes.c_char_p * len(val))(*map(str_to_slice, val))
         return _wrap_in_slice(array, len(val))
-
+    
     if isinstance(inner_type_name, RuntimeType):
         c_repr = [py_to_c(v, c_type=AnyObjectPtr, type_name=inner_type_name) for v in val]
         array = (AnyObjectPtr * len(val))(*c_repr)
--- a/python/test/test_comb.py
+++ b/python/test/test_comb.py
@@ -65,8 +65,8 @@ def test_make_basic_composition_leak():
     # supporting that AnyObject's free recursively frees children
     meas = make_basic_composition([meas])
 
-    # watch for leaked AnyObjects with 1 million i32 values
-    # memory would jump by ~40mb every iteration
+    # watch for leaked AnyObjects with 10 million i32 values
+    # memory would jump significantly every iteration
     for i in range(1000):
         print('iteration', i)
         meas([0] * 10_000_000)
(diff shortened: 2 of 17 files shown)


Shoeboxam:
We are going in a different direction with Polars now.

PR #499, opened 2022-07-22
`make_chain_tm` postprocess combinator

### LINKED ISSUE #498: Postprocess transformations
It is not necessary to add significant new machinery to represent postprocessors-- postprocessors can simply be transformations with maps and metrics stripped away, or rather, with `AgnosticMetric` as the input and output metric, associated distance type `()`, and trivial stability map: `|_| ()`.

To hide these implementation details away from library developers, use a new private constructor `make_postprocess(input_domain, output_domain, function)` to build postprocessors. 

Postprocessors can also be chained after measurements with a tm chainer. This means computation pipelines can look like: 
```T >> T >> M >> T >> T```

The tm chainer still emits a measurement. The tm chainer ignores the metrics and map on the postprocessor, instead reusing the output measure and map from the input measurement.

Shoeboxam:
Closes #498

- Adds `make_chain_tm` and the `Shr` trait impls for `>>` chaining
    - the chainer completely ignores the null metrics and null map in the postprocessor
- Adds `make_postprocess` that takes an input domain, output domain, and function
    - the constructor builds a transformation with null metrics and a null map

### DIFF
--- a/python/src/opendp/combinators.py
+++ b/python/src/opendp/combinators.py
@@ -8,6 +8,7 @@ from opendp.core import *
 __all__ = [
     "make_chain_mt",
     "make_chain_tt",
+    "make_chain_tm",
     "make_basic_composition",
     "make_population_amplification",
     "make_fix_delta",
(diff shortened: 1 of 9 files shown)


andrewvyrros (approved):
Looks good! Should come in handy in the future.

PR #506, opened 2022-07-31
rewrite `make_base_laplace` and `make_base_gaussian`

### LINKED ISSUE #414: Floating-point issue in noise samplers

Shoeboxam:
Closes #414.

This PR uses the discrete laplace and discrete gaussian noise samplers to mimic continuous noise distributions.


rust/src/meas/laplace/mod.rs
```diff
@@ -77,29 +86,38 @@ mod tests {
...
     }
 
     #[test]
     fn test_make_laplace_mechanism() -> Fallible<()> {
-        let measurement = make_base_laplace::<AllDomain<_>>(1.0)?;
+        let measurement = make_base_laplace::<AllDomain<_>>(1.0, None)?;
         let _ret = measurement.invoke(&0.0)?;
 
-        assert!(measurement.check(&1., &1.)?);
+        assert!(measurement.check(&1., &1.00001)?);
```
andrewvyrros, line 110: Could this be tighter?

rust/src/traits/mod.rs
```diff
@@ -104,9 +104,6 @@ pub trait Float:
     + InfExpM1
     + InfPow
     + InfSqrt
-    + SampleUniform
-    + SampleGaussian
-    + SampleLaplace
```
andrewvyrros, line 109: Great we can drop these!

rust/src/traits/samplers/geometric/mod.rs
```diff
@@ -195,18 +195,20 @@ pub(super) fn sample_geometric_buffer(buffer_len: usize, constant_time: bool) ->
 
 
 
-#[cfg(test)]
+#[cfg(all(test, feature="test-plot"))]
 mod test {
```
andrewvyrros, line 199: Need to remember this applies to all tests. Maybe rename the mod?

rust/src/traits/samplers/mod.rs
```diff
@@ -6,15 +6,24 @@ mod cks20;
...
 
 mod geometric;
 pub use geometric::*;
 
 mod uniform;
 pub use uniform::*;
 
+// these samplers are only accessible if the library is compiled in an explicitly unsafe way
+#[cfg(not(feature = "use-mpfr"))]
+mod vulnerable_fallbacks;
```
andrewvyrros, line 22: I like the name of this module!

rust/src/traits/samplers/discretize/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+{
+    fn sample_discrete_laplace_Z2k(shift: Self, scale: Self, k: i32) -> Fallible<Self> {
+        // integerize
+        let mut i = find_nearest_multiple_of_2k(shift.into_rational()?, k);
+
+        // sample from the discrete laplace on ℤ*2^k
+        i += sample_discrete_laplace(scale.into_rational()? >> k)?;
+
+        // postprocess! int -> rational -> T
+        Ok(Self::from_rational(x_mul_2k(i, k)))
```
andrewvyrros, line 30: Wow! So concise and elegant.

andrewvyrros (commented):
This is fantastic! So clean.

andrewvyrros (commented):
This is good to go once #505 is done.

andrewvyrros (approved):

PR #507, opened 2022-08-02
switch to rejection sampling in `SampleDiscreteLaplaceLinear`

Shoeboxam:
- removes the SampleRademacher trait
- adjusts `SampleDiscreteLaplaceLinear` to avoid some small float inaccuracy at the origin

Before, the algorithm would conservatively work out the probability of not adding noise, and then return the origin without adding noise if a bernoulli sample was positive with this probability. While all probabilities are calculated conservatively, the multiplicative difference between the probability of returning the origin or a neighbor is not exactly the same as the multiplicative differences between other shifts.

The simple solution is to sample geometric noise that starts at zero, and to reject the sample if the sign is negative and offset is zero.

### DIFF
--- a/rust/src/traits/samplers/bernoulli/mod.rs
+++ b/rust/src/traits/samplers/bernoulli/mod.rs
@@ -1,5 +1,3 @@
-use std::ops::Neg;
-
 use num::{One, Zero};
 
 use crate::{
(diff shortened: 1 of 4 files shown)


andrewvyrros (commented):
This is good to go once #506 is done.

andrewvyrros (approved):

PR #509, opened 2022-08-02
account for extended exponent range in `inf_*` functions

### LINKED ISSUE #508: Account for extended MPFR exponent range in `inf_*` functions
MPFR has a much higher upper bound on the exponent than native float types do, at the same level of precision. Because of this, `inf_*` operations may evaluate to floating-point values smaller than the smallest subnormal. To fix this, we need to convert from mpfr float types to native types in a way that rounds up (or down, respectively).

Shoeboxam:
Closes #508


andrewvyrros (commented):
Looks good, just waiting on #507.

andrewvyrros (approved):

PR #513, opened 2022-08-05
Port of make_count proof

silviacasac:
Moving the proof for the count transformation from last summer's OpenDP Whitepapers repository (https://github.com/opendp/whitepapers/tree/make_count). The TeX file is the same one.

Port of the whitepapers repository PR: https://github.com/opendp/whitepapers/pull/3

### DIFF
--- a/rust/src/transformations/count/make_count.tex
+++ b/rust/src/transformations/count/make_count.tex
@@ -17,16 +17,12 @@ The length of the vector, of type \texttt{usize}, is exactly casted to a user sp
 If the length is too large to be represented exactly by \texttt{TO}, 
 the cast saturates at the maximum value of type \texttt{TO}.
 
-\section{Vetting history}
+\subsection*{Vetting History}
 \begin{itemize}
     \item \vettingPR{513}
 \end{itemize}
 
-\section{Pseudocode}
-
-\label{sec:pseudocode}
-
-\subsection*{Preconditions}
+\section{Preconditions}
 \begin{itemize}
 
     \item \texttt{TIA} (atomic input type) is a type with trait \rustdoc{traits}{trait}{Primitive}. \texttt{Primitive} implies \texttt{TIA} has the trait bound:
@@ -48,7 +44,7 @@ the cast saturates at the maximum value of type \texttt{TO}.
     \end{itemize}
 \end{itemize}
 
-\subsection*{Implementation}
+\section{Pseudocode}
 \begin{lstlisting}[language = Python, escapechar=|]
 def make_count():
     input_domain = VectorDomain(AllDomain(TIA))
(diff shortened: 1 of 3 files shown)


rust/src/transformations/count/make_count.tex
```diff
@@ -0,0 +1,152 @@
...
+Before proceeding with proving the validity of the stability map, we provide a couple lemmas.
+
+\begin{lemma}
+    \label{dsym-sens}
+    $|\function(u) - \function(v)| \leq |\texttt{len(u)} - \texttt{len(v)}|$, 
+    where \texttt{len} is an alias for \\ \texttt{input\_domain.size}.
+\end{lemma}
+
+\begin{proof}
+    In line \ref{line:try-catch}, we know the argument to \texttt{TO.exact\_int\_cast} is non-negative and integral.
```
zachratliff, line 101: The first sentence of this lemma was not immediately apparent the first time I read it. Maybe we should call out that `exact_int_cast` accepts an argument of type `usize` (or refer back to the pre-condition of the `ExactIntCast` trait) and therefore we know that the argument is non-negative and integral.

rust/src/transformations/count/make_count.tex
```diff
@@ -0,0 +1,152 @@
...
+    Thus, in all cases, the function (from line \ref{line:try-catch}) returns a non-null value of type \texttt{TO}.
+\end{proof}
+
+\begin{proof} \textbf{(Part 2 -- domain-metric compatibility).}
+    Our \texttt{input\_metric} of \rustdoc{metrics}{struct}{SymmetricDistance} is compatible with any domain of the form \texttt{VectorDomain(inner\_domain)}, 
+    and our \texttt{input\_domain} of \\\texttt{VectorDomain(AllDomain(TIA))} is of this form. 
+    Therefore our \texttt{input\_domain} and \texttt{input\_metric} are compatible.
+
+    Our \texttt{output\_metric} of \rustdoc{metrics}{struct}{AbsoluteDistance} is compatible with any domain of the form \texttt{AllDomain(T)} where \texttt{T} has the trait \texttt{InfSub}, 
+    and our \texttt{output\_domain} of \texttt{AllDomain(TO)} is of this form and \texttt{TO} has the necessary trait.
```
zachratliff, line 88: Should the requirement that `TO` has the trait `InfSub` be listed in the pre-conditions? I know that the trait `InfSub` is implied by the trait `Number`, but perhaps we should explicitly list this as a pre-condition since we call it out here.

zachratliff (commented):
Looks good! I made a few minor comments. 

I'm a big fan of the new rustdoc links. It makes parsing the proof a million times easier!

rust/src/transformations/count/make_count.tex
```diff
@@ -0,0 +1,152 @@
...
+\section{Vetting history}
+\begin{itemize}
+    \item \vettingPR{513}
+\end{itemize}
+
+\section{Pseudocode}
+
+\label{sec:pseudocode}
+
+\subsection*{Preconditions}
```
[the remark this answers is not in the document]
Shoeboxam: I added a section at the top of the proof.

rust/src/transformations/count/make_count.tex
Shoeboxam: Good catch!

I'm conflicted on if we want to thoroughly list implied traits in the preconditions. For brevity, you could argue that they are technically unnecessary, and it would be nicer on the proof-writer to omit them completely. But for readability, and for reviewers, listing implied traits is useful.

rust/src/transformations/count/make_count.tex
Shoeboxam: I tied this reasoning to the interface of `CollectionDomain` instead!

andrewvyrros (approved):
Approving with the understanding that this is still under the contrib flag.

PR #515, opened 2022-08-05
`make_impute_constant` proof

silviacasac:
Moving the proof for the make_impute_constant from last summer's OpenDP Whitepapers repository (https://github.com/opendp/whitepapers/tree/impute_constant). The TeX file is the same one.

Port of the whitepapers repository PR: https://github.com/opendp/whitepapers/pull/8


Shoeboxam:
Converted to draft as this constructor needs to be updated to accept domain/metric, and the proof adjusted accordingly.

PR #518, opened 2022-08-09
build LaTeX in PRs

### LINKED ISSUE #331: Build Latex Documents in CI
We need a system for building latex documents as part of the build process for the documentation.
This would allow us to deploy the proofs alongside the library documentation.

Depends on #56, but development on this issue is not blocked.

A few comments:
- it should run as a separate script from `make html` or `make versions`, to make local doc builds easier
- for CI we could use a build server or cloud service to build the latex
- this could be useful: https://github.com/aslushnikov/latex-online#api

Shoeboxam:
Closes #331 

- adds a GH actions workflow that
    1. checks out the full repository
    2. uses a git command to find a list of edited .tex files in this PR
    3. if the diff is non-empty, opens a new job to build the .tex files in the diff
    4. the list of diffed .tex files is passed to the latex-action step, which builds them in-place
    5. uploads all .pdfs to a [new artifacts repository](https://github.com/opendp/artifacts)
    6. adds or updates the github-actions message with links to said .pdfs (see below)
- commits are made to the artifacts repo by the dummy OpenDP user account named `opendp-dev`
    - `opendp-dev` only has permissions to the artifacts repo
    - `opendp-dev` has an authentication token in the opendp repository's secrets
- steps 5 and 6 are handled by [this fork of an existing MIT-licensed action](https://github.com/opendp/pull-request-artifacts)

I rebased the ported whitepapers proofs on this PR, so that they would get rendered pdfs as well. Holding off to rebase the newer interns' PRs until I talk with them.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
-          python-version: 3.6
+          python-version: 3.7
           cache: pip
 
       - name: Assert version
@@ -164,7 +164,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
-          python-version: 3.6
+          python-version: 3.7
           cache: pip
 
       - name: Prepare for libs
@@ -209,7 +209,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
-          python-version: 3.6
+          python-version: 3.7
           cache: pip
 
       - name: Publish Rust crate
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -53,7 +53,7 @@ jobs:
         working-directory: python
     strategy:
       matrix:
-        python-version: [3.6]
+        python-version: [3.7]
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,12 +9,65 @@ Please keep this up to date, following the [instructions](#instructions) below.
 ## [Unreleased](https://github.com/opendp/opendp/compare/stable...HEAD)
 
 ### Added
+- Account for finite data types in aggregators based on our paper [CSVW22](https://arxiv.org/abs/2207.10635)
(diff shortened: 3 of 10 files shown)


.github/workflows/latex.yml
```diff
@@ -0,0 +1,50 @@
...
+        uses: xu-cheng/latex-action@v2
+        with:
+          root_file: ${{ needs.pre-latex-build.outputs.tex }}
+          work_in_root_file_dir: true
+
+      - name: Pull request artifacts
+        uses: Shoeboxam/pull-request-artifacts@main
+        with:
+          commit: ${{ github.event.pull_request.head.sha }}
+          glob-pattern: rust/**/*.pdf
```
andrewvyrros, line 47: Doing this with a glob pattern seems brittle, could pick up spurious files. You have the tex files in hand, can't you do this explicitly?

andrewvyrros (commented):
OK, I've come around to the necessity of putting these in an external repo. 😄

I'm a little concerned about having a custom version of the `pull-request-artifacts` action. I worry about this breaking in the future, and the project needing someone who can debug Javascript to fix things. At the very least, we should put the fork in the opendp org. But ideally, we wouldn't need a custom action.

Also, it's not ideal to have a separate `opendp-shoebot` user, as it's another set of credentials to maintain. Can this run as the original user of the PR?

Shoeboxam:
I can't commit as the original user of the PR, because I won't have account tokens to commit on someone else's behalf. I could spoof the commit information to make it look like someone else made the commit, but I still have to authenticate with the bot account. I think that gives us the worst of both worlds- still have another set of credentials, and I'd rather not spoof commit identification.

There were several changes made to the action-
1. glob star matching. I know how to work around this now and could omit this completely
2. preserving the file structure in the artifacts repo. I think there's a high likelihood of path collisions if this change isn't made
3. formatting changes in the PR comment (we could do without this)

I could open a PR to the original repo with the changes for 2.

Shoeboxam:
I moved the action into the opendp org, dropped the glob matching, and dropped some unnecessary changes. I also switched the bot account from opendp-shoebot to the existing opendp-dev account (we already use this for pypi releases). 

Given the number of changes to the pull-request-artifacts action, I'm not sure if it will be accepted upstream.

Issue #191, opened 2021-07-23
prove discrete gaussian mechanisms


andrewvyrros:
- [ ] make_vector_integer_gaussian
- [ ] make_scalar_integer_gaussian 
Should follow in similar fashion to #1320

Shoeboxam:
Closed in #2134

PR #519, opened 2022-08-13
proofs for cks20 samplers

Shoeboxam:
Dependency for #191.
Dependency for #503.


andrewvyrros (approved):
Approving with the understanding that this is still under the contrib flag. We'll need to make another pass over these for final approval, and resolution of the outdated marker.

PR #520, opened 2022-08-14
arrays and synthetic data generation

### LINKED ISSUE #337: 2DArrays
We need a representation for 2-dimensional homogeneously-typed arrays. 
Assumptions:
- row-axis represents records, column-axis features
- the number of columns is not known at compile-time

Use cases:
- used as an input to covariance
- clamp to a ball and chain with multivariate aggregators for n-D means, variances, etc.
- systems of counting queries/matrix mechanism

I have a fair amount of experience with https://crates.io/crates/ndarray from work on SmartNoise-Core, so this might be a viable approach:

```rust
struct ArrayDomain<D: Domain> {
    element_domain: D,
    columns: usize
}
impl Domain for ArrayDomain<D: Domain> {
    type Carrier = Array2<D::Carrier>;
    fn member(&self, candidate: &Self::Carrier) -> bool {
        candidate.iter().all(|v| D::member(v)) && candidate.ncols() == self.columns
    }
}
```

Open to alternatives, as always. 

I think there is more to explore in Arrow, but I don't see a homogeneous ndarray in arrow. There may be some data unloading into Array2's in situations where the [kernels](https://docs.rs/arrow/5.4.0/arrow/compute/kernels/index.html) don't support our needs.

pcsong33:
Closes #337.
Include changes for creating synthetic data pipeline. 

- created array/mod.rs containing array transformation functions
- created synthetic_data/mod.rs containing synthetic data chained transformations
- updates to domains/mod.rs to allow for RowDomain, Array2Domain, ArrayDDomain and made updates to existing domains
- generalizations for make_row_by_row


Shoeboxam:
The direction of this has changed, we're now moving forward with preparing marginals via Polars.

PR #521, opened 2022-08-14
dataframe subsetting

### LINKED ISSUE #366: Filtering -- Implementation
Many of the SmartNoise-Samples notebooks depend on some kind of filtering operation. In SNC, this was done by adding many transformations that create/manipulate a boolean vector, and then the boolean vector and data are passed into a component that filters the data. SmartNoise Core used row permutation identifiers to ensure that the data and boolean arguments consisted of data that was user-aligned. In SNC, any time a row permutation or resize took place, the permutation identifier associated with the data would change. The filter component checked that the row permutation identifiers matched. It isn't obvious to me how we do this in OpenDP.

I think the ideal input type is some type that implements `Iterable<Item=(T, bool)>`. We're getting into the same situation as the covariance constructor though, in that we don't have a way to feed data in. One approach is to add a dataframe map combinator that applies a transformation to a dataframe column, as well as a zipper selector transformation that behaves like the rust `zip` iterator and python `zip`. 

The eternal problem with zips is that the left and right args are not always the same length. I propose that instead of trying to consider what privacy means in the cases of non-equal lengths, we structurally ensure that lengths are always equivalent. The zipper operates on a dataframes, and dataframes should always have the same number of elements in each column. To keep this invariant we need to ensure that the mapper combinator only accepts transformations that don't change row-ordering or dataset length.

At first I thought we could keep the invariant by only accepting transformations on sized domains, where the input size is equal to the output size. But this still doesn't preserve row ordering, and it requires alternate versions of transformations that work on sized domains. It also doesn't work in situations where size is unknown.

Instead, the map combinator could remain private like `make_row_by_row` and itself be invoked by specialized constructors: `make_df_is_equal`, `make_df_is_lt`, etc. The specialized constructors provide a way to approve a constructor for usage on dataframes. The associated proof document would have to establish that the approved transformation does not change row ordering or row count.

Are there any other implementation ideas?

Shoeboxam:
Closes #366.

Adds three new transformation constructors:
- make_df_is_equal
- make_df_cast_default
- make_subset_by

These transformations give very simple subsetting functionality on dataframes. 
To be used by the proportion CI estimators.


rust/src/traits/cast/mod.rs
```diff
@@ -325,7 +325,7 @@ macro_rules! impl_round_cast_self_string_bool {
             }
         }
         impl RoundCast<$T> for bool {
-            fn round_cast(v: $T) -> Fallible<Self> {Ok(v.is_zero())}
+            fn round_cast(v: $T) -> Fallible<Self> {Ok(!v.is_zero())}
```
andrewvyrros, line 328: Oops!

python/test/test_trans.py
```diff
@@ -299,3 +299,28 @@ def test_lipschitz_mul_float():
...
+def test_df_is_equal():
+    from opendp.trans import make_split_dataframe, make_df_is_equal, make_select_column
+
+    query = (
+        make_split_dataframe(separator=",", col_names=["23", "17"]) >>
+        make_df_is_equal(column_name="17", value="2.") >>
+        make_select_column(key="17", TOA=bool)
+    )
+    assert query("0,0.\n1,1.\n2,2.\n3,3.") == [False, False, True, False]
+    assert query.check(1, 1)
```
andrewvyrros, line 326: How about a test for `make_subset_by`?

rust/src/trans/dataframe/subset/mod.rs
```diff
@@ -0,0 +1,49 @@
...
+                Fallible::Ok(())
+            })?;
+
+            Ok(subsetted)
+        }),
+        SymmetricDistance::default(),
+        SymmetricDistance::default(),
+        StabilityMap::new_from_constant(1),
+    ))
+}
```
andrewvyrros, line 49: Could we get a test here?

andrewvyrros (commented):
Splitting `dataframe` up into submodules was a good call, makes it more manageable. Looks good, just could use tests for `make_subset_by`.

andrewvyrros (approved):
Groovy, thanks for adding the extra tests.

PR #523, opened 2022-08-17
fix 32-bit SMD Curve downcast

### LINKED ISSUE #522: Failed downcast_ref of `SMDCurve<f32>`
The 32-bit SMD curve spews an ugly error when invoked without type parameters, because type inference defaults to f64, and the expected type is not introspected from the curve struct.

Shoeboxam:
Closes #522.

This adjusts the FFI for the SMD curve evaluation to introspect `T` from the curve struct itself. No additional FFI helpers added, invoking the map reuses `object_type` on the curve, and then retrieves the atom.

### DIFF
--- a/python/src/opendp/_data.py
+++ b/python/src/opendp/_data.py
@@ -236,8 +236,7 @@ def bool_free(
 
 def smd_curve_epsilon(
     curve: Any,
-    delta: Any,
-    T: RuntimeTypeDescriptor = None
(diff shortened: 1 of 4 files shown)


andrewvyrros (commented):
Cool, fixes an annoying nit. Will be good once #509 is done.

andrewvyrros (approved):

PR #527, opened 2022-08-22
proof for `make_zCDP_to_approxDP`

### LINKED ISSUE #524: `make_zCDP_to_approxDP` proof
Proof for #504.

Shoeboxam:
Closes #524.

Unblocks #505.


Shoeboxam:
This PR needs to be updated to use the macros and appeal more closely to the trait definitions.

andrewvyrros (approved):
Approving with the understanding that this is still under the contrib flag. We'll need to make another pass over these for final approval, and resolution of the outdated marker.

PR #528, opened 2022-08-22
fix clippy warnings

### LINKED ISSUE #489: Make clippy less unhappy
This is less of a bug than of a tidiness issue.

**Describe the bug**
Running clippy (rust linter) on the rust code yields a bunch of minor warnings. None of the warnings represent real problems, but cleaning those up will make it easier to spot future problems earlier

**To Reproduce**

```
% cd rust
% cargo clippy
```
## Fixing

Looking over what clippy complains about, the fixes are easy and uncontroversial. I am happy to create a branch and submit a PR for those, but there is also a (set of) warnings from code generated from `impl_exact_int_cast_int_float!` in `cast.rs`.

```rust
macro_rules! impl_exact_int_cast_int_float {
    ($int:ty, $float:ty) => (impl ExactIntCast<$int> for $float {
        fn exact_int_cast(v_int: $int) -> Fallible<Self> {
            let v_float = v_int as $float;
            if <$float>::MIN_CONSECUTIVE > v_float || <$float>::MAX_CONSECUTIVE < v_float {
```
Clippy wants (see [manual_range_contains](https://rust-lang.github.io/rust-clippy/master/index.html#manual_range_contains)) us to use `.contains()` for checking whether a value is within a range.

I suspect that this is straight forward under the assumption that I am not too mistaken about when types are checked and enforced wrt to macro expansion. Indeed, the intent of the macro illustrates that I am not too mistake.  But I've never played with Rust macros before. (There was a time when I wrote M4 macros for sendmail.cf configurations, but that felt like several lifetimes ago.)

Shoeboxam:
Closes #489.

Thanks @jpgoldberg!

### DIFF
--- a/rust/build/main.rs
+++ b/rust/build/main.rs
@@ -84,7 +84,7 @@ impl Argument {
         self.c_type.clone().expect("unknown c_type when parsing argument")
     }
     fn c_type_origin(&self) -> String {
-        self.c_type().split("<").next().unwrap().to_string()
+        self.c_type().split('<').next().unwrap().to_string()
     }
 }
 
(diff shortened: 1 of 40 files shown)


andrewvyrros (approved):
Thanks clippy!

PR #532, opened 2022-08-23
change `make_resize`' MO to SymmetricDistance

### LINKED ISSUE #531: Tune default dataset metric
If I change the default output metric of resize from insertdelete to symmetricdistance, and change the default input metric of sized bounded sum to symmetric distance, then
1. there will be one less shuffle in most cases
2. the default input metric of `make_sized_bounded_sum` will be the same as the other statistics
3. the default output metric of `make_bounded_resize` will match `make_resize`

Shoeboxam:
Closes #531


andrewvyrros (approved):
Good call!

PR #541, opened 2022-08-26
optimize notebook execution time

### LINKED ISSUE #253: Tests for example code
To prevent broken examples, we should hook them into the pytest infrastructure:

For notebooks, we can use `nbmake`:

    pytest --nbmake python/example

(In `smoke-test.yml`, will need to pip install nbmake, as well as any notebook requirements.)

For Python scripts, we can use the trick here: https://stackoverflow.com/a/56813896

### LINKED ISSUE #540: Optimize execution time of notebooks
They needlessly take too long to run! Reduce some fidelity in graphs! Prepare for CI testing!

Shoeboxam:
Closes #540 
Closes #253 

Also runs the notebooks in CI.


python/src/opendp/mod.py
```diff
@@ -638,7 +638,13 @@ def exception_predicate(v):
             return False
     exception_bounds = exponential_bounds_search(exception_predicate, T=T)
     if exception_bounds is None:
-        raise ValueError("predicate always fails")
+        err = ""
```
andrewvyrros, line 641: Should this be `error = ""`?

.github/workflows/smoke-test.yml
```diff
@@ -86,9 +86,6 @@ jobs:
           # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
           flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
 
-      - name: Prepare for libs
-        run: mkdir -p src/opendp/v1/lib
-
```
andrewvyrros, line 91: Whoa that's old!

andrewvyrros (commented):
Thanks for doing this, it's been on my mind for a while. It's great to have automated checking of notebooks, will save us a lot of trouble.

Now that we have some infrastructure for notebooks, maybe we should think about stripping the notebook output (especially of images) in the checked-in files, and generating the full versions when publishing to the docs site?

andrewvyrros (approved):

PR #548, opened 2022-09-14
Add Retries For Pulling Pypi Repo

### LINKED ISSUE #537: release_tool.py postflight can fail if PyPI is slow updating its index
`release_tool.py postflight` is one of the steps of generating a release. It sanity checks a new (pre-)release by pulling down the package from PyPI and running a simple program against the published binary (rather than the local build). But if the PyPI index is slow, this can fail because the requested `opendp` version isn't available yet. That prevents you from using the meta-command `release_tool.py {major|minor|patch}`, which automates an entire release from end to end. Instead, you have to run each step manually, which is a hassle because of the long build time.

We should add a pause or retry somewhere, so that `postflight` will do the right thing, and we can use the automation.

michaeleliot:
Closes #537:
* Creates retry command function that will retry any command, and log/report the final failure
* ~~Captures stderr when running command~~
* small name change nit to python version formatting

### DIFF
--- a/tools/release_tool.py
+++ b/tools/release_tool.py
@@ -10,8 +10,6 @@ import time
 
 import semver
 
-ANSI_FAIL = '\033[91m' # Prints Terminal Text Red
-ANSI_END = '\033[0m'
 
 def log(message, command=False):
     prefix = "$" if command else "#"
@@ -24,7 +22,7 @@ def run_command(description, args, capture_output=True, shell=True):
     printed_args = args.join(" ") if type(args) == list else args
     log(printed_args, command=True)
     stdout = subprocess.PIPE if capture_output else None
-    completed_process = subprocess.run(args, stdout=stdout, stderr=stdout, shell=shell, check=True, encoding="utf-8")
+    completed_process = subprocess.run(args, stdout=stdout, shell=shell, check=True, encoding="utf-8")
     return completed_process.stdout.rstrip() if capture_output else None
 
 
@@ -35,7 +33,6 @@ def run_command_with_retries(description, args, retries = 10, wait_time_seconds 
         except Exception as e:
             was_final_attempt = retries == 1
             if was_final_attempt:
-                log(f"{ANSI_FAIL}{e.stderr}{ANSI_END}")
                 raise e
             log(f"Waiting {wait_time_seconds} Seconds | Retries Left: {retries - 1}")
             time.sleep(wait_time_seconds)
--- a/tools/release_tool.py
+++ b/tools/release_tool.py
@@ -6,9 +6,12 @@ import platform
 import subprocess
 import sys
 import types
+import time
 
 import semver
 
+ANSI_FAIL = '\033[91m' # Prints Terminal Text Red
+ANSI_END = '\033[0m'
 
 def log(message, command=False):
     prefix = "$" if command else "#"
@@ -21,7 +24,7 @@ def run_command(description, args, capture_output=True, shell=True):
     printed_args = args.join(" ") if type(args) == list else args
     log(printed_args, command=True)
     stdout = subprocess.PIPE if capture_output else None
-    completed_process = subprocess.run(args, stdout=stdout, shell=shell, check=True, encoding="utf-8")
+    completed_process = subprocess.run(args, stdout=stdout, stderr=stdout, shell=shell, check=True, encoding="utf-8")
     return completed_process.stdout.rstrip() if capture_output else None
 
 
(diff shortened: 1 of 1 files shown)


Shoeboxam:
These look like useful changes, thanks @michaeleliot! Really appreciate the PR.

@andrewvyrros, in your experience, is 15 seconds long enough for the PyPi package index to update? From reading online, I see one person claim it can take anywhere from 5-10 minutes. If this is really the case, we might want to tune the parameters to... 10 retries, with a 60 second backoff?

michaeleliot:
> These look like useful changes, thanks @michaeleliot! Really appreciate the PR.

Thank you! Very interested in the project.

> 
> 
> 
> @andrewvyrros, in your experience, is 15 seconds long enough for the PyPi package index to update? From reading online, I see one person claim it can take anywhere from 5-10 minutes. If this is really the case, we might want to tune the parameters to... 10 retries, with a 60 second backoff?

I mean I wouldn't expect it to be that long, though it depends on the artifact size. Not sure if this is a caching issue either. Happy to implement a backoff if you like.

andrewvyrros:
For sure, thanks @michaeleliot!

> > @andrewvyrros, in your experience, is 15 seconds long enough for the PyPi package index to update? From reading online, I see one person claim it can take anywhere from 5-10 minutes. If this is really the case, we might want to tune the parameters to... 10 retries, with a 60 second backoff?
> 
> I mean I wouldn't expect it to be that long, though it depends on the artifact size. Not sure if this is a caching issue either. Happy to implement a backoff if you like.

I've only hit this a few times, but it seemed to last about a minute. So I'd say 10 retries at 30 second interval is reasonable. If it's still failing at that point, then it's probably time to investigate before proceeding.

tools/release_tool.py
```diff
@@ -193,13 +196,27 @@ def python_version(version):
...
+    except Exception as e:
+      was_final_attempt = retries == 1
+      if was_final_attempt:
+        log(f"{ANSI_FAIL}{e.stderr}{ANSI_END}")
+        raise e
+      log(f"Waiting {wait_time_seconds} Seconds | Retries Left: {retries - 1}")
+      time.sleep(wait_time_seconds)
+    retries -= 1
+
+
```
andrewvyrros, line 219: Could you reformat to match our convention of 4 spaces indentation?

Also, for organization purposes, I'd suggest moving this function up to line 31 (immediately after `run_command`).

Thanks!

andrewvyrros (changes requested):

tools/release_tool.py
```diff
@@ -21,10 +24,24 @@ def run_command(description, args, capture_output=True, shell=True):
     printed_args = args.join(" ") if type(args) == list else args
     log(printed_args, command=True)
     stdout = subprocess.PIPE if capture_output else None
-    completed_process = subprocess.run(args, stdout=stdout, shell=shell, check=True, encoding="utf-8")
+    completed_process = subprocess.run(args, stdout=stdout, stderr=stdout, shell=shell, check=True, encoding="utf-8")
```
andrewvyrros, line 27: Configuring `stderr` this way causes us to lose the compiler output on the terminal. (For comparison, run `python release_tool.py preflight` with and without this change.) Is it necessary to do that? The compiler output is helpful, so it'd be good to retain it if possible.

andrewvyrros (commented):
Thanks for responding the the feedback @michaeleliot !

Apologies for asking for more work, but I just noticed that this change causes us to lose logging from the compiler. It'd be swell if we could avoid that.

tools/release_tool.py
michaeleliot: I debated that as well. It made sense to me given that it's named capture_output to capture both stdout and stderr. The only reason I made the change was because repeating the failed output to the console seemed a bit much. I'm fine with either.

michaeleliot: I'll also note, I didn't find semvar's errors particularly useful since, in many cases though not this one specifically since pip writes to stderr afaik, it only returns if it failed, and not what the error was. Capturing the stderr and returning it allows the code to handle it better in cases where that isn't the case.

michaeleliot:
Also I haven't been able to figure out how to run the full build since I'm not sure what `/tmp/opendp-release` should be, and it doesn't seem to be prepopulated anywhere. I also don't see it via [searching the site](https://docs.opendp.org/en/stable/search.html?q=%2Ftmp%2Fopendp-release), or in the [developer setup](https://docs.opendp.org/en/stable/developer/development-environment.html).

andrewvyrros:
> Also I haven't been able to figure out how to run `python release_tool.py preflight` since I'm not sure what `/tmp/opendp-release` should be, and it doesn't seem to be prepopulated anywhere. I also don't see it via [searching the site](https://docs.opendp.org/en/stable/search.html?q=%2Ftmp%2Fopendp-release), or in the [developer setup](https://docs.opendp.org/en/stable/developer/development-environment.html).

Ah good point! `/tmp/opendp-release` is the working directory when creating a release. It's set up by doing `release_tool.py init`. Creating releases isn't something we'd expect new contributors to tackle, so the docs are lacking on that topic. But getting feedback like this is how we learn where they need work! Thanks for your help!

andrewvyrros (approved):
Looks good, thanks!

PR #555, opened 2022-09-24
port `combinators` module to bootstrap macro

Shoeboxam:


andrewvyrros (commented):
The checked in version of the Python is different from what is generated, seems to lack the `:raises` change in [f44a6d1](https://github.com/opendp/opendp/pull/558/commits/f44a6d1116310d85c3f83d608df46f617af720bd)

andrewvyrros (approved):

PR #556, opened 2022-09-24
port remaining modules to bootstrap macro

Shoeboxam:


andrewvyrros (commented):
Similar to #554, the generated Python files seem to be stale.

andrewvyrros (approved):

PR #559, opened 2022-09-25
Small Onboarding Updates

michaeleliot:
Been getting setup with opendp. I had to make a few adjustments to get things working.
* Install wheel into virtual env
* add feature flag to rust lib.rs

I'll also note I haven't been able to get the python pytest tests to work. After following the [setup guide](https://docs.opendp.org/en/stable/developer/development-environment.html#build-opendp), I get a Module Not Found error. Looking through online, I think it's a pythonpath error, but I can't figure out how to fix it.

```
pytest -v
```
Screen Shot 2022-09-25 at 1 12 12 PM

```
// A common solution online was to manually add src to pythonpath
PYTHONPATH=src pytest -v
```
Screen Shot 2022-09-25 at 1 14 22 PM

### DIFF
--- a/docs/source/developer/development-environment.rst
+++ b/docs/source/developer/development-environment.rst
@@ -128,7 +128,7 @@ Change to the ``python`` directory, install dependencies, and then install the P
 
     cd python
 
-    pip install flake8 pytest
+    pip install flake8 pytest wheel
     pip install -e .
 
 The `-e` flag is significant! It stands for "editable", meaning you only have to run this command once.
--- a/docs/source/developer/development-environment.rst
+++ b/docs/source/developer/development-environment.rst
@@ -33,8 +33,13 @@ If you have not `set up SSH <https://docs.github.com/en/authentication/connectin
 Build OpenDP
 ============
 
-Next, you'll need to build the Rust binaries. 
-This is done by running ``cargo build`` in the ``rust`` subdirectory of the repo.
+Make sure you are on the latest rust version
+
+.. code-block:: bash
+  rustup update
+
+
+Next, you'll need to build the Rust binaries. This is done by running ``cargo build`` in the ``rust`` subdirectory of the repo.
 
 .. code-block:: bash
 
--- a/rust/src/data/mod.rs
+++ b/rust/src/data/mod.rs
@@ -1,5 +1,4 @@
 //! Framework for flexible abstract data type model for DataFrames.
-
 #[cfg(feature="ffi")]
 mod ffi;
 
--- a/rust/src/data/mod.rs
+++ b/rust/src/data/mod.rs
@@ -1,4 +1,5 @@
 //! Framework for flexible abstract data type model for DataFrames.
+
 #[cfg(feature="ffi")]
 mod ffi;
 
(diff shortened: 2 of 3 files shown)


Shoeboxam:
This is really valuable info, thanks!

1. The fact that you need those feature flags indicates to me that your rust version is out-of-date. Would you mind running `rustup update`?
2. With the newer version of rust, you can drop those nightly features (they should no longer be nightly by our MSVC of rust 1.59)
3. `cargo build --features=untrusted,bindings-python` should emit a file `/Users/michaeleliot/Projects/personal/opendp/rust/target/debug/libopendp.dylib`. Can you verify this file exists?
4. Finally, make sure that the folder `./python/src/opendp/lib` does not exist! In a typical developer environment that folder should not exist. If the `lib` folder doesn't exist, the library looks for the dylib in `rust/target/debug/` instead (where cargo puts it!)

Also feel free to reach out to me on Slack-- I'll be able to respond more quickly.

michaeleliot:
> This is really valuable info, thanks! The way the system works is, you first run the cargo command to build a binary library `libopendp.dylib`. When you import `opendp`, Python tries to open this file. Python throws your error when it can't find this file.
> 
>     1. The fact that you need those feature flags indicates to me that your rust version is out-of-date. Would you mind running `rustup update`?

Yep good catch!

> 
>     2. With the newer version of rust, you can drop those nightly features (they should no longer be nightly by our MSVC of rust 1.59)
> 

Can do!

>     3. `cargo build --features=untrusted,bindings-python` should emit a file `/Users/michaeleliot/Projects/personal/opendp/rust/target/debug/libopendp.dylib`. Can you verify this file exists?
> 

Yes that file does exist

>     4. Finally, make sure that the folder `./python/src/opendp/lib` does not exist! In a typical developer environment that folder should not exist. If the `lib` folder doesn't exist, the library looks for the dylib in `rust/target/debug/` instead (where cargo puts it!)
> 

Yes there is no lib folder. I'm a little confused why this would be. It seems to reference lib from the setup.cfg, but I don't see where it would link the compiled rust.

Shoeboxam:
When you installed the package, did you use the`-e` flag? As in, `pip install -e .`? The editable install is important.

The code here is the logic for loading the .dylib:
https://github.com/opendp/opendp/blob/ce8c6c6546c3a1f1cadb2366844be2cf3f4c50cc/python/src/opendp/_lib.py#L15

You'll notice that it starts by checking an environment variable. If the env var is not set it points to the lib folder. 
On line 16, it should enter the conditional, because your lib folder does not exist. Now `lib_dir` should be pointing to your `rust/target/debug` directory. 

Put some print statements in to see if the variable is being set correctly. This code executes when you import anything in opendp.

Shoeboxam:
When I tried running this on my own on a new virtual env, I wasn't able to find pytest unless I ran `python -m pytest -v`. Things otherwise worked. I'm not sure where the issue is, but I'd be interested to hear what paths come up when you add some print statements. ^^

Shoeboxam:
Since this has been quiet for a while, I'm closing in favor of #628. Please send me a message if you find the time to pursue this further! Thanks.

PR #566, opened 2022-10-02
fix approximate composition FFI

### LINKED ISSUE #545: Restructure dispatch in composition FFI
Right now, dispatch in `impl BasicCompositionMeasure for AnyMeasure` is typed by `M::Distance`, but only dispatches to `@floats`. Thus we can't do composition of fixed approx-DP distances.
This isn't as breaking as it seems, as it's generally better to do composition in rho anyways.

Shoeboxam:
Closes #545

### DIFF
--- a/python/test/test_comb.py
+++ b/python/test/test_comb.py
@@ -76,6 +76,19 @@ def test_make_basic_composition_leak():
         meas([0] * 10_000_000)
     
 
+def test_make_basic_composition_approx():
+    from opendp.combinators import make_basic_composition, make_zCDP_to_approxDP, make_fix_delta
+    composed_curve = make_basic_composition([
+        make_zCDP_to_approxDP(make_base_gaussian(1.))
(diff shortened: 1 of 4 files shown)


python/test/test_comb.py
```diff
@@ -88,5 +96,5 @@ def test_cast_zcdp_approxdp():
     print(smd_gaussian.map(1.).epsilon(1e-6))
     
 if __name__ == "__main__":
-    test_cast_zcdp_approxdp()
+    test_make_basic_composition_approx()
```
andrewvyrros, line 100: Leftovers?

andrewvyrros (approved):

python/test/test_comb.py
Shoeboxam: It's guarded by an if, so it only runs if you run the file manually. Seems fine to me if it changes over time

PR #572, opened 2022-10-11
support user-supplied callbacks in client code

### LINKED ISSUE #44: Facility for using callbacks implemented in client code
It'd be very nice to have a facility whereby functions implemented in client code (i.e., outside FFI, in Python) could be passed into the library and used as callbacks. This would allow us to support custom transformations and relations. (This would be available only in an explicit "unsafe" mode.)

### Dependencies
- [x] #43

### LINKED ISSUE #603: Add `FixedSmoothedMaxDivergence` to `opendp.typing`
We're missing a type alias, which makes it annoying to reference this measure.

Shoeboxam:
Closes #44 
Closes #603 

Allows users to write custom transformations, measurements and postprocessors from Python code:

```python
enable_features("contrib", "honest-but-curious")

def make_duplicate(multiplicity):
    """An example user-defined transformation from Python"""
    def function(arg):
        return [elem + 1 for elem in arg] * multiplicity

    def stability_map(d_in):
        return d_in * multiplicity

    return make_default_transformation(
        function,
        stability_map,
        DI=VectorDomain[AllDomain[i32]],
        DO=VectorDomain[AllDomain[i32]],
        MI=SymmetricDistance,
        MO=SymmetricDistance,
    )

trans = (
    make_cast_default(TIA=str, TOA=int)
    >> make_duplicate(2)
    >> make_clamp((1, 2))
    >> make_bounded_sum((1, 2))
    >> make_base_discrete_laplace(1.0)
)

print(trans(["0", "1", "2", "3"]))
print(trans.map(1))
```

### DIFF
--- a/python/src/opendp/_convert.py
+++ b/python/src/opendp/_convert.py
@@ -61,6 +61,9 @@ def py_to_c(value: Any, c_type, type_name: Union[RuntimeType, str] = None):
     if isinstance(value, c_type):
         return value
 
+    if c_type == CallbackFn:
+        return _wrap_py_func(value, type_name)
+
     # check that the type name is consistent with the value
     if type_name is not None:
         RuntimeType.assert_is_similar(RuntimeType.parse(type_name), RuntimeType.infer(value))
@@ -364,3 +367,41 @@ def _slice_to_hashmap(raw: FfiSlicePtr) -> Dict[Any, Any]:
 
 def _wrap_in_slice(ptr, len_: int) -> FfiSlicePtr:
     return FfiSlicePtr(FfiSlice(ctypes.cast(ptr, ctypes.c_void_p), len_))
+
+
+#                            (output         , input       )
+CallbackFn = ctypes.CFUNCTYPE(ctypes.c_void_p, AnyObjectPtr)
+
+def _wrap_py_func(func, TO):
+    from opendp._convert import c_to_py, py_to_c
+
+    def wrapper_func(c_arg):
+        try:
+            # 1. convert AnyObject to Python type
+            py_arg = c_to_py(c_arg)
+            # don't free c_arg, because it is owned by Rust
+            c_arg.__class__ = ctypes.POINTER(AnyObject)
+
+            # 2. invoke the user-supplied function
+            py_out = func(py_arg)
+
+            # 3. convert back to an AnyObject
+            c_out = py_to_c(py_out, c_type=AnyObjectPtr, type_name=TO)
+            # don't free c_out, because we are giving ownership to Rust
+            c_out.__class__ = ctypes.POINTER(AnyObject)
+
+            # 4. pack up into an FfiResult
+            lib.ffiresult_ok.argtypes = [ctypes.c_void_p]
+            lib.ffiresult_ok.restype = ctypes.c_void_p
+            return lib.ffiresult_ok(ctypes.addressof(c_out.contents))
+
+        except Exception:
+            import traceback
+            lib.ffiresult_err.argtypes = [ctypes.c_char_p, ctypes.c_char_p]
+            lib.ffiresult_err.restype = ctypes.c_void_p
+            return lib.ffiresult_err(
+                ctypes.c_char_p(f"Continued stack trace from Exception in user-defined function".encode()),
+                ctypes.c_char_p(traceback.format_exc().encode()),
+            )
+
+    return CallbackFn(wrapper_func)
--- a/python/src/opendp/_convert.py
+++ b/python/src/opendp/_convert.py
@@ -369,6 +369,8 @@ def _wrap_in_slice(ptr, len_: int) -> FfiSlicePtr:
     return FfiSlicePtr(FfiSlice(ctypes.cast(ptr, ctypes.c_void_p), len_))
 
 
+# The result type cannot be an `ctypes.POINTER(FfiResult)` due to:
+#   https://bugs.python.org/issue5710#msg85731
 #                            (output         , input       )
 CallbackFn = ctypes.CFUNCTYPE(ctypes.c_void_p, AnyObjectPtr)
 
--- a/python/src/opendp/typing.py
+++ b/python/src/opendp/typing.py
@@ -411,6 +411,7 @@ class PrivacyMeasure(RuntimeType):
 
 MaxDivergence = PrivacyMeasure('MaxDivergence')
 SmoothedMaxDivergence = PrivacyMeasure('SmoothedMaxDivergence')
+FixedSmoothedMaxDivergence = PrivacyMeasure('FixedSmoothedMaxDivergence')
 ZeroConcentratedDivergence = PrivacyMeasure('ZeroConcentratedDivergence')
 
 class Carrier(RuntimeType):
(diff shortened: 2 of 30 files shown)


python/src/opendp/_convert.py
```diff
@@ -364,3 +367,41 @@ def _slice_to_hashmap(raw: FfiSlicePtr) -> Dict[Any, Any]:
 
 def _wrap_in_slice(ptr, len_: int) -> FfiSlicePtr:
     return FfiSlicePtr(FfiSlice(ctypes.cast(ptr, ctypes.c_void_p), len_))
+
+
+#                            (output         , input       )
+CallbackFn = ctypes.CFUNCTYPE(ctypes.c_void_p, AnyObjectPtr)
```
andrewvyrros, line 373: Shouldn't this return `FFIResult` (rather than a pointer) like other FFI functions?

python/src/opendp/_convert.py
```diff
@@ -364,3 +367,41 @@ def _slice_to_hashmap(raw: FfiSlicePtr) -> Dict[Any, Any]:
...
+
+            # 3. convert back to an AnyObject
+            c_out = py_to_c(py_out, c_type=AnyObjectPtr, type_name=TO)
+            # don't free c_out, because we are giving ownership to Rust
+            c_out.__class__ = ctypes.POINTER(AnyObject)
+
+            # 4. pack up into an FfiResult
+            lib.ffiresult_ok.argtypes = [ctypes.c_void_p]
+            lib.ffiresult_ok.restype = ctypes.c_void_p
+            return lib.ffiresult_ok(ctypes.addressof(c_out.contents))
```
andrewvyrros, line 396: Aren't you leaking the `FFIResult` here?

rust/src/data/ffi.rs
```diff
@@ -403,6 +403,35 @@ pub extern "C" fn opendp_data__to_string(this: *const AnyObject) -> FfiResult<*m
         FfiResult::Ok)
 }
 
+/// wrap an AnyObject in an FfiResult::Ok(this)
+#[no_mangle]
+pub extern "C" fn ffiresult_ok(this: *const AnyObject) -> *const FfiResult<*const AnyObject> {
+    util::into_raw(FfiResult::Ok(this))
+}
```
andrewvyrros, line 410: Why have a function for this? `FfiResult` is a C union, Python can deal with it fine. It's intended to be returned on the stack, not as a pointer, so Python doesn't have to deal with memory management for return values. (I think these functions leak in the callback code.)

rust/src/combinators/chain/ffi.rs
```diff
@@ -7,7 +7,8 @@ use crate::ffi::any::{AnyMeasurement, AnyTransformation};
 
 #[bootstrap(
     features("contrib"),
-    arguments(measurement1(rust_type = b"null"), transformation0(rust_type = b"null"))
+    arguments(measurement1(rust_type = b"null"), transformation0(rust_type = b"null")),
+    dependencies("$get_dependencies(measurement1)", "$get_dependencies(transformation0)")
```
andrewvyrros, line 11: I'm not convinced this will always do the right thing. In most of the combinators, the arguments get cloned. So the dependency here would be on the original argument, not what actually ends up being used. It works (I think) because internally, the clone is implemented by bumping the count on the `Rc`, so it's the same object. Seems like a better approach would be for Rust to declare the dependencies (though not sure how to do that).

The Apache Arrow memory model may be instructive here.

rust/src/combinators/user_defined/ffi.rs
```diff
@@ -0,0 +1,281 @@
...
+/// 
+/// # Arguments
+/// * `function` - A function mapping data from `DI` to `DO`.
+/// * `stability_map` - A function mapping distances from `MI` to `MO`.
+/// * `DI` - Input Domain. See Supported Domains
+/// * `DO` - Output Domain. See Supported Domains
+/// * `MI` - Input Metric. See Supported Metrics
+/// * `MO` - Output Metric. See Supported Metrics
+#[no_mangle]
+pub extern "C" fn opendp_combinators__make_default_transformation(
```
andrewvyrros, line 152: It feels a little awkward to call these `make_default_xxx`, as the default part seems incidental. The emphasis here is on the user part, so how about `make_user_xxx`? If we later have a way to pass in domains etc, we could have that be an optional argument, or make a different entry point.

andrewvyrros (commented):
This is super exciting, something we've imagined for a long time!

The dependencies thing is a cool trick, but I'm a little concerned about its robustness. Might be worth talking through the mechanics.

python/src/opendp/typing.py
```diff
@@ -470,3 +473,9 @@ def get_first(value):
 
 def parse_or_infer(type_name: RuntimeType, example):
     return RuntimeType.parse_or_infer(type_name, example)
+
```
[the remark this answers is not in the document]
Shoeboxam: I failed to push that commit up. Nice.

python/src/opendp/_convert.py
Shoeboxam: That would be nice, but we're running into a limitation of Python here:
https://bugs.python.org/issue5710#msg85731

I've added a comment to this effect.

python/src/opendp/_convert.py
Shoeboxam: It shouldn't, because Rust takes ownership of the `FfiResult` here:

https://github.com/opendp/opendp/pull/572/files#diff-d59f16e57c55ad474cc8d65373f333ede766806b239443d5bdfed23f44ff42a4R31-R33

rust/src/data/ffi.rs
Shoeboxam: This was my first approach, but I can't seem to escape the python refcounter. The C union needs to outlive the python stack frame it was defined in. Even storing the payload and ffiresult in globals is not enough. I ended up using this to ensure that Rust has ownership and can free it.

rust/src/combinators/chain/ffi.rs
Shoeboxam: The dependencies in `measurement1` are the c wrappers for the function and privacy map. So the Rc we're keeping from going to zero is not a Rust Rc, it's a Python one. This approach is to appease the python refcounter only.

rust/src/combinators/user_defined/ffi.rs
Shoeboxam: I was trying to guard the `make_transformation` namespace. This reflects the limitation that the domains you choose must impl `Default`. Keep in mind, in the future, when we change the arguments to accept custom domains, the arguments will get reordered. Does this context influence your naming suggestion?

python/src/opendp/_convert.py
andrewvyrros: Dang it!

python/src/opendp/_convert.py
andrewvyrros: Ah yes, I see it now, thanks!

rust/src/data/ffi.rs
andrewvyrros: Ah right, you can't get Python to give you stack value in this situation.

rust/src/combinators/chain/ffi.rs
andrewvyrros: After our discussion, I see how this is OK: Cloning could only ever reduce the number of dependencies, so this should always be safe.

andrewvyrros (approved):
Nice!

PR #579, opened 2022-10-13
discrete accuracy utilities

Shoeboxam:
Closes https://github.com/opendp/opendp/issues/577

### DIFF
--- a/rust/src/accuracy/accuracy_to_discrete_gaussian_scale.tex
+++ b/rust/src/accuracy/accuracy_to_discrete_gaussian_scale.tex
@@ -1,0 +1,22 @@
+\documentclass{article}
+\input{../lib.sty}
+
+\title{\texttt{fn accuracy\_to\_discrete\_gaussian\_scale}}
+\author{Michael Shoemate}
+
+\begin{document}
+\maketitle
+
+This document contains materials associated with \rustdoc{accuracy/fn}{accuracy\_to\_discrete\_gaussian\_scale}.
+
+By \rustdoc{accuracy/fn}{discrete\_gaussian\_scale\_to\_accuracy}, 
+the relationship between $\alpha$, $a$ and $scale$, is:
+
+\begin{equation*}
+    \alpha = 2 \frac{e^{(1 - a)/s}}{e^{1/s} + 1}
+\end{equation*}
+
+There unfortunately isn't a closed-form expression for s.
+The algorithm conducts a binary search, 
+with an upper bound provided by \rustdoc{accuracy/fn}{accuracy\_to\_laplacian\_scale}.
+\end{document}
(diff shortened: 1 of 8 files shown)


rust/src/accuracy/mod.rs
```diff
@@ -110,6 +268,20 @@ pub mod test {
...
+    fn test_comparison() -> Fallible<()> {
+        let alpha = 0.05;
+
+        let scale = 20.;
+        println!("cont accuracy: {}", laplacian_scale_to_accuracy(scale, alpha)?);
+        println!("disc accuracy: {}", discrete_laplacian_scale_to_accuracy(scale, alpha)?);
+
+        let accuracy = 20.;
+        println!("cont scale: {}", accuracy_to_laplacian_scale(accuracy, alpha)?);
+        println!("disc scale: {}", accuracy_to_discrete_laplacian_scale(accuracy, alpha)?);
```
andrewvyrros, line 281: How about gaussian?

Can we make any assertions of discrete vs continuous?

rust/src/accuracy/mod.rs
```diff
@@ -237,8 +409,8 @@ pub mod test {
         let accuracy = 1.0;
         let theoretical_alpha = 0.05;
         let scale = accuracy_to_laplacian_scale(accuracy, theoretical_alpha)?;
-        let base_laplace = make_base_laplace::<AllDomain<f64>>(scale, None)?;
-        let n = 50_000;
+        let base_laplace = make_base_laplace::<AllDomain<f64>>(scale, Some(-100))?;
+        let n = 500_000;
```
andrewvyrros, line 413: These are a little slow at 500K, do you want this so big?

Similar for `test_empirical_discrete_gaussian_accuracy`.

andrewvyrros (commented):
Code looks good, but we may want to have someone else look at the proofs.

rust/src/accuracy/accuracy_to_discrete_gaussian_scale.tex
```diff
@@ -0,0 +1,27 @@
...
+the relationship between $\alpha$, $a$ and $scale$, is:
+
+\begin{align*}
+    1 - \alpha = \frac{\sum_{y=0}^{a - 1} (1 + 1[y = 0]) e^{-(y/s)^2/2}}{\sum_{z \in \mathbb{Z}} e^{-(z/s)^2/2}}
+\end{align*}
+
+A closed-form expression for s doesn't exist, 
+so we use a numerical approach by a binary search.
+An upper bound is provided by \rustdoc{accuracy/fn}{accuracy\_to\_laplacian\_scale}.
+The binary search finds the smallest $s$ such that
```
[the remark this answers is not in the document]
Shoeboxam: Fixed that link to refer to the gaussian. 

No dumb questions. I guess there is no "the" upper bound, as there can always be more than one, but there is only one tightest upper bound. But in this case, it referring specifically to "the" upper bound used in the binary search.

andrewvyrros (dismissed):

andrewvyrros (approved):

PR #596, opened 2022-10-26
Do crate version munging in release_tool, so changes can be committed.

andrewvyrros:
Because we need to munge the version to publish the dependency crates, we ran into an issue with cargo complaining about a dirty repo: https://github.com/opendp/opendp/actions/runs/3325358361/jobs/5497961034

We could use the --allow-dirty flag, but that feels a little dicey.

Instead, this change moves the munging from `publish_tool.py` to `release_tool.py`, which is already dealing with git changes on the release branch.

Side note: We could've tested this locally outside GitHub, and avoided a lot of hassle, if not for the fact that you can't use --dry-run with workspaces. Possible solution here: https://github.com/rust-lang/cargo/issues/1169#issuecomment-1254260875

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -32,7 +32,7 @@ rust-version.workspace = true
 exclude = ["windows/*"]
 
 [dependencies]
-opendp_derive = { path = "opendp_derive" }
+opendp_derive = { path = "opendp_derive", version = "0.0.0+development" }
 rand = "0.7.3"
 num = "0.3.1"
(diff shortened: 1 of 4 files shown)


Shoeboxam:
We already need to go in and set `/VERSION`, `rust/Cargo.toml`'s workspace version, and `python/setup.cfg`'s version. Should we handle setting those versions differently here?

Shoeboxam (approved):

PR #617, opened 2022-11-30
Doc update

alexWhitworth:


Shoeboxam (approved):
Thanks for the updates! I'm observing that even if a prior doctest has enabled features, it is not necessarily still enabled when a later test runs, so the testsetup block is still necessary. This is surprising, because variables and imports seem to be preserved across doctests. 

I'm seeing a number of locations where similar changes should happen, when globally searching for `testsetup`. Those changes don't necessarily need to be in this PR.

PR #618, opened 2022-12-04
queryables

### LINKED ISSUE #80: Interactive Measurement Design

Shoeboxam:
Closes #80.

Based on many discussions with @andrewvyrros, @SalilVadhan, Marco Gaboardi and Michael Hay, as well as Andy's Python prototypes. Expect further changes.

# `struct Queryable<Q, A>`

In this implementation, queryables are modeled as:
```rust
pub struct Queryable<Q: ?Sized, A>(
    Rc<RefCell<dyn FnMut(&Self, Query<Q>) -> Fallible<Answer<A>>>>,
);
```
That is, a queryable is a closure that captures mutable variables. These captured mutable variables are the queryable state.

* Why is the queryable generic over `Q: ?Sized`? `Q` represents the type of a user's query. `?Sized` means the compiler is allowed to _not_ know the size of `Q` at compile-time (in terms of the number of bytes needed to store a value in memory). Unlike most traits, `Sized` is enabled by default on almost all types. Allowing dynamically-sized types (DSTs) allows `Q` to be `dyn Any`, which makes it possible to wrap a `Queryable<dyn Any, _>` in a queryable with a concrete type for `Q`, without requiring `Q: Clone`. This is particularly useful because if `Q: Clone`, then it is no longer possible to use type-erased queryables in IMs that transform the answers of descendants. 

* Why `FnMut`? I prefer using `FnMut` over separate `state` and `transition` because it eliminates the need to unbox, downcast, unpack, repack, upcast and re-box the state. It is more ergonomic to reference mutable variables that have been closed over. `FnMut` also helps to hide the representation of state from the queryable definition.

* Why `Rc`? The parent should survive as long as any of its descendants. The initial approach involved [lifetimes](https://doc.rust-lang.org/rust-by-example/scope/lifetime.html), where a child may hold a reference to its parent. This meant that children are generic WRT a lifetime parameter that signifies that they may live no longer than the parent (not the ideal behavior). A second issue that surfaced with lifetimes is that `Any` must be `'static`, so the `Context` query (which itself contained a reference/lifetime) must be static.

* Why `RefCell`? Now that we've established why we need `Rc`, realize that `DerefMut` is not implemented for `Rc`, to eliminate the possibility of multiple mutable borrows. `RefCell` is a pattern for interior mutability that enforces at most one mutable borrow exists at a time. Because of this, expect a panic if the transition function attempts to query `&Self`. If we wanted to support computational concurrency, we could swap these types to an `Arc<Mutex<dyn FnMut ...>>` that would block instead. I chose `RefCell` over `Cell` because the `FnMut` is not `Copy`.

## `FnMut` Arguments
1. What is `&Self`? Interactive DP algorithms typically need a way to communicate with the queryable that spawns it (sequential composition, odometers, etc). To make this possible, `&Self` (an immutable reference to itself, a `Queryable<Q, DA>`) is passed as the first argument to the `FnMut`. The parent queryable uses this to pass a `Context` query containing a reference to the parent queryable to a child queryable. Note that it is necessary to wrap oneself in a type-erasure queryable when passing a reference to oneself to the child (for `Context` internal queries), so that `Context` doesn't leak the `Q` and `DA` types of the parent.

2. What is `Query<Q>`? It is defined as:
```rust
pub(crate) enum Query<'a, Q: ?Sized> {
    External(&'a Q),
    Internal(&'a dyn Any),
}
```
It either holds a reference to a variable a user may pass of type `Q`, or a reference to some other piece of data passed internally (by the framework). These two kinds of queries are handled by one transition function so that the state is shared. The lifetime parameter `'a` is elided from the queryable struct. The un-elided form is `... dyn for<'a, 'b> FnMut(&'a Self, Query<'b, Q> ...`, where the higher-ranked trait bound represents that queries with any lifetime are valid.
If a queryable anticipates internal queries, the body of the transition function will attempt to downcast this to a small number of expected concrete types (like `CheckPermission`, `Context`, and `CheckDescendantChange`).

## `FnMut` Return
* `Fallible<T>`: represents a value of type `T` or an error. 

* What is `Answer<A>`? It is defined as:
```rust
pub(crate) enum Answer<A> {
    External(A),
    Internal(Box<dyn Any>),
}
```
It either holds an answer to a question a user asked of type `A`, or an answer to some query passed internally by the framework. 

# `struct PolyQueryable`

`PolyQueryable` is a type alias:
```rust
pub type PolyQueryable = Queryable<dyn Any, Box<dyn Any>>;
```

`PolyQueryable` is just like any other queryable, but the values of `Q` and `A` are parameterized with dynamic types to avoid the use of generics. This is appealing because:
1. The output type of compositors can be set to`PolyQueryable`, for composition of queries with diverse/varying output types. Compositors can only function for one specific setting of `Q` and `A`, but `Q` and `A` can be "erased" dyn types for which methods are called via dynamic dispatch.
2. Sequential composition can wrap the type-erased version of any descendant queryable without needing to know their concrete type.

* Why is `Q` parameterized as `dyn Any`? The natural first choice here was `Box<dyn Any>`, that is, a query is first "boxed" (put behind a pointer) like `Box::new(query)`, of type `Box<Q>`, and then the type is erased `Box::new(query) as Box<dyn Any>`, which constructs a vtable that supports downcasting the type back to a `Box<Q>` later. The issue here starts to surface in that queries are passed in as `&Q`, not `Q`, so as not to consume memory owned by other programming languages. So if `Q` were `Box<dyn Any>`, then the downcast from a `PolyQueryable` to a `Queryable<Q, A>` would require constructing an `&Box<dyn Any>` from an `&Q`. This is unfortunately not possible, because you can't box a value of type `Q` without having ownership. The dynamically-sized type `dyn Any` doesn't have this downside, as you can easily/freely cast from `&Q` to `&dyn Any` and back, making it easy to erase types and downcast respectively. The drawback is that `dyn Any` does not have a known size at compilation time, so the relaxation `Q: ?Sized` is needed, to tell the compiler that `Q` is allowed to be parameterized as a type with unknown size. This implementation uses `Q` in ways that abide by additional restrictions the compiler has for dynamically sized types.

* Why is `A` parameterized as `Box<dyn Any>`? In contrast to the query side, answers are owned, so upcasting/downcasting is not a problem. The box makes the return type `Sized`, and the underlying data is stored on the heap.


Shoeboxam:
* main
  * **PR #648** <a href="https://app.graphite.dev/github/pr/opendp/opendp/648" target="_blank">Graphite</a> 
    * **PR #618** <a href="https://app.graphite.dev/github/pr/opendp/opendp/618" target="_blank">Graphite</a>  👈
      * **PR #456** <a href="https://app.graphite.dev/github/pr/opendp/opendp/456" target="_blank">Graphite</a> 
        * **PR #641** <a href="https://app.graphite.dev/github/pr/opendp/opendp/641" target="_blank">Graphite</a> 
        * **PR #653** <a href="https://app.graphite.dev/github/pr/opendp/opendp/653" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/opendp/opendp/618?utm_source=stack-comment).

Shoeboxam:
* main
  * **PR #648** <a href="https://app.graphite.dev/github/pr/opendp/opendp/648" target="_blank">Graphite</a> 
    * **PR #618** <a href="https://app.graphite.dev/github/pr/opendp/opendp/618" target="_blank">Graphite</a>  👈
      * **PR #456** <a href="https://app.graphite.dev/github/pr/opendp/opendp/456" target="_blank">Graphite</a> 
        * **PR #641** <a href="https://app.graphite.dev/github/pr/opendp/opendp/641" target="_blank">Graphite</a> 
        * **PR #653** <a href="https://app.graphite.dev/github/pr/opendp/opendp/653" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/opendp/opendp/618?utm_source=stack-comment).

rust/src/interactive/mod.rs
```diff
@@ -0,0 +1,253 @@
...
+}
+
+impl PolyQueryable {
+    pub fn into_downcast<Q: ?Sized, A>(self) -> Queryable<Q, A>
+    where
+        Queryable<Q, A>: FromPolyQueryable,
+    {
+        Queryable::<Q, A>::from_poly(self)
+    }
+}
```
andrewvyrros, line 253: This doesn't seem to be used in any upcoming branches. Is it intended for user code?

andrewvyrros (commented):
It's fantastic that we're on the verge of completing interactive measurements! This was a huge amount of work, and you've been a good sport as we've debated different approaches and iterated to contain the complexity.

I think where we've landed looks super nice, and will be a great foundation for interesting future work.

rust/src/interactive/mod.rs
```diff
@@ -0,0 +1,253 @@
...
+                // evaluate the query and wrap the answer
+                wrap(wrap_logic.as_map(), || inner_qbl.eval_query(query))
+            })
+        }))
+    }
+}
+
+impl<Q: ?Sized, A> Queryable<Q, A>
+where
+    Self: IntoPolyQueryable + FromPolyQueryable,
```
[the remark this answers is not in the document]
Shoeboxam: I could remove these bounds from `new_raw`. Otherwise `IntoPolyQueryable`/`FromPolyQueryable` are necessary to construct queryables.

rust/src/interactive/mod.rs
Shoeboxam: It is used for Queryable::new. It could be used for downcasting poly queryables, but people will usually have `Box<dyn Any>` and downcast directly to `Queryable<Q, A>`. Removed.

andrewvyrros (approved):

PR #623, opened 2023-01-12
Clean up default builds

### LINKED ISSUE #615: Clean up default builds
Doing a build with default features (without setting `--features untrusted,ffi` that we usually enable) has some issues. It's not critical, but it'd be nice if this were cleaned up, so you could do a `cargo build` on a fresh clone.

* `cargo build`:
```
warning: /Users/av/Repositories/opendp/rust/Cargo.toml: version requirement `0.0.0+development` for dependency `opendp_derive` includes semver metadata which will be ignored, removing the metadata is recommended to avoid confusion
warning: /Users/av/Repositories/opendp/rust/Cargo.toml: version requirement `0.0.0+development` for dependency `opendp_tooling` includes semver metadata which will be ignored, removing the metadata is recommended to avoid confusion
warning: /Users/av/Repositories/opendp/rust/opendp_derive/Cargo.toml: version requirement `0.0.0+development` for dependency `opendp_tooling` includes semver metadata which will be ignored, removing the metadata is recommended to avoid confusion
...
   Compiling opendp v0.0.0+development (/Users/av/Repositories/opendp/rust)
warning: unused macro definition: `enclose`
   --> src/lib.rs:148:14
    |
148 | macro_rules! enclose {
    |              ^^^^^^^
    |
    = note: `#[warn(unused_macros)]` on by default

warning: unused import: `std::any::Any`
  --> src/domains/mod.rs:11:5
   |
11 | use std::any::Any;
   |     ^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default
```

* `cargo test`:
Same as above, plus:
```
error[E0425]: cannot find function `make_identity` in module `transformations`
  --> src/combinators/mod.rs:57:26
   |
57 |         transformations::make_identity(AllDomain::<T>::new(), SymmetricDistance::default()).unwrap_test()
   |                          ^^^^^^^^^^^^^ not found in `transformations`

warning: `opendp` (lib) generated 2 warnings
warning: unused import: `crate::error`
  --> src/combinators/mod.rs:41:9
   |
41 |     use crate::error::*;
   |         ^^^^^^^^^^^^
```

andrewvyrros:
Fixes #615

This tweaks a few things so you can run `cargo build` and `cargo test` with the default features, and things will work OK.

* Adds a few '#[cfg(feature="ffi")]'
* Changes a few imports
* Removes versions from local crate dependencies
* Adjusts `release_tool.py` to add local versions, so crate builds continue to work

### DIFF
--- a/rust/src/domains/mod.rs
+++ b/rust/src/domains/mod.rs
@@ -8,7 +8,8 @@
 //! Each metric (see [`crate::metrics`]) is associated with certain data domains. 
 //! The [`Domain`] trait is implemented for all domains used in OpenDP.
 
-#[cfg(feature="ffi")]
+// Once we have things using `Any` that are outside of `contrib`, this should specify `feature="ffi"`.
+#[cfg(feature="contrib")]
 use std::any::Any;
 use std::collections::HashMap;
 use std::hash::Hash;
--- a/rust/src/domains/mod.rs
+++ b/rust/src/domains/mod.rs
@@ -8,6 +8,7 @@
 //! Each metric (see [`crate::metrics`]) is associated with certain data domains. 
 //! The [`Domain`] trait is implemented for all domains used in OpenDP.
 
+#[cfg(feature="ffi")]
 use std::any::Any;
 use std::collections::HashMap;
 use std::hash::Hash;
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -145,7 +145,8 @@
 #![recursion_limit="512"]
 
 // create clones of variables that are free to be consumed by a closure
-#[cfg(feature="ffi")]
+// Once we have things using `enclose!` that are outside of `contrib`, this should specify `feature="ffi"`.
+#[cfg(feature="contrib")]
 macro_rules! enclose {
     ( $x:ident, $y:expr ) => (enclose!(($x), $y));
     ( ($( $x:ident ),*), $y:expr ) => {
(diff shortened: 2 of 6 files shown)


rust/src/lib.rs
```diff
@@ -145,6 +145,7 @@
 #![recursion_limit="512"]
 
 // create clones of variables that are free to be consumed by a closure
+#[cfg(feature="ffi")]
```
Shoeboxam, line 148: This is still needed when not in ffi. It's used when building `'static' closures for functions and maps.

rust/src/domains/mod.rs
```diff
@@ -8,6 +8,7 @@
 //! Each metric (see [`crate::metrics`]) is associated with certain data domains. 
 //! The [`Domain`] trait is implemented for all domains used in OpenDP.
 
+#[cfg(feature="ffi")]
```
Shoeboxam, line 11: This is used by `DataDomain` outside of ffi.

Shoeboxam (changes requested):
The ffi conditional compilation attributes break the build!

rust/src/lib.rs
andrewvyrros: Thanks, good catch! Per discussion, I'll change this to `feature="ffi"`.

rust/src/domains/mod.rs
andrewvyrros: Thanks, good catch! Per discussion, I'll change this to `feature="ffi"`.

Shoeboxam (approved):
Great to have those warnings gone! Thanks.

PR #629, opened 2023-02-02
fix accuracy edge cases

### LINKED ISSUE #626: `discrete_gaussian_scale_to_accuracy` edge case
`discrete_gaussian_scale_to_accuracy` returns an accuracy one too large when the scale is on the lower edge.

Reported by @alexWhitworth:

```python
import opendp.accuracy as dp_a
acc = 2.
dp_a.discrete_gaussian_scale_to_accuracy(
    dp_a.accuracy_to_discrete_gaussian_scale(accuracy= acc, alpha=0.1)
    , alpha= 0.1
) 
# returns 3.0, not 2.0
```

Thanks for the report!

Solution:
The continuation condition in the loop should be `total > 0.`, not `total.is_sign_positive()`. That is, the loop should break if the total is negative **or exactly zero**.

Shoeboxam:
Closes #626


andrewvyrros (approved):
Nice catch. Thanks @alexWhitworth!

PR #633, opened 2023-02-07
Add sleep so rustup can settle on windows

### LINKED ISSUE #631: 0.6.2 release preparation

andrewvyrros:
Fixes #631.

* Add sleep so rustup can settle on windows
* Remove explicit cargo path on windows
* Remove cargo login

### DIFF
--- a/tools/publish_tool.py
+++ b/tools/publish_tool.py
@@ -21,12 +21,9 @@ def run_command(description, args, capture_output=True, shell=True):
 
 def rust(args):
     log(f"*** PUBLISHING RUST LIBRARY ***")
-    os.environ["CARGO_REGISTRY_TOKEN"] = os.environ["CRATES_IO_API_TOKEN"]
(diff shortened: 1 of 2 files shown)


Shoeboxam:
Why sleep, and not `rustup self update`?

andrewvyrros:
> Why sleep, and not `rustup self update`?

It's failing with "/c/Users/runneradmin/.cargo/bin/rustup: Device or resource busy", so `rustup` isn't able to run at all! The report in https://github.com/rust-lang/rustup/issues/3189 shows similar "Permission denied" and "required file not found" errors. To me, those sound like filesystem issues. My theory is that the initial `rustup` that the GitHub image runs is not really finished yet, and the binary can't run because it hasn't been flushed to the filesystem.

So I brute forced it with a sleep, which seems to work. It's horrible, but not sure what else to do!

Shoeboxam (approved):
Weird, but not worth losing any more time over.

PR #647, opened 2023-02-28
remove output domain from Measurement

### LINKED ISSUE #646: Remove output domain from Measurement

Shoeboxam:
Closes #646


andrewvyrros (commented):
This is a long time coming. It was never in the framework, but I originally gave Measurements an output domain for symmetry with Transformations and Functions. It seemed like the right thing at the time! But now we know better, and this is a useful cleanup for interactive.

andrewvyrros (approved):

PR #650, opened 2023-03-02
integrate supporting elements

### LINKED ISSUE #649: use domain/metric/measure FFI to unify Python and Rust API

Shoeboxam:
Closes #649


andrewvyrros (approved):
This is a nice factoring of the interface to Domains and Metrics. The previous way this stuff was wrapped was a little clunky, and this makes it cleaner and coherent with Rust.

I don't love having `Clone` on `D::Carrier` (and `AnyObject`). It feels a little wrong to have to clone the data just to change the Metric. But it gets the job done nicely here. Big picture, I think it's worth having a discussion about the interface to data, but probably better to do that in the context of interactive, which may change our perspective.

andrewvyrros (approved):

PR #691, opened 2023-04-09
optimize backtrace handling

Shoeboxam:
Closes #670 

RelationDebug doesn't have significant performance issues, resolving backtraces does. This PR changes the default behavior to only resolve backtraces in `test` mode, not debug mode.

- remove backtrace crate in favor of Rust 1.65's stabilized std backtrace
- disable backtraces via an environment variable when conducting binary searches 
- avoids a performance cliff when calling the ffi for `atom_domain`, as the dispatch used to rely on error handling
    - added a trait that makes it possible to return `None` from `dispatch!`
- avoids a performance cliff when calling `cdp_epsilon` when `rho` is infinity (scale is zero). Instead of overflowing and resolving a backtrace, the function now simply returns `epsilon = infinity`

PR #708, opened 2023-04-18
remove backtrace dependency

### LINKED ISSUE #670: Deprecate `RelationDebug`
Some relations pass richer debug information back to the user via errors. Unfortunately, symbol resolution for error construction is very slow, especially on Linux. In most cases, this doesn't matter. But in relations, we should be more careful not to throw exceptions, because relations are used in hot code paths.

An alternative is to only resolve symbols if the error variant isn't RelationDebug.

Shoeboxam:
Closes #670

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -36,7 +36,6 @@ opendp_derive = { path = "opendp_derive" }
 rand = "0.7.3"
 num = "0.3.1"
 thiserror = "1.0.24"
-backtrace = "0.3"
 statrs = "0.13.0"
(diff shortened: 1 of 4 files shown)


Shoeboxam:
* main
  * **PR #708** <a href="https://app.graphite.dev/github/pr/opendp/opendp/708" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/opendp/opendp/708?utm_source=stack-comment).

Shoeboxam:
Closing, as we're indeed preserving other perf adjustments from #691.

PR #730, opened 2023-05-11
Python `prelude` module

### LINKED ISSUE #711: Add a prelude for OpenDP Python
When you want to use the library, you oftentimes need a large set of imports:

```python
from opendp.combinators import *
from opendp.transformations import *
from opendp.measurements import *
from opendp.domains import *
from opendp.metrics import *
from opendp.measures import *
from opendp.typing import *
from opendp.accuracy import *
```

It would be convenient if you could access all of these from a prelude module:
```python
import opendp.prelude as odp
```
or
```python
from opendp.prelude import *
```

Shoeboxam:
Closes #711 

This PR addresses comments left in the issue.

The conventional library import would be:
```python
import opendp.prelude as dp
```

I'm not placing these imports in the `__init__.py` because adding a new module...
* gives more flexibility to preserve module structure but shorten names
	* I think we should keep module distinctions for transformations, measurements and combinators
	* The prelude makes these available under `t`, `m`, `c` respectively: `dp.t.make_clamp(...)`
* doesn't pollute the `opendp` namespace
	* `smartnoise` and the higher-level api both want entries in the `opendp` namespace

The name of the new module is `prelude` because...
* "prelude" is already a common name for when you want to shorten imports
* It implies that imports under this path are aliases, not a separate copy of the code
	* For example, naming it `core` or `base` might give the wrong impression


Shoeboxam:
* main
  * **PR #730** <a href="https://app.graphite.dev/github/pr/opendp/opendp/730" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/opendp/opendp/730?utm_source=stack-comment).

andrewvyrros (approved):
I'm liking the `dp` prefix with the prelude!

andrewvyrros (approved):

PR #729, opened 2023-05-11
free supporting elements

Shoeboxam:
Closes #727

PR #731, opened 2023-05-11
free supporting elements (fix symbols)

### LINKED ISSUE #727: Add free functions for supporting elements

Shoeboxam:
Closes #727 

Fixes a bug in #729 that wasn't caught by testing, likely because `__del__` isn't guaranteed to run.


andrewvyrros (approved):
Doh! So weird that CI didn't catch this.

PR #756, opened 2023-06-02
update sparse histogram api

### LINKED ISSUE #755: update ALP APIs
Before we make the approximate laplace projections available to Python, there needs to be some adjustments to the API. 

* use the dom/met convention
* adjust the constructor names to be clearer about what it is building
* adjust the names and types of constructor arguments
* optionally pull domain size in from the input domain

Shoeboxam:
Closes #755

The constructors are now named:
* `make_alp_state_with_hashers`
* `make_alp_state`
* `make_alp_queryable`

Just like before, each constructor calls the previous one. The `make_alp_queryable` constructor is the one meant to be used from Python (implemented in #747)

Generics have been renamed:
* `C` -> `CI` for "input count"
* `T` -> `CO` for "output count"

Arguments have been renamed/reordered to be more consistent with other constructors, to highlight the repetition in limit/upper bound arguments, and to make optional arguments trailing. 

These constructors now also follow the (new) library convention of passing the input domain and input metric in as the first two arguments of the constructor, and read what they can from the input.

What should be the only substantive change to the core algorithm is on line 103-- it seems that if the cast of x is overflowing a u64, it should be clamped down to `u64::MAX`, not filled in with the default of zero. Will need to validate that this is the correct behavior.

### DIFF
--- a/rust/src/domains/mod.rs
+++ b/rust/src/domains/mod.rs
@@ -117,6 +117,9 @@ impl<T: CheckAtom> AtomDomain<T> {
         }
         Ok(())
     }
+    pub fn bounds(&self) -> Option<&Bounds<T>> {
+        self.bounds.as_ref()
+    }
 }
 impl<T: CheckAtom + InherentNull> AtomDomain<T> {
     pub fn new_nullable() -> Self {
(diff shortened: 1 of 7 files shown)


Shoeboxam:
* main
  * **PR #756** <a href="https://app.graphite.dev/github/pr/opendp/opendp/756" target="_blank">Graphite</a>  👈
    * **PR #747** <a href="https://app.graphite.dev/github/pr/opendp/opendp/747" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/opendp/opendp/756?utm_source=stack-comment).

ChristianLebeda (dismissed):
Looks good. Thanks for updating the api!

Good catch on line 103. The overflow should indeed be clamped down.

andrewvyrros (approved):
Looks good pending rebase.

andrewvyrros (approved):

PR #763, opened 2023-06-08
Polars: add `SeriesDomain`

Shoeboxam:
Closes #762

PR #769, opened 2023-06-09
Polars: `lazyframe_domain` ffi

### LINKED ISSUE #768: add `lazyframe_domain` ffi

Shoeboxam:
Closes #768

### DIFF
--- a/python/src/opendp/_convert.py
+++ b/python/src/opendp/_convert.py
@@ -2,7 +2,7 @@ from typing import Sequence, Tuple, List, Union, Dict, cast
 from inspect import signature
 
 from opendp._lib import *
-from opendp.mod import UnknownTypeException, OpenDPException, Transformation, Measurement, SMDCurve, Queryable
+from opendp.mod import Domain, UnknownTypeException, OpenDPException, Transformation, Measurement, SMDCurve, Queryable
 from opendp.typing import RuntimeType, RuntimeTypeDescriptor, Vec
 
 ATOM_MAP = {
@@ -193,6 +193,12 @@ def _slice_to_py(raw: FfiSlicePtr, type_name: Union[RuntimeType, str]) -> Any:
         if type_name == "String":
             return _slice_to_string(raw)
         
+        if type_name == "LazyFrame":
+            return _slice_to_lazyframe(raw)
+        
+        if type_name == "DataFrame":
+            return _slice_to_dataframe(raw)
+            
         if type_name == "Series":
             return _slice_to_series(raw)
 
@@ -233,6 +239,12 @@ def _py_to_slice(value: Any, type_name: Union[RuntimeType, str]) -> FfiSlicePtr:
         
         if type_name == "Series":
             return _series_to_slice(value)
+        
+        if type_name == "LazyFrame":
+            return _lazyframe_to_slice(value)
+        
+        if type_name == "DataFrame":
+            return _dataframe_to_slice(value)
     
     if isinstance(type_name, RuntimeType):
         if type_name.origin == "Vec":
@@ -326,6 +338,13 @@ def _vector_to_slice(val: Sequence[Any], type_name: RuntimeType) -> FfiSlicePtr:
         def str_to_slice(val):
             return ctypes.c_char_p(val.encode())
         array = (ctypes.c_char_p * len(val))(*map(str_to_slice, val))
+        return _wrap_in_slice(array, len(val))
+    
+    if inner_type_name == "SeriesDomain":
+        # define the ctype of an array of domains
+        domain_array_type = (Domain * len(val)) # type: ignore[operator]
+        # create an instance of a ctype array of domains
+        array = domain_array_type(*val) # type: ignore[operator]
         return _wrap_in_slice(array, len(val))
 
     if inner_type_name not in ATOM_MAP:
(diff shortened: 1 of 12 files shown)


Shoeboxam:
* main
        * **PR #769** <a href="https://app.graphite.dev/github/pr/opendp/opendp/769" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/opendp/opendp/769?utm_source=stack-comment).

python/src/opendp/_convert.py
```diff
@@ -327,6 +339,10 @@ def str_to_slice(val):
             return ctypes.c_char_p(val.encode())
         array = (ctypes.c_char_p * len(val))(*map(str_to_slice, val))
         return _wrap_in_slice(array, len(val))
+    
+    if inner_type_name == "SeriesDomain":
+        array = (Domain * len(val))(*val) # type: ignore[operator]
```
mccalluc, line 344: `Domain * len(val)` is a callable? More explanation, in general, and about the type ignore, would be nice.

python/test/test_polars.py
```diff
@@ -1,16 +1,24 @@
...
     pl = pytest.importorskip("polars")
     return [
-        dp.series_domain("A", dp.atom_domain(T=float)),
-        dp.series_domain("B", dp.atom_domain(T=int)),
-        dp.series_domain("C", dp.option_domain(dp.atom_domain(T=str))),
-        dp.series_domain("D", dp.atom_domain(T=int)),
+        dp.series_domain("A", dp.atom_domain(T=dp.f64)),
+        dp.series_domain("B", dp.atom_domain(T=dp.i32)),
+        dp.series_domain("C", dp.option_domain(dp.atom_domain(T=dp.String))),
+        dp.series_domain("D", dp.atom_domain(T=dp.i32)),
```
mccalluc, line 21: If the python types also work, it'd be good to test them. If they don't, I'd worry about usability -- Not necessarily something to address now, but a comment pointing to an issue would be good if it's not something for right now.

python/test/test_polars.py
```diff
@@ -22,6 +30,31 @@ def test_infer():
...
+    return dp.with_margin(domain, by=by, public_info="lengths", max_partition_size=50)
+
+
+def test_lazyframe_domain_infer():
+    _, lf = example_lf()
+    assert dp.infer_lazyframe_domain(lf).member(lf)
+
+
+def test_domains():
+    with optional_dependency("polars"):
```
mccalluc, line 56: `pytest.importskip` is probably preferable to this, because it will explicitly flag that the test was skipped, rather than giving us a pass when the test is a no-op.

python/test/test_polars.py
```diff
@@ -30,3 +63,11 @@ def test_series_ffi(domain, series):
     """ensure that series can be passed to/from Rust"""
     t_ident = (domain, dp.symmetric_distance()) >> dp.t.then_identity()
     assert t_ident(series).equals(series)
+
+
+def test_lazyframe_ffi():
+    """ensure that lazyframes can be passed to/from Rust"""
+    with optional_dependency("polars"):
```
mccalluc, line 70: ditto: prefer `pytest.importskip`

rust/src/data/ffi.rs
```diff
@@ -175,11 +177,45 @@ pub extern "C" fn opendp_data__slice_as_object(
...
+        
+        Ok(AnyObject::new(DataFrame::new(series)?))
+    }
 
+    #[cfg(feature = "polars")]
+    pub fn deserialize_raw<T>(
+        raw: &FfiSlice, name: &str
+    ) -> Fallible<T> where for<'de> T: Deserialize<'de> {
+        let slice = unsafe { slice::from_raw_parts(raw.ptr as *const u8, raw.len) };
+        // https://github.com/pola-rs/pyo3-polars/blob/5150d4ca27c287ff4be5cafef243d9a878a8879d/pyo3-polars/src/lib.rs#L147-L153
```
mccalluc, line 196: What does this URL do for us? It looks similar to what we're doing, so even if we're just citing sources, wouldn't hurt to make that explicit:
```suggestion
        // Error checking based on pyo3-polars:
        // https://github.com/pola-rs/pyo3-polars/blob/5150d4ca27c287ff4be5cafef243d9a878a8879d/pyo3-polars/src/lib.rs#L147-L153
```

rust/src/domains/polars/frame/ffi.rs
```diff
@@ -0,0 +1,202 @@
...
+};
+
+use super::{Frame, FrameDomain, LazyFrameDomain};
+use polars::prelude::*;
+
+#[bootstrap(
+    arguments(series_domains(rust_type = "Vec<SeriesDomain>")),
+    returns(c_type = "FfiResult<AnyDomain *>")
+)]
+#[allow(dead_code)]
```
mccalluc, line 23: explain why it's needed?

rust/src/domains/polars/frame/ffi.rs
```diff
@@ -0,0 +1,202 @@
...
+    let max_influenced_partitions = if let Some(x) = util::as_ref(max_influenced_partitions) {
+        Some(*try_!(x.downcast_ref::<u32>()))
+    } else {
+        None
+    };
+
+    let public_info = if let Some(public_info) = try_!(to_option_str(public_info)) {
+        Some(match public_info {
+            "keys" => MarginPub::Keys,
+            "lengths" => MarginPub::Lengths,
```
mccalluc, line 127: This corresponds to this test helper?
```
def _with_margin(domain, by):
    return dp.with_margin(domain, by=by, public_info="lengths", max_partition_size=50)
```

I feel weird about passing in strings when we really just need to convey a boolean.

Also: It looks like "lengths" is exercised in the tests. Should we have some coverage for "keys"?

python/test/test_polars.py
Shoeboxam: An annoying aspect is that Polars defaults to `i64`, but the default inferred type of `int` is `i32`. So if you don't specify the types exactly, then the domains will each expect `i32` but the data (a few lines below) is actually `i64`.

rust/src/domains/polars/frame/ffi.rs
Shoeboxam: removed!

rust/src/domains/polars/frame/ffi.rs
Shoeboxam: There's actually three states: no public info, known keys, and known lengths. I think the stringly-typed argument is clearer. An enum would be best, but I don't have FFI for it.

Shoeboxam: Added another test for initializing with keys.

mccalluc (dismissed):
Thanks. Changes look good.

PR #795, opened 2023-06-13
Polars: add `ExprDomain`

### LINKED ISSUE #772: add `ExprDomain`
More consideration is needed! This also doesn't gel with discussions had with @andrewvyrros (will need to discuss more).

An `ExprDomain` consists of the set of dataframes reachable upon applying a given expression within a context:

```rust
struct ExprDomain {
    lazy_frame_domain: LazyFrameDomain,
    context: Context,
    active_column: String
}

enum Context {
    Select,
    GroupBy(Vec<String>),
    Agg(Vec<String>),
    Filter,
    WithColumns
}
```

The `ExprDomain`'s carrier type is a tuple, consisting of the seed `LazyFrame` and an `Expr` to be applied to it:

```rust
impl Domain for ExprDomain {
    type Carrier = (LazyFrame, Expr);
    fn member(&self, val: &LazyFrame) -> Fallible<bool> {
        let frame = match self.context {
            Context::Select => val.0.select([val.1]),
            Context::GroupBy(grouping_cols) => val.0,
            Context::Agg(grouping_cols) => val.0.group_by(grouping_cols).agg(val.1),
            Context::Filter => val.0.filter(val.1),
            Context::WithColumns => val.0.with_columns([val.1]),
            _ => unimplemented!()
        };
        self.lazy_frame_domain.member(frame)
    }
}
```
In the member check, you apply the expr, then run the lazy_frame_domain check.

ankke:
Closes #772

### DIFF
--- a/python/src/opendp/_convert.py
+++ b/python/src/opendp/_convert.py
@@ -201,6 +201,9 @@ def _slice_to_py(raw: FfiSlicePtr, type_name: Union[RuntimeType, str]) -> Any:
             
         if type_name == "Series":
             return _slice_to_series(raw)
+        
+        if type_name == "Expr":
+            return _slice_to_expr(raw)
 
     if isinstance(type_name, RuntimeType):
         if type_name.origin == "Vec":
@@ -245,6 +248,9 @@ def _py_to_slice(value: Any, type_name: Union[RuntimeType, str]) -> FfiSlicePtr:
         
         if type_name == "DataFrame":
             return _dataframe_to_slice(value)
+            
+        if type_name == "Expr":
+            return _expr_to_slice(value)
     
     if isinstance(type_name, RuntimeType):
         if type_name.origin == "Vec":
@@ -391,6 +397,14 @@ def _tuple_to_slice(val: Tuple[Any, ...], type_name: Union[RuntimeType, str]) ->
     if len(inner_type_names) != len(val):
         raise TypeError("type_name members must have same length as tuple")
 
+    if inner_type_names == ['LogicalPlan', 'Expr']:
+        lf_slice = _lazyframe_to_slice(val[0])
+        expr_slice = _expr_to_slice(val[1])
+        array = (ctypes.c_void_p * len(val))(ctypes.cast(lf_slice, ctypes.c_void_p), ctypes.cast(expr_slice, ctypes.c_void_p))
+        out = _wrap_in_slice(ctypes.pointer(array), 2)
+        out.depends_on(lf_slice, expr_slice)
+        return out
+
     for t in inner_type_names:
         if t not in ATOM_MAP:
             raise TypeError(f"Tuple members must be one of {ATOM_MAP.keys()}. Got {t}")
@@ -414,6 +428,12 @@ def _slice_to_tuple(raw: FfiSlicePtr, type_name: RuntimeType) -> Tuple[Any, ...]
     void_array_ptr = ctypes.cast(raw.contents.ptr, ctypes.POINTER(ctypes.c_void_p))
     # list of void*
     ptr_data = void_array_ptr[0:raw.contents.len]
+
+    if inner_type_names == ['LogicalPlan', 'Expr']:
+        lp_slice = ctypes.cast(ptr_data[0], FfiSlicePtr)
+        expr_slice = ctypes.cast(ptr_data[1], FfiSlicePtr)
+        return _slice_to_lazyframe(lp_slice), _slice_to_expr(expr_slice)
+    
     # tuple of instances of Python types
     return tuple(ctypes.cast(void_p, ctypes.POINTER(ATOM_MAP[name])).contents.value # type: ignore
                  for void_p, name in zip(ptr_data, inner_type_names))
@@ -467,6 +487,21 @@ def _slice_to_lazyframe(raw: FfiSlicePtr):
     slice_array = ctypes.cast(raw.contents.ptr, ctypes.POINTER(ctypes.c_uint8))
     lf.__setstate__(bytes(slice_array[0:raw.contents.len]))
     return lf
+
+
+def _expr_to_slice(val) -> FfiSlicePtr:
+    state = val.__getstate__()
+    raw = _wrap_in_slice(state, len(state))
+    raw.depends_on(state)
+    return raw
+
+
+def _slice_to_expr(raw: FfiSlicePtr):
+    pl = import_optional_dependency('polars')
+    expr = pl.all()
+    slice_array = ctypes.cast(raw.contents.ptr, ctypes.POINTER(ctypes.c_uint8))
+    expr.__setstate__(bytes(slice_array[0:raw.contents.len]))
+    return expr
 
 def _dataframe_to_slice(val) -> FfiSlicePtr:
     slices = list(_series_to_slice(s) for s in val.get_columns())
--- a/python/src/opendp/domains.py
+++ b/python/src/opendp/domains.py
@@ -210,7 +210,6 @@ def domain_type(
 
 def expr_domain(
     lazyframe_domain: Domain,
-    context: Optional[str] = None,
     grouping_columns: Optional[List[str]] = None
 ) -> Domain:
     r"""Construct an ExprDomain from a LazyFrameDomain.
@@ -219,9 +218,7 @@ def expr_domain(
 
     :param lazyframe_domain: the domain of the LazyFrame to be constructed
     :type lazyframe_domain: Domain
-    :param context: used when the constructor is called inside a lazyframe context constructor
-    :type context: str
-    :param grouping_columns: used when the constructor is called inside a groupby context constructor
+    :param grouping_columns: set when creating an expression that aggregates
     :type grouping_columns: List[str]
     :rtype: Domain
     :raises TypeError: if an argument's type differs from the expected type
@@ -233,15 +230,14 @@ def expr_domain(
     # No type arguments to standardize.
     # Convert arguments to c types.
     c_lazyframe_domain = py_to_c(lazyframe_domain, c_type=Domain, type_name=None)
-    c_context = py_to_c(context, c_type=ctypes.c_char_p, type_name=None)
     c_grouping_columns = py_to_c(grouping_columns, c_type=AnyObjectPtr, type_name=RuntimeType(origin='Option', args=[RuntimeType(origin='Vec', args=[String])]))
 
     # Call library function.
     lib_function = lib.opendp_domains__expr_domain
-    lib_function.argtypes = [Domain, ctypes.c_char_p, AnyObjectPtr]
+    lib_function.argtypes = [Domain, AnyObjectPtr]
     lib_function.restype = FfiResult
 
-    output = c_to_py(unwrap(lib_function(c_lazyframe_domain, c_context, c_grouping_columns), Domain))
+    output = c_to_py(unwrap(lib_function(c_lazyframe_domain, c_grouping_columns), Domain))
 
     return output
 
--- a/python/src/opendp/domains.py
+++ b/python/src/opendp/domains.py
@@ -22,6 +22,7 @@ __all__ = [
     "domain_carrier_type",
     "domain_debug",
     "domain_type",
+    "expr_domain",
     "infer_lazyframe_domain",
     "lazyframe_domain",
     "map_domain",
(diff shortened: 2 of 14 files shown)


rust/src/domains/polars/expr/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+
+#[derive(Clone, PartialEq)]
+pub struct ExprDomain {
+    pub lazy_frame_domain: LazyFrameDomain,
+    pub context: Context,
+    pub active_column: Option<String>,
+}
+
+impl Domain for ExprDomain {
+    type Carrier = (Expr, LazyFrame);
```
Shoeboxam, line 24: Do you think this is easier to read if it's `(LazyFrameDomain, Expr)`?

rust/src/domains/polars/expr/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+        let lazy_frame = val_lazy_frame.clone();
+
+        let result_frame = match self.context {
+            Context::Select => lazy_frame.select([expr]),
+            Context::Agg { ref columns } => {
+                let column_expr: Vec<_> = columns
+                    .clone()
+                    .into_iter()
+                    .map(move |c| col(c.as_str()))
+                    .collect();
```
Shoeboxam, line 37: Does this work?
```rust
let column_exprs: Vec<_> = columns.iter().map(|c| col(c.as_ref())).collect();
```
The `as_ref()` takes you from `String` to `&str`.

rust/src/domains/polars/expr/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+            .field("lazy_frame_domain", &self.lazy_frame_domain)
+            .finish()
+    }
+}
+
+impl<D: DatasetMetric> MetricSpace for (ExprDomain, D) {
+    fn check(&self) -> bool {
+        true
+    }
+}
```
Shoeboxam, line 60: This is tricky, it actually isn't always a valid pairing. In the case where `D` is `ChangeOneDistance` or `HammingDistance`, there should be a margin. We should also define what the margin means-- are the grouping values guaranteed to be exhaustive (no unique values that are not listed in the margin)? 
If yes/is exhaustive, then we don't need a join in the groupby, and the check here can be a simple check to see if margins are non-empty. 

Seems like we should continue with the definition that the margin should be exhaustive, as it will help keep things simpler.

Shoeboxam: If the margin isn't exhaustive, then it would only be a valid pairing with bounded metrics if there were a margin with zero grouping keys, denoting the size of the entire dataframe.

Shoeboxam (changes requested):
Great work, thanks for opening a PR.

rust/src/domains/polars/expr/mod.rs
Shoeboxam: Pending feedback from FSO.

PaulineMauryL: The margin would be exhaustive (with ideally the exception of potential null values if possible).

python/src/opendp/_convert.py
```diff
@@ -391,6 +397,14 @@ def _tuple_to_slice(val: Tuple[Any, ...], type_name: Union[RuntimeType, str]) ->
     if len(inner_type_names) != len(val):
         raise TypeError("type_name members must have same length as tuple")
 
+    if inner_type_names == ['LogicalPlan', 'Expr']:
```
mccalluc, line 400: In the rest of this function, the order doesn't seem to matter. If we saw `[ 'Expr', 'LogicalPlan']` would that just be a bug? Maybe compare as a set?

python/src/opendp/_convert.py
```diff
@@ -414,6 +428,12 @@ def _slice_to_tuple(raw: FfiSlicePtr, type_name: RuntimeType) -> Tuple[Any, ...]
     void_array_ptr = ctypes.cast(raw.contents.ptr, ctypes.POINTER(ctypes.c_void_p))
     # list of void*
     ptr_data = void_array_ptr[0:raw.contents.len]
+
+    if inner_type_names == ['LogicalPlan', 'Expr']:
```
mccalluc, line 432: ditto. Or pull this out, if it's reused.

python/src/opendp/domains.py
```diff
@@ -207,6 +208,44 @@ def domain_type(
...
 
 
+def expr_domain(
+    lazyframe_domain: Domain,
+    context: Optional[str] = None,
+    grouping_columns: Optional[Any] = None
+) -> Domain:
+    r"""Construct an ExprDomain from a LazyFrameDomain.
+
+    Must pass either `context` or `grouping_columns`.
```
mccalluc, line 218: This is generated code, I think? If it were manual, I'd expect there to be a check if both `context` and `grouping_columns` have been provided. I guess that would come from the Rust code?

python/src/opendp/domains.py
```diff
@@ -207,6 +208,44 @@ def domain_type(
     return output
 
 
+def expr_domain(
+    lazyframe_domain: Domain,
+    context: Optional[str] = None,
+    grouping_columns: Optional[Any] = None
```
mccalluc, line 214: Expecting something like `Optional[List[str]]`... but I also wonder if it really needs to be optional: Since it seems to be a list, maybe an empty list can handle that?

python/test/test_polars.py
```diff
@@ -30,6 +30,7 @@ def test_infer():
     assert dp.RuntimeType.infer(series) == "Series"
     assert dp.RuntimeType.infer(pl.DataFrame(series)) == "DataFrame"
     assert dp.RuntimeType.infer(pl.LazyFrame(series)) == "LazyFrame"
+    assert dp.RuntimeType.infer(pl.col("A")) == "Expr"
```
mccalluc, line 33: As this gets longer, it feels more like a test that should be parameterized:
```
@pytest.mark.parametrize("expected_type,obj", {
    'Series': series,
    'DataFrame': pl.DataFrame(series),
    'LazyFrame': pl.LazyFrame(series),
    'Expr': pl.col("A")
}.items())
```

python/test/test_polars.py
```diff
@@ -71,3 +72,11 @@ def test_lazyframe_ffi():
     lf_domain, lf = example_lf()
     t_ident = (lf_domain, dp.symmetric_distance()) >> dp.t.then_identity()
     assert t_ident(lf).collect().equals(lf.collect())
+
+def test_expr_ffi():
+    """ensure that lazyframes can be passed to/from Rust"""
+    pl = pytest.importorskip("polars")
+    lf_domain, lf = example_lf()
+    expr_domain = dp.expr_domain(lf_domain, "select")
+    t_ident = (expr_domain, dp.symmetric_distance()) >> dp.t.then_identity()
```
mccalluc, line 81: It's not obvious that the assertion depends on these lines. Maybe stronger assertions? Or more explanation? Or drop these?

mccalluc (commented):
Another half review. Finish this up in a bit.

rust/src/domains/polars/expr/ffi.rs
```diff
@@ -0,0 +1,89 @@
...
+
+    let context = if let Some(context_str) = try_!(to_option_str(context)) {
+        match context_str.to_lowercase().as_str() {
+            "select" => ExprContext::Select,
+            "filter" => ExprContext::Filter,
+            "with_columns" => ExprContext::WithColumns,
+            _ => {
+                return err!(
+                FFI,
+                "context must be one of `select`, `filter` or `with_columns`, or pass `grouping_columns`"
```
mccalluc, line 56: Should there be a check earlier that either `context` or `grouping_columns` but not both have been passed? This message could then probably be simplified:

```suggestion
                "context must be one of `select`, `filter` or `with_columns`"
```

mccalluc: Or would it be simpler to just have two separate functions, in Rust and in Python? We could get rid of all the run-time error checking, and we wouldn't have to worry about Python users having to deal with Rust errors.

rust/src/domains/polars/expr/mod.rs
```diff
@@ -0,0 +1,214 @@
...
+use crate::traits::ProductOrd;
+use crate::transformations::DatasetMetric;
+use crate::{core::Domain, error::Fallible};
+
+use super::{Frame, FrameDomain, LogicalPlanDomain, NumericDataType, SeriesDomain};
+
+#[cfg(feature = "ffi")]
+mod ffi;
+
+// TODO: remove this allow marker later
```
mccalluc, line 20: Link to issue so we don't lose track of this?

rust/src/domains/polars/expr/mod.rs
```diff
@@ -0,0 +1,214 @@
...
+            ExprContext::Filter => frame.filter(expr),
+            ExprContext::WithColumns => frame.with_columns([expr]),
+            ExprContext::Aggregate(columns) => frame
+                .group_by(&columns.iter().map(|n| col(n)).collect::<Vec<_>>())
+                .agg([expr]),
+        }
+        .logical_plan
+    }
+
+    pub fn break_alignment(&self) -> Fallible<()> {
```
mccalluc, line 49: I think I've seen this referenced in other PRs, but if we want each PR to stand on its own, then should there be some reference to this method here?

rust/src/domains/polars/expr/mod.rs
```diff
@@ -0,0 +1,214 @@
...
+            _ => BTreeSet::new(),
+        }
+    }
+}
+
+/// # Proof Definition
+/// `ExprDomain` is the domain of all polars expressions that can be applied to a `LazyFrame` where:
+/// * `lazy_frame_domain` - `LazyFrameDomain`.
+/// * `context` - Context in which expression is to be applied.
+/// * `active_column` - Column to which expression is to be applied.
```
mccalluc, line 71: Should this correspond to the code in the struct? It has `frame_domain` instead of `lazy_frame_domain`, and doesn't mention `active_column`.

mccalluc (commented):
Some Rust testing would be good to have.

python/src/opendp/_convert.py
Shoeboxam: I specifically only worked out FFI for the `(LogicalPlan, Expr)` tuple, in that order, because that is the carrier type for `ExprDomain`. I don't really see a need to add FFI for other variations at this point.

python/src/opendp/domains.py
Shoeboxam: Yep, those checks happen in the Rust code.

python/src/opendp/domains.py
Shoeboxam: I added another note to the bootstrap. This is trying to map to the ExprContext enum- if context is set, grouping columns shouldn't be.

python/test/test_polars.py
Shoeboxam: I don't think works with optional pl?

python/test/test_polars.py
Shoeboxam: The primary purpose of this is to ensure that the FFI is working properly. The identity function tests both data loading and unloading.

rust/src/domains/polars/expr/ffi.rs
Shoeboxam: Places where expressions can be used in Polars:

1. To build a sorting key
2. Fill null/fill NaN
3. Filter
4. Select
5. Group by keys
6. Group by agg
7. Join keys
8. With column
9. Explode
10. Drop nulls
11. Top/bottom k
12. Shift

For our use, I think these fall into four categories:
1. Not useful on their own for DP (11, 12)
2. Must be leaf nodes, like only col or lit (2, 5, 7, 9, 10)
3. Must be row-by-row (1, 3, 8)
4. Aggregates (4, 6)

We don't need Expression contexts for categories one or two, leaving only row-by-row and aggregates. I'll change the API to take an optional list of grouping keys. If set, then you are in category four (aggregates), otherwise in category three (row-by-row). When the grouping keys are empty, you have selection, otherwise groupby/agg.

Shoeboxam: I forgot-- delineating how exactly the expression will be used is necessary for the member check.

rust/src/domains/polars/expr/mod.rs
Shoeboxam: I'll just leave it as a warning for now.

python/test/test_polars.py
mccalluc: good point -- could wrap this in a class to group all the tests that need the `pl`, but not worth it.

python/src/opendp/domains.py
mccalluc: Looks like it's updated. Thanks!

mccalluc (approved):
Spot checks on comments look good. Not able to select the most recent changes, so I won't re-read everything: going ahead and approving. Let me know if anything should get a closer look.

PR #774, opened 2023-06-12
Lp metric

Shoeboxam:
Closes #773

Generalizes the LpDistance metric to be higher-order. See notes in issue.

PR #771, opened 2023-06-12
Polars: add `CsvDomain`, `make_scan_csv`

Shoeboxam:
Closes #770

PR #797, opened 2023-06-13
Polars: add `make_expr_col`

ankke:


ankke:
* main
        * **PR #769** <a href="https://app.graphite.dev/github/pr/opendp/opendp/769" target="_blank">Graphite</a> 
          * **PR #771** <a href="https://app.graphite.dev/github/pr/opendp/opendp/771" target="_blank">Graphite</a> 
            * **PR #795** <a href="https://app.graphite.dev/github/pr/opendp/opendp/795" target="_blank">Graphite</a> 
              * **PR #797** <a href="https://app.graphite.dev/github/pr/opendp/opendp/797" target="_blank">Graphite</a>  👈
            * **PR #774** <a href="https://app.graphite.dev/github/pr/opendp/opendp/774" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/opendp/opendp/797?utm_source=stack-comment).

Shoeboxam:
Can you open an issue for this PR and add the issue to the Polars milestone? In the future, we also try to prefix branch names with the issue number they resolve. Thanks!

rust/src/transformations/polars/make_col/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+
+    #[test]
+    fn test_make_col_expr_no_wildcard() -> Fallible<()> {
+        let (expr_domain, lazy_frame) = get_test_data();
+        let selected_col = "B";
+
+        let expression = make_col(expr_domain, SymmetricDistance::default(), selected_col)?;
+        let error_res = expression
+            .invoke(&(col(selected_col), lazy_frame))
+            .map(|v| v.0)
```
Shoeboxam, line 165: I don't think you need this map

rust/src/transformations/polars/make_col/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+            },
+        ),
+        input_metric,
+        SymmetricDistance::default(),
+        StabilityMap::new_from_constant(1),
+    )
+}
+
+#[cfg(test)]
+mod test_make_col {
```
Shoeboxam, line 78: Great tests!

Shoeboxam (changes requested):
This looks great! I made a few comments but it's definitely on the right track.

One thing you could try adding is a generic, `M: DatasetMetric`, so that the col can be used with other dataset metrics. Be sure to add a trait bound `(ExprDomain, M): MetricSpace`. If you miss this `MetricSpace` bound, the compiler tends to give really strange/confusing errors.

rust/src/transformations/polars/make_col/mod.rs
Shoeboxam: You do need this map, I'm crazy.

rust/src/transformations/polars/make_col/mod.rs
```diff
@@ -0,0 +1,191 @@
...
+use polars::prelude::*;
+use std::collections::BTreeSet;
+
+use crate::core::{Function, MetricSpace, StabilityMap, Transformation};
+use crate::domains::{Context, ExprDomain, ExprMetric, LazyFrameDomain};
+use crate::error::*;
+use crate::traits::DistanceConstant;
+
+#[bootstrap(ffi = false)]
+/// Make a Transformation that returns a col(<column_name>) expression for a Lazy Frame.
```
Shoeboxam, line 12: Use backticks instead of <> because <> tend to cause formatting issues downstream, and backticks trigger nice rendering in docs.

rust/src/transformations/polars/make_col/mod.rs
```diff
@@ -0,0 +1,191 @@
...
+///
+/// # Arguments
+/// * `input_domain` - Domain of the expression to be applied.
+/// * `input_metric` - The metric space under which neighboring LazyFrames are compared.
```
Shoeboxam, line 16: Omit space!

rust/src/transformations/polars/make_col/mod.rs
```diff
@@ -0,0 +1,191 @@
...
+                        FailedFunction,
+                        "make_col has to be the first expression in the expression chain"
+                    );
+                }
+                Ok((frame.clone(), col(&*col_name)))
+            },
+        ),
+        input_metric.clone(),
+        input_metric,
+        StabilityMap::new_from_constant(M::Distance::one()),
```
Shoeboxam, line 86: Minor nit, but can you change this to `StabilityMap::new(Clone::clone)`? Then the trait bound above can just be `M::Distance: Clone,`

rust/src/transformations/polars/mod.rs
```diff
@@ -26,4 +26,40 @@ pub use scan_parquet::*;
...
+#[cfg(feature = "contrib")]
+pub use make_col::*;
+
+#[cfg(test)]
+mod test {
+    use crate::domains::{AtomDomain, ExprDomain, LazyFrameDomain, SeriesDomain, LazyFrameContext};
+    use crate::error::*;
+    use polars::prelude::*;
+
+    pub fn get_test_data() -> Fallible<(ExprDomain<LazyFrameContext>, LazyFrame, LazyFrameDomain)> {
```
Shoeboxam, line 41: How do you feel about dropping `LazyFrameDomain` from the return? If you ever need it, can do `expr_domain.lazy_frame_domain`.

rust/src/transformations/polars/make_col/mod.rs
```diff
@@ -0,0 +1,191 @@
...
+    M::Distance: DistanceConstant<M::Distance> + One + Clone,
+    (ExprDomain<C>, M): MetricSpace,
+{
+    if input_domain
+        .lazy_frame_domain
+        .column(col_name.as_str())
+        .is_none()
+    {
+        return fallible!(MakeTransformation, "{} is not in dataframe", col_name);
+    }
```
Shoeboxam, line 38: We have another helper for this now- `input_domain.lazy_frame_domain.try_column(col_name.as_str())?`

Shoeboxam: I think there should also be a check that the input domain's selected column is `None`.

rust/src/transformations/polars/make_col/mod.rs
```diff
@@ -0,0 +1,191 @@
...
+    use crate::transformations::polars::test::get_test_data;
+
+    use super::*;
+
+    #[test]
+    fn test_make_col_expr() -> Fallible<()> {
+        let (expr_domain, lazy_frame, _) = get_test_data()?;
+        let selected_col = "B";
+        let transformation =
+            make_col(expr_domain, SymmetricDistance, selected_col.to_string()).unwrap_test();
```
Shoeboxam, line 103: Replace `.unwrap_test()` with `?` throughout.

Shoeboxam (changes requested):

rust/src/transformations/make_stable_expr/expr_col/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+    let col_name = col_name.to_string();
+
+    let mut output_domain = input_domain.clone();
+    output_domain
+        .frame_domain
+        .series_domains
+        .retain(|v| v.field.name == col_name);
+
+    // ensure that column exists in the domain
+    output_domain.active_series()?;
```
mccalluc, line 37:
```suggestion
    output_domain.check_one_column()?;
```

Looking at the definition, it looks like there's a separate function for just the checking? Better to use a function that is explicitly a check, rather than relying on a side-effect.

rust/src/transformations/make_stable_expr/expr_col/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+    Transformation::new(
+        input_domain,
+        output_domain,
+        Function::new_fallible(
+            // in most other situations, we would use `Function::new_expr`, but we need to return a Fallible here
+            move |(plan, expr): &(LogicalPlan, Expr)| -> Fallible<(LogicalPlan, Expr)> {
+                if expr != &Expr::Wildcard {
+                    return fallible!(
+                        FailedFunction,
+                        "make_col has to be the first expression in the expression chain"
```
mccalluc, line 48: Not sure why `Wildcard` implies `make_col`. Might be obvious if I know polars better.

rust/src/transformations/make_stable_expr/expr_col/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+    fn test_make_col_expr() -> Fallible<()> {
+        let (lf_domain, lf) = get_test_data()?;
+        let expr_domain = lf_domain.select();
+        let expr_exp = col("A");
+        let t_col: Transformation<_, _, _, SymmetricDistance> = expr_exp
+            .clone()
+            .make_stable(expr_domain.clone(), SymmetricDistance)?;
+        let expr_res = t_col.invoke(&(lf.logical_plan, all()))?.1;
+
+        assert_eq!(expr_res, expr_exp);
```
mccalluc, line 77:
```suggestion
        let t_col: Transformation<_, _, _, SymmetricDistance> = expr_exp
            .clone()
            .make_stable(expr_domain.clone(), SymmetricDistance)?;
        let expr_result = t_col.invoke(&(lf.logical_plan, all()))?.1;

        assert_eq!(expr_result, col("A"));
```

- Spell out names
- Generally like to keep definitions close to where they are used... or in this case it seems like it could just be inlined?

rust/src/transformations/make_stable_expr/expr_col/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    #[test]
+    fn test_make_col_expr_no_wildcard() -> Fallible<()> {
+        let (lf_domain, lf) = get_test_data()?;
+        let expr_domain = lf_domain.select();
+
+        let expr_exp = col("A");
+        let t_col: Transformation<_, _, _, SymmetricDistance> =
+            expr_exp.make_stable(expr_domain.clone(), SymmetricDistance)?;
```
mccalluc, line 89:
```suggestion
        let t_col: Transformation<_, _, _, SymmetricDistance> =
            col("A").make_stable(expr_domain.clone(), SymmetricDistance)?;
```

If `exp` is `expected`, it's a little odd because we don't end up using it in a comparison.

rust/src/transformations/make_stable_expr/expr_col/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+        let expr_domain = lf_domain.select();
+
+        let expr_exp = col("nonexistent");
+        let error_res = expr_exp
+            .make_stable(expr_domain.clone(), SymmetricDistance)
+            .map(|_: Transformation<_, _, _, SymmetricDistance>| ())
+            .unwrap_err()
+            .variant;
+
+        assert_eq!(error_res, ErrorVariant::FailedFunction);
```
mccalluc, line 112: Do we get a different error message here and above? Consider adding a message check to the test, or make finer-grained error classes, if we don't want to compare strings.

rust/src/transformations/make_stable_expr/mod.rs
```diff
@@ -94,11 +97,56 @@ where
...
+    pub fn get_test_data() -> Fallible<(LazyFrameDomain, LazyFrame)> {
+        let lf_domain = LazyFrameDomain::new(vec![
+            SeriesDomain::new("A", AtomDomain::<i32>::default()),
+            SeriesDomain::new("B", AtomDomain::<f64>::new_closed((0.5, 2.5))?),
+            SeriesDomain::new("C", AtomDomain::<i32>::default()),
+        ])?
+        .with_margin::<&str>(&[], Margin::new().with_max_partition_length(3u32))?
+        .with_margin(&["A"], Margin::new().with_max_partition_length(2u32))?
+        .with_margin(&["B"], Margin::new().with_max_partition_length(2u32))?
+        .with_margin(&["C"], Margin::new().with_max_partition_length(1u32))?;
```
mccalluc, line 132: I don't think "C" is referenced elsewhere -- If it's doing something interesting, add tests, if it's not, then remove?

"A" and "B" are both referenced: assuming the differences between then are significant, maybe have the names reflect that, so if a test fails for one and not the other, you have an clue about how they are different, without reading this code, or going into the debugger?

Something like:
`A` -> `open_partition_2`
`B` -> `closed_partition_2`
`C` -> `open_partition_1`

rust/src/transformations/make_stable_expr/expr_col/mod.rs
Shoeboxam: Adjusted the error in a way that may clarify.

rust/src/transformations/make_stable_expr/expr_col/mod.rs
Shoeboxam: How about this? `expr_exp` (now `expected`) is used in two places.

rust/src/transformations/make_stable_expr/mod.rs
```diff
@@ -70,3 +70,58 @@ where
         }
     }
 }
+
+#[cfg(test)]
+pub mod polars_test {
```
Shoeboxam, line 75: @mccalluc Good place to look!

rust/src/transformations/make_stable_expr/mod.rs
```diff
@@ -45,23 +47,21 @@ pub trait StableExpr<MI: Metric, MO: Metric> {
     ) -> Fallible<Transformation<ExprDomain, ExprDomain, MI, MO>>;
 }
 
-pub trait DatasetOuterMetric: OuterMetric {}
```
Shoeboxam, line 48: @mccalluc This became unnecessary after refactoring aggregators.

rust/src/transformations/make_stable_expr/mod.rs
```diff
@@ -70,3 +70,58 @@ where
...
+            Margin::new()
+                .with_public_keys()
+                .with_max_partition_length(200),
+        )?
+        .with_margin(
+            &["chunk_(..10u32)"],
+            Margin::new()
+                .with_public_keys()
+                .with_max_partition_length(100),
+        )?;
```
mccalluc, line 114: Are these margins chosen for particular tests? Maybe leave these off, and let the individual tests add margins to the domain?

rust/src/transformations/make_stable_expr/mod.rs
```diff
@@ -70,3 +70,58 @@ where
...
+    use crate::error::*;
+    use polars::prelude::*;
+
+    pub fn get_test_data() -> Fallible<(LazyFrameDomain, LazyFrame)> {
+        let lf_domain = LazyFrameDomain::new(vec![
+            SeriesDomain::new("chunk_2_bool", AtomDomain::<bool>::default()),
+            SeriesDomain::new("cycle_5_alpha", AtomDomain::<String>::default()),
+            SeriesDomain::new("const_1f64", AtomDomain::<f64>::default()),
+            SeriesDomain::new("chunk_(..10u32)", AtomDomain::<u32>::default()),
+            SeriesDomain::new("cycle_(..100i32)", AtomDomain::<i32>::default()),
```
mccalluc, line 87: Wait to add these until you have a test that uses them?

rust/src/transformations/make_stable_expr/expr_col/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+    use crate::transformations::{polars_test::get_test_data, StableExpr};
+
+    use super::*;
+
+    #[test]
+    fn test_make_col_expr() -> Fallible<()> {
+        let (lf_domain, lf) = get_test_data()?;
+        let expr_domain = lf_domain.row_by_row();
+        let expected = col("const_1f64");
+        let t_col: Transformation<_, _, _, SymmetricDistance> = expected
```
mccalluc, line 72: Not feasible to infer this type information?

mccalluc (commented):
I can look more on Monday, but a couple comments.

rust/src/transformations/make_stable_expr/expr_col/mod.rs
Shoeboxam: Neat, it can be dropped now.

rust/src/transformations/make_stable_expr/mod.rs
Shoeboxam: I came up with a combination of margins and dtypes that would cover most use-cases. They get used in specific tests later on. I'm not super keen on breaking their attribution apart across the relevant PRs...

PR #814, opened 2023-06-16
dom/met signature: `make_identity`

Shoeboxam:


rust/src/transformations/manipulation/ffi.rs
```diff
@@ -1,95 +1,23 @@
...
-            }
-            dispatch!(monomorphize, [
-                (T, @numbers)
-            ], (M))
-        }
-        _ => err!(FFI, "Monomorphizations for the identity function are only available for VectorDomain<AtomDomain<_>> and AtomDomain<_>").into()
-    }
+    let domain = try_as_ref!(domain).clone();
+    let metric = try_as_ref!(metric).clone();
+    super::make_identity(domain, metric).into()
```
andrewvyrros, line 20: Wow, good to get rid of all that! The issue with `MetricSpace` for `(AnyDomain, AnyMetric)` is tricky, not worth solving here. We don't really use `make_identity` anywhere except the sanity test, so I think just tagging it with a warning and `honest-but-curious` is a reasonable mitigation.

andrewvyrros (approved):

PR #824, opened 2023-06-21
Polars: add `make_with_columns`

ankke:

PR #819, opened 2023-06-19
Polars: add `make_expr_sum`

PaulineMauryL:

PR #829, opened 2023-06-22
Polars: add `make_expr_laplace`

PaulineMauryL:


PaulineMauryL:
* main
        * **PR #769** <a href="https://app.graphite.dev/github/pr/opendp/opendp/769" target="_blank">Graphite</a> 
          * **PR #771** <a href="https://app.graphite.dev/github/pr/opendp/opendp/771" target="_blank">Graphite</a> 
            * **PR #795** <a href="https://app.graphite.dev/github/pr/opendp/opendp/795" target="_blank">Graphite</a> 
              * **PR #819** <a href="https://app.graphite.dev/github/pr/opendp/opendp/819" target="_blank">Graphite</a> 
                * **PR #829** <a href="https://app.graphite.dev/github/pr/opendp/opendp/829" target="_blank">Graphite</a>  👈
              * **PR #797** <a href="https://app.graphite.dev/github/pr/opendp/opendp/797" target="_blank">Graphite</a> 
                * **PR #824** <a href="https://app.graphite.dev/github/pr/opendp/opendp/824" target="_blank">Graphite</a> 
            * **PR #774** <a href="https://app.graphite.dev/github/pr/opendp/opendp/774" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/opendp/opendp/829?utm_source=stack-comment).

rust/src/measurements/polars/make_laplace/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+///
+/// # Arguments
+/// * `input_domain` - ExprDomain
+/// * `input_metric` - The metric space under which neighboring LazyFrames are compared
+/// * `scale` - Noise scale parameter for the laplace distribution. `scale` == sqrt(2) * standard_deviation.
+pub fn make_laplace_expr<
+    C: Context,
+    MI,
+    TA: Float + CastInternalRational + SampleDiscreteLaplaceZ2k + DataTypeFrom + Send + Sync,
+    QI: Number,
```
Shoeboxam, line 27: TA and QI are the same, because L1 distances are always integers on integral data. Therefore QI can be removed. Later on, when doing the gaussian, TA and QI can be different (because the L2 distance can be float, while the data type is integer).

rust/src/measurements/polars/make_laplace/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+>(
+    input_domain: ExprDomain<C>,
+    input_metric: MI,
+    scale: TA,
+    k: Option<i32>,
+) -> Fallible<Measurement<ExprDomain<C>, Expr, MI, MaxDivergence<TA>>>
+where
+    (ExprDomain<C>, MI): MetricSpace,
+    MI: ExprMetric<C, InnerMetric = AbsoluteDistance<QI>, Distance = TA>,
+    ChunkedArray<TA::Polars>: ToVec<TA>,
```
Shoeboxam, line 37: Since we have `.into_no_null_iter`, can this trait bound of `ToVec` be removed?

rust/src/measurements/polars/make_laplace/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+
+    Measurement::new(
+        input_domain,
+        Function::new_fallible(move |(_frame, expr): &(C::Value, Expr)| -> Fallible<Expr> {
+            Ok(expr.clone().map(
+                move |s: Series| {
+                    let noisy_vec: Vec<TA> = s
+                        .unpack::<TA::Polars>()?
+                        .into_no_null_iter()
+                        .map(|value| TA::sample_discrete_laplace_Z2k(scale.clone(), value, k))
```
Shoeboxam, line 63: This PR now sits above the threading PR, so you should be able to use the function from `lap_meas` in here.

rust/src/measurements/polars/make_laplace/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+    MI: ExprMetric<C, InnerMetric = AbsoluteDistance<QI>, Distance = TA>,
+    ChunkedArray<TA::Polars>: ToVec<TA>,
+    i32: ExactIntCast<TA::Bits>,
+    TA::Polars: PolarsNumericType<Native = TA>,
+    Series: NamedFrom<Vec<TA>, [TA]>,
+{
+    if scale.is_sign_negative() {
+        return fallible!(MakeMeasurement, "scale must not be negative");
+    }
+
```
Shoeboxam, line 45: Should adjust to retrieve the expected number of unique values to noise. This number will get embedded within the input domain for `lap_meas`. This way `lap_meas` can calibrate noise according to how many values it must round.

Shoeboxam (changes requested):

python/src/opendp/_lib.py
```diff
@@ -61,18 +61,19 @@ def get_lib_path(name_template, architecture):
...
         except Exception as e:
             raise Exception("Unable to load OpenDP shared library", lib_path, e)
+        
 
     elif os.environ.get('OPENDP_HEADLESS', "false") != "false":
         return None
 
     else:
         raise ValueError("Unable to find lib directory. Consider setting OPENDP_LIB_DIR to a valid directory.")
+
```
mccalluc, line 74:
```suggestion
        
    elif os.environ.get('OPENDP_HEADLESS', "false") != "false":
        return None

    else:
        raise ValueError("Unable to find lib directory. Consider setting OPENDP_LIB_DIR to a valid directory.")
```
Probably don't need double linebreaks withing function... but not something that really matters, either.

python/src/opendp/_lib.py
```diff
@@ -124,6 +125,30 @@ def next_raw(_voidp):
 except ImportError:  # pragma: no cover
     pass
 
+try:
+    pl = import_optional_dependency("polars")
+    from polars.plugins import register_plugin_function
```
mccalluc, line 130: I'd prefer to keep the `except ImportError` out of the code, now that it's cleaner.

python/src/opendp/_lib.py
```diff
@@ -124,6 +125,30 @@ def next_raw(_voidp):
...
+
+    @pl.api.register_expr_namespace("dp")
+    class DPNamespace(object):
+        def __init__(self, expr):
+            self._expr = expr
+
+        def laplace(self, scale=float("nan")):
```
mccalluc, line 137: Something in polars would prevent the usual `None` default value?

mccalluc (commented):
Another partial review. Should look at this some more tomorrow.

rust/src/measurements/make_private_expr/expr_laplace/mod.rs
```diff
@@ -0,0 +1,320 @@
...
+            // apply Laplace noise to the non-null values
+            .try_apply_nonnull_values_generic(|v| sample_discrete_laplace_Z2k(v, scale, k))?
+            // convert the resulting chunked array back to a series
+            .into_series())
+    }
+
+    match series.dtype() {
+        // the u8 and u16 features are not enabled in the Polars crate dependency features
+        // DataType::UInt8 => laplace_impl_integer::<UInt8Type>(series, scale),
+        // DataType::UInt16 => laplace_impl_integer::<UInt16Type>(series, scale),
```
mccalluc, line 217: Thanks! I remember seeing this gap somewhere else, and the explanation helps.

rust/src/measurements/make_private_expr/expr_laplace/mod.rs
```diff
@@ -0,0 +1,320 @@
...
+        DataType::UInt32 => laplace_impl_integer::<UInt32Type>(series, scale),
+        DataType::UInt64 => laplace_impl_integer::<UInt64Type>(series, scale),
+        DataType::Int8 => laplace_impl_integer::<Int8Type>(series, scale),
+        DataType::Int16 => laplace_impl_integer::<Int16Type>(series, scale),
+        DataType::Int32 => laplace_impl_integer::<Int32Type>(series, scale),
+        DataType::Int64 => laplace_impl_integer::<Int64Type>(series, scale),
+        DataType::Float32 => laplace_impl_float::<Float32Type>(series, scale),
+        DataType::Float64 => laplace_impl_float::<Float64Type>(series, scale),
+        _ => Err(PolarsError::ComputeError(
+            "Expected numeric data type.".into(),
```
mccalluc, line 227: If a user had tried to use u8 or u16, and go a message "Expected numeric data type", they might be confused. Rather than making this message more complicated, add matches for u8 and u16 with a more precise error?

rust/src/measurements/make_private_expr/expr_laplace/mod.rs
```diff
@@ -0,0 +1,320 @@
...
+///
+/// Ensures that the input field is numeric.
+pub(crate) fn laplace_type_udf(input_fields: &[Field]) -> PolarsResult<Field> {
+    let [field] = <&[Field; 1]>::try_from(input_fields).map_err(|_| {
+        PolarsError::InvalidOperation("Identity expects a single input field".into())
+    })?;
+    use DataType::*;
+    if !matches!(
+        field.data_type(),
+        UInt8 | UInt16 | UInt32 | UInt64 | Int8 | Int16 | Int32 | Int64 | Float32 | Float64
```
mccalluc, line 242: but u8 and u16 are good here?

rust/src/measurements/make_private_expr/expr_laplace/mod.rs
```diff
@@ -0,0 +1,320 @@
...
+        let (lf_domain, lf) = get_test_data()?;
+        let expr_domain = lf_domain.select();
+        let scale: f64 = 0.0;
+
+        let m_lap = col("B")
+            .clip(lit(1.), lit(2.))
+            .sum()
+            .dp()
+            .laplace(f64::NAN)
+            .make_private(
```
mccalluc, line 282: huh: Is this something we'd expect an end-user to do? Maybe there are separate things to exercise in more granular tests.
- laplace with a non-NaN scale should work on its own?
- make_private without laplace might work on its own?
- what does laplace do with NaN, but without make_private?

rust/src/measurements/make_private_expr/expr_laplace/mod.rs
```diff
@@ -0,0 +1,320 @@
...
+                MaxDivergence::default(),
+                scale,
+            )?;
+
+        let meas_res = m_lap.invoke(&(lf.logical_plan.clone(), all()))?;
+
+        let df_act = lf.clone().select([meas_res]).collect()?;
+        let df_exp = lf.clone().select([sum("B")]).collect()?;
+
+        assert_eq!(df_act, df_exp);
```
mccalluc, line 294: So I guess we're not actually comparing DP results here, so I'm confused about what this test is doing.

rust/src/core/polars/mod.rs
```diff
@@ -72,3 +192,28 @@ impl_from_for_scalar!(u32, U32);
...
+
+/// Helper trait for Rust users to access differentially private expressions.
+pub trait PrivacyNamespaceHelper {
+    fn dp(self) -> DPNamespace;
+}
+impl PrivacyNamespaceHelper for Expr {
+    fn dp(self) -> DPNamespace {
+        DPNamespace(self)
+    }
+}
```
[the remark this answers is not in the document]
Shoeboxam: yep!

rust/src/measurements/make_private_expr/expr_laplace/mod.rs
Shoeboxam: * it does work on its own
* if there is another expression that qualifies as a mechanism, it could still work. But it must match an expression that makes things private.
* it fails to run!

rust/src/measurements/make_private_expr/expr_laplace/mod.rs
Shoeboxam: With zero scale, the test becomes deterministic.

rust/src/measurements/make_private_expr/expr_laplace/mod.rs
Shoeboxam: Added another catch. Thx!

rust/src/measurements/make_private_expr/expr_laplace/mod.rs
```diff
@@ -0,0 +1,320 @@
...
+    fn get_options(&self) -> FunctionOptions {
+        FunctionOptions {
+            collect_groups: ApplyOptions::ElementWise,
+            fmt_str: "laplace",
+            ..Default::default()
+        }
+    }
+}
+
+// allow the LaplaceArgs struct to be stored inside an AnonymousFunction, when used from Rust directly
```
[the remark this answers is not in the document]
Shoeboxam: The Rust and Python APIs behave the same to the end-user, but I had to implement them in different ways. This causes some logic duplication. Left a comment.

rust/src/measurements/make_private_expr/expr_laplace/mod.rs
```diff
@@ -0,0 +1,320 @@
...
+    // PT stands for Polars Type
+    fn laplace_impl_integer<PT: 'static + PolarsDataType>(
+        series: &Series,
+        scale: f64,
+    ) -> PolarsResult<Series>
+    where
+        // the physical (rust) dtype must be able to be converted into a big integer
+        for<'a> IBig: From<PT::Physical<'a>>,
+        // a big integer must be able to be converted into the physical (rust) dtype
+        for<'a> PT::Physical<'a>: SaturatingCast<IBig>,
```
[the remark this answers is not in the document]
Shoeboxam: Polars only works with native types that have fixed widths, like u32, u64, etc. This trait bound requires that it must be possible to construct an IBig from the physical Polars type.

Shoeboxam:
I believe the pyo3 dependency adds a linker dependency on Python itself that was until now completely compiled out. Since this PR implements a third-party Polars plugin that exercises this dependency, the PyO3 linker dep appears in the staticlib sent to the R CI. I've added a separate compilation step to build an R staticlib without this linker dep.

python/src/opendp/_lib.py
```diff
@@ -124,6 +124,42 @@ def next_raw(_voidp):
 except ImportError:  # pragma: no cover
     pass
 
+pl = import_optional_dependency("polars", raise_error=False)
+if pl is not None:
```
mccalluc, line 128: May have come up elsewhere, but it might be nice to have a little context manager for places like this. Will link an issue to comment.

rust/src/measurements/make_private_expr/expr_laplace/mod.rs
```diff
@@ -0,0 +1,333 @@
...
+            // convert the resulting chunked array back to a series
+            .into_series())
+    }
+
+    use DataType::*;
+    match series.dtype() {
+        UInt32 => laplace_impl_integer::<UInt32Type>(series, scale),
+        UInt64 => laplace_impl_integer::<UInt64Type>(series, scale),
+        Int8 => laplace_impl_integer::<Int8Type>(series, scale),
+        Int16 => laplace_impl_integer::<Int16Type>(series, scale),
```
mccalluc, line 226: but signed 8 and 16bit are supported? might be the case, but seems odd.

rust/src/measurements/make_private_expr/expr_laplace/mod.rs
```diff
@@ -0,0 +1,333 @@
...
+        Int32 => laplace_impl_integer::<Int32Type>(series, scale),
+        Int64 => laplace_impl_integer::<Int64Type>(series, scale),
+        Float32 => laplace_impl_float::<Float32Type>(series, scale),
+        Float64 => laplace_impl_float::<Float64Type>(series, scale),
+        UInt8 | UInt16 => {
+            polars_bail!(InvalidOperation: "u8 and u16 not supported in the OpenDP Polars plugin. Please use u32 or u64.")
```
mccalluc, line 232:
```suggestion
            polars_bail!(InvalidOperation: "u8 and u16 not supported by Polars. Please use u32 or u64.")
```
Point to the root of the problem, if it is a problem?

mccalluc (commented):
Some comments -- Several little questions, but not sure I'm getting the big picture.

python/src/opendp/_lib.py
```diff
@@ -124,6 +124,42 @@ def next_raw(_voidp):
...
+    @pl.api.register_expr_namespace("dp")
+    class DPNamespace(object):
+        def __init__(self, expr):
+            self._expr = expr
+
+        def laplace(self, scale=None):
+            """Add Laplace noise to the expression.
+
+            `scale` must not be negative or inf.
+            Scale may be left NaN, to be filled later by [`make_private_expr`] or [`make_private_lazyframe`].
```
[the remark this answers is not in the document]
Shoeboxam: TODO: update all other APIs (make_private_lazyframe, make_private_expr, postprocessors and RNM gumbel) to take Option<f64> instead of just f64.

rust/src/measurements/make_private_expr/expr_laplace/mod.rs
```diff
@@ -0,0 +1,333 @@
...
+    let [input] = <&[_; 1]>::try_from(input.as_slice())
+        .map_err(|_| err!(MakeMeasurement, "Laplace expects a single input expression"))?;
+
+    Ok(Some((input, scale)))
+}
+
+// When using the plugin API from other languages, the LaplaceArgs struct is serialized inside a FunctionExpr::FfiPlugin.
+// When using the Rust API directly, the LaplaceArgs struct is stored inside an AnonymousFunction.
+//
+/// Arguments for the Laplace noise expression
```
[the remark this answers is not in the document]
Shoeboxam: This feels better suited as a comment than as documentation. Neither Rust nor Python users need to know this, only developers.

chikeabuah (dismissed):

rust/src/measurements/make_private_expr/expr_laplace/mod.rs
Shoeboxam: Yep, just following along.

PR #831, opened 2023-06-23
remove openssl dependency

Shoeboxam:


Shoeboxam:
* main
  * **PR #831** <a href="https://app.graphite.dev/github/pr/opendp/opendp/831" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/opendp/opendp/831?utm_source=stack-comment).

python/src/opendp/measurements.py
```diff
@@ -1098,10 +1164,18 @@ def then_report_noisy_max_gumbel(
...
-    r"""Partial constructructor which uses the preceding output domain and metric as its input domain and metric. See also:
+):  
+    r"""partial constructor of make_report_noisy_max_gumbel
 
-    :py:func:`opendp.measurements.make_report_noisy_max_gumbel`
+    .. seealso:: 
+      Delays application of `input_domain` and `input_metric` in :py:func:`opendp.measurements.make_report_noisy_max_gumbel`
+
+    :param scale: Higher scales are more private.
+    :type scale: Any
```
mccalluc, line 1174: surprised this isn't `float`?

python/src/opendp/measurements.py
```diff
@@ -1098,10 +1164,18 @@ def then_report_noisy_max_gumbel(
...
+    :param optimize: Indicate whether to privately return the "Max" or "Min"
```
mccalluc, line 1175: I'm not sure this is in sync with the rust, or at least there seems to be some upper/lowercase inconsistency:
```
    let optimize = match try_!(to_str(optimize)) {
        i if i.to_lowercase().starts_with("min") => Optimize::Min,
        i if i.to_lowercase().starts_with("max") => Optimize::Max,
        _ => return err!(FFI, "optimize must start with \"min\" or \"max\"").into(),
    };
```
... but I wonder if we should even have string arguments here: A boolean or enum seems more natural. Probably not something to block this PR: Suggest filing a new issue from this comment.

python/src/opendp/measurements.py
```diff
@@ -607,10 +645,17 @@ def then_base_geometric(
...
+    r"""partial constructor of make_base_geometric
+
+    .. seealso:: 
+      Delays application of `input_domain` and `input_metric` in :py:func:`opendp.measurements.make_base_geometric`
 
-    :py:func:`opendp.measurements.make_base_geometric`
+    :param scale: 
+    :param bounds: 
+    :type bounds: Any
+    :param QO:
```
mccalluc, line 657: Add description? (Upstream change?)

mccalluc (dismissed):
I feel like some unrelated changes got merged into this. The PRNG changes seem fine, though the details are beyond me. I'll file a follow-up issue for max vs. Max or str vs. bool. Approved.

PR #843, opened 2023-06-26
Add aarch64-unknown-linux-gnu target

### LINKED ISSUE #817: Support aarch64-linux
**Describe the bug**
It appears that the wheel ships with precompiled binaries for x86_64 except for the Apple .dylib which is a universal fat binary.

Unfortunately this means `opendp` won't work on Arm64 (aarch64) linux nor will it run in a container on Apple Silicon since macOS will run aarch64 linux.

Are you able to provide aarch64 linux support? The same thing will eventually be an issue on Windows Arm64 too.

**To Reproduce**
pip install on linux on an arm64 cpu.

output from file tool:
```
file libopendp.so
libopendp.so: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, BuildID[sha1]=a4467d4313afcd5f01ff0a4253ebc53810f68cef, with debug_info, not stripped

file libopendp.dylib
libopendp.dylib: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit x86_64 dynamically linked shared library, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL|NO_REEXPORTED_DYLIBS|HAS_TLV_DESCRIPTORS>] [arm64]

file opendp.dll
opendp.dll: PE32+ executable (DLL) (console) x86-64, for MS Windows
root@backend-0:/usr/local/lib/python3.11/site-packages/opendp/lib#
```

andrewvyrros:
Fixes #817.

* Refactor `release.yml` to include target `aarch64-unknown-linux-gnu`
  * The main grief was getting `gmp` to cross compile to aarch64 from x86_64.
    The configure script just wouldn't work with the toolchain in the standard `manylinux_2014` images,
    but `messense/manylinux2014-cross` seems to have a better setup.
  * Because linux doesn't have fat binaries, we need to have separate `libopendp-XXX.so` files with tagged names
  * Also did some cleanup of action versions
* Tweak rust_build.sh and python_build.sh
* Add logic to _lib.py to load arch-specific libraries

### DIFF
--- a/python/src/opendp/_lib.py
+++ b/python/src/opendp/_lib.py
@@ -1,8 +1,9 @@
 import ctypes
 import os
+import platform
+import re
 import sys
 from typing import Optional, Any
-import re
 
 
 # list all acceptable alternative types for each default type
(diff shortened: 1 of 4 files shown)


andrewvyrros:
* main
  * **PR #843** <a href="https://app.graphite.dev/github/pr/opendp/opendp/843" target="_blank">Graphite</a>  👈
    * **PR #853** <a href="https://app.graphite.dev/github/pr/opendp/opendp/853" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/opendp/opendp/843?utm_source=stack-comment).

python/src/opendp/_lib.py
```diff
@@ -14,30 +15,62 @@
...
+    elif os.environ.get('OPENDP_HEADLESS', "false") != "false":
+        return None
+
+    else:
+        raise ValueError("Unable to find lib directory. Consider setting OPENDP_LIB_DIR to a valid directory.")
 
-else:
-    raise ValueError("Unable to find lib directory. Consider setting OPENDP_LIB_DIR to a valid directory.")
 
+lib = _load_library()
```
Shoeboxam, line 73: This should make it easier to reload the binary in notebooks.

Shoeboxam (approved):
Great to add support for these targets.

Comparing against [Rust's available targets](https://doc.rust-lang.org/nightly/rustc/platform-support.html#tier-1-with-host-tools), excluding 32-bit builds, we now have binaries for all Tier-1 targets except for `x86_64-pc-windows-msvc`. This is due to our dependency on GMP/MPFR. We also have support for one tier-2 target, `aarch64-apple-darwin`.

PR #845, opened 2023-06-26
polish edge-cases in maps

### LINKED ISSUE #654: `new_from_constant` should check that `d_in >= 0`
We have bounds checks in many other places, but should have bounds checks _everywhere_.
This can be added by changing `new_from_constant` to convert to QO first, and then apply the signed-ness check. Add another `Signed` trait to `DistanceConstant` and then implement `Signed` for distance types.

EG:
```rust
pub trait Sign {
    fn is_negative(&self) -> bool;
}


macro_rules! impl_is_negative {
    ($($ty:ty)+) => ($(impl Sign for $ty {
        fn is_negative(&self) -> bool {
            self < &Self::zero()
        }
    })+)
}

impl_is_negative!(f32 f64 u8 u16 u32 u64 u128 usize i8 i16 i32 i64 i128 isize);

```

### LINKED ISSUE #656: Adjust privacy maps to return zero `d_out` if `d_in` is zero, regardless of scale
This is a small QOL tweak.

If `d_in` is zero (including relaxation parameter if applicable), return a `d_out` of zero.
Else if `scale` is zero, return infinity.

On many measurements, the current behavior returns infinity if `scale` is zero, even if `d_in` is zero.

### LINKED ISSUE #844: Handle edge case in zCDP_to_approxDP curve when δ=0
This gives an infinite loop:

```python
curve = dp.c.make_zCDP_to_approxDP(dp.m.make_base_gaussian(*input_space, scale=1.)).map(d_in=1.)
curve.epsilon(0.)
```

Shoeboxam:
Closes #844 (infinite loop in zCDP_to_approxDP when δ=0)
Closes #656 (privacy maps return zero when d_in=0)
Closes #654 (new_from_constant enforces c>= 0)

### DIFF
--- a/python/test/test_meas.py
+++ b/python/test/test_meas.py
@@ -1,3 +1,4 @@
+import pytest
 import opendp.prelude as dp
 
 dp.enable_features('floating-point', 'contrib')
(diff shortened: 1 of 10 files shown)


Shoeboxam:
* main
  * **PR #845** <a href="https://app.graphite.dev/github/pr/opendp/opendp/845" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/opendp/opendp/845?utm_source=stack-comment).

andrewvyrros (commented):
Good catch!

Seems like we could use at least a sanity check test for `cdp_epsilon`.

andrewvyrros (approved):
Thanks for adding the tests!

PR #870, opened 2023-07-12
make Queryable from Python UDF

Shoeboxam:
Closes #851

PR #871, opened 2023-07-12
add `UserDomain`

### LINKED ISSUE #848: user-defined domain: `ExtrinsicDomain`
The `ExtrinsicDomain` is specifically for user-defined functions in Python:

```rust
type ExtrinsicObject = *const c_void;

struct ExtrinsicDomain {
    descriptor: String
    member: Function<ExtrinsicObject, bool>
}
impl Domain for ExtrinsicDomain {
    Carrier = ExtrinsicObject;
    fn member(&self, val: &ExtrinsicObject) -> Fallible<bool> {
        self.member.eval(val)
    }
}
```

* The struct holds a descriptor. This is meant to be a string description of the domain. We could define a JSON standard here.
    * Ideally, the descriptor should make it possible to impl traits used for combinators on `ExtrinsicDomain`, like retrieving an optional dataset size
* The `member` function in the struct holds a user-defined Python function that checks if the value is in the domain. On the python side, the internal python function is stored in the ExtrinsicDomain struct's dependencies.
* The carrier type is a pointer to a `ExtrinsicObject`. This pointer can be retrieved [for any given Python object via ctypes](https://docs.python.org/3/library/ctypes.html#ctypes.py_object).

The purpose of this is to allow users to define their own domains/metrics from Python, and to avoid the conversion into Rust datatypes at each transformation/transformation or transformation/measurement boundary.

For instance, the following transformation transforms a numpy array to a vector of sklearn models, for use in PATE:

```python
from sklearn.base import is_classifier
import numpy as np
import opendp.prelude as dp

trans = dp.c.make_user_transformation(
    dp.extrinsic_domain({"name": "numpy_array", "shape": [1000, 7]}, member=lambda x: isinstance(x, np.ndarray)),
    dp.vector_domain(dp.extrinsic_domain(descriptor={"name": "sklearn_classifiers"}, member=is_classifier))
    partition_and_fit_sklearn_classifiers,
    dp.symmetric_distance(),
    dp.symmetric_distance(),
    lambda x: x
)

models = trans.invoke(np.random.uniform(shape=[1000, 7]))
```

Shoeboxam:
Closes #848


Shoeboxam:
* main
  * **PR #870** <a href="https://app.graphite.dev/github/pr/opendp/opendp/870" target="_blank">Graphite</a> 
    * **PR #871** <a href="https://app.graphite.dev/github/pr/opendp/opendp/871" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/opendp/opendp/871?utm_source=stack-comment).

python/src/opendp/typing.py
```diff
@@ -268,13 +269,15 @@ def infer(cls, public_example: Any) -> Union["RuntimeType", str]:
...
         if isinstance(public_example, list):
             if public_example:
-                inner_type = cls.infer(public_example[0])
-                for inner in public_example[1:]:
-                    other_type = cls.infer(inner)
+                inner_type = cls.infer(public_example[0], py_object=py_object)
+                for inner in public_example:
+                    other_type = cls.infer(inner, py_object)
+                    if inner_type == "ExtrinsicObject":
+                        break
```
andrewvyrros, line 280: Issue here: We want to be able to detect inadvertent heterogeneous list, but also allow intentional heterogeneous list in the case of ExtrinsicObject.

From discussion:
 * Change `infer` to return Vec<ExtrinsicObject> if heterogeneous and py_object==True; no error message here in that case.
 * Change `assert_is_similar` to take example, do the right thing in case of heterogeneous list.

andrewvyrros (changes requested):

rust/src/data/ffi.rs
```diff
@@ -135,6 +142,7 @@ pub extern "C" fn opendp_data__slice_as_object(
     }
     match T.contents {
         TypeContents::PLAIN("String") => raw_to_string(raw),
+        TypeContents::PLAIN("ExtrinsicObject") => raw_to_extrinsic_object(raw),
```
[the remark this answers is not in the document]
Shoeboxam: I think that's the wrong direction to go-- IMO we shouldn't even have VEC in there.

python/src/opendp/typing.py
Shoeboxam: I got rid of assert_is_similar completely, instead folding it directly into the loaders.

Shoeboxam:
Reworked ExtrinsicObject to be a struct containing a void* and a function to increment/decrement refcounts.

python/test/test_trans.py
```diff
@@ -1,7 +1,7 @@
 import opendp.prelude as dp
 from opendp.mod import enable_features
 
-enable_features('contrib')
+enable_features('contrib', 'honest-but-curious')
```
andrewvyrros, line 4: Is `honest-but-curious` needed here? Seems not.

python/src/opendp/_convert.py
```diff
@@ -239,43 +274,41 @@ def _slice_to_string(raw: FfiSlicePtr) -> str:
 
 
 def _vector_to_slice(val: Sequence[Any], type_name: RuntimeType) -> FfiSlicePtr:
-    assert type_name.origin == 'Vec'
-    assert len(type_name.args) == 1, "Vec only has one generic argument"
-    inner_type_name = type_name.args[0]
-
-    # input is numpy array
```
andrewvyrros, line 246: You don't need this anymore?

python/src/opendp/typing.py
```diff
@@ -338,51 +346,6 @@ def parse_or_infer(
...
-        # allow extra flexibility around options, as the inferred type of an Option::<T>::Some will just be T
-        def is_option(type_name):
-            return isinstance(type_name, RuntimeType) and type_name.origin == "Option"
-        if is_option(expected):
-            expected = expected.args[0]
-            if is_option(inferred):
-                if isinstance(inferred.args[0], UnknownType):
-                    return
-                else:
-                    inferred = inferred.args[0]
```
andrewvyrros, line 363: Did you lose this handling of options?

andrewvyrros (commented):
It was super fun to come up with an approach for `ExtrinsicObject` in realtime!

python/src/opendp/typing.py
Shoeboxam: Now `_convert.py` handles this in `py_to_c`:

```python
    if type_name is not None:
        # exit early with a null pointer if trying to load an Option type with a None value
        if isinstance(type_name, RuntimeType) and type_name.origin == "Option":
            if value is None:
                return
            type_name = type_name.args[0]
```

andrewvyrros: Ah OK thanks!

andrewvyrros (approved):

PR #874, opened 2023-07-14
support multithreading

### LINKED ISSUE #842: Support multi-threading in library primitives
Polars' `Expr.map` has a bound that requires the closure to be `Send + Sync`, which `Rc` isn't. `Function` holds an `Rc<dyn Fn>`, or a reference-counted function. When polars does multithreading, the `Rc` can't guarantee that rc increments and decrements are atomic- so one thread could stomp on another, eventually resulting in a double-free of the memory or leak.

To support multi-threaded computation, need to replace `Rc` with `Arc`, and supply `Send + Sync` bounds as needed throughout the lib.

Shoeboxam:
Closes #842


Shoeboxam:
* main
  * **PR #874** <a href="https://app.graphite.dev/github/pr/opendp/opendp/874" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/opendp/opendp/874?utm_source=stack-comment).

rust/src/combinators/composition/sequential/ffi.rs
```diff
@@ -78,3 +114,54 @@ pub extern "C" fn opendp_combinators__make_sequential_composition(
...
+                            Ok(Answer::Internal(a))
+                        }
+                    })
+                },
+            ),
+            self.input_metric.clone(),
+            self.output_measure.clone(),
+            self.privacy_map.clone(),
+        )
+        .expect("AnyDomain is not checked for compatibility")
```
andrewvyrros, line 165: Surprising that this message is specific to AnyDomain, when we don't know here if that's always what would cause this to panic.

rust/src/ffi/any.rs
```diff
@@ -461,6 +345,109 @@ impl Debug for AnyMetric {
...
+    move |arg: &AnyObject| -> Fallible<AnyObject> {
+        into_owned(func(arg as *const AnyObject))?.into()
+    }
+}
+
+#[allow(dead_code)]
+pub enum Either<L, R> {
+    Left(L),
+    Right(R),
+}
```
andrewvyrros, line 361: Is this used anymore?

rust/src/measures/ffi.rs
```diff
@@ -188,3 +188,53 @@ pub extern "C" fn opendp_measures__user_divergence(
...
+        Self {
+            measure: self.measure.clone(),
+            marker: self.marker.clone(),
+        }
+    }
+}
+
+impl<Q> Debug for TypedMeasure<Q> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("TypedMetric")
```
andrewvyrros, line 227: Wrong debug name

andrewvyrros (commented):
Great way to resolve this issue with a reasonable amount of packing/unpacking.

andrewvyrros (approved):

Issue #851, opened 2023-06-28
bindings for constructing Queryable from Python UDFs


Shoeboxam:
Since there are no bindings for constructing a Queryable from Python, it is not possible to build interactive measurements from Python.

Simple example:

```python
def make_rr_vec_meas(prob, size):
    rr_meas = dp.m.make_randomized_response_bool(prob)
    
    def function(data):
        idx = 0
        def transition(_this, _query, _internal):
            nonlocal idx
            idx += 1
            return rr_meas(data[idx - 1])
        
        return new_queryable(transition, TO=dp.Queryable[Any, bool])
    
    return dp.c.make_user_measurement(
        dp.vector_domain(dp.atom_domain(T=bool), size=size),
        function,
        dp.symmetric_distance(),
        dp.max_divergence(T=float),
        lambda d_in: d_in * rr_meas.map(int(d_in > 0))
    )
```

PR #884, opened 2023-07-20
update udf constructors

### LINKED ISSUE #794: move user transformation/measurement constructors
The Analysis API only reads from the transformations/measurements modules. The user-defined trans/meas constructors should live in these modules, because then they'll get picked up in the Analysis API. It also makes intuitive sense, as these constructors don't take trans/meas as arguments.

Shoeboxam:
Closes #794

* Move `make_user_transformation` into `transformations`
	* so that it will appear in the HLAPI. It also isn't a combinator.
* Move `make_user_measurement` into `measurements`
 	* so that it will appear in the HLAPI. It also isn't a combinator.
* Move `make_user_postprocessor` into `core` and rename to `new_function`
    * consistent with `new_queryable` in #851 
    * naming was too specific before- will use it for other things than just postprocessors
* Reorder arguments to `make_user_transformation` and `make_user_measurement` to dom/met convention
    * so that you can do `>> then_user_transformation(DO, MO, func, map)`, etc


Shoeboxam:
* main
  * **PR #884** <a href="https://app.graphite.dev/github/pr/opendp/opendp/884" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/opendp/opendp/884?utm_source=stack-comment).

rust/src/measurements/make_user_measurement/mod.rs
```diff
@@ -0,0 +1,70 @@
...
+    privacy_map: CallbackFn,
+) -> Fallible<Measurement<AnyDomain, AnyObject, AnyMetric, AnyMeasure>> {
+    let _ = (
+        input_domain,
+        input_metric,
+        output_measure,
+        privacy_map,
+        function,
+    );
+    panic!("this signature only exists for code generation")
```
andrewvyrros, line 49: Nice trick!

rust/src/measurements/make_user_measurement/mod.rs
andrewvyrros, line 1: The module name feels a little odd. Wouldn't `user_defined` or `user_measurement` (or just `user`) fit the pattern better? (Same for transformation.)

rust/src/measurements/make_user_measurement/mod.rs
```diff
@@ -0,0 +1,70 @@
...
+};
+
+#[bootstrap(
+    name = "make_user_measurement",
+    features("contrib", "honest-but-curious"),
+    arguments(
+        input_domain(hint = "Domain"),
+        input_metric(hint = "Metric"),
+        function(rust_type = "$pass_through(TO)"),
+        output_measure(hint = "Measure"),
```
andrewvyrros, line 18: Minor: The order of these is swapped vs the function.

rust/src/measurements/make_user_measurement/mod.rs
Shoeboxam: Anna/Pauline started naming modules after the constructor function, and I really like it.

andrewvyrros: OK, if the crew is liking this convention, I'm cool with it.

andrewvyrros (approved):

andrewvyrros (approved):

PR #905, opened 2023-08-05
Fix notebook dependencies

### LINKED ISSUE #903: Ensure that notebooks can run in Binder out of the box
(Follow-up from discussion https://github.com/opendp/opendp/discussions/898)

The first cell of `first-look-at-DP.ipynb` failed for me on Binder because pandas is not installed. Should fix it by adding pandas to `.binder/requirements.txt`.

(... but `smoke-test.yml` seems like it should be testing notebooks with `pytest --nbmake source/**/*.ipynb -n=auto`, so maybe that test isn't working as intended?)

I'll try starting a PR, and will update here with what I learn. 

Ideally, close when all notebooks work in Binder if cells are evaluated in order, and CI tests fail if this is not true.

mccalluc:
- Remove requirements_notebooks.txt and move contents to .binder/requirements.txt
- Change reference from pinned release to local directory, so CI will be testing the current code
- Update .github/workflows/smoke-test.yml
- Fix #903

### DIFF
--- a/.binder/requirements.txt
+++ b/.binder/requirements.txt
@@ -1,3 +1,9 @@
 # For https://mybinder.org
-# Pin to matching version
-opendp==0.8.0.dev0
+--editable python/
+
+# Used by examples in notebooks:
+matplotlib
+numpy
+pandas
+seaborn
+scipy
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -140,16 +140,13 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           python -m pip install flake8 pytest nbmake pytest-xdist
-          python -m pip install -r requirements_notebooks.txt
+          python -m pip install -r .binder/requirements.txt
 
       - name: Download libs
         uses: actions/download-artifact@v3
         with:
           name: libs
           path: python/src/opendp/lib
-
-      - name: Install package
-        run: (cd ../python && python -m pip install -e .)
 
       - name: Test
         run: |
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -140,7 +140,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           python -m pip install flake8 pytest nbmake pytest-xdist
-          python -m pip install -r .binder/requirements.txt
+          python -m pip install -r ../.binder/requirements.txt
 
       - name: Download libs
         uses: actions/download-artifact@v3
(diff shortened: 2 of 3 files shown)


Shoeboxam:
I did some digging on this. I don't think editable installs on binder will work, because the package needs a compiled Rust binary to function. On the other hand, [we do push nightly builds up to PyPi now](https://github.com/opendp/opendp/pull/880), and we do have branches and tags where the dependencies are automatically maintained:
* [nightly](https://github.com/opendp/opendp/tree/nightly)
* [beta](https://github.com/opendp/opendp/tree/beta)
* [stable](https://github.com/opendp/opendp/tree/stable)

And we also have tags for each version:
https://github.com/opendp/opendp/tags

I think what needs to happen to fix this is just to fix the binder links to either point to the tag, or to one of the above branches, depending on which version of the docs is rendered. 

Now this wouldn't fix historical versions (0.7 or older) but just fixing binder links *will* work going forward from 0.8. Maybe we could just hide binder links for versions 0.7 and older?

Given that this metadata will be checked into the repo on the 0.8 tag, this isn't something that has to be fixed before 0.8 is released.

Shoeboxam:
The code that inserts the link to binder is here:
https://github.com/opendp/opendp/blob/v0.6.1/docs/source/conf.py#L198-L209

Shoeboxam:
It actually already works as-is for tagged versions:

The generated binder link is only wrong on nightly and beta now. If you adjust the link from `v0.8.0` to `nightly`, I think it works:
https://mybinder.org/v2/gh/opendp/opendp/nightly?filepath=docs/source/user/programming-framework/a-framework-to-understand-dp.ipynb

That being said, it's incredibly slow...

Shoeboxam:
After thinking about it a bit, I think the ideal approach is to:
1. Don't show a binder link for versions 0.7 or below (their .binder/requirements are broken)
2. Don't show a binder link for beta or nightly (the images take too long to build and constantly get rebuilt)
    * if you want a nightly build, you're better off setting things up locally
4. Adjust .binder/requirements to -r install the existing `../docs/requirements_notebooks.txt`
    * because .binder is not the source of truth for notebook dependencies

mccalluc:
Thank you for thinking this through, and sorry not to have it on the right path.

At this point, maybe the least confusing thing is to close this PR, and add the criteria above to #903. I would also add a note where we reference the pinned version of opendp to emphasize that the test is not about the current state of the code, but to ensure that the documentation of released versions is consistent.

Closing this PR, and updating the issue.

Shoeboxam:
@mccalluc No need to be sorry! I really appreciate you opening a PR.

PR #912, opened 2023-08-08
Only show binder link above 0.7.0

mccalluc:
Possible fix for the remaining part of:
- #903

I've run `make html`, but I'm getting an error with `make versions` locally, though I don't think it's related to these particular changes.

I'm ambivalent about this PR:
- It adds complexity, for a pretty minor change in behavior.
- I'm not confident that 0.7.0 is the right bound: would less-than 0.8.0 be better?
- Even if this is the right direction, maybe it's better to do the conditional inside the template?

### DIFF
--- a/.binder/requirements.txt
+++ b/.binder/requirements.txt
@@ -1,3 +1,5 @@
 # For https://mybinder.org
 # Pin to matching version
 opendp==0.8.0.dev0
+
+-r docs/requirements_notebooks.txt
--- a/.binder/requirements.txt
+++ b/.binder/requirements.txt
@@ -1,5 +1,3 @@
 # For https://mybinder.org
 # Pin to matching version
 opendp==0.8.0.dev0
-
--r docs/requirements_notebooks.txt
--- a/docs/README.md
+++ b/docs/README.md
@@ -12,6 +12,7 @@ The steps below assume the use of [Homebrew] on a Mac.
 [Homebrew]: https://brew.sh
 
 ```shell
+cd docs
 brew install pandoc
 python3 -m venv venv
 source venv/bin/activate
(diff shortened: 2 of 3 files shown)


andrewvyrros:
Thanks for the submission!

> I've run `make html`, but I'm getting an error with `make versions` locally, though I don't think it's related to these particular changes.

Are you using a Mac? There's a bug in `sphinx` or `nbsphinx` (not sure which) that causes relative paths of images to fail (because `/var` is a symlink to `/private/var` on macOS).

You can hack around it by editing `site-packages/sphinx/environment/collectors/asset.py`. In `process_doc()`, after `node['uri'], _ = app.env.relfn2path(imguri, docname)`, add the following:

```
                if '/nbsphinx/' in node['uri']:
                    node['uri'] = "../" + node['uri']
```

> * I'm not confident that 0.7.0 is the right bound: would less-than 0.8.0 be better?

Yes, probably < 0.8.0 would be safer.

> * Even if this is the right direction, maybe it's better to do the conditional inside the template?

Yes, I don't think this will work as written. `sphinx-multiversion` somehow caches config variables, then runs the build for the different versions in a different context, so anything that's version-specific must be resolved dynamically in the `nbsphinx_prolog` Jinja template. That's the reason `frag` is computed there. You might be able to get that to work, but I don't think you'll have access to `semver` there, so it'd probably be a pain to do the version check safely.

If you're determined to solve this, then I salute you! But at this point, it's probably not worth the hassle for old releases.

andrewvyrros (commented):
Might be made to work, but probably not worth the hassle.

mccalluc:
I'm happy to close this. I learned something from this: Thanks for the pointers.

PR #925, opened 2023-08-16
Update issue template links, and use markdown in template

mccalluc:
- Fix #904 
- Also adds link to Slack, and uses MD syntax for links
- Does not address other references to `latest`

Issue #904, opened 2023-08-05
Links in "New Contribution" template are dead


mccalluc:
https://github.com/opendp/opendp/blob/main/.github/ISSUE_TEMPLATE/new-contribution.md

The links in this template are dead. Might instead reference https://opendp.org/contribute or https://docs.opendp.org/en/beta/contributor/contribution-process.html ?

> How exciting!
> 
> Please review the following link for information to include in your issue:
> https://docs.opendp.org/en/latest/contributor/intro.html#code-and-proof-contributions
> 
> We will try to get back with you shortly.
> 
> In the meantime, you might want to review the contribution process:  
> https://docs.opendp.org/en/latest/contributor/contribution-process.html  
> (and start getting your development environment set up)

Shoeboxam:
This was caused by https://github.com/opendp/opendp/pull/880.
I think both links should be the same, but point to nightly:

- https://docs.opendp.org/en/nightly/contributor/intro.html#code-and-proof-contributions
- https://docs.opendp.org/en/nightly/contributor/contribution-process.html

Might be a good place to advertise the Slack as well:
https://join.slack.com/t/opendp/shared_invite/zt-zw7o1k2s-dHg8NQE8WTfAGFnN_cwomA

(of which you are welcome to join!)

Thanks for catching this. More generally, all docs.opendp.org/en/latest/ links are now broken.

Shoeboxam:
Yikes, thanks for double-checking my links. Another benefit of your suggestion is it would remove some un-tested links.

PR #926, opened 2023-08-16
"latest" to "nightly"

mccalluc:
- Addresses ["More generally, all docs.opendp.org/en/latest/ links are now broken."](https://github.com/opendp/opendp/issues/904#issuecomment-1667025539)
- Extends #925
- Also changes references to `latest` branch to `nightly`

Does not:
- touch `new-contribution.md`, so you can avoid merge conflicts
- touch any of the `https://docs.rs/opendp/latest/opendp/...` references in Python docstrings: These still work, and changing them to `nightly` doesn't, but that might also be a bug in the documentation generation process?

The https://github.com/opendp/opendp/tree/latest branch could probably be deleted on github to reduce confusion?

### DIFF
--- a/docs/source/contributor/maintainer-notes.rst
+++ b/docs/source/contributor/maintainer-notes.rst
@@ -180,10 +180,10 @@ These are the GitHub workflows that support the release process.
 sync-branches.yml
 ^^^^^^^^^^^^^^^^^
 
-* Keeps the tracking branches ``nightly`` and ``stable`` in sync with their targets.
+* Keeps the tracking branches ``latest`` and ``stable`` in sync with their targets.
   This is used when generating docs, so that we have a consistent path to each category.
 * Triggered on every push to ``main``, or when release is published.
-* Whenever there's a push to ``main``, it advances ``nightly`` to the same ref.
+* Whenever there's a push to ``main``, it advances ``latest`` to the same ref.
 * Whenever a release is created, it advances ``stable`` to the release tag.
 
 release.yml
--- a/docs/source/contributor/maintainer-notes.rst
+++ b/docs/source/contributor/maintainer-notes.rst
@@ -180,10 +180,10 @@ These are the GitHub workflows that support the release process.
 sync-branches.yml
 ^^^^^^^^^^^^^^^^^
 
-* Keeps the tracking branches ``latest`` and ``stable`` in sync with their targets.
+* Keeps the tracking branches ``nightly`` and ``stable`` in sync with their targets.
   This is used when generating docs, so that we have a consistent path to each category.
 * Triggered on every push to ``main``, or when release is published.
-* Whenever there's a push to ``main``, it advances ``latest`` to the same ref.
+* Whenever there's a push to ``main``, it advances ``nightly`` to the same ref.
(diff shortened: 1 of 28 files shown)


docs/source/contributor/maintainer-notes.rst
```diff
@@ -180,10 +180,10 @@ These are the GitHub workflows that support the release process.
 sync-branches.yml
 ^^^^^^^^^^^^^^^^^
 
-* Keeps the tracking branches ``latest`` and ``stable`` in sync with their targets.
+* Keeps the tracking branches ``nightly`` and ``stable`` in sync with their targets.
   This is used when generating docs, so that we have a consistent path to each category.
 * Triggered on every push to ``main``, or when release is published.
-* Whenever there's a push to ``main``, it advances ``latest`` to the same ref.
+* Whenever there's a push to ``main``, it advances ``nightly`` to the same ref.
```
andrewvyrros, line 186: This whole doc is out of date with the new release process, filed #927.

docs/source/user/getting-started.ipynb
```diff
@@ -344,7 +344,7 @@
       "    \n",
       "    **Proof Definition:**\n",
       "    \n",
-      "    [(Proof Document)](https://docs.opendp.org/en/latest/proofs/rust/src/transformations/clamp/make_clamp.pdf)\n",
+      "    [(Proof Document)](https://docs.opendp.org/en/nightly/proofs/rust/src/transformations/clamp/make_clamp.pdf)\n",
```
andrewvyrros, line 347: Probably better to reference `stable` here.

python/src/opendp/_lib.py
```diff
@@ -209,7 +209,7 @@ def versioned(function):
 
         docs_ref = get_docs_ref(version)
         function.__doc__ = function.__doc__.replace(
-            "https://docs.opendp.org/en/latest/",
+            "https://docs.opendp.org/en/nightly/",
```
andrewvyrros, line 212: This is as actually as intended, the codegen in the `opendp_tooling` crate creates the link using `latest`. (We're going to clean this up.)

python/src/opendp/accuracy.py
andrewvyrros, line 1: This is a generated file, is as expected by the `@versioned` decorator.

python/src/opendp/measurements.py
andrewvyrros, line 1: This is a generated file, is as expected by the `@versioned` decorator.

python/src/opendp/transformations.py
andrewvyrros, line 1: This is a generated file, is as expected by the `@versioned` decorator.

andrewvyrros (commented):
Thanks for finding these! Note that some of this is as intended, and the whole thing should be cleaned up with #928.

docs/source/contributor/maintainer-notes.rst
mccalluc: I'll revert this then, so it doesn't conflict with other work.

mccalluc:
I think latest commits address the comments... but if it all will be covered by #928 it's fine to close this without merging. Whatever is best.

raprasad:
Closing for now.

PR #934, opened 2023-09-25
rename `base_discrete_exponential` and `LInfDiffDistance`

### LINKED ISSUE #933: Adjust naming for exponential mechanism
`make_base_discrete_exponential` as implemented is not specific enough.
* there are variations of the exponential mechanism with discrete, but infinite support. The implemented mechanism requires finite support
* there are already variations of the exponential mechanism in the library that have specific scope (laplace mechanism)

Should instead name the mechanism `make_gumbel_max`, which describes exactly what it does... and opens the door to `make_exponential_max` and `make_laplace_max`

-----

`LInfDiffDistance` would be better named `RangeDistance`. The name is much simpler and better shows the equivalence with https://arxiv.org/pdf/1909.13830.pdf#subsection.2.1

Shoeboxam:
Closes #933

A change from the solution in that issue:
Many mechanisms would benefit from having both the l-infinity distance and range distance. Since there are no known situations where the range distance and infinity distance differ by a factor different than two, we can instead have a metric for *just* the l-infinity distance, and the metric stores a monotonicity boolean inside the metric's struct. The benefit to storing the bool inside the metric's struct, instead of in the distance type, is that this allows you to still conduct binary searches.

This PR changes the library to only have an `LInfDistance`, which stores a monotonicity bool inside. If the monotonicity bool is set, then the distance between score vectors with non-monotonic changes is considered to be infinity. A helper is added to the struct to translate l-infty distances to range distances.

Also renames temperature to scale, for consistency with other mechanisms.


Shoeboxam:
* main
  * **PR #934** <a href="https://app.graphite.dev/github/pr/opendp/opendp/934" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/opendp/opendp/934?utm_source=stack-comment).

python/test/test_meas.py
```diff
@@ -189,13 +189,19 @@ def test_gaussian():
...
+    meas = (input_domain, input_metric) >> dp.m.then_gumbel_max(1., "maximize")
     print(meas(list(range(10))))
-    print(meas.map(2))
+    assert meas.map(2) == 4
 
+    input_metric = dp.linf_distance(monotonic=True, T=dp.usize)
+    meas = (input_domain, input_metric) >> dp.m.then_gumbel_max(1., "maximize")
+    print(meas(list(range(10))))
+    assert meas.map(2) == 2
+test_gumbel_max()
```
andrewvyrros, line 204:
```suggestion

```

rust/src/metrics/mod.rs
```diff
@@ -455,45 +455,77 @@ impl<T: CheckAtom> MetricSpace for (AtomDomain<T>, DiscreteDistance) {
...
+    /// ```
+    pub fn range_distance(&self, d_in: Q) -> Fallible<Q> {
+        if self.monotonic {
+            Ok(d_in)
+        } else {
+            d_in.inf_add(&d_in)
+        }
+    }
+
+    pub fn monotonic() -> Self {
```
andrewvyrros, line 488: Would be more general to have a `new(monotonic: bool)`

andrewvyrros (approved):
I think the switch to report_noisy_max_gumbel is good.

PR #938, opened 2023-10-02
add "Above Threshold" mechanism

Shoeboxam:
Closes #937

PR #684, opened 2023-04-04
add sparse vector technique

Shoeboxam:
Closes #683

This SV IM's input is a queryable that emits a stream of bounded-sensitivity aggregates, but we don't have any interactive transformations at the moment, much less any that behave this way. 

May consider refactoring this IM to take a sensitivity in the constructor arguments. If a sensitivity is passed in the constructor arguments, then the input could be a dataset, and the query type a measurement.

PR #951, opened 2023-10-03
feat: add `make_report_noisy_max` with exponential noise

### LINKED ISSUE #950: Laplace max mechanism
The Laplace max mechanism dominates the Gumbel max in pure-DP, but vice-versa in zCDP.

Shoeboxam:
Closes #950


Shoeboxam:
* main
  * **PR #934** <a href="https://app.graphite.dev/github/pr/opendp/opendp/934" target="_blank">Graphite</a> 
    * **PR #940** <a href="https://app.graphite.dev/github/pr/opendp/opendp/940" target="_blank">Graphite</a> 
      * **PR #948** <a href="https://app.graphite.dev/github/pr/opendp/opendp/948" target="_blank">Graphite</a> 
        * **PR #951** <a href="https://app.graphite.dev/github/pr/opendp/opendp/951" target="_blank">Graphite</a>  👈
      * **PR #938** <a href="https://app.graphite.dev/github/pr/opendp/opendp/938" target="_blank">Graphite</a> 
    * **PR #684** <a href="https://app.graphite.dev/github/pr/opendp/opendp/684" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/opendp/opendp/951?utm_source=stack-comment).

rust/src/measurements/exponential_max/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+                r_scale.clone(),
+            )
+        }),
+        input_metric.clone(),
+        MaxDivergence::default(),
+        PrivacyMap::new_fallible(report_noisy_max_exponential_map(scale, input_metric)),
+    )
+}
+
+pub(crate) fn report_noisy_max_exponential_map<QI, QO>(
```
Shoeboxam, line 81: This code is a duplicate of the gumbel map, however, I anticipate the gumbel map changing.

mccalluc (commented):
The issue addressed here:
- #950

suggests how this contributes to the bigger picture. Should that be translated into docs? I'm not sure if it would fit in the API docs here, or if it would be better to file a followup issue if many pieces are coming together.

Usual concerns about bare `print` statements in logs. I think you have some expectation of what is expected, but someone looking at this in the future will need help. Even just a label on the output would be something, so if more than one test is being run, someone looking at the STDOUT knows which output came from which statement.

rust/src/measurements/exponential_max/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+) -> Fallible<Measurement<VectorDomain<AtomDomain<TIA>>, usize, LInfDistance<TIA>, MaxDivergence>>
+where
+    TIA: Number,
+    FBig: TryFrom<TIA> + TryFrom<f64>,
+    f64: InfCast<TIA>,
+{
+    if input_domain.element_domain.nullable() {
+        return fallible!(MakeMeasurement, "values must be non-null");
+    }
+
```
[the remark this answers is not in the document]
Shoeboxam: Weird, I wonder how that got in there.

rust/src/traits/samplers/psrn/exponential/test.rs
```diff
@@ -0,0 +1,28 @@
...
+#[test]
+fn test_exponential_psrn() -> Fallible<()> {
+    let (shift, scale) = (FBig::ZERO, FBig::ONE);
+    let exponential = ExponentialRV::new(shift, scale)?;
+
+    let samples = (0..1000)
+        .map(|_| PartialSample::new(exponential.clone()).value())
+        .collect::<Fallible<Vec<f64>>>()?;
+    println!("{:?}", samples);
+    Ok(())
```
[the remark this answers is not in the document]
Shoeboxam: Changed to kolmogorov-smirnov test.

mccalluc (approved):
Changes look good to me; Still wonder how and where we explain this to users, but if a more systematic docs update is on your radar, that works.

rust/src/traits/samplers/psrn/exponential/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+impl ExponentialRV {
+    pub fn new(shift: FBig, scale: FBig) -> Fallible<Self> {
+        if let Sign::Negative = scale.sign() {
+            return fallible!(FailedFunction, "scale ({}) must be non-negative", scale);
+        }
+        Ok(ExponentialRV { shift, scale })
+    }
+}
+
+impl InverseCDF for ExponentialRV {
```
tudorcebere, line 26: 🔥

python/test/test_meas.py
```diff
@@ -200,17 +200,36 @@ def test_gaussian():
...
 
     input_metric = dp.linf_distance(monotonic=True, T=dp.usize)
     meas = (input_domain, input_metric) >> dp.m.then_report_noisy_max_gumbel(1., "max")
-    print(meas(list(range(10))))
+    # fails with very small probability
+    assert meas([0, 0, 20, 0]) == 2  # because score 2 is by far the greatest
+    assert meas.map(2) == 2
+
+
+def test_report_noisy_max_exponential():
```
tudorcebere, line 220: Not sure where this should live, but I think it would be great to have somewhere a benchmark showing where it makes sense to use either one of them to discuss maybe somewhere the tradeoff between the two, so that people reading the source code get some intuition why there are 2 mechanisms for argmax currently.

rust/src/measurements/gumbel_max/mod.rs
```diff
@@ -125,7 +131,22 @@ where
     }
 }
 
-pub fn select_score<TIA>(
+pub(crate) trait NewArgmaxRV: InverseCDF + Sized {
```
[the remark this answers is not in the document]
Shoeboxam: Only took me a few minutes to swap things around.

python/test/test_meas.py
Shoeboxam: There isn't really a good reason yet until bounded-range is added, maybe we can do a pass then.

rust/src/measurements/report_noisy_max/test.rs
```diff
@@ -0,0 +1,79 @@
...
+    let input_metric = LInfDistance::default();
+    let de = make_report_noisy_max(
+        input_domain,
+        input_metric,
+        RangeDivergence,
+        1.,
+        Optimize::Max,
+    )?;
+    let release = de.invoke(&vec![1., 2., 3., 2., 1.])?;
+    println!("{:?}", release);
```
mccalluc, line 17: as with the python tests, add little explanation for the future dev about what they should expect to see.

rust/src/measurements/report_noisy_max/test.rs
```diff
@@ -0,0 +1,79 @@
...
+    )?;
+    assert_eq!(m_rnm.invoke(&input)?, expected);
+    Ok(())
+}
+
+#[test]
+fn test_max_vs_min_gumbel() -> Fallible<()> {
+    check_rnm_outcome(RangeDivergence, 0., Optimize::Max, vec![1, 2, 3], 2)?;
+    check_rnm_outcome(RangeDivergence, 0., Optimize::Min, vec![1, 2, 3], 0)?;
+    check_rnm_outcome(RangeDivergence, 1., Optimize::Max, vec![1, 1, 100_000], 2)?;
```
mccalluc, line 44: I'm hoping order of candidates doesn't matter. Add a test to check?

rust/src/measurements/report_noisy_max/test.rs
```diff
@@ -0,0 +1,79 @@
...
+#[test]
+fn test_rnm_exponential() -> Fallible<()> {
+    let input_domain = VectorDomain::new(AtomDomain::default());
+    let input_metric = LInfDistance::default();
+    let de = make_report_noisy_max(input_domain, input_metric, MaxDivergence, 1., Optimize::Max)?;
+    let release = de.invoke(&vec![1., 2., 3., 2., 1.])?;
+    println!("{:?}", release);
+
+    Ok(())
+}
```
mccalluc, line 64: Seems pretty similar to `check_rnm_outcome`. Put these next to each other and highlight the distinction?

rust/src/measurements/report_noisy_max/test.rs
```diff
@@ -0,0 +1,79 @@
...
+
+#[test]
+fn test_max_vs_min_exponential() -> Fallible<()> {
```
mccalluc, line 67: Ditto: Point out the differences.

mccalluc (commented):
Just looking at the last diff, a couple wording tweaks. Unless there's something particular I should look at, will approve on the next round.

mccalluc (approved):
Looks good to me. Some comments you can take or leave. There might be one extra line in the notebook, where the cell ends by evaluating a measure, but it wouldn't do any harm, if that's intentional.

rust/src/measurements/report_noisy_max/mod.rs
```diff
@@ -0,0 +1,204 @@
...
+        optimize(c_type = "char *", rust_type = "String"),
+        output_measure(c_type = "AnyMeasure *", rust_type = b"null"),
+    ),
+    generics(MO(suppress), TIA(suppress))
+)]
+/// Make a Measurement that takes a vector of scores and privately selects the index of the highest score.
+///
+/// # Arguments
+/// * `input_domain` - Domain of the input vector. Must be a non-nullable `VectorDomain`
+/// * `input_metric` - Metric on the input domain. Must be `LInfDistance`
```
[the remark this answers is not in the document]
Shoeboxam: 1. mainly for uniformity with the convention that the first two arguments are the input domain and input metric, so that it works in the context API
2. it isn't completely trivial as it carries a monotonicity bool

docs/source/api/user-guide/plugins/code/selecting-grouping-columns.txt
```diff
@@ -38,7 +38,7 @@
 ...     """Create a measurement that selects a set of grouping columns from `candidates`."""
 ...     return (
 ...         make_grouping_cols_score(candidates, min_bin_size)
-...         >> dp.m.then_report_noisy_max_gumbel(scale, optimize="max")
+...         >> dp.m.then_report_noisy_max(dp.max_divergence(), scale, optimize="max")
```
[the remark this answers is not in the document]
Shoeboxam: We have a pretty strong convention to have the first two arguments be the input domain and input metric. I could foresee the third argument for measurement constructors being the output measure, so that the expected measure could be passed in. This is also how `make_private_lazyframe` is, and how all-in-one constructors for stats in the framework API would have looked, had I ever gotten around to it.

I then try to put positional parameters that don't directly control the privacy tradeoff next. You want the positional parameters that directly control the privacy tradeoff as late as possible (but before arguments with defaults). This is because you are more likely to omit those parameters and use the Context API to solve for them later. This lets you pass arguments positionally in the Context API.

Finally comes arguments with defaults, which have to be last. In this case, the optimize argument is meant to have a default of "max" because it is consistent with the name of the function (granted, it only now has it).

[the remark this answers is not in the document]
Shoeboxam: Also, the order of arguments isn't changing.

Shoeboxam:
Closing in favor of #2220.

PR #1010, opened 2023-11-15
Move JSON out of docs and add to .vscode-suggested

### LINKED ISSUE #996: Move configurations from docs to `.vscode-suggested`
https://docs.opendp.org/en/stable/contributor/development-environment.html has a suggested `task.json`. Instead of asking devs to copy from their screen to a new file, it would be simpler to have a `.vscode-suggested` directory which can be copied to the git-ignored `.vscode`. (We do want to keep on git-ignoring: Devs should be free to tweak their local settings even if it's not something that should be contributed upstream.)

mccalluc:
- Fix #996


raprasad (approved):
Doc updates. Tests are passing

PR #1013, opened 2023-11-15
In venv example, use conventional gitignored name

### LINKED ISSUE #1012: In docs, venv instructions produce a directory which is not gitignored
https://docs.opendp.org/en/stable/contributor/development-environment.html#python-setup

```
python3 -m venv opendp
source opendp/bin/activate
```

While the venv directory could be anywhere, I think this will usually be run in the checkout. (Or at least that's what I did.)
The problem is that the resulting opendp directory is not in the gitignore list.

Suggest tweaks to the docs:
- should be run from root of checkout
- name the environment `.venv`, or something else conventional.

mccalluc:
- Fix #1012


raprasad (approved):
Approved

PR #1014, opened 2023-11-15
Replace all instances of SKIP with ELLIPSIS, adding "..." as appropriate

### LINKED ISSUE #1000: Remove `doctest: +SKIP` annotations in docs
Because these lines aren't run, we can miss examples that depend on undefined functions, or old APIs. One example is `combinators.rst`:

```
    >>> input_space = dp.vector_domain(dp.atom_domain(bounds=(0., 10.)), size=10), dp.symmetric_distance()
    >>> meas = input_space >> dp.t.then_mean() >> dp.m.then_laplace(scale=0.5)
    >>> print("standard mean:", amplified([1.] * 10)) # -> 1.03 # doctest: +SKIP
```

`amplified` has not been defined, and so this fails when a user tries to run it.

One approach is to replace `SKIP` with `ELLIPSIS`, and add `...` on the next line where output is expected. This way we can make sure every line is run, but we don't make assertions on the output.

mccalluc:
- Fix #1000 

Also remove some extra ellipsis, and replace with just a blank line for clarity.


docs/source/user/combinators.rst
```diff
@@ -223,7 +223,8 @@ The resulting measurement expects the size of the input dataset to be 10.
 
     >>> input_space = dp.vector_domain(dp.atom_domain(bounds=(0., 10.)), size=10), dp.symmetric_distance()
     >>> meas = input_space >> dp.t.then_mean() >> dp.m.then_laplace(scale=0.5)
-    >>> print("standard mean:", amplified([1.] * 10)) # -> 1.03 # doctest: +SKIP
+    >>> print("standard mean:", meas([1.] * 10)) # -> 1.03 # doctest: +ELLIPSIS
```
raprasad, line 226: nice catch

raprasad (approved):
Approved!

PR #1015, opened 2023-11-15
Instead of loading PUMS from local filesystem, use wget

### LINKED ISSUE #1002: Ensure that notebooks work without checkout
Across many of the notebooks, the sample PUMS data is loaded from the local filesystem:
```
data_path = os.path.join('..', 'data', 'PUMS_california_demographics_1000', 'data.csv')
```

This won't work if the user only has a notebook rather than a full checkout. I think there is a consensus that we want to move to something like:
```
!wget https://raw.githubusercontent.com/opendp/opendp/main/docs/source/data/PUMS_california_demographics_1000/data.csv
```

This does have downsides:
- Tests will have a network dependency where before they were reading local files. This could be slower and more fragile.
- Tests will depend on the state of the repo outside the branch. The file hasn't been touched in a long time... but if it were, tests run in the branch wouldn't see changes until merged to `main`.

One slight elaboration instead of plain `wget` would be:
```
![ -e data.csv ] || wget https://raw.githubusercontent.com/opendp/opendp/main/docs/source/data/PUMS_california_demographics_1000/data.csv
```

If notebooks are run from the same directory, this would at least reduce the number of downloads needed to run tests, and it's pretty transparent.

mccalluc:
- Fix #1002: See caveats there.
- `import os` can be dropped.
- It checks for the existence of `data.csv` before downloading. May prevent some re-downloads, but it is a little more complex.
- Assumes wget binary is available.
- Styling: I've given each wget its own cell. Maybe it should share the python cell?


raprasad (approved):
Approving.

PR #1016, opened 2023-11-15
Motivate partial constructors by pointing out redundancy

### LINKED ISSUE #1008: Motivate partial constructors
https://docs.opendp.org/en/stable/user/getting-started.html#Partial-Constructors

Suggest that instead we point out a problem, and then offer a solution. Right now we just begin by saying partial application is possible, and the motivation comes later.

We can update the example in the previous section to use input_domain, input_metric for a fair comparison

> You may notice some redundancy in the example: The output domain of a transformation will always match the input of its successor. We can make this shorter by using the `then_*` constructors: These delay application of the input_domain and input_metric arguments, so we can rewrite the example above in a more concise form:
> 
> ```
> dp_sum = (
>   (input_domain, input_metric) >>
>   dp.t.then_clamp((0., 5.)) >>
>   dp.t.then_sum() >>
>   dp.m.then_laplace(5.)
> )
> ```
>

mccalluc:
- Fix #1008

I think this is more direct, but it is subjective.


raprasad (approved):
tested locally

Issue #997, opened 2023-11-15
Less meta-programming for context methods


mccalluc:
Currently in `context.py` the `__getattr__` on `Query` is handling everything. This means 
- There is no generated documentation
- There is no contextual help in the IDE
- because the signature is very loose, ie `*args, **kwargs`, bad parameter errors come late and can be confusing.

For all the methods a typical user would need, we want to explicitly define them here. Some care should be taken to avoid large blocks of boilerplate code. Most of the lines should be the doc strings.

- [ ] laplace
- [ ] gaussian
- [ ] private_mean
  - clamps
  - either splits between a sum and count, then postprocess or just mean, depending on if data size known
  - noise addition- if self.privacy_measure is max divergence then laplace, else gaussian
- [ ] private_count
- [ ] private sum
- [ ] private_variance
- [ ] private_std
- [ ] private_histogram
- [ ] any other private_* things?
- [ ] some transformations... 
- [ ] and remove `__get_attr__` when it's no longer needed?

mccalluc:
Started work in [997-explicit-query-methods](https://github.com/opendp/opendp/compare/997-explicit-query-methods?expand=1)

mccalluc:
With the polars-based API this won't be necessary; Closing.

PR #1017, opened 2023-11-20
997 concrete mean and laplace on query

mccalluc:
- Towards #997

Rather than making a PR against `main` I've created a new branch, https://github.com/opendp/opendp/tree/997-concrete-query-methods-IN-PROGRESS, with the idea that smaller PRs could be reviewed and merged into it, rather than leaving `main` half-baked. Is this approach ok?

Questions on this bit:

- Since the Context API is designed to be used end-to-end, I'm not sure that there is a good minimal example. Is a long example like this too much clutter?
- Are there other things we do definitely want in the docs? Exceptions?
- In the draft, you had proposed `Query.extend`, which I'm interpreting as a new static method. It looks like `new_with` already does what we need: Ok to just use that?
- Are you imagining that the Laplace helper will do the binary search to determine the scale? If anything, maybe the user would supply a confidence level?
- Am I right in thinking that down-the-road, usage of both of these would be discouraged?

### DIFF
--- a/python/src/opendp/context.py
+++ b/python/src/opendp/context.py
@@ -447,7 +447,6 @@ class Query(object):
         >>> dp_mean = context.query().clamp((0.0, 10.0)).mean().laplace().release()
 
         :param scale: Noise scale parameter for the laplace distribution. `scale` == standard_deviation / sqrt(2).
-        :param k: The noise granularity in terms of 2^k.
         """
         # TODO: When they exist, add this note to docstring:
         # > The `private_mean`, `private_count`, etc. methods apply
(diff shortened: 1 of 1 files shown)


mccalluc:
Marking as draft: I changed the signature of `laplace`, but it needs to remain the same so there is no change in behavior. Tests are failing.

mccalluc:
Tests are passing: Ready for review.

Copy and pasting `-1074` doesn't seem great, but I've filed a separate issue to clean that up: 
- #1026

The `PartialChain` logic confused me for a while, but I think I've got it now.

raprasad (approved):
Approved for the docstring updates.

PR #1018, opened 2023-11-20
Add example of calling the function member

### LINKED ISSUE #1006: Give an example of calling the function member
https://docs.opendp.org/en/stable/user/programming-framework/supporting-elements.html#function

Without context I was wondering if I needed `.function` to access the member. A short example could clarify this.

mccalluc:
- Fix #1006

It wasn't obvious to me on first reading that you could just treat the Transformation itself as a function, but this example may be too long-winded. Up to you.


raprasad (approved):
Approved. (tested locally)

PR #1019, opened 2023-11-20
GH Pages 404

### LINKED ISSUE #986: Friendlier 404s
[This issue](https://github.com/opendp/opendp/issues/985) identifies a broken link from the project site to the docs site. What is currently shown is a default GitHub pages 404. I believe we can specify our own 404 page, and on it we might encourage users to use the site search.

mccalluc:
- Fix #986

[GH Pages docs](https://docs.github.com/en/pages/getting-started-with-github-pages/creating-a-custom-404-page-for-your-github-pages-site) make it sound like it's as simple as having a `404.html` at the top level.

I've run the simple doc build locally to confirm that the RST is translated into HTML... but that wouldn't be the top level in production, so I added a step to the workflow: Haven't run this, but seems plausible.

With a reorganization of the docs coming up, I'm a little worried about our 404 rate going up, if there are any deep links into our docs from outside. If we see that happening, we could enrich this to help users: Maybe a "greatest hits" list of documentation pages, or JS injected into the page that could redirect to the current page if someone arrives with an old link.


raprasad (approved):
Approved

PR #1020, opened 2023-11-20
Update list of allowed default int types, and test

### LINKED ISSUE #1004: Drop `isize` from list of possible default types
https://docs.opendp.org/en/stable/api/python/opendp.typing.html#opendp.typing.set_default_int_type

> must be one of [u8, u16, u32, u64, u128, usize, i8, i16, i32, i64, i128, isize]

`isize` doesn’t work
```
opendp.typing.set_default_int_type('isize')
…
AssertionError: T must be one of ['u8', 'u16', 'u32', 'u64', 'i8', 'i16', 'i32', 'i64', 'usize']
```

@Shoeboxam confirms that `isize` should be removed from the list.

mccalluc:
- Fix #1004

This began with isize, but when I added a test I realized that i128 and u128 also do not work as default int types. If that's not the intended behavior, then this PR might be extended to fix the Rust code.


raprasad (approved):
reviewed

Issue #256, opened 2021-08-30
Add a workflow to automatically check code style on PRs


andrewvyrros:
We should add a workflow that checks code style via rustfmt, pep8, etc, and leaves comments on PRs when submitted. (This is less onerous than using pre-commit hooks, and the feedback feels friendlier when it comes from a robot than a human reviewer.

Suggested implementation: https://github.com/reviewdog/reviewdog#github-action-reviewdogaction-setup

mccalluc:
This could be useful, but it's probably several parallel efforts:

## python

Flake8 (a superset of pep8) has been in place for a while: https://github.com/opendp/opendp/pull/62/files ... but we're only requiring a subset of its rules to pass: any other violations are just warnings:

```
      - name: Lint with flake8
        run: |
          # stop the build if there are Python syntax errors or undefined names
          flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
          # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
          flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
```
("`--select errors`: Limit the reported error codes to codes prefix-matched by this list.")

It would be low-cost to see how many other rules we are already in compliance with, and require those to pass going forward.

### ... inside notebooks?

Options include https://github.com/nbQA-dev/nbQA and https://github.com/s-weigand/flake8-nb.

## rust

rustfmt is mentioned in the instructions, but it is not used in CI, as far as I can tell.

clippy is complementary to rustfmt: Also mentioned in instructions, but not used in CI as far as I can tell.

## r

[lintr](https://lintr.r-lib.org/) seems to be the default. I don't think there is yet any R code in `main`, but this might be something to have in place when it is merged to `main`.

## reviewdog

It looks nice: I think the first step would be to just to get the cli tools running in ci, and once something is there, we could use this for a prettier UI.

raprasad:
- closing and opened #1342

PR #1027, opened 2023-11-21
Add `rustfmt` in CI

mccalluc:
- Towards #256

Filing as draft. There are a lot of warnings, locally and on CI. Since there are lots of PRs out right now, making changes to the formatting in this branch would just mean a lot of merge conflict headaches. Instead I'd suggest that the Rust PRs that are out now should be formatted before being merged, then this branch can be updated, and hopefully we'll have just a a few remaining formatting issues to address.


mccalluc:
@Shoeboxam -- It's probably in your court to decide whether this would be useful or not. For the other CI checks, I've been checking in a long list of ignores, but:
- You might not be interested in any of these automated suggestions
- There don't seem to be [config options](https://rust-lang.github.io/rustfmt/?version=master&search=) to just ignore most formatting rules -- Instead the options let you pick from a set.

An alternative or complement would be to run clippy in CI. Your call?

Shoeboxam (changes requested):
This has come up in some previous discussions, the downside being that it's another step for new contributors, but I think this is a particularly small step. It will also help reduce reformatting diffs in unrelated PRs. So I'm generally positive on adding rustfmt to CI. I don't think we should add clippy at this point.

* Can you move the check to before the build step, so that potential failures happen more quickly?
* I think a reformatting PR should be merged first, so that this PR's CI can pass.

Shoeboxam:
Two more changes to improve end-user experience:
1. Use `cargo fmt --all -- --check` instead. It's clearer that `cargo fmt --all` will work.
2. Update the step name to something more descriptive, like `Check formatting (cargo fmt --all)`

mccalluc:
... was actually expecting a failure... unless this is a different set of styling rules? Need to investigate.

Shoeboxam:
> ... was actually expecting a failure... unless this is a different set of styling rules? Need to investigate.

It succeeded because the `||`: if the fmt check fails, then the echo will still succeed, returning exit status 0.

mccalluc:
[Twenty minutes ago](https://github.com/opendp/opendp/actions/runs/7183438533/job/19562225105#step:4:9) version was 1.74.0.

[Latest run](https://github.com/opendp/opendp/actions/runs/7183664151/job/19562948334#step:4:9) version is 1.74.1.

raprasad (approved):
approved

Issue #1022, opened 2023-11-21
Configure pytest so one invocation runs all python tests?


mccalluc:
Right now, I think `smoke-test.yml` runs python tests in three different places:

Normal tests:
```
      - name: Test
        run: pytest -v
```

Doctests in docs:
```
      - name: Test docs
        run: |
          cd ../docs
          python -m pip install -r requirements.txt
          make doctest-python
```

Check that notebooks run without exceptions:
```
      - name: Test
        run: |
          shopt -s globstar
          pytest --nbmake source/**/*.ipynb -n=auto
```

It may be a little simpler to just run pytest once from the top level with appropriate configs to cover all of this. Replicating the doctests locally right now is slow, because they follow the full documentation build in `make doctest-python`. We might also be missing things: because we're running the tests in different places, we don't get one count for the total number of tests run... if we misconfigured something and a set of tests stopped running, we don't have a count or a coverage measure that could show that something changed.

mccalluc:
I think this was closed inadvertently -- #1029 is related, but doesn't address this completely.

Shoeboxam:
The notebooks are run in a separate task on purpose, so that both sets of tests can be executed simultaneously. It shaves up to ~3 minutes of time off of CI.

PR #1029, opened 2023-11-21
CI fails if python code is untested, unless explicitly ignored

### LINKED ISSUE #1028: Check test coverage just for Python
Higher level issue: 
- #1024

mccalluc:
- Fix #1028
- Since this adds another dev dependency, formalize it with a `requirements-dev.txt`, and loosely pin dependencies.
- Intersects with #1022 -- That may expand our test coverage.
  - If that's resolved first, confirm that all of the ignores here are still needed.
  - If this is resolved first, then add a note there to confirm that all ignores are needed before merging.

Filing as draft.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -98,7 +98,10 @@ jobs:
         run: python -m pip install -e .
 
       - name: Test
-        run: pytest -v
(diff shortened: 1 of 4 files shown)


raprasad (commented):
Approved

raprasad (approved):
approved

Issue #1025, opened 2023-11-21
Python API Measures: Links from Python to Rust are broken


raprasad:
**Describe the bug**
**Page**: https://docs.opendp.org/en/stable/api/python/opendp.measures.html
**What**:
- Links under each function titled "[_function name_] in Rust documentation" are broken
- For example, the page has a link titled: "fixed_smoothed_max_divergence in Rust documentation" which goes to a broken link: 
   - https://docs.rs/opendp/0.8.0/opendp/measures/fn.fixed_smoothed_max_divergence.html
   - However, it should go here:
      - https://docs.rs/opendp/0.8.0/opendp/measures/struct.FixedSmoothedMaxDivergence.html
   - `fn.fixed_smoothed_max_divergence` should be `struct.FixedSmoothedMaxDivergence`
- Other Python Measures API links to the Rust API have the same issue

mccalluc:
Slightly different diagnosis:
> Construct an instance of the `FixedSmoothedMaxDivergence` measure.
>
> [fixed_smoothed_max_divergence in Rust documentation.](https://docs.rs/opendp/0.8.0/opendp/measures/fn.fixed_smoothed_max_divergence.html)

- It would be nice if the class name in the description were a working link to the python class docs,
  - but although `from opendp.typing import FixedSmoothedMaxDivergence` works, it is not in the [python typing docs](https://docs.opendp.org/en/stable/api/python/opendp.typing.html)
  - ... if it were, those docs could link to the corresponding rust docs 
- It's simplest if the link to the rust documentation points to a Rust element that parallels the Python element. `function -> function` seems simple; `function -> struct` seems complicated. We could instead:
  - expose rust documentation for functions?
  - not generate links for rust functions?

mccalluc:
Michael notes that R PR should fix this:
> I made changes in the R PR to omit links in cases where there is no corresponding rust function. This should catch most of those cases

mccalluc:
Still not working. Two issues:

Release naming convention: `latest` goes to the last release, not nightly, because we're suffixing the previous release (I think)

Object naming conventions: 
- docs.rs has: https://docs.rs/opendp/0.9.2-nightly.20240328.1/opendp/measures/ffi/fn.fixed_smoothed_max_divergence.html
- while we link to: https://docs.rs/opendp/latest/opendp/measures/fn.fixed_smoothed_max_divergence.html

We're missing a `ffi` in the path.

mccalluc:
[...]

Issue #983, opened 2023-11-01
Fix broken link and scan for others


mccalluc:
Problem: On https://opendp.org/tools there is a link to [Developer Guide](https://docs.opendp.org/en/stable/developer/index.html), but it is 404.

Suggestions:
- Scan the project site for other broken links, and make sure whatever tool is used identifies at least this one.
- Fix this link or just remove it.
- Decide how we can keep on top of dead links in the future.

Prerequisite: Google Search Console can help identify dead links, so consider enabling that before tackling this:
- #969

PR #1043, opened 2023-11-27
Run linkchecker in CI

mccalluc:
Fix some broken links, and add checks so we'll notice new ones. There isn't an issue filed for exactly this work.

Related issues:
- #983: Originally filed for links on the project site, not the docs site.
- #1025: There are actually a lot of broken links to the rust docs, all following the same pattern. No attempt to fix them here, but at least we know the scope of the problem 

Sphinx also offers [linkchecking](https://www.sphinx-doc.org/en/master/usage/builders/index.html#sphinx.builders.linkcheck.CheckExternalLinksBuilder) ... but it seems safer to use a separate tool for this, rather than putting more eggs in that basket.

Separately, there is also a sphinx [nitpick](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-nitpicky) option: I'm not clear exactly what it does.


raprasad (approved):
Approved. (linkchecker!)

PR #1044, opened 2023-11-27
extrinsic descriptor in user_domain

### LINKED ISSUE #994: `user_domain` descriptor
`user_domain` currently allows the carrier type to be a `void*`, but doesn't have a way to define descriptors.

Shoeboxam:
Closes #994 

Allows user-defined domains to carry arbitrary data in the native language.
Example usage:

```python
from datetime import datetime

def datetime_domain(months=None):
    """The domain of datetimes, restricted by user-defined months"""
    if months is None:
        months = set(range(1, 13)) # any month
    
    return dp.user_domain(
        identifier=f"DatetimeDomain(months={months})", 
        member=lambda x: isinstance(x, datetime) and x.month in months,
        descriptor=months)

domain = datetime_domain(months={1, 2, 3, 4})
assert str(domain).startswith("DatetimeDomain")

element = datetime.strptime('03/17/20 4:32:34', '%m/%d/%y %H:%M:%S')
assert domain.member(element)
assert not domain.member("A")
# can retrieve the descriptor for use in further analysis
assert domain.descriptor == {1, 2, 3, 4}
```

### DIFF
--- a/python/src/opendp/_convert.py
+++ b/python/src/opendp/_convert.py
@@ -7,7 +7,7 @@ from opendp.typing import RuntimeType, RuntimeTypeDescriptor, Vec
 try:
     import numpy as np # type: ignore[import-not-found]
 except ImportError: # pragma: no cover
-    np = None  # type: ignore[assignment]
+    np = None # type: ignore[assignment]
 
 ATOM_MAP = {
     'f32': ctypes.c_float,
--- a/python/src/opendp/_convert.py
+++ b/python/src/opendp/_convert.py
@@ -71,9 +71,6 @@ def py_to_c(value: Any, c_type, type_name: RuntimeTypeDescriptor = None) -> Any:
 
     if isinstance(type_name, str):
         type_name = RuntimeType.parse(type_name)
-    
-    if type_name == ExtrinsicObject:
-        type_name = "ExtrinsicObject"
 
     if isinstance(value, c_type):
         return value
@@ -83,6 +80,12 @@ def py_to_c(value: Any, c_type, type_name: RuntimeTypeDescriptor = None) -> Any:
     
     if c_type == TransitionFn:
         return _wrap_py_transition(value, type_name)
+
+    if c_type == ExtrinsicObjectPtr:
+        # since the memory is allocated by python, 
+        #    don't actually return an ExtrinsicObjectPtr, 
+        #    which would call rust to free the Python-allocated ExtrinsicObject
+        return ctypes.pointer(ExtrinsicObject(ctypes.py_object(value), c_counter))
 
     # check that the type name is consistent with the value
     if type_name is not None:
@@ -163,10 +166,6 @@ def c_to_py(value: Any) -> Any:
         value_contents = value.contents.value
         bool_free(value)
         return value_contents
-
-    if isinstance(value, (Transformation, Measurement)):
-        # these types are meant to pass through
-        return value
 
     if isinstance(value, ctypes.POINTER(ExtrinsicObject)):
         return value.contents.ptr
--- a/python/src/opendp/_convert.py
+++ b/python/src/opendp/_convert.py
@@ -271,8 +271,7 @@ def _refcounter(ptr, increment):
 c_counter = ctypes.CFUNCTYPE(ctypes.c_bool, ctypes.py_object, ctypes.c_bool)(_refcounter)
 
 def _extrinsic_to_slice(val) -> FfiSlicePtr:
-    val = ctypes.py_object(val)
-    return _wrap_in_slice(ctypes.pointer(ExtrinsicObject(val, c_counter)), 1)
+    return _wrap_in_slice(ctypes.pointer(ExtrinsicObject(ctypes.py_object(val), c_counter)), 1)
 
 def _slice_to_extrinsic(raw: FfiSlicePtr):
     return ctypes.cast(raw.contents.ptr, ctypes.POINTER(ExtrinsicObject)).contents.ptr
--- a/python/src/opendp/_convert.py
+++ b/python/src/opendp/_convert.py
@@ -57,7 +57,7 @@ def check_similar_scalar(expected, value):
 def check_c_int_cast(v, type_name):
     lower, upper = INT_SIZES[type_name]
     if not (lower <= v <= upper):
-        raise ValueError(f"value is not representable by {type_name}")
+        raise ValueError(f"{v} is not representable by {type_name}")
 
 
 def py_to_c(value: Any, c_type, type_name: RuntimeTypeDescriptor = None) -> Any: # type: ignore[assignment]
@@ -71,6 +71,9 @@ def py_to_c(value: Any, c_type, type_name: RuntimeTypeDescriptor = None) -> Any:
 
     if isinstance(type_name, str):
         type_name = RuntimeType.parse(type_name)
+    
+    if type_name == ExtrinsicObject:
+        type_name = "ExtrinsicObject"
 
     if isinstance(value, c_type):
         return value
@@ -164,6 +167,9 @@ def c_to_py(value: Any) -> Any:
     if isinstance(value, (Transformation, Measurement)):
         # these types are meant to pass through
         return value
+
+    if isinstance(value, ctypes.POINTER(ExtrinsicObject)):
+        return value.contents.ptr
 
     if isinstance(value, ctypes.c_void_p):
         # returned void pointers are interpreted as None
@@ -265,7 +271,8 @@ def _refcounter(ptr, increment):
 c_counter = ctypes.CFUNCTYPE(ctypes.c_bool, ctypes.py_object, ctypes.c_bool)(_refcounter)
 
 def _extrinsic_to_slice(val) -> FfiSlicePtr:
-    return _wrap_in_slice(ctypes.pointer(ExtrinsicObject(ctypes.py_object(val), c_counter)), 1)
+    val = ctypes.py_object(val)
+    return _wrap_in_slice(ctypes.pointer(ExtrinsicObject(val, c_counter)), 1)
 
 def _slice_to_extrinsic(raw: FfiSlicePtr):
     return ctypes.cast(raw.contents.ptr, ctypes.POINTER(ExtrinsicObject)).contents.ptr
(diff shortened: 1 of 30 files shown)


Shoeboxam:
* main
  * **PR #1044** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1044?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #1045** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1045?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/opendp/opendp/1044?utm_source=stack-comment).

rust/src/core/ffi.rs
```diff
@@ -240,7 +240,8 @@ pub extern "C" fn opendp_core___error_free(this: *mut FfiError) -> bool {
 #[bootstrap(
     name = "transformation_input_domain",
     arguments(this(rust_type = b"null")),
-    returns(c_type = "FfiResult<AnyDomain *>", do_not_convert = true)
+    returns(c_type = "FfiResult<AnyDomain *>", do_not_convert = true),
+    dependencies("this")
```
Shoeboxam, line 244: Document this on bootstrap!

rust/src/domains/ffi.rs
```diff
@@ -309,15 +309,56 @@ pub extern "C" fn opendp_domains__map_domain(
     }.into()
 }
 
+pub struct ExtrinsicElement {
```
Shoeboxam, line 312: More docs!

rust/src/domains/ffi.rs
```diff
@@ -309,15 +309,56 @@ pub extern "C" fn opendp_domains__map_domain(
...
+    }
+}
+impl Drop for ExtrinsicElement {
+    fn drop(&mut self) {
+        (self.value.count)(self.value.ptr, false);
+    }
+}
+
+unsafe impl Send for ExtrinsicElement {}
+unsafe impl Sync for ExtrinsicElement {}
```
Shoeboxam, line 351: Leave a comment here.

rust/src/domains/ffi.rs
```diff
@@ -338,29 +379,51 @@ impl Domain for UserDomain {
...
+    descriptor: *mut AnyObject,
 ) -> FfiResult<*mut AnyDomain> {
-    let descriptor = try_!(to_str(descriptor)).to_string();
-    let func = move |arg: &ExtrinsicObject| -> Fallible<AnyObject> {
-        let res = member(AnyObject::new_raw(arg.clone()));
-        util::into_owned(res)?.into()
-    };
+    let identifier = try_!(to_str(identifier)).to_string();
+    let descriptor = try_!(try_as_ref!(descriptor).downcast_ref::<ExtrinsicObject>()).clone();
+    let descriptor = ExtrinsicElement::new(identifier, descriptor);
```
Shoeboxam, line 406: Tune the naming here to not shadow.

Shoeboxam (commented):
From our review session. Thanks all!

python/src/opendp/mod.py
```diff
@@ -492,6 +497,11 @@ def _depends_on(self, *args):
         """Extends the memory lifetime of args to the lifetime of self."""
         setattr(self, "_dependencies", args)
 
+    def __getattr__(self, name: str) -> Any:
+        from opendp.domains import _user_domain_descriptor
+        descriptor = _user_domain_descriptor(self)
+        return getattr(descriptor, name)
```
mccalluc, line 503:
```suggestion
        return getattr(self.descriptor, name)
```

python/src/opendp/typing.py
```diff
@@ -125,8 +125,8 @@ def __init__(self, origin, args=None):
     def __eq__(self, other):
         if isinstance(other, str):
```
mccalluc, line 126: It's not a change that is coming in with this PR, but it feels weird that the result of `==` could vary, depending on whether a string is on the LHS or RHS. I feel a bit uncomfortable with the flexibility on RuntimeType arguments, but this PR isn't the place to try to make changes.

python/test/test_core.py
```diff
@@ -192,10 +192,34 @@ def test_new_domain():
...
     from datetime import datetime
 
-    domain = dp.user_domain("all datetimes", lambda x: isinstance(x, datetime))
-    assert str(domain) == 'UserDomain("all datetimes")'
-    assert domain.member(datetime.now())
+    def datetime_domain(months=None):
+        """The domain of datetimes, restricted by user-defined months"""
+        if months is None:
+            months = set(range(1, 13))  # any month
+
```
mccalluc, line 199:
```suggestion
    def datetime_domain(months):
        """A domain of datetimes, restricted to the specified months"""
```
We don't need default behavior for the test, so simplify.

python/src/opendp/mod.py
mccalluc: I also wonder if this is really a feature we want. It's a little bit of a convenience, but I could imagine it causing confusion: For example, if the user thinks they are accessing an attribute of `descriptor` but they actually get an attribute of this object.

mccalluc (commented):
I guess my biggest question is just one of process: Since we've looked through this before, should review be restricted to confirming that fixes from the last round of comments make sense? Or should I be looking at it with fresh eyes? I don't want you to be stuck in a loop, and I don't want the main effort to be hung up on quibbles.

python/src/opendp/_data.py
```diff
@@ -49,6 +50,34 @@ def bool_free(
...
 
 
+@versioned
+def extrinsic_object_free(
+    this
+):
+    r"""Internal function. Free the memory associated with `this`, a string.
+    Used to clean up after the type getter functions.
+    
+    [extrinsic_object_free in Rust documentation.](https://docs.rs/opendp/latest/opendp/data/fn.extrinsic_object_free.html)
```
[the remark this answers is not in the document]
Shoeboxam: This is an old, and frustrating issue- the codegen is in Rust, and there wasn't an easy way to run pandoc from Rust to convert markdown to rst. So the conversion is handled in a function in conf.py, instead. 

Potential fixes:
* Convert to RST from Rust
* Configure sphinx to run in a "markdown mode", which last I checked, was not fully featured (although it's been a couple years now)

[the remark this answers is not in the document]
Shoeboxam: I think the conf.py hook is a bit too magical, but it's been a convenient patch for this in the interim.

python/src/opendp/_data.py
```diff
@@ -49,6 +50,34 @@ def bool_free(
...
+    c_this = this
+    
+    # Call library function.
+    lib_function = lib.opendp_data__extrinsic_object_free
+    lib_function.argtypes = [ExtrinsicObjectPtr]
+    lib_function.restype = FfiResult
+    
+    output = c_to_py(unwrap(lib_function(c_this), ctypes.c_void_p))
+    
+    return output
```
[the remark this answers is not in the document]
Shoeboxam: This is an artifact of the codegen. There's sometimes some code inserted between these steps.

python/src/opendp/_convert.py
```diff
@@ -71,6 +71,9 @@ def py_to_c(value: Any, c_type, type_name: RuntimeTypeDescriptor = None) -> Any:
 
     if isinstance(type_name, str):
         type_name = RuntimeType.parse(type_name)
+    
+    if type_name == ExtrinsicObject:
+        type_name = "ExtrinsicObject"
```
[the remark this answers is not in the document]
Shoeboxam: This is a great comment. For context, as implemented, the parsed representation of an atomic type (like "f64") is still a string ("f64"), which is why the latter code worked.

I made some adjustments to simplify the process. Mainly, ExtrinsicObjects are now passed directly over Ffi, instead of undergoing the process of being packed up into an AnyObject.

python/src/opendp/typing.py
Shoeboxam: I mentioned in another comment how parsed primitive/atomic types are still strings, which is why you get this behavior. 
To recap, the result of == should only vary if the LHS is not parsed: like if LHS is "Vec<u64>". The RHS gets promoted, but the LHS does not.

We could revisit this. The equivalent parser in the R bindings wraps primitive types (it made things much easier in that setting).

python/src/opendp/mod.py
Shoeboxam: How do you feel about a shorter accessor instead?

```python
@property
def d(self):
    return self.descriptor
```

There's a lot of code that accesses attributes downstream: `input_domain.d.norm`, `input_domain.d.origin`, etc.

python/src/opendp/_data.py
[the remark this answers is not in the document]
mccalluc: @Shoeboxam -- I remember why the broken link isn't causing a CI problem: We had too many false positives, and more generally didn't want to have tons of network dependencies during smoke-test, so the external links are now only checked by the weekly cronjob. Sorry not to remember.

python/src/opendp/mod.py
mccalluc: I feel a little queasy about the single letter names that are already in the API, but that's just me.

For me, I'm less concerned about saving keystrokes and more interested in how we talk about the object. If this is just a shortcut I'm uncomfortable with it, but if `descriptor` is really an implementation detail we want to hide, then great, let's do that!

mccalluc (approved):
My last thought on `descriptor` is that I'm not really comfortable with the `getattr` if it's just saving keypresses, but if conceptually the `descriptor` is an implementation detail, and we really do want to talk about a `domain` having a `norm`, then the direct access makes sense. Your call. Approved!

python/test/test_typing.py
```diff
@@ -49,6 +49,7 @@ def test_sensitivity():
     assert isinstance(L1Distance[int], SensitivityMetric)
     assert not isinstance(RuntimeType.parse('(f32)'), SensitivityMetric)
     assert str(L1Distance[int]) == "L1Distance<i32>"
+    assert L1Distance[int] != {}
```
mccalluc, line 52: The assert just above would seem to guarantee that this would also pass: Not sure what it adds?

mccalluc (approved):
Approved!

python/test/test_typing.py
Shoeboxam: It makes it so that code coverage is complete- a RuntimeType should not generally be equal to non-RuntimeTypes.

PR #1048, opened 2023-11-27
Add missing context methods so we no longer need `__getattr__`

mccalluc:


mccalluc:
With the shift to the Polars API, and this issue closed:
- #997

we won't be going in this direction. Closing.

PR #1049, opened 2023-11-28
In `then_*` docs, point to `make_*`

### LINKED ISSUE #935: Add `seealso::` directive to `then_*` Sphinx documentation
The `then_*` docs should contain a stub and link to corresponding `make_*` function.

mccalluc:
- Fix #935 

Spent some time trying to get the restructured text syntax right, but decided that that wasn't the most important thing.

Most of the changes here come from a perl one-liner:
```
perl -e 'undef $/; $_ = <>; s/then_(\w+)\([^)]+\):\n/$&    r"""Partial constructructor which uses the preceding output domain and metric as its input domain and metric. See also:\n\n    :py:func:`opendp.transformations.make_$1`\n    """\n/g; print' python/src/opendp/transformations.py > python/src/opendp/transformations.new.py
```

And at the end there's a very parameterized test that checks that the documentation is in place -- Possible that we'll add more `then_*` functions, and we'd want to be sure they got the same documentation.

Happy to take suggestions on the RST syntax, and the content of this string.

### DIFF
--- a/python/src/opendp/measurements.py
+++ b/python/src/opendp/measurements.py
@@ -121,8 +121,7 @@ def then_alp_queryable(
     alpha = 4,
     CO: Optional[RuntimeTypeDescriptor] = None
 ):
-    r"""Partial constructructor which uses the preceding output domain and metric
-    as its input domain and metric. See also:
+    r"""Partial constructructor which uses the preceding output domain and metric as its input domain and metric. See also:
 
     :py:func:`opendp.measurements.make_alp_queryable`
     """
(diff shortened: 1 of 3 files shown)


mccalluc:
Just reading through conf.py now, I see that we have custom logic for handling docstrings... possible that something there is obstructing normal RST processing? I could look into it more, if that would be useful.

Shoeboxam:
@mccalluc In retrospect, this was a PR that made edits to an auto-generated file. This means the diffs in this file will get blown anytime the rust library is built! It's a great idea though. Would need edits in `rust/opendp_tooling/src/codegen/python.rs`.

mccalluc:
(Just reopened the original issue. Sorry!)

PR #1055, opened 2023-11-29
Banner links:  Getting Started / API / Theory / Contributing

mccalluc:
- Towards #1053: make a minimal set of changes to reorganize the docs under the new banner links
- "user/" -> "user-guide/" for consistency with title
- Get rid of commons page here, and instead point to the Tools page on the project site
- New index pages where needed
- Split examples between Getting-started and Theory, depending on focus
- move Resources to the Theory section
- "contributor guide" -> "contributing": Make it a gerund, like "getting started"
- Move Contact to Contributing

Target of PR is a new feature branch I've just made from `main`. At the very least the flow of the text in the new arrangement needs to be checked before merging to `main`, but I wanted to keep this PR on just the reorganization.

PR #1056, opened 2023-11-29
Remove twitter and discussion from banner

mccalluc:
- Adds to #1055 (Thought this was in there, but must have reverted it by mistake.)
- twitter and github can just be on the resources page: They don't need this real estate.


raprasad (approved):
approved

Issue #1054, opened 2023-11-29
Pin python doc build dependencies


mccalluc:
I'm just getting into the doc generation, and I see that most of the dependencies are unpinned; pydata-sphinx-theme in particular had a 0.14 release on September 15, and [they note](https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/index.html):
> This theme is still under active development, and we make no promises about the stability of any specific HTML structure, CSS variables, etc. Make these customizations at your own risk, and pin versions if you’re worried about breaking changes!

I also am seeing small differences in the rendering when I do my local build, vs the deployed:
<img width="1127" alt="Screenshot 2023-11-29 at 9 46 29 AM" src="https://github.com/opendp/opendp/assets/730388/eb0040cf-7246-43d6-8b34-3fed31450156">

Proposal:
- [ ] pin `pydata-sphinx-theme==13.3`, last of that major version... Unless we know we want something later?
- [ ] pin other dependencies in docs/requirements.txt
- [ ] pin transitive dependencies, probably with pip-tools, maybe Pipfile.

PR #1057, opened 2023-11-29
Pin docs/requirements.txt

mccalluc:
Pinned dependencies to what `pip freeze` gave me, except for `pydata-sphinx-theme` which I downgraded to the last major version, set up a new venv, and ran `build html` -- got a pixel-perfect match for the current docs site, with none of the little differences I was seeing earlier.

- Towards #1054
- Before closing that issue, I would like to pin transitive dependencies as well. Will make a separate PR.
- Ambivalent about upgrading `pydata-sphinx-theme` -- Feel free to file an issue if you'd like that to happen.


mccalluc:
CI fails: Running a different python locally than on CI...

mccalluc:
It's expecting `python-test (3.8)` to run, but this PR bumps up to 3.9. My guess is that this is configured as a [required workflow](https://docs.github.com/en/actions/using-workflows/required-workflows) in github, outside the source code for this project. I'd suggest that going forward we make sure that _every_ test passes, and stop using this feature -- Github has deprecated it:

> Note: GitHub is deprecating support for required workflows for GitHub Actions. Upgrade to GitHub Enterprise Cloud and use repository rulesets instead. For more information, see "[Upgrading your account's plan](https://docs.github.com/en/billing/managing-the-plan-for-your-github-account/upgrading-your-accounts-plan)."

If we want to make sure that we still have test coverage for 3.8, then the best path is probably to split the doc build (which doesn't need to run on all versions) from the rest of the tests.

raprasad:
Note: python-test (3.8) didn't run b/c the test matrix has been updated to Python 3.9

PR #1058, opened 2023-11-29
Pin transitive doc dependencies

### LINKED ISSUE #1054: Pin python doc build dependencies

mccalluc:
- Builds on #1057 
- Completes: Fix #1054 

There are other approaches to this:
- [pipenv](https://pipenv.pypa.io/en/latest/)
- just running `pip freeze` and checking in the result.
- ... or just don't worry about the transitive dependencies: it is fairly low risk.

It's also not essential that `pip-tools` itself be part of the requirements. We could just assume the user has it installed already.


mccalluc:
On the backburner: See if #1057 is merged, and then figure out what's wrong here.

mccalluc:
Close in favor of #1117.

PR #1059, opened 2023-11-29
Add mypy and check python type annotations

### LINKED ISSUE #1023: Check python type annotations?
While python type annotations are used in places (grep for `List[`, `Any`, `->`), I don't see that we're checking the annotations anywhere: None of `mypy`, `pyre`, or `pytype` seem to be be run in CI, or mentioned in developer docs.

If there aren't other preferences, using mypy in CI seems good. I don't know that it needs to be explicitly mentioned in developer docs, unless we want to have a policy that new code should be typed? Some files (for instance `_convert.py`) seem to have fairly good typing coverage.

mccalluc:
- Fix #1023
- Builds on #1029

Even with the 13 error codes we ignore here, this is protecting us against a number of other possible errors: I think there are 47 mypy error codes total:
```
wget https://raw.githubusercontent.com/python/mypy/master/docs/source/error_code_list.rst
grep '.. _code-' error_code_list.rst | perl -pne 's/.*_code-//; s/://' | sort | wc -l
      47
```

If this much is good, can file a follow-up issue to address the errors in the ignore list: They might be the result of misusing the tool and not real errors at all, for all I know.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -88,6 +88,10 @@ jobs:
           # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
           flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
 
+      - name: Type check with mypy
+        run: |
+          mypy .
+
       - name: Download libs
         uses: actions/download-artifact@v3
         with:
--- a/python/.mypy.ini
+++ b/python/.mypy.ini
@@ -1,16 +1,33 @@
 [mypy]
 
-disable_error_code = 
+# To generate list:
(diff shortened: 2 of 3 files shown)


mccalluc:
Notes on some errors it reports, if I comment out individual lines in the `ini`:
```
src/opendp/mod.py:278: error: Too many arguments for "__call__"  [call-arg]
```
```python
        if isinstance(other, PartialConstructor):
            return self >> other(self.output_domain, self.output_metric)
```

Haven't been able to make a minimal reproduce: This seems analogous, but doesn't produce a mypy error:
```python
from typing import Union, Any

class MyCallable(object):
    def __call__(self, a, b):
        print(a, b)

def demo(mystery_obj: Union["MyCallable", Any]):
    if isinstance(mystery_obj, MyCallable):
        mystery_obj("foo", "bar")
```

raprasad (approved):
approved

PR #1060, opened 2023-11-30
Fix easy typing errors

mccalluc:
- Builds on #1059

Innocuous changes so we can let mypy report on more errors, and limit most of the ignores to the individual files where we need them.


raprasad (approved):
approved

PR #1063, opened 2023-11-30
Proposal: Docstrings on autogenerated modules

### LINKED ISSUE #1032: Docstrings on modules?
Inspired by:
- #1031

Having something human readable at the top of the each module in the API docs would be nice: What is this collection of functions and classes about? See if there's a way to have it flow from the Rust code, or alternatively, it might be possible to insert the docstrings either in the generated python, or after `sphinx-apidoc` runs.

mccalluc:
- Fix #1032

Eventually this may be expanded, but something simple for now. We want to point users from the narrative docs to the API... and eventually we also want to point them back from the API docs to the narrative for more context.

Alternatives:
- Add module level docs in rust: Not a bad thing, but it would prevent us from introducing python specifics.
- Post process generated docs: This doesn't have to be in the rust code... but adding an additional step after `cargo build --features bindings-python` feels like it just adds complexity.
- Put it in the python doc build... but it is nice to have this in the actual python code, so it's available to the IDE, for example.

It looks like `_data.py` may have been hand editted... probably by me. There may be a linting error to fix...

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -107,6 +107,7 @@ jobs:
           cd ../docs
           python -m pip install -r requirements.txt
           make doctest-python
+          linkchecker -f linkchecker.cfg build/html/index.html
 
 
   python-test-notebooks:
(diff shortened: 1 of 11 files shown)


raprasad (approved):
Nice work with `python.rs`

PR #1064, opened 2023-12-01
cp data.csv, add header row

mccalluc:
My sense is that the new polars APIs will be able to read headers when creating the dataframe, and that's what we'd like to demonstrate? But for older APIs we'd still want a CSV without the header rows? 

(Note that the target of this PR is the docs feature branch, so it's not going right to `main`, but having an answer to these questions would be good.)


raprasad (approved):
approved

Issue #1070, opened 2023-12-05
Fix links to PDFs


mccalluc:
With
- #1043 
 
in place, we are checking links, but ignoring existing 404s until they can be fixed: This includes 10 404s for PDFs. Fix these, remove the ignore from `linkchecker.cfg`, and make sure there are no unreferenced PDFs.

Do not confuse this issue with 
- #1047

Checking links _in_ PDFs vs links _to_ PDFs.

PR #1071, opened 2023-12-05
Update README and Makefile to enable PDF build

mccalluc:
Update instructions for PDF build. This works for me... but I'm not great with LaTeX so this may not be the best approach... and maybe there was something wrong in my environment and it should work even without this PR.

(Actually fixing the links to PDFs #1070 turns out to be trickier, because the links are absolute.)


raprasad (approved):
approved

PR #1073, opened 2023-12-05
Enable Sphinx nitpicks, and fix most

mccalluc:
- Fix #1068
- Fix two references, and manually confirm that the links work now.
- Filed a follow-up issue which is also fixed #1072 ... The ignores which remain are explained, and they aren't just simple typos.

If you'd like a smaller PR, I could make a new branch with just the first couple commits, which would add the check, without doing the search and replace across the codebase.

PR #1074, opened 2023-12-05
Only enable Spinx nitpicks

### LINKED ISSUE #1068: Turn on `nitpicky` for Sphinx
[sphinx docs](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-nitpicky)

>  `nitpicky`: If true, Sphinx will warn about all references where the target cannot be found.

My guess is that this might help us catch cases where we intended to reference a function, but it can't find the target: It doesn't construct a link, and so these cases aren't handled by:

- #1043

... or I might be misunderstanding this feature: Try it, and if it's not useful close this issue without any action.

mccalluc:
- Fix #1068
- Extended by #1073... but that got too big, and I wanted a simple PR with just the basics.

### DIFF
--- a/docs/source/user/transformations.rst
+++ b/docs/source/user/transformations.rst
@@ -387,7 +387,7 @@ provide the option to choose the output metric: ``L1Distance[TOA]`` or ``L2Dista
 These default to ``L1Distance[TOA]``, which chains with L1 noise mechanisms like :func:`opendp.measurements.make_base_discrete_laplace` and :func:`opendp.measurements.make_base_laplace`.
 If you set the output metric to ``L2Distance[TOA]``, you can chain with L2 mechanisms like :func:`opendp.measurements.make_gaussian`.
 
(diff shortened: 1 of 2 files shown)


docs/source/user/transformations.rst
```diff
@@ -387,7 +387,7 @@ provide the option to choose the output metric: ``L1Distance[TOA]`` or ``L2Dista
 These default to ``L1Distance[TOA]``, which chains with L1 noise mechanisms like :func:`opendp.measurements.make_base_discrete_laplace` and :func:`opendp.measurements.make_base_laplace`.
 If you set the output metric to ``L2Distance[TOA]``, you can chain with L2 mechanisms like :func:`opendp.measurements.make_gaussian`.
 
-The constructor :func:`opendp.measurements.make_count_by` does a similar aggregation as :func:`opendp.transformations.make_count_by_categories <make_count_by_categories>`,
+The constructor :func:`opendp.transformations.make_count_by` does a similar aggregation as :func:`opendp.transformations.make_count_by_categories`,
```
raprasad, line 390: Nice catch

raprasad (approved):
approved

Issue #1062, opened 2023-11-30
Ensure popular docs urls work even after reorg


mccalluc:
Take the list from [search console](https://search.google.com/search-console/links/drilldown?resource_id=sc-domain%3Aopendp.org&type=EXTERNAL&target=&domain=&hl=en-US): The docs pages that have a lot of outside links should still resolve after the reorg, even if it's only a JS redirect:

| url | # links | # sites |
| -- | -- | -- |
| https://docs.opendp.org/ | 52 | 11 |
| https://docs.opendp.org/en/stable/api/python/index.html | 45 | 8 |
| https://docs.opendp.org/en/stable/user/index.html | 42 | 9 |
| https://docs.opendp.org/en/stable/user/getting-started.html | 38 | 7 |
| https://docs.opendp.org/en/stable/api/python/opendp.mod.html | 4 | 1 |

(and this is the whole list for docs: Wish there were more, but it's good that it won't be a huge job.)

mccalluc:
- https://docs.opendp.org/en/stable/api/python/index.html - Works!
- https://docs.opendp.org/en/stable/user/index.html - JS redirect from 404
- https://docs.opendp.org/en/stable/user/getting-started.html - no redirect
- https://docs.opendp.org/en/stable/api/python/opendp.mod.html - Works!

Let me add one redirect, and this can be closed.

Issue #987, opened 2023-11-01
Decide if redirects should be used to mitigate dead links


mccalluc:
We anticipate [some reorganization](https://github.com/opendp/opendp/issues/985) of the docs, and URLs that used to work might stop working. It might be possible to handle this by adding some kind of redirect in place of any page we remove ... but this would add to our work, and may not be worth the hassle. If most of our traffic is internal navigation, we don't need to worry about someone hypothetically coming from a link stored somewhere else.

Deliverable: A consensus decision about whether we want to maintain this kind of backward compatibility. (My preference: Don't worry about it.)

See also:
- #986 
- #983
- #985

PR #1077, opened 2023-12-06
JS redirects on the 404 page

mccalluc:
- Follow-up to #987
- Towards #1062

If approved, this can be merged to main, and main can be merged to the docs feature branch, and when that is stable and we know exactly what redirects are needed these TODOS can be filled in before the feature branch is merged to main -- Only then do we close the issue.

We also still need to confirm that the 404 is working on GH pages:
- waiting on #986


raprasad (approved):
approved

PR #1078, opened 2023-12-06
Expand test matrix: test with Python 3.12

### LINKED ISSUE #1036: Expand python version test matrix
We are currently only testing against Python 3.8, while 3.12 was released last month. I'd be surprised if there are big breaking changes across these versions, but I've been surprised in the past. 

- We should rearrange the workflow before taking this on, ie #1022 -- doctests currently run after doc build, but no need to build the docs for every version. We also don't need to lint under different versions.
- Checking the endpoints of the range is probably sufficient, ie `python-version: [3.8, 3.12]`: Particularly in `typing.py` there is code that will run differently in 3.8 and 3.9.

Related: The end-of-support date for 3.8 is October of next year -- Would we want to also end support for 3.8 at that time, or is extended backward compatibility a goal?

mccalluc:
- Fix #1036

Unless there's a reason we've chosen not to test 3.12? Testing for 3.9, 3.10, and 3.11 feels like it would add a lot to CI, without corresponding benefits... but maybe it would be something to check before release?

- See also #1022:
  - It would be cleaner if the version matrix didn't need to be maintained in two places
  - We only need to run the doctests under different versions -- I don't think we care so much about running the doc build itself under different versions -- but maybe not worth refactoring.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -52,7 +52,7 @@ jobs:
         working-directory: python
     strategy:
       matrix:
-        python-version: [3.8]
+        python-version: [3.8, 3.12]
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
@@ -118,7 +118,7 @@ jobs:
         working-directory: docs
     strategy:
       matrix:
-        python-version: [3.8]
+        python-version: [3.8, 3.12]
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


mccalluc:
Hitting `Could not find a version that satisfies the requirement ipython==8.18.1`. This is the same error I hit trying to pin dependencies for the docs build:
- #1057

In that case, I fixed it by moving up to 3.9. I don't know how important it is to continue supporting 3.8, but dropping support would simplify some things. (There's some special case logic for typing that works different on 3.8 vs. 3.9 and above.)

Can talk about it tomorrow.

mccalluc:
I'm going to try fixing this first:
- #1022

ipython is only needed for highlighting in notebooks, so if we can separate the tests, which need to run on every version, from the doc build, which only needs to run on a version of our choosing, we won't have the same dependency problem.

mccalluc:
- Replaced by #1099

Issue #979, opened 2023-11-01
Simplify block diagram


mccalluc:
Problem: Some aspects of the architecture are not clear in the block diagram, and the hypothetical future projects make the diagram harder to read.

A revised diagram might highlight:
- What blocks belong to which repos
- What blocks constitute the OpenDP Commons
- New R interface
- DPCreator

It might leave out:
- Hypothetical projects
- FFI layer

mccalluc:
Proposal:

[![diagram](https://docs.google.com/drawings/d/e/2PACX-1vTrq1N3dX2Z926blC_LFsaDudxGZZG99cpQu8yX1UFG7Vv_yISuKNUhl29QXzyTFXIwU_eCnHTrufes/pub?w=852&h=295)](https://docs.google.com/drawings/d/1lFEOmpAdnMOAZ7H6rXNdD833WnBFhpOj_C_8HM7TRFc/edit)

One substantive change: The current diagram suggests that SmartNoise depends directly on the Rust code, but glancing at the code, both SmartNoise packages are python. Am I confused?

SalilVadhan:
I like the look of this diagram!  However, the OpenDP Commons should
include the OpenDP Library, but not SmartNoise.  It is more about community
ownership and governance rather than the nature of the software itself.

mccalluc:
Thank you for clarification. We're mulling over displaying https://github.com/opendp/opendp-logger as a part of the commons. Not sure if there are other good examples.

mccalluc:
Revised
- show that not everything in the org is in the commons
- emphasize that opendp library itself is in the commons
- move the "?" boxes outside entirely to suggest it is software that anyone can build on.

It feels less tidy than the first version, but more accurate.

![[v1](https://docs.google.com/drawings/d/1VRePWbnjlzAbW-fO4b9lvkjNkCzCGQRv3Z5-1j8LYCM/edit)](https://docs.google.com/drawings/d/e/2PACX-1vSDUceajUYQ_TRTQwSUUZAG_03nKy0kctQDVU6-jBJcHwWHKV_TbagZFfVWXxIpSQl4BsUYnkYwsQcR/pub?w=600)

mccalluc:
The docs now on `main` (deployed on `nightly`) have made things more clear by just focusing on OpenDP the library (rather than OpenDP the project), so we no longer need a version of this diagram here. If it has a place in the project docs, that can be handled separately.

PR #1081, opened 2023-12-06
Remove unused diagrams

mccalluc:
Unless some outside resource depends on these images being deployed with our docs, these seem to be unreferenced.

I don't know the history of the example diagrams -- If they look useful, an alternate approach would be to recover the docs that originally referenced them. One advantage of notebooks is that any images needed are embedded. (Though that also mixes binary data into what's really a big JSON file, but that's not something we can change.)

The `doc-IN-PROGRESS` branch, but not yet `main`, has no reference to the cake diagram. It wasn't my favorite: Right now my plan is that the OpenDP _library_ docs will not try to explain OpenDP _project_ or _commons_. 

Separate PRs will add new diagrams.

- Related: #979


raprasad (approved):
Approved for the `docs-IN-PROGRESS` branch

PR #1084, opened 2023-12-06
Revise core structures docs

### LINKED ISSUE #1083: Revise core structures docs
- Add a diagram
- Discuss Transformations and Measurements in parallel, but highlight where they differ.
- Transformation first, then Measurement, since that's the usual flow.

mccalluc:
- Fix #1083

Feedback is welcome, particularly about the diagram: Not sure it makes things more clear, or if it's a pile of rectangles and arrows.


raprasad (approved):
Helpful text updates.
The diagram definitely assists in visualizing the concepts.

PR #1085, opened 2023-12-06
Code tabs

### LINKED ISSUE #1052: Review multi-programming-language documentation tools
We're moving in a direction where we want to maintain and display several sets of examples in parallel:
- new Polars Python API
- classic Python API
- R API
- Rust API?

The UI  we want is tabs for each of the language options. For projects that do something like this, are they using a tool we should use? Other considerations...

- Support for restructured text, and plays well with sphinx, or offers a migration path.
- Support for python doctest syntax, but if the `>>>` is displayed, it would be nice if it's not copied and pasted
- Support for [R doctest](https://cran.r-project.org/web/packages/doctest/vignettes/doctest.html)? Not a built-in / there might be other options.
- If this replaces notebooks, make sure that variables can be defined in one "cell" and reused in later ones.
- Tabs between examples in sync: If I'm just interested in R, I don't want to have to make a mouse click to see every example.
- For a snippet in language X, we may also want to link to the API reference for language X in the same tab: ie, tabs are not just for showing code.

mccalluc:
- Fix #1052 

Adds an extension for tabs in sphinx. Chose [https://sphinx-tabs.readthedocs.io](sphinx-tabs) over [https://sphinx-inline-tabs.readthedocs.io](sphinx-inline-tabs):

- sphinx-tabs: seems a little more mature: 74 closed PRs vs. 7; 28 contributors vs. 6
- Default styling seems better: sphinx-inline-tabs defaults to no line around the tab content ... They explain this choice as minimizing clutter.
- sphinx-tabs markup is one layer deeper: sphinx-inline-tabs makes a group from successive adjacent tags, with no top-level tag. It does seem efficient, but it might also get confusing.
- sphinx-tabs comes with code highlighting integration... though I'm not sure how to use that and have doctests at the same time.

All that said, the simplicity of sphinx-inline-tabs is not a bad thing either: Either seems reasonable, and there may be other options I've missed.

I've added a little custom CSS: Without that, it's too many boxes inside of boxes. Here's how this PR renders:

 
Screenshot 2023-12-06 at 6 12 12 PM

### DIFF
--- a/docs/source/_static/css/custom.css
+++ b/docs/source/_static/css/custom.css
@@ -13,3 +13,11 @@ details summary p { display: inline }
 
 /* fix rendering of pandas dataframe tables in chrome and safari */
 div.rendered_html table { width: 100% }
+
+/* default <pre> blocks in tab looks clunky */
+.sphinx-tabs-panel div.highlight pre {
+    border: unset;
+    background-color: unset;
+    padding: 0;
+    margin: -1em 0;
+}


raprasad (approved):
Thanks for the research and example

Issue #1089, opened 2023-12-07
Workflow dry run caused update to `nightly` branch


mccalluc:
Trying to debug the 404 copy problem that is breaking the doc workflow right now, I started a workflow from the github UI, (helped by the directions here: #1082,) selecting Dry Run and Fake Build. The channel pull down requires a value, so I just left it on nightly.

It looks like the nightly branch was updated: when I update locally:
```
$ git pull
...
   211cd62b..5f140d75  nightly    -> origin/nightly
```
and [this commit](https://github.com/opendp/opendp/commit/5f140d753be0670105eef7797cf5ed005dcbb6d6) is in the branch.

Is this a mistake on my part? or should there be a change to the workflow so that dry runs don't push anything to github?

mccalluc:
No: just replicated it now, and that box is not checked. The UI looked like this:

<img width="306" alt="Screenshot 2023-12-12 at 3 25 46 PM" src="https://github.com/opendp/opendp/assets/730388/8cfac2b8-8291-4a26-a1ca-002dbbd6c8d5">

But it can see that a new commit has just been made to `nightly`:
```
commit aad41ef85e06c8df5f96a3ba62c5dea83fa65c96 (HEAD -> nightly, origin/nightly)
Author: mccalluc <mccalluc@users.noreply.github.com>
Date:   Tue Dec 12 20:26:26 2023 +0000

    Initialize and configure channel
```

mccalluc:
The only inputs to [`prepare.yml`](https://github.com/opendp/opendp/blob/main/.github/workflows/prepare.yml) are `channel`, `sync`, and `counter`. 

`sync` is used early in the workflow:
```
      - name: Initialize channel
        env:
          SYNC_FLAG: ${{ inputs.sync && '-s' || '-ns' }}
        run: python tools/channel_tool.py initialize -c ${{ inputs.channel }} $SYNC_FLAG
```
but the update to github at the end is unconditional:
```
      - name: Push changes
        run: |
          git commit -a -m "Initialize and configure channel"
          git push --force origin ${{ inputs.channel }}
```

- It will update the branch even when `sync` is not selected.
- It would ignore `dry-run`, even if it were set: It's just not an input.

I'll try to prepare a PR, but my own idea of how it should work may never have been the goal.

PR #1090, opened 2023-12-07
Fix workflow cp error

mccalluc:
- Running a workflow now to confirm that this runs without error: https://github.com/opendp/opendp/actions/runs/7133384025  ... and it is all green!
- Towards #986... but wait to close until we see it working on the docs site.
- Dry runs seem to update the channel branch, and that surprised me. Separate issue: #1089


raprasad (approved):
Approved

PR #1091, opened 2023-12-07
Apply rustfmt

mccalluc:
In https://github.com/opendp/opendp/pull/1027#pullrequestreview-1770875864 it's suggested that we rustfmt the code in a separate PR, and once that is merged to main, and main is merged to that feature branch, we can confirm that the new CI check is working.

With all the work that is is process, I'd worry that this will cause merge conflicts down the line, so if it makes more sense to close this now, and come back to it when there are fewer outstanding PRs, that would be fine. Whatever makes the most sense.


Shoeboxam (approved):
`cargo fmt` looks good!

Issue #739, opened 2023-05-18
Simulate smoke-test locally


Shoeboxam:
We should have a script to simulate the smoke-test locally, that can run as part of a pre-commit hook. This can help to avoid situations where CI failures cause lost time.

mccalluc:
This would be hard: On the Python side, we could look at something to like tox to manage different environments, but that's only part of the picture. 

I'd suggest looking at the places where we have been surprised by failures in CI, and focus on improving those; The notebook failures when python tests pass has been a problem recently, and I can try making an improvement there.

PR #1095, opened 2023-12-08
Run doctests with pytest (v2)

### LINKED ISSUE #1022: Configure pytest so one invocation runs all python tests?

mccalluc:
(Earlier PR needed to be rebased: #1094)

- Fix #1022, or at least enough to close the issue.
 
@Shoeboxam notes that it makes sense to keep nbmake separate to keep tests fast. Besides that, it seems that turning on nbmake turns off the other tests -- I couldn't find a setting that ran notebooks _and_ other tests. (If we wanted to go farther with this, there are other libraries we could look at: [pytest-notebook](http://pytest-notebook.readthedocs.io/) seems to be more like doctest, where it checks that the outputs are the same not just that the notebook run. In some ways the stricter check is nice, but we would need a lot of configuration to help it play well with our non-deterministic behavior, on top of the hassle of just using a different library.)

With this PR, `pytest` now does three things:
- unit tests
- doctests on the doc source code
- doctests on the python source code

Removed the post-sphinx-build doctest to avoid confusion about how to run doctests

Motivations:
- We want to run tests under different python version, but we're running into dependency conflicts on ipython with python 3.12. With this change, pytest is covering everything except the notebooks, so we can run pytest on different versions without worrying about the notebook dependencies on different versions.
- Towards #739 -- Most of the python testing is now covered by just running `pytest`.
- Fail faster: Don't need to go through the whole docs build to run doc tests.
- Protection against regressions: I've added `test_tests.py` -- since the tests are now being pulled from different locations, a reorganization might cause us to lose tests, and we wouldn't notice. Now there's a check on the number of tests found, and if that drops suddenly, we'll notice.


raprasad (approved):
approved

Issue #902, opened 2023-08-04
Update the OpenDP security policy


raprasad:
- Add a Security.md file with appropriate process in place regarding contacts
  - Use the [SmartNoise information ](https://github.com/opendp/smartnoise-sdk/blob/main/SECURITY.md) as a template
   - Consider adding instructions for encrypting the bug report with a public key
- Enableme CodeQL and code scanning (https://github.com/opendp/opendp/settings/security_analysis#code_scanning_settings)

mccalluc:
Main items are done: CodeQL is running, and SECURITY.md is in place.

- Filed #1122 for remaining item.

PR #1098, opened 2023-12-11
Enable CodeQL

mccalluc:
- Towards #902
- This is the version generated by github: it filled in `python` in the matrix. No modifications.
- CodeQL does not support Rust. (Github's list of tools does include Clippy.)

We haven't tackled all the warnings from flake8 and mypy yet, and I'm not sure how much overlap there is between these tools. -- It's probably low cost to merge this and then see if it gives us useful information.


raprasad (approved):
approving to see feedback

PR #1099, opened 2023-12-11
pull out linting and typing as separate jobs

mccalluc:
- Builds on #1095

With more tooling in place, we probably want to pull some of it into separate jobs. Not because we're expecting to save any CI time from parallelizing -- these checks run fast -- but to save developer time: It's annoying to fix an issue in linting ... and then CI tells you there's a problem with typing ... and then there's a problem with tests.

It does add a lot of lines to the YAML.
- Making more artifacts seems like overkill
- Most of the steps could be reduced to a single line, at some cost to readability
- use YAML shortcuts to minimize copy and paste.

Filing as draft for now.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -49,17 +49,14 @@ jobs:
     defaults:
       run:
         working-directory: python
-    strategy:
-      matrix:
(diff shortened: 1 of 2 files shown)


raprasad (approved):
Nice improvement.

PR #1106, opened 2023-12-12
Use longer param names for flags

mccalluc:
- Builds on #1105
- More transparent
- Less fragile
- Less code, when we let the `dest` be implicit.


raprasad (approved):
approved

PR #1107, opened 2023-12-12
Upgrade pydata-sphinx-theme; use default search UI

### LINKED ISSUE #990: Docs search appears not to work in mobile / narrow
To reproduce:
- Go to https://docs.opendp.org/
- Make the window narrow
- Click on the magnifying glass

Nothing seems to happen: Focus would move to the search box in the sidebar... but the sidebar remains closed. About a third of our traffic to opendp.org is on mobile, so this is a problem.

Proposed fix: 
- Make the magnifying glass a link to https://docs.opendp.org/en/stable/search.html instead. 
- Remove the search box from the sidebar: It's redundant.

Alternatively:
- Make the sidebar visible when the searchbox should have focus.

mccalluc:
- Fix #990

After updating dependencies and rebuilding, the search issue is resolved. There are several issues in their repo related to search UI. If I leave the search in the sidebar, we still have the same problem as before, even with the upgrade.

Screenshot 2023-12-12 at 5 59 34 PM


raprasad (approved):
approved

PR #1108, opened 2023-12-13
pass dry_run to prepare.yml

### LINKED ISSUE #1089: Workflow dry run caused update to `nightly` branch

mccalluc:
- Fix #1089

In release.yaml, `dry_run` is never referenced alone. I'll make a branch from this and see if I can make a top-level variable to capture `inputs.dry_run || inputs.fake`, rather than repeating it across the file.

### DIFF
--- a/.github/workflows/prepare.yml
+++ b/.github/workflows/prepare.yml
@@ -23,6 +23,10 @@ on:
         type: number
         required: false
         default: 0
+      dry_run:
+        type: boolean
+        required: false
+        default: false
     outputs:
       date:
         value: ${{ jobs.prepare.outputs.date }}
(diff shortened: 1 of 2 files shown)


raprasad (approved):
Thanks for adding this variable

PR #1109, opened 2023-12-13
Define a `dry_run` variable that incorporates `fake`

mccalluc:
- Extends #1108

I almost forgot to make `fake` imply `dry_run` in the last PR. This may make that mistake less likely in the future, but it does add complexity in another way.


raprasad (approved):
approved

Issue #1082, opened 2023-12-06
Document how to create a "one-off" PyPI release


raprasad:
- This was recently used for the DP-PCA
- Document it for other team members

Shoeboxam:
`Actions > Show More Workflows... > Release > Run Workflow`

Then choose the branch you want to release, and check dry run.

Here's how all the options work:
- The target channel controls how the release is tagged, and what semantic version is given to the release
- The "sync the Channel from upstream?" checkbox is for when you want to update the `nightly`, `beta` or `stable` branches. For when you want to commit to a new official nightly/beta/stable release.
- Update the version counter when you want to release another nightly in a single day.
- Dry runs get sent to test-pypi, and don't update the docs
- Fake is for developer convenience when debugging CI: it skips compilation and inserts dummy binaries instead

PR #1110, opened 2023-12-13
Link directly to release.yml

mccalluc:
- Refinement of #1082

Just link directly to the workflow on github, instead of explaining how to navigate the UI


raprasad (approved):
approved

PR #1117, opened 2023-12-13
Pin transitive doc dependencies v2

### LINKED ISSUE #1054: Pin python doc build dependencies

mccalluc:
- Fix #1054
- Replaces #1058

Conventionally, `requirements.in` doesn't specify versions unless we know of a constraint: For a reproducible build, everything is pinned in `requirements.txt`, and we can try recompiling occasionally to pick up bug fixes.


raprasad (approved):
Quite thorough

PR #1118, opened 2023-12-13
Add manual external link check

### LINKED ISSUE #1102: Lower frequency check for external links
Binder has been returning 500s, and in general, and it's not great to have every test run depend on multiple outside resources. Instead we could have a work that we run manually, or on a scheduled basis, perhaps weekly or monthly.

mccalluc:
- Fix #1102
- Also link checks in markdown
- Update workflows README

The workflow itself does not pass right now, but it is catching more problems, which is what we want it for.

Until it's in main, it can't be run with the github UI, but it can be run with the CLI:
```
gh workflow run weekly-link-check.yml --ref 1102-external-link-check
```


raprasad (approved):
approved

PR #1124, opened 2023-12-15
pseudocode to markdown, so CodeQL does not hit parse errors

### LINKED ISSUE #1123: CodeQL hits parse error
CodeQL is running, and is not raising any alarms, but it is not able to parse some files:

> A parse error occurred while processing /home/runner/work/opendp/opendp/rust/src/transformations/quantile\_score\_candidates/pseudocode/compute\_score.py, and as a result this file could not be analyzed. Check the syntax of the file using the python -m py\_compile command and correct any invalid syntax.

I'll reformat that pseudocode as markdown so it doesn't try to parse.

mccalluc:
- Fix #1123


raprasad (approved):
approved

Issue #1130, opened 2023-12-18
New `env.dry_run` is breaking nightly


mccalluc:
Nightly fails: https://github.com/opendp/opendp/actions/runs/7243370103/workflow
Running release.yml alone: https://github.com/opendp/opendp/actions/runs/7250654521/workflow
```
The workflow is not valid. .github/workflows/release.yml (Line: 81, Col: 16): Unrecognized named-value: 'env'. Located at position 1 within expression: env.dry_run .github/workflows/release.yml (Line: 104, Col: 16): Unrecognized named-value: 'env'. Located at position 1 within expression: env.dry_run
```

Introduced in https://github.com/opendp/opendp/pull/1113

Revert should be easy... but the idea of the variable still seems good, to make sure we always get a dry run when that's intended. So, once it's fixed, do a little research to see if this could work, if done differently.

Retrospective: Exercise code before release, and for workflows, link from PR.

mccalluc:
Hopefully fixed by #1131  -- but windows build problem prevent confirmation of fix. Keep this open until we know it works.

mccalluc:
Nightly seems to be passing, and [I've just started a release with `dry_run`](https://github.com/opendp/opendp/actions/runs/7933803785) -- assuming that works, I think this can be closed.

mccalluc:
and it ran, so I'll close this.

PR #1131, opened 2023-12-18
get rid of `env` that was causing error

mccalluc:
- Toward #1130 

It looks like I had run `release.yml` on the branch with the change... but it didn't _really_ run: https://github.com/opendp/opendp/actions/runs/7197498991 -- It just ran `debug` and exited in 10s. Lesson: green checkmark isn't enough -- confirm that it is actually doing the right thing.

Update: The workflow run is still failing on this branch: The new `--counter` flag is not being recognized… Looking up to the top:
```
  publish-python:
    ...
    steps:
      - name: Checkout repository
        uses: actions/checkout@v3
        with:
          ref: ${{ inputs.channel }}
```
So it’s running the workflow from `nightly`, not from this branch.
Normally, `prepare.yml` would have updated nightly with the changes in this PR… but since it’s a dry-run, it doesn’t.

Not sure how best to move this forward: Maybe run it without dry-run on nightly? Or do a close code review, merge, and then try running after that?

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -91,7 +91,7 @@ jobs:
           REPOSITORY: ${{ inputs.dry_run && 'testpypi' || 'pypi' }}
         run: |
           pip install -r tools/requirements-publish_tool.txt
-          python tools/publish_tool.py python -r $REPOSITORY --counter ${{ inputs.channel }}
+          python tools/publish_tool.py python -r $REPOSITORY --counter ${{ inputs.counter }}
 
   publish-rust:
     needs: publish-python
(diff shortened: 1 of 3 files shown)


mccalluc:
- Fixed an underscore-vs-dash problem
- next run failed on test.pypi because of name collision: bumped version number and re-ran.
- It fails with same error: Not sure if version number is making it to this command: `python -m twine upload -r testpypi --verbose python/wheelhouse/opendp-0.9.0a20231213001-py3-none-any.whl`

mccalluc:
https://github.com/opendp/opendp/actions/runs/7251877896/job/19755219251#step:5:111
```
publish_tool.py: error: unrecognized arguments: --counter nightly
```
Latest commit fixes this typo.

raprasad (approved):
approved

PR #1134, opened 2023-12-18
fix typo in discrete Laplace density proof

### LINKED ISSUE #1133: typo in discrete laplace density proof

Shoeboxam:
Closes #1133

### DIFF
--- a/rust/src/traits/samplers/cks20/sample_discrete_laplace.tex
+++ b/rust/src/traits/samplers/cks20/sample_discrete_laplace.tex
@@ -14,6 +14,7 @@ This proof is an adaptation of \href{https://arxiv.org/pdf/2004.00010.pdf#subsec
 \subsection*{Vetting history}
 \begin{itemize}
     \item \vettingPR{519}
+    \item \vettingPR{1134}
 \end{itemize}
 
 \section{Hoare Triple}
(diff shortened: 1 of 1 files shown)


Shoeboxam:
* `main`
  * **PR #1134** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1134?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/opendp/opendp/1134?utm_source=stack-comment).

mccalluc (commented):
Maybe double check? I'll looking at the cited reference https://arxiv.org/pdf/1709.05396.pdf and I has hoping to find a line where it defines "discrete laplace" and it would match this, but no luck. Under section 4.1 there is this

Screenshot 2024-02-01 at 10 39 04 AM

... which is more similar to the original, than to what we have after the change:

Screenshot 2024-02-01 at 10 42 39 AM

... but opposite signs on the exponents.

In any case, should the implementation also be checked? If it's based on the original form here, it might be wrong too.

Obviously out of my depth, but if you just want to confirm that it's double checked, I'll mark as approved.

Shoeboxam:
The two-sided geometric is the same distribution as the discrete Laplace. After the typo fix in this PR, the representation in  https://arxiv.org/pdf/1709.05396.pdf/CKS20/GRS12, and this proof doc are equivalent: 
If you multiply by e^(1/s) / e^(1/s) (which is just one), then you get the form in the papers. 

I've updated the notation to more closely match the papers.

mccalluc (approved):
Thanks. Approved.

PR #1135, opened 2023-12-18
pass counter down to publish_tool.py

mccalluc:
- Needed to unblock #1131

Since it is stand-alone, I've filed a separate PR... but I'm actually going to merge this branch to that other one, and if that is approved and merged, this PR can just be closed.

Given that uncertainty, filing as draft.


mccalluc:
Covered by #1131. Closing.

PR #1137, opened 2023-12-18
More idiomatic doctests

mccalluc:
Small change: makes the doctests a little more idiomatic, but it's subjective.


raprasad (approved):
approved

Issue #1035, opened 2023-11-22
Improve python test coverage


mccalluc:
Follow up to
- #1028

With machinery to check test coverage in place, we could ask if it should be improved. For most of the code, I don't have an intuition about whether a gap is a matter of concern, or not -- Might find a moment to grep for `pragma: no cover` and the other ignores, and file some more granular issues as needed.

(I don't think actual 100% coverage needs to be the goal, but it would be good to have a shared understanding about how much testing is enough.)

PR #1140, opened 2023-12-19
Improve test coverage

mccalluc:
This is primarily about gaining familiarity with the codebase. I've skipped sections where the correct behavior wasn't clear -- I don't want to add tests that blindly assert the current behavior, without understanding whether that behavior is correct or not.

- Towards #1035


raprasad (approved):
approved

Issue #1148, opened 2024-01-09
Missing text in Rust mod.rs docs


mccalluc:
The documentation for mod.rs seems like it's a skeleton: examples and content are missing. Fill in the gaps, or just tidy up and remove the empty headers for now?

> ## `Vec<Type>` and the Dispatch Pattern
>
> To work through a simple example, imagine we had a generic function like this:
>
> ## Dispatch Macro
>
> ...
> * All generic parameters must be passed as references.
>
> In order
>
> # Combinators
>
> ...
>
> This becomes an issue with the OpenDP combinators. (TODO: link to module after moving combinators to separate module.)
>
> ## Glue Structs
> ##
>
> # Memory Management

PR #1150, opened 2024-01-09
Make ffi public

### LINKED ISSUE #1149: Make FFI public?
- Working on #1128

I've gotten [this far](https://github.com/opendp/opendp-rust-example/pull/1/commits/0a81fee) and seem to be blocked because ffi is not public:

```
error[E0603]: module `ffi` is private
   --> src/lib.rs:11:13
    |
11  | use opendp::ffi::any::AnyDomain;
    |             ^^^  --- module `any` is not publicly re-exported
    |             |
    |             private module
    |
note: the module `ffi` is defined here
   --> /Users/chuck/.cargo/registry/src/index.crates.io-6f17d22bba15001f/opendp-0.8.0/src/lib.rs:170:1
    |
170 | mod ffi;
    | ^^^^^^^
```

Michael notes:
> seems like we might want that to be public, so people can write their own FFI
>
> There are functions you can use to access the AnyDomain inside of an AnyTransformation or AnyMeasurement. But I guess you can't make an instance of one yourself unless it's made public (edited)

mccalluc:
- Fix #1149
- Note that the FFI docs could use some work: #1148 -- That page was rendered as part of the documentation before, but as part of the public API cleaning it up may be a higher priority.
  - (Michael explains that we are rendering all docs, even for code that is not public, so proofs can be linked.)
- Should anything else change, assuming this is the right direction?

### DIFF
--- a/rust/src/ffi/mod.rs
+++ b/rust/src/ffi/mod.rs
@@ -3,6 +3,8 @@
 //! # Overview
 //!
 //! This module contains utilities necessary for the FFI bindings.
+//! It is public so you can write your own lightweight FFI for quick one-off language integrations,
+//! but its use should be rare: More often, you should use the Rust API directly.
 //!
 //! # Generic Functions
 //!
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -167,7 +167,7 @@ macro_rules! enclose {
 
 #[cfg(feature = "ffi")]
 #[macro_use]
-mod ffi;
+pub mod ffi;
 #[cfg(feature = "ffi")]
 #[macro_use]
 extern crate lazy_static;


Shoeboxam (changes requested):
One little pub, and the world opens for Rust crates pretending to be C. 

On second thought, there isn't much reason for people to use these. If you want to bind to the C API, you're practically not going to do it from Rust- you'd use the Rust API. 

You could argue that having these APIs marked pub (without the lock) will make it more clear that people can use these structs. But most structs aren't `#[repr(C)]`, so the rust compiler will swap the order of struct fields, and pack things differently than in C. This is part of the reason why there's helper functions for everything.

Therefore I think this should stay private, and any example code showing how to bind to it should behave in a similar way that Python does. In your example crate, instead of adding opendp as a dependency, link to the built `.so` or `.dylib` file produced by building the opendp library (see `rust/target/debug/libopendp.dylib`).

Then create secondary structs for `AnyObject`, `AnyMeasurement`, etc that hold the pointer to the true structs.

```rust
struct AnyObject2 {
    ptr: *const c_char
}


impl Debug for AnyObject2 {
    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
        let res = opendp_data__to_string(self.ptr);
        ...
	}
}
```

A downside to this is that you are missing the `.h` file... But that file is getting generated and added in the R PR.

Shoeboxam:
Reopening because, if the `ffi` module is public, people could write their own lightweight FFI for quick one-off language integrations.

Shoeboxam (approved):
Thanks for opening this! I had forgotten about this potential reason to want the module public.

PR #1158, opened 2024-01-11
write then_* comments

Shoeboxam:


Shoeboxam:
* `main`
  * **PR #1158** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1158?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/opendp/opendp/1158?utm_source=stack-comment).

python/src/opendp/measurements.py
```diff
@@ -123,10 +123,19 @@ def then_alp_queryable(
...
+):  
+    r"""partial constructor of make_alp_queryable
 
-    :py:func:`opendp.measurements.make_alp_queryable`
+    .. seealso:: 
+      Delays application of `input_domain` and `input_metric` in :py:func:`opendp.measurements.make_alp_queryable`
+
+    :param scale: Privacy loss parameter. This is equal to epsilon/sensitivity.
+    :param total_limit: Either the true value or an upper bound estimate of the sum of all values in the input.
+    :param value_limit: Upper bound on individual values (referred to as β). Entries above β are clamped.
```
mccalluc, line 134:
```suggestion
    :param value_limit: Optional upper bound on individual values (referred to as β). Entries above β are clamped.
```
In isolation, the reference to beta doesn't make things more clear, but might make sense in context -- your call.

python/src/opendp/measurements.py
```diff
@@ -123,10 +123,19 @@ def then_alp_queryable(
...
+    :param size_factor: Optional multiplier (default of 50) for setting the size of the projection.
+    :param alpha: Optional parameter (default of 4) for scaling and determining p in randomized response step.
+    :param CO:
```
mccalluc, line 137: add description, particularly since the name may not be meaningful to someone who starts here.

python/src/opendp/measurements.py
```diff
@@ -202,10 +211,15 @@ def make_base_discrete_gaussian(
...
-    r"""Partial constructructor which uses the preceding output domain and metric as its input domain and metric. See also:
+):  
+    r"""partial constructor of make_base_discrete_gaussian
+
+    .. seealso:: 
+      Delays application of `input_domain` and `input_metric` in :py:func:`opendp.measurements.make_base_discrete_gaussian`
 
-    :py:func:`opendp.measurements.make_base_discrete_gaussian`
+    :param scale: Noise scale parameter for the gaussian distribution. `scale` == standard_deviation.
+    :param MO: Output measure. The only valid measure is `ZeroConcentratedDivergence<QO>`, but QO can be any float.
```
mccalluc, line 221: - If this is processed as RST, do you want double backticks here?
- "but QO can be any float" confuses me. `QO` is a type, right? Something like `ZeroConcentratedDivergence<1.5>` doesn't make sense.
- rephrase, unless there is actually `standard_deviation` parameter?
```suggestion
    :param scale: Noise scale parameter for the gaussian distribution. Equal to the standard deviation.
    :param MO: Output measure. The only valid measure is `ZeroConcentratedDivergence<QO>`, but QO can be any float.
```

python/src/opendp/measurements.py
```diff
@@ -282,10 +296,15 @@ def make_base_discrete_laplace(
...
-    r"""Partial constructructor which uses the preceding output domain and metric as its input domain and metric. See also:
+):  
+    r"""partial constructor of make_base_discrete_laplace
 
-    :py:func:`opendp.measurements.make_base_discrete_laplace`
+    .. seealso:: 
+      Delays application of `input_domain` and `input_metric` in :py:func:`opendp.measurements.make_base_discrete_laplace`
+
+    :param scale: Noise scale parameter for the laplace distribution. `scale` == standard_deviation / sqrt(2).
+    :param QO: Data type of the output distance and scale. `f32` or `f64`.
```
mccalluc, line 306: Offer a fully qualified name for `f32`?
```suggestion
    :param QO: Optional data type of the output distance and scale. `f32` or `f64`.
```

mccalluc (approved):
I started making copy-edit changes, but that's probably not helpful, though I am curious where the param descriptions come from now. Probably just walk me through it.

Shoeboxam:
Copy-edit changes are welcome, but agreed they are out-of-scope for this PR. 

The text comes from the documentation on the corresponding Rust function.

mccalluc (approved):
Consensus that copy-edits can wait.

PR #1180, opened 2024-01-29
update rug dependency

### LINKED ISSUE #1179: Build failing because of backward incompatible update to rug
```
error[E0599]: no function or associated item named `new_custom` found for struct `ThreadRandState` in the current scope
   --> src/traits/samplers/uniform/mod.rs:174:46
    |
174 |             let mut state = ThreadRandState::new_custom(&mut rng);
    |                                              ^^^^^^^^^^ function or associated item not found in `ThreadRandState<'_>`
```

Michael Shoemate:
> Yeah I noticed this today as well, it's because Cargo considers minor versions to be a "minimum", and rug pushed an update that makes "std" an optional feature that new_custom is behind
> (a backwards-incompatible update on a minor version bump)

mccalluc:
- Fix #1179

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -37,7 +37,7 @@ rand = "0.7.3"
 num = "0.3.1"
 thiserror = "1.0.24"
 statrs = "0.13.0"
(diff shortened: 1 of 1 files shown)


Shoeboxam (approved):

Shoeboxam:
Thanks Chuck!

PR #1186, opened 2024-01-30
Remove Rust doc TODO

### LINKED ISSUE #1160: Link locally generated Rust docs into sphinx structure?
This goes back several years, to PR #300 -- https://github.com/opendp/opendp/blame/008b0e20356f49ff358eb58c69f9aff2a512b788/docs/Makefile#L32

> ```
> 	# TODO: Figure out how to link locally generated Rust docs into sphinx structure.
> ```

Does it still make sense to try to have the rust documentation inside the sphinx directory? Or are we happy linking to docs.rs? Possible advantages of rust docs inside the sphinx output directory:
- Relative links in and out of rust docs.
- Link checking against all docs.
- Possibly, text search across all docs.

Might get a better sense of whether this is a good or bad idea if I actually try to implement it... but feel free to comment.

mccalluc:
- Fix #1160 (though, at the end of that comment thread I don't have the sense that we're 100% confident about this.)

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -33,6 +33,7 @@ html:
 	@echo
 	@echo "Build finished. The HTML pages are in $(BUILDDIR)/html."
 
+# In contrast to https://docs.rs/opendp, this will document all functions, including private functions.
 html-rustdoc:
 	cd ../rust && cargo rustdoc --features untrusted,derive -- --html-in-header katex.html --document-private-items
 
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -29,8 +29,6 @@ clean:
 html:
 	@set -e
 	$(SPHINXAPIDOC) -f -F -e -H "OpenDP" -A "The OpenDP Project" -V $(VERSION) -o source/api/python ../python/src/opendp --templatedir source/_templates
-	# TODO: Figure out how to link locally generated Rust docs into sphinx structure.
-	# 	cd ../rust && $(CARGO) doc --no-deps --target-dir ../docs/source/api/rust
 	$(SPHINXBUILD) $(SPHINXOPTS) -D version=$(VERSION) -D 'html_sidebars.**'=sidebar-nav-bs.html source $(BUILDDIR)/html
 	@echo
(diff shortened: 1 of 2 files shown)


docs/Makefile
```diff
@@ -29,12 +29,11 @@ clean:
 html:
 	@set -e
 	$(SPHINXAPIDOC) -f -F -e -H "OpenDP" -A "The OpenDP Project" -V $(VERSION) -o source/api/python ../python/src/opendp --templatedir source/_templates
-	# TODO: Figure out how to link locally generated Rust docs into sphinx structure.
-	# 	cd ../rust && $(CARGO) doc --no-deps --target-dir ../docs/source/api/rust
 	$(SPHINXBUILD) $(SPHINXOPTS) -D version=$(VERSION) -D 'html_sidebars.**'=sidebar-nav-bs.html source $(BUILDDIR)/html
 	@echo
 	@echo "Build finished. The HTML pages are in $(BUILDDIR)/html."
 
+# In contrast to https://docs.rs/opendp, this will document all functions, including private functions.
```
Shoeboxam, line 36: docs.rs does document private items because of this:
https://github.com/opendp/opendp/blob/main/rust/Cargo.toml#L81-L89

Shoeboxam (changes requested):

docs/Makefile
mccalluc: Thanks -- Maybe I was trying to find the corresponding docs for a function that is in `main` but hasn't been released yet? Anyway, less confused now. Will change.

mccalluc:
@Shoeboxam -- I think these lines in `conf.py` are referring to the same idea of building the rust docs along with the rest of sphinx docs, so they can also be deleted?

Short of parsing the TOML, there's not an option that runs `rustdoc` with the values specified in `rustdoc-args`? If there were, we could have a little bit less redundancy, but I don't think it's worth worrying about.

Shoeboxam (approved):
Yeah I don't see a way to de-dupe either, unless we start parsing the toml, but now it's only in the suggested tasks. Thanks for cleaning these up!

PR #1187, opened 2024-01-30
Fix namespace pollution

### LINKED ISSUE #1163: `from opendp.context import *` pollutes prelude
The prelude.py should only import explicit names, since it makes it possible to access certain constructors directly from `dp`.

mccalluc:
- Fix #1163

The issue suggests explicit imports. I implemented that, and then had second thoughts, and added `__all__` to `context.py` instead. Either would work, but this way:
- If functions are added or removed from `context.py`, the corresponding change is in the same file, rather than in `prelude.py`
- This is consistent with the way the auto-generated files work.

This does leave `core` as an outlier: `prelude` only imports `new_function`, `new_user_queryable`, but `__all__` lists 30 items.


Shoeboxam (approved):
Changes to core will require changes in the codegen, which we could consider in another PR.

PR #1210, opened 2024-02-06
fix R docs download path in CI

Shoeboxam:
Closes #1209

PR #1213, opened 2024-02-07
Change "R" to "r" in paths

### LINKED ISSUE #1211: Rename `R` directory to `r`
Lower case is consistent with other directories, and with paths in the docs. (See #1210.)

MacOS Filesystem doesn't actually see `R` and `r` as different names, so it might need to be a two step process.

mccalluc:
- Fix #1211

I'm not 100% sure this is the right direction: We can be consistent with our other directory names, or we can be consistent with R conventions.

The prompt for this is https://github.com/opendp/opendp/pull/1210/files:
```
          path: en/stable/api/r
```
I was worried about having different capitalizations in different paths.

Consider this an experiment: After seeing the changes involved, closing this and the related issue would be fine... Or maybe it is the best course?


mccalluc:
So I think the first `R` in the path is something we could change, but I think the second `R` is more conventional. I think this was a useful experiment, but I don't think it needs to go forward. Closing PR and issue.

Issue #1069, opened 2023-12-05
Make decisions about flake8 ignores


mccalluc:
With this PR merged
- #1021

we have a number of flake8 checks in place... but there are a number we are still ignoring. I'd suggest making a few separate PRs.

- Before starting a PR for whitespace, make sure enforcing the style rule has buy-in from the team. -- It's fine to say we don't care about a particular formatting issue.
- There are also checks that could help us avoid bugs: These may be higher priority, but may also be harder to fix. (`E722: do not use bare 'except'`, `E741: ambiguous variable name`, `F405: '...' may be undefined, or defined from star imports`, `F821: undefined name '...'`)

mccalluc:
Update: We're down to
```
extend-ignore =
    # E261: at least two spaces before inline comment
    E261,
    # E306: expected 1 blank line before a nested definition, found ...
    E306,
    # E501: line too long (... > 79 characters)
    E501,
    # F821: undefined name '...'
    F821,
    # W291: trailing whitespace
    W291,
    # W293: blank line contains whitespace
    W293,
```
I'll look at the undefined named... might be tricky because of pseudocode.

For the rest, I don't feel strongly -- If you'd ask for a fix in code review, let's automate, but I'm fine ignoring some formatting quirks.

PR #1217, opened 2024-02-07
fix unneeded f-strings; stricter linting

mccalluc:
- towards #1069

Not critical, but it's something that could come up in a code review and could be automated.


Shoeboxam (dismissed):
More automation, fewer ignores!

Shoeboxam (approved):

PR #1226, opened 2024-02-08
remove reference to poly

### LINKED ISSUE #1225: Rust build warning: unused import: `poly::*`
I'm up to date on main, ran `rustup` and then `cargo build --features untrusted,bindings`: got one warning:

```
warning: unused import: `poly::*`
  --> src/domains/mod.rs:30:9
   |
30 | pub use poly::*;
   |         ^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: `opendp` (lib) generated 1 warning (run `cargo fix --lib -p opendp` to apply 1 suggestion)
```

I'll make a PR with the suggestion.

mccalluc:
- Fix #1225

Did a build from scratch with this change and there are no warnings.

### DIFF
--- a/rust/src/domains/mod.rs
+++ b/rust/src/domains/mod.rs
@@ -26,8 +26,6 @@ use std::fmt::{Debug, Formatter};
 
 #[cfg(feature = "contrib")]
 mod poly;
-#[cfg(feature = "contrib")]
-pub use poly::*;
 
 /// # Proof Definition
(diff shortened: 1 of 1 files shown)


Shoeboxam (approved):
I think this module used to have real exports, but now that it is only impls the pub use became empty? Or maybe a newer Rust compiler added a warning?

PR #1229, opened 2024-02-08
Explain relationship between `bindings` and `derive`

### LINKED ISSUE #1228: Explain `bindings` feature / fix typo
https://docs.opendp.org/en/stable/contributor/development-environment.html currently says:

> Enables the `ffi` and `derive` feature and regenerates sources in the Python and R packages.

- Why does `bindings` imply `derive`? That just seems to be a documentation configuration, but maybe it's more? Clarify.
- And make it `features` with an `s`.

mccalluc:
- Fix #1228

### DIFF
--- a/docs/source/contributor/development-environment.rst
+++ b/docs/source/contributor/development-environment.rst
@@ -79,11 +79,11 @@ Setting a feature changes how the crate compiles:
    * - ``floating-point``
      - Enable to include transformations/measurements with floating-point vulnerabilities.
    * - ``bindings``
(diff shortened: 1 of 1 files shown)


Shoeboxam (changes requested):
Could you also add an entry for the partials feature? Partials also enables derive, it's a default feature, and it's what makes it possible to call then_* functions in Rust.

Shoeboxam (approved):
Thanks a bunch!

PR #1231, opened 2024-02-08
Update and fix typos on dev instructions

### LINKED ISSUE #1223: Mention R install at top of development-environment.html
My sense is that R should be considered a peer of Python, and should also be installed by developers? There is an "R Setup" section, but it doesn't mention installing R itself, or whether there is a minimum version requirement.

mccalluc:
- Fix #1223 (Actually, move every language install down to the appropriate section.)
- Correct envvar
- Add missing double backticks
- Second level headers


Shoeboxam (approved):
Reads much cleaner!

PR #1233, opened 2024-02-09
Python and R examples in tabs on quickstart

### LINKED ISSUE #1232: Make quickstart examples for R
There's a bigger documentation project in draft:
- #1177

but we wanted to wait on that until polars is in.

Before then, it would be good to demonstrate how sphinx tabs will work, and apply it to the quickstart page.

- first: get the tabs up, with R examples
- second, figure out something like doctests for R, so we're sure the code works (might be split to a separate issue)

mccalluc:
- Fix #1232

I think this much is probably useful on its own, but it wouldn't be unreasonable to ask that we figure out something like doctests for R examples before this goes in. Your call.

Screenshot 2024-02-08 at 7 31 45 PM

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -255,9 +255,12 @@ jobs:
           working-directory: R/opendp
           extra-packages: devtools,local::.,RcppTOML
 
-      - name: Test
+      - name: Unit tests
         working-directory: R/opendp/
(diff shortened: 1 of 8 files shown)


mccalluc:
- #1262 handled most of this
- #1287 is filed for the remaining cleanup

Think it will be most clear to close this, and do any remaining work in a new PR.

PR #1235, opened 2024-02-09
Link directly to R function reference

### LINKED ISSUE #1224: R docs landing page
Currently, the link from https://docs.opendp.org/en/stable/api/index.html to [R API](https://docs.opendp.org/en/stable/api/r) brings you to a rendering of the top-level readme. To get to any R specifics you need to go up to the "Reference" menu.

- We should land on a page that just addresses the needs of R users.
- Link to any subsidiary pages from the main body of the page, not just the header.

See also:
- #835
- #1041

mccalluc:
- Fix #1224

I should understand the R docs build process better, but for now, this will save users one click.

### DIFF
--- a/docs/source/api/index.rst
+++ b/docs/source/api/index.rst
@@ -9,5 +9,5 @@ If you are looking for long-form, higher-level documentation, then please consul
    :maxdepth: 1
 
    Python API <python/index>
-   R API <https://docs.opendp.org/en/stable/api/r>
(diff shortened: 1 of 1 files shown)


Shoeboxam (changes requested):
When the README is updated, then we'd rather point users to the top-level, right?

mccalluc:
yep. This is obsolete now. Closing.

PR #1240, opened 2024-02-11
Avoid panic in ALP histogram

Shoeboxam:
I think the map on this is backwards, could you check @ChristianLebeda? Thanks. This is a sign we need better proof coverage!

I also adjusted the code to not panic when scale is zero.


ChristianLebeda:
I think the map is correct. I will take a closer look at the code this week to make sure.

However, the variable should properly be renamed to avoid confusion with the "scale" parameter of e.g. the Laplace mechanism. 
Here the scale parameter does not describe the magnitude of noise. It is related to the length of a bitstring. The longer the bitstring is, the more accurate (and less private) the output is.

Shoeboxam:
I took a closer look on my flight, and agree with you- it's right but a better name is needed.

I'll reduce this PR to just the panic fix.

ChristianLebeda (approved):
Thanks for taking a closer look!
The panic fix looks good.

Issue #1244, opened 2024-02-13
R docs formatting


mccalluc:
It's not going to be the same as the sphinx docs, but some changes could make it easier to use:
- Don't copy project README down to this level -- If we don't have anything to say to R users, it would be better to keep it short rather than give them something redundant. Fixed in #1247
- Include the opendp logo in the navbar, and consider which links are most useful. Fixed in #1246
- [It looks](https://pkgdown.r-lib.org/reference/build_reference.html#topic-matching) like there are options for grouping functions based on prefixes. This could help us to navigate what is a long list. Fixed in #1245

mccalluc:
Three child PRs merged. Closing.

PR #1245, opened 2024-02-13
Subheadings in R docs

mccalluc:
- Towards #1244... but keep it open for now.

Tweaked the Rust codegen to include `@concept` in the generated docstrings. With this, R pkgdown can be configured to include headings.

If this much is useful, there are a couple of follow-ups which could be filed.
- Generate `_pkgdown.yml` in Rust, so the list stays consistent.
- Split up the `Other` catch all by adding `@concept` on the non-generated files.


Shoeboxam (approved):
This is great, I was disappointed that everything was dumped into the same bin, this is so much more navigable.

PR #1247, opened 2024-02-13
R doc README

mccalluc:
Towards #1244

Just repeating the top-level README doesn't tell R users much, but I think there are a few things we could point out to them before letting them loose on the generated docs.

We are a little constrained: It should work both as a readme for this directory, and as a source for generated docs. Considered relative link to the API reference... but that wouldn't work when the README is viewed in github.

### DIFF
--- a/R/opendp/README.md
+++ b/R/opendp/README.md
@@ -6,4 +6,5 @@ These hints will help you apply the Python examples to R.
 
 - After invoking `library(opendp)` all functions are available: There are no namespaces like those of the Python library.
 - The chaining operator in R is `|>`, as opposed to `>>` in Python.
+- Type arguments like `.T` are prefixed with a period in R to avoid name conflicts, while in Python it is plain `T`.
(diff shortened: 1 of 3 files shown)


R/opendp/README.md
```diff
@@ -0,0 +1,9 @@
+# OpenDP R API
+
+The OpenDP R API largely mirrors the Python API, with some key differences.
+Most Python examples have not yet been translated to R:
+These hints will help you apply the Python examples to R.
+
+- After invoking `library(opendp)` all functions are available: There are no namespaces like those of the Python library.
+- The chaining operator in R is `|>`, as opposed to `>>` in Python.
+- All portions of the Python API that are derived from Rust have parallels in the R API, but packages like `context` that exist only in Python have not yet been implemented in R.
```
Shoeboxam, line 9: Can you also mention that type arguments are prefixed with a `.`? Like `T` becomes `.T`.

Shoeboxam (approved):

Shoeboxam (approved):
Thanks!

PR #1248, opened 2024-02-13
Make docs dependency on r-docs explicit

### LINKED ISSUE #1242: Clarify R docs build process
I think there are places where the goals and implementation of the R docs build could be clarified. These are some things that could make it simpler and more consistent in my view, but I may not have the big picture.

- [ ] #1235 changes the link target, but it should also be relative: Unlike the Rust docs, the R docs are in the same directory structure as the python docs, so a relative link would suffice.

In `doc.yml`, the `docs` job does not explicitly list `r-docs` under `needs`, but it does rely on the `r-docs` artifact. My guess is that the R docs run is fast enough that the artifact is there, but a race condition is possible. 

- [ ] Just add `r-docs` to the dependency list for `docs`?
- [ ] Or split `docs` into multiple jobs, so the python and R portions could run in parallel?
- [ ] Or add `r_stage.sh -d` to the Makefile, so the whole docs directory is generated with one command.

mccalluc:
make the dependency on r-docs explicit, at the cost of slowing down the docs build

- Possible fix #1242, but there are other approaches, including doing nothing.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -47,6 +47,8 @@ graph TD
 
     subgraph docs.yml
         dcc[credential-check] --> docs
+        dcc[credential-check] --> r-docs
+        r-docs --> docs
     end
     docs.yml -.- rdocs[docs]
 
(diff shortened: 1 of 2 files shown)


Shoeboxam (approved):
This was an oversight. Thanks for catching it.

PR #1249, opened 2024-02-14
reverse ternary, so it is not confused by false-y empty string

### LINKED ISSUE #1234: publish-r still pushes to `r-universe-dev` on stable
I think something might be wrong with `BRANCH=r-universe${{ (inputs.channel == 'stable') && '' || '-dev' }}`, because the action still pushed to `r-universe-dev`, even when the channel was stable.

mccalluc:
- Fix #1234

If you already have something in the works, just close this.


Shoeboxam (approved):
Thanks!

Issue #1236, opened 2024-02-09
Investigate R doctests


mccalluc:
https://hughjonesd.github.io/doctest/

We currently have precisely one `@examples` tag (in `mod.R`) but I think the doctest package would make sure it runs, and let us add assertions, if we want.

---

On second thought, I'm not sure what this gives us beyond what we get from [pkgdown](https://pkgdown.r-lib.org/articles/pkgdown.html#reference):

> pkgdown executes all examples, inserting the rendered results in the generated HTML files.

I'm going to try adding a syntax error to the one example we have, and confirm that CI catches it.

mccalluc:
Update: `Rscript -e 'devtools::check()'` is run by CI, and it will fail on a syntax error in examples... but the error message is not helpful:

```
❯ checking for unstated dependencies in examples ... ERROR
  Warning: parse error in file 'lines':
  1: unexpected symbol
  32: ###
  33: cleanEx
      ^
```

Can we do better?

mccalluc:
Maybe one big issue for R dev tooling?

mccalluc:
Filing two big R issues:
- #2961 
- #2962

closing.

PR #1250, opened 2024-02-14
Test examples in R docs

mccalluc:
- Towards #1236

This confirms that the examples are at least parsed... but the [error message](https://github.com/opendp/opendp/actions/runs/7903643328/job/21572262951#step:8:145) wouldn't point the user in the right direction:
```
 Error:   Warning: parse error in file 'lines':
  1: unexpected symbol
  32: ###
  33: cleanEx
```

I haven't figured out a way to get more useful feedback. Leave this in draft for a little while, and then close it to get it out of the way if I still have no fix.


mccalluc:
Not an active area of work. Closing.

PR #1252, opened 2024-02-14
Generated changes to R conf

mccalluc:
I think that running `devtools::check` produced these changes; Certainly not manual. They seem reasonable, but if they aren't, I should understand what generated them so I can avoid committing generated code changes by mistake.


Shoeboxam (approved):
Document generates code- it updates the description based on collate, and apparently updates its own version number in the description file. 

Check the changelogs for Roxygen 7.3 and you'll see it now exports s3 methods that are missing an export tag:
https://www.tidyverse.org/blog/2024/01/roxygen2-7-3-0/

This would also explain why we started getting warnings from document for hashtab.

PR #1253, opened 2024-02-14
Error on warning from devtools::check

### LINKED ISSUE #1251: R warning: "needs @export or @exportS3method tag"
`Rscript -e 'devtools::document()'` is working, but early in the log there are a pair of warnings:

```
ℹ Updating opendp documentation
ℹ Loading opendp
Registered S3 method overwritten by 'opendp':
  method        from
  print.hashtab utils
✖ mod.R:390: S3 method `to_str.default` needs @export or @exportS3method tag.
✖ mod.R:391: S3 method `to_str.hashtab` needs @export or @exportS3method tag.
```

[Same warning in CI](https://github.com/opendp/opendp/actions/runs/7903643328/job/21572262951?pr=1250#step:7:12)

The build still works, so not critical, but it would be good not to have warnings in the log.

mccalluc:
- Elevate check warnings to errors, just to be safe.
- Does not address #1251, sadly: `devtools::document` does not support `error_on` and I haven't found an alternative.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -263,4 +263,4 @@ jobs:
         run: cd R/opendp/ && Rscript -e 'devtools::document()'
 
       - name: Check
-        run: cd R/opendp/ && Rscript -e 'devtools::check()'
+        run: cd R/opendp/ && Rscript -e 'devtools::check(error_on="warning")'


Shoeboxam (commented):
I don't understand why the CI passes, even though it errors on warning, and has a warning?

Shoeboxam:
Answered my question: in the CI the documentation step is run separately, and warnings from documenting aren't raised as errors.

Shoeboxam (approved):

PR #1255, opened 2024-02-14
471 binary search continued

### LINKED ISSUE #471: Binary search utilities in Rust

mccalluc:
- New PR replaces #472 (Normally, would make that branch the target of this PR, but after merging main there are hundreds of changes, so this seems simpler.)
- Fixes #471

In the python, the bounds are optional, and if missing, `exponential_bounds_search` is called. In rust `exponential_bounds_search` is currently unused. Preference between making `bounds` an `Option`, or creating pairs of bounded/unbounded functions?

Also: what's the motivation for the `signed_*` functions? Is this additional information more useful in the Rust context than it is in Python and R?


mccalluc:
Regarding this question:
> Preference between making bounds an `Option`, or creating pairs of bounded/unbounded functions?

`Option` is the way to go.
- see #1308

mccalluc:
back to draft: add optional bounds args.

mccalluc:
@Shoeboxam :

Fails locally with:
```
error: reached the recursion limit while instantiating `search::signed_fallible_binary_search::<i32, {closure@src/utils/search.rs:71:21: 71:33}>`
```

There's mutual recursion between these different functions: Not sure if the logic should change, or if there's a way to tell Rust that the recursion is ok?

Also, not sure what we should have instead of `expect` here:
```
None => fallible_exponential_bounds_search(&predicate)?.expect("TODO"),
```

mccalluc:
Mike can look at this and suggest a way forward.

mccalluc:
Not an active area of work. Closing PR, but will keep the branch around, if it can be useful.

PR #1177, opened 2024-01-29
Documentation reorg

mccalluc:
I have been merging smaller feature branches into this, and just merged `main` into it to bring it up to date.

~~Waiting to move out of draft until after the 0.9 release.~~ Talked with Raman: Given the uncertain timeline on polars, and the trouble we'll have trying to merge doc changes into this, going to document the system as it currently exists, and try to merge this sooner than later.

--- 

Update:
- Last child branches reviewed and merged
- up to date
- Built docs locally and skimmed every page
- Fixed some typos, and filed a half dozen new, hopefully small issues that shouldn't block this.

This is a big update, and not without risk: Most of the urls on our docs site are changing, and links from outside are likely to break. -- but, I think this is a lot cleaner, and however people get here, they are more likely to find what they need. Most of the changes here were their own PRs that already were reviewed.

PR #1262, opened 2024-02-16
Python and R examples in tabs on quickstart (just CI)

### LINKED ISSUE #1232: Make quickstart examples for R

mccalluc:
- Fix #1232
- Branching from last passing commit on #1233

Not immediately obvious to me what I did wrong pulling the CI logic into a script in #1233, and just this much is valuable, so filing a new issue here.

If this much is good, there's still the question of what the next step should be: I'd be reluctant to start adding more tabs, because that's going to cause a lot of conflicts with 
- #1177

but that is in a holding pattern because the original plan was to wait until the polars API is ready; Maybe we should revisit that plan?

Screenshot 2024-02-08 at 7 31 45 PM


raprasad (approved):
Approved. Reviewed with @mccalluc

PR #1263, opened 2024-02-16
Remove `@versioned` from generated code

### LINKED ISSUE #1145: Remove vestigial `@versioned` decorator
- From #1144 
> `@proven` is vestigial, but hasn't been removed yet. We've made updates that makes this un-necessary, but a careful, close look is necessary to extricate this.

mccalluc:
- Remove one line from Rust code and regenerate Python
- Since it's now unused, remove function
- Fix #1145

(If you're looking at that issue, note that it started from a misunderstanding on my part: it's `@versioned` rather than `@proven` which is vestigial.)

### DIFF
--- a/python/src/opendp/_data.py
+++ b/python/src/opendp/_data.py
@@ -23,7 +23,6 @@ __all__ = [
 ]
 
 
-@versioned
 def bool_free(
     this
 ):
@@ -49,7 +48,6 @@ def bool_free(
     return output
 
 
-@versioned
 def extrinsic_object_free(
     this
 ):
@@ -75,7 +73,6 @@ def extrinsic_object_free(
     return output
 
 
-@versioned
 def ffislice_of_anyobjectptrs(
     raw: Any
 ) -> Any:
@@ -102,7 +99,6 @@ def ffislice_of_anyobjectptrs(
     return output
 
 
-@versioned
 def fill_bytes(
     ptr,
     len: int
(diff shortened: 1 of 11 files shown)


mccalluc:
mypy failures:
```
src/opendp/context.py:684: error: Missing positional arguments "input_metric", "scale" in call to "make_base_laplace"  [call-arg]
src/opendp/context.py:684: error: Argument 1 to "make_base_laplace" has incompatible type "float"; expected "Domain"  [arg-type]
test/test_comb.py:58: error: Unexpected keyword argument "D" for "make_base_discrete_laplace"  [call-arg]
src/opendp/measurements.py:230: note: "make_base_discrete_laplace" defined here
Found 3 errors in 2 files (checked 69 source files)
```
I can reproduce locally on this branch, and also confirm that the errors are not there on `main`

Looking at the first error: 
> Missing positional arguments "input_metric", "scale" in call to "make_base_laplace"

```
    if from_to == ("ZeroConcentratedDivergence", "MaxDivergence"): # pragma: no cover
        space = atom_domain(T=T), absolute_distance(T=T)
        scale = binary_search_param(
            lambda eps: make_pureDP_to_zCDP(make_base_laplace(*space, eps)),
            d_in=constant,
            d_out=d_from,
            T=float,
        )
        return make_base_laplace(scale).map(constant) # <------- here!
```
- It does look like it's missing arguments.
- `pragma: no cover`: Probably no test coverage, so nothing checks that these lines even run.
- Guessing that the decorator masked `make_base_laplace`, so mypy couldn't see the signature.
- Good to surface this problem! I'll file a separate issue -- this is blocked till resolved -- but don't want to put it all in one huge PR.

PR #1265, opened 2024-02-16
Fix masked mypy errors

### LINKED ISSUE #1264: `@versioned` is masking mypy errors in *_laplace calls
See https://github.com/opendp/opendp/issues/1263#issuecomment-1948973263

When the `@versioned` decorator is removed, we get a few new mypy errors on calls to `make_base_discrete_laplace` and `make_base_laplace`

#1263 is blocked until this is fixed. Either make a branch from that branch, or temporarily remove the generated annotation just to see the mypy errors locally.

- [ ] src/opendp/context.py:684: error: Missing positional arguments "input_metric", "scale" in call to "make_base_laplace"  [call-arg]
- [ ] src/opendp/context.py:684: error: Argument 1 to "make_base_laplace" has incompatible type "float"; expected "Domain"  [arg-type]
- [ ] test/test_comb.py:58: error: Unexpected keyword argument "D" for "make_base_discrete_laplace"  [call-arg]
- [ ] src/opendp/measurements.py:230: note: "make_base_discrete_laplace" defined here

Improving the test coverage here would also be good.

mccalluc:
I'm not sure of the metric in `test_comb.py`: @Shoeboxam, can you make a suggestion there?

My sense is that `context` might go away, to be replaced by Polars, so we don't need to invest more time there? Though if it is staying, the test coverage could be improved.

- Fix #1264
- Unblock #1263

### DIFF
--- a/python/test/test_comb.py
+++ b/python/test/test_comb.py
@@ -55,7 +55,9 @@ def test_make_basic_composition_leak():
 
     # choose a vector-valued mechanism that should run quickly for large inputs
     # we want to add as little noise as possible, so that execution time is small
-    meas = dp.m.make_base_discrete_laplace(scale=1e-6, D=dp.VectorDomain[dp.AtomDomain[int]])
+    domain = dp.VectorDomain[dp.AtomDomain[int]]
+    metric = ... # TODO!
+    meas = dp.m.make_base_discrete_laplace(domain, metric, 1e-6)
 
     # memory usage remains the same when this line is commented,
     # supporting that AnyObject's free recursively frees children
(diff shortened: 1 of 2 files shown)


Shoeboxam:
Context stays, but the `Query` class will get less use in common workloads than before. Test coverage can definitely be improved.

python/test/test_comb.py
```diff
@@ -55,7 +55,9 @@ def test_make_basic_composition_leak():
 
     # choose a vector-valued mechanism that should run quickly for large inputs
     # we want to add as little noise as possible, so that execution time is small
-    meas = dp.m.make_base_discrete_laplace(scale=1e-6, D=dp.VectorDomain[dp.AtomDomain[int]])
+    domain = dp.VectorDomain[dp.AtomDomain[int]]
+    metric = ... # TODO!
```
mccalluc, line 59: If this is otherwise ok, file an issue to fill this in.

raprasad (approved):
Approved.  Reviewed with @mccalluc

PR #1267, opened 2024-02-16
Now if "stable" and "dry-run" are selected, will append "-dev"

### LINKED ISSUE #1266: publish-r doesn't respect dry-run
Did a dry-run to confirm that #1130 could be closed, and it worked. Great!

... but when I next updated locally, I saw:
```
 + a812a67a...7d15c621 r-universe-dev -> origin/r-universe-dev  (forced update)
```

and now looking at the `publish-r` job, I can confirm that there is no reference to `inputs.dry_run`, unlike `publish-{python,rust,github}`, where it is checked.

mccalluc:
If we want to treat `r-universe-dev` like `testpypi` then this seems appropriate... but if we want to make sure that nothing gets pushed to any branch on github from a dry-run, should try something different.

- fix #1266

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -109,7 +109,7 @@ jobs:
           git config --global user.name "$GITHUB_ACTOR"
 
           # create brand new history each time
-          BRANCH=r-universe${{ (inputs.channel != 'stable') && '-dev' || '' }}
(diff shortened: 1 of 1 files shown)


Shoeboxam (approved):
Great, we don't want to be hitting the stable universe on dry-runs.

PR #1268, opened 2024-02-16
manylinux2014 -> manylinux_2_24

### LINKED ISSUE #1172: Manylinux 2014 EOL
If I'm reading [this](https://github.com/pypa/manylinux) correctly, it hits EOL this June. Reference a newer version here, or file an issue to follow up?

_Originally posted by @mccalluc in https://github.com/opendp/opendp/pull/1141#discussion_r1462578920_

mccalluc:
- Fix #1172

There is also a 2_28 out, but maybe be safe? I've confirmed on quay and dockerhub that these images exist. 

[Dry-run of build running now.](https://github.com/opendp/opendp/actions/runs/7935903813)

Also: It doesn't look like `build_tool.py` is referenced anywhere. I've done some linting on it recently, but before that the last commit was
```
commit ae35ac4e7457da687611117c98f279da37a9c3a2
Author: Andrew Vyrros <6577271+andrewvyrros@users.noreply.github.com>
Date:   Tue Oct 25 14:06:42 2022 -0700

    * Try manylinux2014_x86_64 (#592)

    * Switch args to toml.dump()
```

Avoiding vendor lock-in is a good goal, but I don't think we've maintained this, and it may just cause confusion now. Ok to delete?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -154,7 +154,7 @@ jobs:
       matrix:
         architecture: [x86_64, aarch64]
     env:
-      DOCKER_IMAGE: messense/manylinux2014-cross:${{ matrix.architecture }}
+      DOCKER_IMAGE: messense/manylinux_2_24-cross:${{ matrix.architecture }}
       TARGET: ${{ matrix.architecture }}-unknown-linux-gnu
     steps:
       - name: Checkout repository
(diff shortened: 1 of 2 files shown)


mccalluc:
[Dry run failed](https://github.com/opendp/opendp/actions/runs/7935903813/job/21670723313#step:5:156) -- I think because I had already used version `1` in an earlier dry-run today. Instead of asking the user to track numbers, would it make sense to use a time stamp or hash instead to avoid collisions?

mccalluc:
And the [dry-run passes](https://github.com/opendp/opendp/actions/runs/7936158705): Moving out of draft.

raprasad (approved):
Approved

Issue #1274, opened 2024-02-21
Interleave Python and R


mccalluc:
In this PR:
- #1262

tabbed Python and R examples are introduced. I looked for a bit, but couldn't find a way to have the runnable R examples inline. So instead the R examples are included (with `literalinclude`), and the Python doctests are still in the `.rst`.

This works well enough, but it is hard to confirm that the python and R examples are running in parallel: It would be easy to tweak one, and forget the other.

Proposal: Interleave the Python and R in one `.R` file: Protect the python doctests by putting them in multi-line R strings. We'll need to make sure the `.R` files are scanned for python doctests, but otherwise should work smoothly. Something like:

```R
'
# p-1
>>> from opendp.mod import enable_features
>>> enable_features('contrib')

# p-1-end
'

# r-1
library(opendp)
enable_features("contrib")
# r-1-end

'
# p-2
>>> import opendp.prelude as dp
>>> base_laplace = dp.space_of(float) >> dp.m.then_base_laplace(scale=1.)
>>> dp_agg = base_laplace(23.4)

# p-2-end
'

# r-2
space <- c(atom_domain(.T = "f64"), absolute_distance(.T = "f64"))
base_laplace <- space |> then_base_laplace(1.)
dp_agg <- base_laplace(arg = 23.4)
# r-2-end

```

I'm not sure about this -- both approaches are somewhat awkward. Feedback welcome.

mccalluc:
Decided interleaving won't be sustainable with more languages... and even with examples of the the two python APIs in parallel: Instead, pull out python examples.

PR #1270, opened 2024-02-19
updated First-Look-at-DP notebook

Shoeboxam:
Closes #1269

PR #1278, opened 2024-02-21
Translate fladp notebook to RST

mccalluc:
The new notebook in 
- #1270 

is great, but we want to move to RST and doctests, so we can have R examples in parallel.

I've also rearranged the content, moving the explanation of the context vs the framework to the top, and removing the introductory section, since that is covered on the previous page.

Doing Python in RST for now, so I only need to worry about one file, but if anyone really likes the proposal in 
- #1274

that could be done here.

R examples could/should also be provided, but don't think that needs to block this.


raprasad (approved):
approved

PR #1279, opened 2024-02-22
mechanism: randomized response bitvec

abigail-gentle:
Added a rust implementation for [basic rappor](https://research.google/pubs/rappor-randomized-aggregatable-privacy-preserving-ordinal-response/) for locally private frequency estimation. This includes the randomiser, a de-bias function, and a proof (work in progress).

This function requires that each input is one-hot encoded, or hashed onto a bloom filter. It then noises each vector by xor'ing each bit with Bern(f/2). If you are using a bloom filter you must specify $m$ as the number of hash functions used.

Any feedback, changes, and proof checking is greatly appreciated!

Abigail


rust/src/measurements/rappor/mod.rs
```diff
@@ -0,0 +1,111 @@
+use crate::error::Fallible;
+use crate::metrics::DiscreteDistance;
+use crate::measures::MaxDivergence;
+use crate::domains::{AtomDomain, VectorDomain};
+use crate::core::{Function, Measurement, PrivacyMap};
+use crate::traits::{InfSub, InfDiv, InfLn, InfMul, samplers::SampleBernoulli};
+
+
+/// Make a Measurement that implements Basic RAPPOR
```
mccalluc, line 9: Should `#[bootstrap(...)]` be added here, unless there's a reason not to generate Python and R bindings? Or maybe it would be good, but it could be a follow-up issue?

rust/src/measurements/rappor/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+    f: f64,
+    m: i32,
+    constant_time: bool
+) -> Fallible<Measurement<VectorDomain<AtomDomain<bool>>, Vec<bool>, DiscreteDistance, MaxDivergence<f64>>> {
+    if input_domain.size.is_none() {
+        return fallible!(MakeMeasurement, "RAPPOR requires a known number of categories.")
+    }
+
+    if !(0.0..=1.0).contains(&f) {
+        return fallible!(MakeMeasurement, "f must be in (0, 1].")
```
mccalluc, line 30:
```suggestion
        return fallible!(MakeMeasurement, "f must be in (0, 1]")
```

(I won't flag the rest of these below.)

rust/src/measurements/rappor/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+            if a {
+                *c += 1.0;
+            }
+        });
+    });
+
+    Ok(counts.into_iter().map(|y_i| {
+        (y_i - ((f / 2.0) * n)) / (1.0 - f)
+    }).collect())
+
```
mccalluc, line 89:
```suggestion
```

rust/src/measurements/rappor/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+            VectorDomain::new(AtomDomain::default()).with_size(10),
+            DiscreteDistance::default(),
+            0.5,
+            1,
+            false
+        )?;
+        rappor.invoke(&vec![true, false, true, false, true, false, true, false, true, false])?;
+        assert_eq!(rappor.map(&1)?, 2.1972245773362196);
+        Ok(())
+    }
```
mccalluc, line 109: Add a test for `debias_basic_rappor`?

rust/src/measurements/rappor/make_rappor.tex
```diff
@@ -0,0 +1,98 @@
...
+\section{Hoare Triple}
+\subsection{Preconditions}
+\begin{itemize}
+	\item Variable 
+\end{itemize}
+
+
+\section{Proof}
+$f/2$ refers to the probability of 
+\begin{theorem}~\cite{42852}
```
mccalluc, line 29: Maybe
```suggestion
\begin{theorem}~\cite{rappor}
```
and a corresponding change below, unless the number has meaning for you.

rust/src/measurements/rappor/references.bib
```diff
@@ -0,0 +1 @@
+@inproceedings{42852,title	= {RAPPOR: Randomized Aggregatable Privacy-Preserving Ordinal Response},author	= {Úlfar Erlingsson and Vasyl Pihur and Aleksandra Korolova},year	= {2014},URL	= {https://arxiv.org/abs/1407.6981},booktitle	= {Proceedings of the 21st ACM Conference on Computer and Communications Security},address	= {Scottsdale, Arizona}}
```
mccalluc, line 1: format with one key per line?

mccalluc (commented):
Some trivial formatting suggestions. @Shoeboxam will be able to review soon and he'll get into the details.

If it does make sense to generate a Python and R interface for this, would it also be good to explain how to use this in the documentation? (This would be a follow-up issue, and shouldn't block this PR.) Right now, local DP is only mentioned in `randomized-response.ipynb` -- I could imagine renaming that file, and after the introduction to randomized response, add a second section which demos RAPPOR. Does that seem like a good direction?

Looking at the CI failures: `cargo fmt` should fix one set of problems, the the latex build isn't able to match the citation.

Shoeboxam:
If `m` is the number of bits that can change, how do you feel about changing the input metric to the Hamming distance, and then `d_in` corresponds to `m`?

abigail-gentle:
Greatly appreciate all the notes! To be honest I am unsure how local protocols relate to the distance metrics, but Hamming distance likely makes sense, two inputs should be at most `2m` apart. Or alternatively the input domain should be all bit-strings with hamming distance `<= m` from the zero vector. 

To clarify, `m` is the number of bits set in the non-private input, mostly just needed to ensure privacy (if you have 10/20 bits set, you have a lower `eps` than someone with 2/20 even with the same f). 

I'll finish the proof document and update the docs soon, I think a jupyter notebook would be a great idea to introduce concepts of LDP, and allow people to experiment.

Shoeboxam:
Ah, based on your comment, neither of my suggestions were quite correct. Any two adjacent data sets could have completely disjoint set bits- the ideal metric is still the discrete metric. However, the domain is restricted to those input vectors with only at most m (or h, as in the paper) bits set. If m can be less than the vector's size, then we might want to add another descriptor to `VectorDomain`, alongside `size`.

rust/src/measurements/rappor/mod.rs
abigail-gentle: Would love to have this, just wasn't sure how to go about it. Will check the docs and have a try at making it work soon

abigail-gentle (commented):
Added a function to test debias RAPPOR, and expanded proof documentation (there is a utility section, which I was unsure how to organise) remaining priorities are to figure out the right fit for distance metrics, and to make the documentation clearer

abigail-gentle:
I've been reading the documentation for domains, and [issue-1600121842](https://github.com/opendp/opendp/pull/645#issue-1600121842) and am wondering if the best bet is to create a domain for [balls on metric spaces](https://en.wikipedia.org/wiki/Ball_%28mathematics%29). This would take a domain, a distance metric `d`, a centre-point `p`, a distance `r`(and whether to include the boundary, whether the ball is open or closed). Checking membership would just be `d(p,x) <[=] r`

This seems preferable to adding anything inside the vector domain struct, which is why I propose it, but I am open to other ideas.

In the case of RAPPOR the domain would have:
 - `D`: `VectorDomain<Bool>` with size `k` 
 - `d`: Hamming distance
 - `p`: The zero vector
 - `r`:  The maximum number of bits set in each input (`m` in the proof or the number of hash functions used to map inputs onto a bloom filter `h` in the original RAPPOR paper)
 - On a closed ball (including the boundaries)

rust/src/measurements/rappor/mod.rs
Shoeboxam: I was going to handle this part myself!

Shoeboxam:
I think we should use the bitvec crate and introduce a new domain for it. 

* Rust `Vec<bool>` does not optimize for space: https://stackoverflow.com/questions/48875251/rust-seems-to-allocate-the-same-space-in-memory-for-an-array-of-booleans-as-an-a
* We want a specialized descriptor for this case (and we also can't specialize, so it's un-ergonomic to implement the member check on VectorDomain)

The de-facto crate for bit vectors is this one: https://crates.io/crates/bitvec

abigail-gentle:
Agreed! Bitvecs are great and seem necessary in this case. Also simplifies doing anything with the other domains. That seems like the go-to crate, so that would be nice to set up. I've had a look at the documentation and I think writing a few transformations (for one-hot encoding , Vec<Bool> -> BitVec etc.) would help a lot (as the constructors provided are quite easy to misunderstand)

rust/src/measurements/rappor/make_rappor.tex
```diff
@@ -79,15 +119,46 @@ \section{Proof}
...
+	\intertext{Therefore the estimator $\hat{N_i}$, given by}
+\hat{N_i} &= \frac{Y_i-n\frac{f}{2}}{1-f}\\
+\intertext{is unbiased as,}
+\mathbb{E}[\hat{N_i}] &= N_i
+\end{align*}
+\end{proof}
+\begin{theorem}
+	\rustdoc{measurements/fn}{debias\_basic\_rappor} has average squared error
+	\begin{equation}
+		l_2^2(N-\hat{N}) = kn\left(\frac{f}{2}-\frac{f^2}{4}\right)
```
abigail-gentle, line 148: Mising a 1/(1-f) factor here, will fix soon

abigail-gentle:
Ok I have created a `BitVectorDomain`, which has a `max_weight` (i.e. maximum hamming weight), and also created a pub type `BitVector` which aliases `BitVec<usize, Lsb0>`, which is apparently the most performant variety in the module. I placed the alias in domains, but that doesn't seem like the right place for it.

I also updated the RAPPOR code to use these new implementations, but I wonder if I should implement Hamming distance for BitVectors and use that rather than DiscreteDistance?

rust/src/measurements/rappor/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+        input_domain,
+        Function::new_fallible(move |arg: &BitVector| {
+            let n = arg.len();
+            let noise_vector = (1..n)
+                .into_iter()
+                .map(|_| sample_bernoulli_float(f_2, constant_time))
+                .collect::<Fallible<BitVector>>()?;
+            // I wanted to avoid cloning here but the closure makes it necessary
+            // Shouldn't use much memory anyway given bit-vecs
+            Ok(arg.clone() ^ noise_vector) // xor on bit vectors
```
Shoeboxam, line 58: At some point I'd like to adjust the signature of functions to take data by ownership, instead of by reference. I think this will result in less cloning overall.

rust/src/domains/mod.rs
```diff
@@ -445,6 +448,57 @@ impl<D: Domain> Domain for VectorDomain<D> {
...
+/// If a size is specified then it
+#[derive(Clone, PartialEq)]
+pub struct BitVectorDomain {
+    pub max_weight: Option<u32>,
+}
+impl Debug for BitVectorDomain {
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), std::fmt::Error> {
+        let weight_str = self
+            .max_weight
+            .map(|max_weight| format!("weight={:?}", max_weight))
```
Shoeboxam, line 465:
```suggestion
            .map(|max_weight| format!("max_weight={:?}", max_weight))
```

rust/src/measurements/rappor/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+use crate::traits::{samplers::sample_bernoulli_float, InfDiv, InfLn, InfMul, InfSub};
+
+use bitvec::prelude::{bitvec, Lsb0};
+
+/// Make a Measurement that implements Basic RAPPOR
+///
+/// # Citations
+/// * [RAPPOR: Randomized Aggregatable Privacy-Preserving Ordinal Response](https://arxiv.org/abs/1407.6981)
+///
+/// # Arguments
```
Shoeboxam, line 15: Argument docs are a bit dated.

rust/src/measurements/rappor/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+            }
+            if d_in > 1 {
+                return fallible!(FailedFunction, "d_in must be 0 or 1.");
+            }
+            Ok(epsilon)
+        }),
+    )
+}
+
+pub fn debias_basic_rappor(answers: Vec<BitVector>, f: f64) -> Fallible<Vec<f64>> {
```
Shoeboxam, line 74: Please document!

rust/src/metrics/mod.rs
```diff
@@ -295,6 +295,11 @@ impl<D: Domain> MetricSpace for (VectorDomain<D>, HammingDistance) {
         })
     }
 }
+impl MetricSpace for (BitVectorDomain, HammingDistance) {
+    fn check_space(&self) -> Fallible<()> {
+        Ok(())
+    }
+}
```
Shoeboxam, line 302: While this is true, if you end up not using it, then it can be removed. Similarly for (VectorDomain, DiscreteDistance) below.

rust/src/transformations/bit_vec_encode/mod.rs
```diff
@@ -0,0 +1,3 @@
+// TODO: 
+//  - one-hot encoding
+//  - other BitVector encodings
```
Shoeboxam, line 3: These could be added in a separate PR!

Shoeboxam:
I would only implement the case where the input metric could be the Hamming distance if you anticipate settings where adjacent bit vectors may differ in only at most d_in bits: IE, changing any one individual's answer can change only at most d_in bits, instead of the entire bitvec. 

For instance, [1, 1, 1, 0] would not be considered an adjacent bit vector to [1, 0, 0, 0] when d_in=1 under Hamming distance, as an individual could only influence up to one bit. But in this local-dp setting, I'd anticipate the entire bit vector comes from one individual.

You're more familiar with these algorithms than I, but my guess would be that this is unlikely?

Shoeboxam (approved):
Awesome updates!

Shoeboxam (dismissed):
Whoops, I always hit approve too soon.

Shoeboxam:
I took a first pass at the FFI. For the most part it was all pretty straightforward, until I got to how bit vectors are represented in Python and R. I might iterate on the internal rust type for the bit buffer while I look for the ideal Python and R bit vector data types. 

For now, ignore the failing CI check for R linting.

abigail-gentle:
Touched up the docs, hope they're clear enough :) 

I did end up deleting the transformations folder, but thinking on it, implementing some basic machinery for converting a bounded AtomDomain or a hash set to a one-hot encoding seems worth doing so that a basic Jupyter notebook can be set up. It still could be another issue though, I'm just pretty happy it's all working.

raprasad:
Reviewing with @Shoeboxam 
Items left:
 - [ ] Proof for @vikrantsinghal 
 - [ ] R FFI for bit vector (@Shoeboxam)

vikrantsinghal (changes requested):
Here is my initial feedback on the proof doc. I might have more comments once the following edits have been considered and reviewed.
1. I would describe in the doc what `m` is.
2. I would also explicitly state the assumption that `2m+1<=k`, and why we are making that assumption.
3. On page 3, can't the fourth equality follow directly from the first inequality? Why is the third equality a necessary step? Is it adding any new info? I would use `\tag` in the `\align*` environment to describe that. If the third equality isn't doing anything, then might as well get rid of it.
4. What is each user sending, and what is the final function being computed? Is each user a sequence of `k` bits? Is a vector the final quantity that is being computed? I would clearly describe the nature of the input and the output in the doc.
5. Is `N` the total number of users? Or is it the vector of total counts for each bit position or coordinate?
6. The proof of Theorem 2.5 seems not so correct to me in terms of the notation. The first equality in the `\align*` environment equates the L1 norm to the L2 norm squared.
7. Also, in statistics, people generally tend to use the term "Mean Squared Error (MSE)" as opposed to "Average Squared Error". Might want to switch to that?

Cheers!

rust/src/measurements/rappor/make_rappor.tex
```diff
@@ -0,0 +1,167 @@
...
+\author{Abigail Gentle}
+\begin{document}
+
+\maketitle
+
+
+
+\contrib
+
+Proves soundness of \rustdoc{measurements/fn}{make\_rappor} in \asOfCommit{mod.rs}{304ef9c2}.
```
mccalluc, line 15: This is really a process question for @Shoeboxam : After the squash merge, is it safe to continue to reference a commit from a feature branch? I think github will keep it around, but if we auto-delete the branch now, I don't know that future checkouts would have this history. 

If we do need this, it might be a reason to turn auto-delete off again.

rust/src/measurements/rappor/make_rappor.tex
```diff
@@ -0,0 +1,167 @@
...
+
+TODO: algorithm outline
```
mccalluc, line 17: done?

rust/src/measurements/rappor/make_rappor.tex
```diff
@@ -0,0 +1,167 @@
...
+        raise Exception("RAPPOR requires a maximum number of set bits")
+    eps = (2*m)*log((2-f)/f)
+    def privacy_map(d_in: IntDistance): |\label{line:map}|
+        return eps
+    def function(arg: BitVector) -> BitVector: |\label{line:fn}|
+    	k = len(arg)
+    	noise_vector = [bool.sample_bernoulli(f/2, constant_time) for _ in range(k)]
+    	return xor(arg, noise_vector)
+    
+    return Measurement(input_domain, function, input_metric, output_measure, privacy_map)
```
mccalluc, line 47: We prefer to have pseudocode as a separate `.py` file, and include it into the `.tex`. We do some linting of the pseudocode files to catch typos.

rust/src/measurements/rappor/make_rappor.tex
```diff
@@ -0,0 +1,167 @@
...
+			}\nonumber\\
+			%&=\max_{y\in S}\frac{%
+			%	\prod\limits_{i=1}^m \left(\frac{1}{2}f\right)^{1-y_i}\left(1-\frac{1}{2}f\right)^{y_i}%
+			%	\times \prod\limits_{i=m+1}^{2m}\left(\frac{1}{2}f\right)^{y_i} \left(1-\frac{1}{2}f\right)^{1-y_i}%
+			%	\times \prod\limits_{i=2m+1}^{k}\left(\frac{1}{2}f\right)^{y_i} \left(1-\frac{1}{2}f\right)^{1-y_i}
+			%}{%
+				%\prod\limits_{i=1}^m \left(\frac{1}{2}f\right)^{y_i}\left(1-\frac{1}{2}f\right)^{1-y_i}%
+				%\times \prod\limits_{i=m+1}^{2m}\left(\frac{1}{2}f\right)^{1-y_i} \left(1-\frac{1}{2}f\right)^{y_i}%
+				%\times\prod\limits_{i=2m+1}^{k}\left(\frac{1}{2}f\right)^{y_i}\left(1-\frac{1}{2}f\right)^{1-y_i}
+			%}\nonumber\\
```
mccalluc, line 109: Remove commented out blocks... or explain, if there's a good reason to keep it.

```suggestion
```

rust/src/measurements/rappor/mod.rs
```diff
@@ -0,0 +1,154 @@
...
+                "RAPPOR requires a maximum number of set bits!"
+            )
+        }
+    };
+
+    if f <= 0.0 || f > 1.0 {
+        return fallible!(MakeMeasurement, "f must be in (0, 1]");
+    };
+
+    // priv = 2mln((2-f)/f)
```
mccalluc, line 44:
```suggestion
    // epsilon = 2 * m * ln((2-f)/f)
```

mccalluc (changes requested):
I'm afraid I'm not the best to understand the details, but I hope this is useful. Thanks for contributing this!

abigail-gentle:
Greatly appreciate the feedback!

> 1. I would describe in the doc what `m` is.

I added a clarification through two examples, one for categorical data and one for strings hashed into a Bloom filter.

> 2. I would also explicitly state the assumption that `2m+1<=k`, and why we are making that assumption.

Still need to make this clear, and possibly even enforce it as a requirement in the Measurement. The assumption comes from the fact that a Bloom filter would be massively inefficient with that many hash functions.

> 3. On page 3, can't the fourth equality follow directly from the first inequality? Why is the third equality a necessary step? Is it adding any new info? I would use `\tag` in the `\align*` environment to describe that. If the third equality isn't doing anything, then might as well get rid of it.

I added a ref which points out that the cancellation in that line (if I'm looking at the same one) shows that privacy does not depend on `k`.

> 4. What is each user sending, and what is the final function being computed? Is each user a sequence of `k` bits? Is a vector the final quantity that is being computed? I would clearly describe the nature of the input and the output in the doc.
> 5. Is `N` the total number of users? Or is it the vector of total counts for each bit position or coordinate?

Updated the descriptions of both, would love further feedback on readability though. Each user (currently represented by a row in a database though) is a sequence of `k` bits. In order to perform frequency estimation on categorical data this should be a one-hot encoding. To do frequency estimation on strings (or other large domains) a bloom filter or other sketch should be used to decrease the domain size.

> 6. The proof of Theorem 2.5 seems not so correct to me in terms of the notation. The first equality in the `\align*` environment equates the L1 norm to the L2 norm squared.

Correct! Fixed that thank you

> 7. Also, in statistics, people generally tend to use the term "Mean Squared Error (MSE)" as opposed to "Average Squared Error". Might want to switch to that?

Good point!

Thanks so much, I have pushed some updates which I hope clarify the usage. Could probably still do with further expansion.

abigail-gentle:
There were some major errors in the MSE proof which I have now corrected 🤞, I hope I have also made the wording and notation clearer, but as I was changing some things I may have missed some symbols, I will give it a read with fresh eyes later.

Shoeboxam:
The git history was messed up, I think due to the merge from main-- couldn't rebase. I pushed a copy of the old history to the implement-rappor-backup2 branch and rewrote this history.

vikrantsinghal (commented):
The proof itself looks good now. I would just suggest making some changes in the presentation of the document to make it a bit cleaner. If you prefer, I could do that and make the commit, otherwise, I can suggest them here. Up to you!

Shoeboxam:
* **#1279** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1279?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @abigail-gentle and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

python/src/opendp/_convert.py
```diff
@@ -286,6 +292,23 @@ def _slice_to_string(raw: FfiSlicePtr) -> str:
     return ctypes.cast(raw.contents.ptr, ctypes.POINTER(ctypes.c_char_p)).contents.value.decode()  # type: ignore[reportOptionalMemberAccess,union-attr]
 
 
+def _bitvector_to_slice(val: Sequence[Any]) -> FfiSlicePtr:
```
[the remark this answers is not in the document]
Shoeboxam: I dropped this type annotation for the time being.

python/src/opendp/_convert.py
```diff
@@ -286,6 +292,23 @@ def _slice_to_string(raw: FfiSlicePtr) -> str:
...
+    if not isinstance(val, (bytes, bytearray)):
+        raise TypeError("Expected type is BitVector but input data is not bytes or bytearray.")
+
+    array = (ctypes.c_uint8 * len(val)).from_buffer_copy(val) # type: ignore[operator]
+    return _wrap_in_slice(array, len(val))
+
+
+def _slice_to_bitvector(raw: FfiSlicePtr) -> bytes:
+    n_bytes = (raw.contents.len + 4) // 8
+    buffer = ctypes.cast(raw.contents.ptr, ctypes.POINTER(ctypes.c_uint8))[0:n_bytes] # type: ignore
```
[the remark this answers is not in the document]
Shoeboxam: You're right, technically it should be 7, not 4. However, since the input is always padded, any number from 0 and 7 would work. I've adjusted to use ceiling division (which is equivalent to 7).

python/test/test_meas.py
```diff
@@ -1,28 +1,30 @@
 import pytest
 import opendp.prelude as dp
 
-dp.enable_features('floating-point', 'contrib')
+dp.enable_features("floating-point", "contrib")
```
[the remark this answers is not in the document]
Shoeboxam: Reverted styling. I shouldn't have let that format make it into the diffs.

rust/src/measurements/rappor/make_rappor.tex
Shoeboxam: Ultimately these should point to commits in main. I was thinking updating these hashes would be part of a stabilization PR down the line that also drops contrib.

Shoeboxam:
I split the FFI into a separate PR and updated it to work with R as well (see above comment).

I also made a few adjustments:
* changed the type of max_weight to usize
* Vikrant made some changes to the proof
* I pulled the code out of the proof and updated it
* removed a MetricSpace impl
* temporary edit to the `setup.cfg` to resolve an unrelated CI issue having to do with the release of Numpy 2.0
* renamed the mechanism (pending)

How do you feel about renaming the mechanism to just `make_randomized_response_bitvec`? I'd imagine that a separate constructor, `make_rappor` would just chain the bloom filter and this generalized rr mechanism.

rust/src/measurements/randomized_response_bitvec/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+            0.5,
+            false,
+        )?;
+        m_rr.invoke(&bitvec![u8, Lsb0;
+            1, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        ])?;
+        assert_eq!(m_rr.map(&1)?, 2.1972245773362196);
+        Ok(())
+    }
+    #[test]
```
mccalluc, line 133:
```suggestion

    #[test]
```

rust/src/measurements/randomized_response_bitvec/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+    fn test_debias_rr_bitvec() -> Fallible<()> {
+        let f = 0.1;
+        let mut answer = vec![0.0; 10];
+        answer[0] = 1.0;
+
+        let answers = vec![bitvec![u8, Lsb0; 1, 0, 0, 0, 0, 0, 0, 0, 0, 0]; 10];
+
+        let high = 10.555555555555555;
+        let low = -0.5555555555555556;
```
mccalluc, line 142: More clear to express high as a 95/9?

mccalluc (approved):

abigail-gentle:
Hey Michael, greatly appreciate the updates, and huge thank you to Vikrant for the notes on the proof, I agree with all of them, and will make the changes recommended. 

I also agree with the name change, and think that leaving it in the form it is right now would be best. The current view on using Bloom filters seems to be that using them can incur more error than privacy [(see this follow up noting that it requires a lot of domain specific tuning)](https://arxiv.org/abs/2001.03618), and you still need to have a known domain (which you then also need to perform a Lasso regression against, which seems beyond the scope of OpenDP, but probably has implementations in Python already). It still seems good to keep the implementation and proof in full generality, parameterised by `m` so that a user can implement Bloom filters as they wish (as no other protocol right now can handle this form of input as far as I understand), but for the reasons stated above it may be best to push forward with the implementation as it stands right now, and not expand the scope any further.

One question, is there a transformation which can turn a categorical value into a `BitVec` right now? Seems like quickly wrapping `make_find`  and `make_index` for the back-and-forth would work, unless I'm missing an existing implementation.

If that is all good, then the changes to the proof seem like the last thing missing and hopefully we can close all of this out. I will make these changes Wednesday I've got the day off, and hope we can wrap this up :)

Shoeboxam:
The bloom filter/lasso could certainly be handled in a Python transformation, and then chained with this mechanism. 

If someone has a single categorical value, wouldn't they be better off using `make_randomized_response`?

Are there other variants of local-DP algorithms that would be useful to support?

abigail-gentle:
I'm thinking that given `k` categories, it would be nice to map them to their indices in `[k]` and then have an easy transform to the standard basis vectors so they can be fed into `rr_bitvec`

In frequency estimation it would be nice to have subset selection (poor communication, great performance) and projective geometry response (great communication, good performance). Performance here is of course optimal for both, but small variation in constants and rounding.

Shoeboxam (commented):
My thumb is up for merging this, but will wait for a last round from Vikrant before doing so. I'm looking forward to seeing this added to the library!

vikrantsinghal (approved):

PR #1288, opened 2024-02-22
Proposal: Interleave python and R

### LINKED ISSUE #1274: Interleave Python and R

mccalluc:
- Fix #1274

Feedback from you both would be great. Here are the pluses and minus I can come up with:

Plus
- RST is simple and predictable; Could even imagine writing an extension so all you'd need is  ``.. examples:: init`` and ``.. examples:: demo`` here.
- Python and R snippets in one file, so it's easy to make sure they are in maintained in parallel

Minus
- No code in the RST source: examples decoupled from text
- Big initial lift to get the python examples into R
- Need to be sure the python code doesn't use `"` and prematurely end string.
- No obvious way to add Rust examples... Maybe `///` inside R strings, `grep -v` to remove everything else, and then pipe to `rustdoc --test`?

### DIFF
--- a/docs/source/quickstart.R
+++ b/docs/source/quickstart.R
@@ -1,7 +1,27 @@
-# 1-library
+"
+# py-init
+>>> from opendp.mod import enable_features
+>>> enable_features('contrib')
+
+# /py-init
+"
+
+# r-init
 library(opendp)
 enable_features("contrib")
-# 2-use
+# /r-init
+
+"
+# py-demo
+>>> import opendp.prelude as dp
+>>> base_laplace = dp.space_of(float) >> dp.m.then_base_laplace(scale=1.)
+>>> dp_agg = base_laplace(23.4)
+
+# /py-demo
+"
+
+# r-demo
 space <- c(atom_domain(.T = "f64"), absolute_distance(.T = "f64"))
 base_laplace <- space |> then_base_laplace(1.)
 dp_agg <- base_laplace(arg = 23.4)
+# /r-demo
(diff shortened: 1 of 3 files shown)


mccalluc:
Research other notebook formats.

mccalluc:
Talking with Michael, realized that interleaved examples would not make life easier
- Multi-language support would only get harder with more target languages
- And already for Python, we want examples of the Framework and Context API in tabs: Doctest wouldn't be able to distinguish one from the other, if they are in the same file.

mccalluc:
... so I'm closing.

PR #1297, opened 2024-02-23
add link to Reference page from R docs landing page

### LINKED ISSUE #1296: On R docs, "Reference" is hard to notice
Looking at it with fresh eyes, and if I landed on this page, I wouldn't know to click "Reference" in the header. There should be a link from the body.

I didn't do this originally because the link wouldn't work with the file in place in github, but I think that's a secondary usecase.

... Maybe there's a CSS class that github uses to hide content, so we could hide it there, but still have it show when rendered for rdocs?

mccalluc:
- Fix #1296

This link won't work on github, but I think it's worth the trade-off. Without this, there's just a lot on the landing page, and I'd worry about users not noticing the link in the header.

But this is subjective.


raprasad (approved):
Approved

PR #1298, opened 2024-02-23
R favicons

### LINKED ISSUE #1276: Favicon on R docs
https://pkgdown.r-lib.org/reference/build_favicons.html

It looks like it's really just a wrapper around a third-party API, so my inclination would be to run it and check in the artifacts, but alternatively it could be done as part of the build.

mccalluc:
- Fix #1276

R docs now have a favicon.

Screenshot 2024-02-23 at 11 21 17 AM

Steps:
- Crop `opendp-logo.png` to just the star, and saved it to `R/opendp/logo.png`
- Run `build_favicon('R/opendp', TRUE)`
- Remove `R/opendp/logo.png`: With it in place, we would get the big star on the readme, and that's not really needed since we have it in the banner already.
- Check in the remaining files. (We don't really control the HTML that references these, so safest just to leave them all in place, even if the multiplicity of resolutions is a bit much.)

The call to the 3rd party service is pretty slow, so I think checking in the files makes more than than doing it at doc build time.


raprasad (approved):
approved

PR #1299, opened 2024-02-23
A concept on every R function

### LINKED ISSUE #1256: Add `concepts` on non-generated R
- follow-up to #1245

Add a `@concept` matching the file basename on each of the functions in the non-generated R, and update `_pkgdown.yml` so these also appear in sidebar. Also add a `desc` there.

(If one non-generated file has a lot of functions, we could let it be the fall-through default, to save some hand edits.)

This will make the documentation a little nicer... but it does have a maintenance cost. We might judge that this isn't worth the cost.

mccalluc:
Some manual (mod and typing), some generated (the `then` functions).
- fix #1256

("Concepts" are what provide grouping in R docs.)

### DIFF
--- a/R/opendp/_pkgdown.yml
+++ b/R/opendp/_pkgdown.yml
@@ -77,7 +77,7 @@ reference:
 
 - title: Typing
   desc: |
-    The `typing` module provides utilities that bridge between Python and Rust types.
+    The `typing` module provides utilities that bridge between R and Rust types.
     OpenDP relies on precise descriptions of data types to make its security guarantees:
     These are more natural in Rust with its fine-grained type system, but they may feel out of place in R.
     These utilities try to fill that gap.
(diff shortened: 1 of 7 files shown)


R/opendp/_pkgdown.yml
```diff
@@ -63,12 +63,28 @@ reference:
...
+
 - title: Transformations
-  desc: The ``transformations`` module provides functions that deterministicly transform datasets.
+  desc: The `transformations` module provides functions that deterministicly transform datasets.
   contents:
   - has_concept("transformations")
 
+- title: Typing
+  desc: |
+    The `typing` module provides utilities that bridge between Python and Rust types.
```
Shoeboxam, line 80:
```suggestion
    The `typing` module provides utilities that bridge between R and Rust types.
```

Shoeboxam (changes requested):
Small nit.

Shoeboxam (approved):
Didn't realize you could do this, but it's pretty essential!

PR #1300, opened 2024-02-26
Remove build_tool.py

### LINKED ISSUE #1272: Get rid of build_tool.sh
Raman concurs that it is not needed. Came up in the process of #1268.

mccalluc:
- fix #1272

I think the idea of avoiding vendor lock-in is great, but this doesn't seem to be maintained.

If we did want to revive this, one path might be to have a collection of smaller scripts that would be called either by CI, or locally by something like this.


Shoeboxam (approved):
It's nice that we do not need this anymore!

PR #1301, opened 2024-02-26
replace list of "rm" with "git clean"

### LINKED ISSUE #1259: In `r_stage.sh` replace long list of `rm` with `git clean -dX --dry-run`
I don't think I've used it before but [git clean](https://git-scm.com/docs/git-clean) looks useful:
```
git clean -dX --dry-run --exclude=...
```

I know that at least `R/opendp/README.md` is changing from a gitignored file to one that is checked in, so it might be good to simplify this list of files we list for deletion.

mccalluc:
- Fix #1259

Unless there's something subtle here, this seems like a simpler way to do what we want.


raprasad (commented):
~~approved~~

raprasad (approved):
approved

PR #1302, opened 2024-02-26
do not generate NEWS.md

### LINKED ISSUE #1257: Don't generate news/changelog for R docs
After this PR:
- #1246

We no longer link to the changelog from the header of the R docs. (Linking to the changelog does sense on a standalone project, but this is just the docs for R, and we don't need to duplicate what we have elsewhere.)

So let's stop generating this file as part of the R docs build.

mccalluc:
- Fix #1257

### DIFF
--- a/tools/r_stage.sh
+++ b/tools/r_stage.sh
@@ -105,10 +105,6 @@ function docs() {
   run cp -r R/opendp R/opendp-docs
   run rm -rf R/opendp-docs/src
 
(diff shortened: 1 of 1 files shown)


Shoeboxam (approved):
I guess I'm fine with removing this. A little less redundancy.

+1 for the branch name!

Shoeboxam (approved):

PR #1303, opened 2024-02-26
no more bare excepts

### LINKED ISSUE #1281: flake8: Do not use bare `except`
We should catch some particular error. Two files have per-file ignores: `_convert.py` and `test_typing.py`

_Originally posted by @raprasad in https://github.com/opendp/opendp/pull/1261#discussion_r1499107557_

mccalluc:
- Fix #1281

- Drop special handing for python < 3.8 -- If we want to offer some support for versions < 3.8, we should add it to the test matrix. 
- Catch `Exception` instead of implicit `BaseException` -- The tests don't currently exercise this line, so I'm not confident about the change. Since any exception here would be surprising, would we like to log it?

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -49,12 +49,6 @@ extend-ignore =
 per-file-ignores =
     # Use globs here so it works whether it's run from root or from python/.
 
-    # TODO: Bare 'except' should be fixed, but we'll typically want to be more strict,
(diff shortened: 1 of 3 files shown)


Shoeboxam (approved):
No need to handle versions older than 3.8 now!

PR #1311, opened 2024-02-26
Factor out r_examples.sh

### LINKED ISSUE #1287: Factor R example runner out into standalone script
With #1262 merged, we have R examples in the docs, and they are being run in CI. It would be nice if devs could easily run the same examples locally. Might take [this commit](https://github.com/opendp/opendp/pull/1233/commits/b5a647c273ef62f305fb5ef4d5da21db0387656d) as a starting point... but it would be even better if there is an R built-in that can do this, instead of wrapping our own.

mccalluc:
- Fix #1287

Runs locally.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -257,13 +257,7 @@ jobs:
         run: Rscript -e 'devtools::test(reporter = c("summary", "fail"))'
 
       - name: Documentation examples
-        run: |
-          STATUS=0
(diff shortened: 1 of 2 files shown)


mccalluc:
hmm: Same failure as before. Not clear that the shell script is even running. Add echo statements, and comment out all the complicated parts to debug on CI. Draft for now.

mccalluc:
- Replaced by python script in #1466

Issue #1024, opened 2023-11-21
Check test coverage


mccalluc:
We don't seem to be checking test coverage. While test coverage isn't everything (generally, we're only considering whether we hit a line of code, not all the possible states and execution paths), it can be helpful. Other considerations:

- Coverage for each language would need to be handled separately:
  - [ ] Rust
  - [ ] Python
  - [ ] R
- For a particular language, running all tests from one command would give us the best sense of coverage: #1022
- 100% coverage isn't essential, but where code (new or old) is not tested, I think it's better for us to add explicit ignore tags on individual blocks, rather than setting a lower, squishy bound across the entire codebase.
- There are third parties tools that can give a pretty display of the results. No objection, but I've found the plain text readable enough.

PR #1322, opened 2024-02-26
Check Rust test coverage

mccalluc:
- Towards #1024

Draft for now. 
- Want to assert that coverage should be 100%, and then add ignores as necessary, like we do in Python.
- Not sure if there should be corresponding updates in `rust_build.sh` and `development_environment.rst`


mccalluc:
Tarpaulin [warns](https://docs.rs/crate/cargo-tarpaulin/0.6.11):
> If you don't run nightly by default replace all calls to `cargo tarpaulin` with `cargo +nightly tarpaulin`

but we also need to install the nightly toolchain on CI if we go that direction -- not sure it's the best route. Current error:
```
error: toolchain 'nightly-x86_64-unknown-linux-gnu' is not installed
```

mccalluc:
There are at least two very different approaches to test coverage in Rust, and the tools are under active development. An old PR like this won't be useful.

PR #1332, opened 2024-02-27
python tabs from stand-alone file

mccalluc:
Proposal for including python examples the same way we include R examples.
- simple open-close convention with `/` on the closing tag.
- could distinguish python _Context API_ examples with `.context.py` extension.
- alternative to #1288 / #1274
- could imagine a sphinx extension to simplify the RST -- but I don't think that needs to block any thing.
- I wouldn't do more with this until #1177 is in -- conversely, I wouldn't block that PR to apply this convention.

### DIFF
--- a/docs/source/quickstart.py
+++ b/docs/source/quickstart.py
@@ -1,0 +1,12 @@
+'''
+# init
+>>> from opendp.mod import enable_features
+>>> enable_features('contrib')
+# /init
+
+# dp-demo
+>>> import opendp.prelude as dp
+>>> base_laplace = dp.space_of(float) >> dp.m.then_base_laplace(scale=1.)
+>>> dp_agg = base_laplace(23.4)
+# /dp-demo
+'''
--- a/docs/source/quickstart.py
+++ b/docs/source/quickstart.py
@@ -2,11 +2,13 @@
 # init
 >>> from opendp.mod import enable_features
 >>> enable_features('contrib')
+
 # /init
 
 # dp-demo
 >>> import opendp.prelude as dp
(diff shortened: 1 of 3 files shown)


docs/source/quickstart.py
```diff
@@ -0,0 +1,14 @@
+'''
+# init
+>>> from opendp.mod import enable_features
+>>> enable_features('contrib')
```
Shoeboxam, line 4: I'm surprised to see this all inside a string! Do you think it's worth keeping the code inside strings to get the `>>>` formatted doctests?

mccalluc: I do like the doctest style, but could reconsider.

for doctest:
- Particularly if moving away from notebooks, it would be a loss to not show any output.
- Could add `assert` but I find that a little clunkier, particularly if the output is longer. 
- If it was just python code, we'd need machinery to run it, I'm not sure it would work as smoothly with pytest.
- Sphinx renders this nicely, and you can copy and paste without getting the `>>>` prefix.

for plain python:
- syntax highlighting during development
- runnable without doctest

mccalluc: Maybe there's a way to get syntax highlighting in doctests?
Or maybe the file doesn't need to be `.py` if it's just a string?

mccalluc:
Good conversation, and I think we're on the same page. I'll close this and implement what we've learned on 
- #1177

namely:
- We should keep doctests
- they should be pulled into standalone files (particularly where we want to document two different APIs in parallel tabs-- we want to avoid namespace pollution between the two.)
- Rather than wrapping it as a multiline python string, the first line can just start with `>>>`
- There doesn't seem to be a convention for standalone doctests, but will probably just use `.doctest`

PR #1337, opened 2024-02-28
Just one favicon shared by sphinx and R

### LINKED ISSUE #1331: share favicon
Currently in conf.py:
```
html_favicon = u'favicon.ico'
```

Change this to point at the new R favicon, and get rid of the old sphinx one. (I haven't seen a configuration on the R side -- It must just use a conventional location.)

mccalluc:
- Fix #1331

Confirmed that it works locally.


Shoeboxam (commented):
Since the assets can't be put in a neutral place (because R doesn't allow configuring the location, and other langs likely won't), then I'd rather not make R the source of truth. IMO we should leave this as-is, as it won't necessarily scale to more langs.

mccalluc:
ok! Closing.

PR #1344, opened 2024-03-01
R linting

### LINKED ISSUE #1342: Add a lintr for R
**R**
[lintr](https://lintr.r-lib.org/) seems to be the default. I don't think there is yet any R code in main, but this might be something to have in place when it is merged to main.

mccalluc:
- Towards #1342

Probably keep issue open until we can remove most ignores, and get down to a subset that are either stylistic choices we disagree with, or things that would be harder to fix. (and a low priority issue could be filed for followups in that case.) I don't trust my own judgement on R code: I'd suggest that we should actually try to be more stringent when linting R than with a language we use more confidently. 

Right now I've split off the linting as a separate job, but I'm not committed to that -- It is nice to have everything run in parallel, rather than only seeing the first error if they are run serially -- but if the R code is not under active development this is extra CI work and extra noise in the UI.


raprasad (approved):
Approved. Reviewed with @mccalluc

mccalluc:
(No substantive changes, but merge from main cleared earlier approval.)

raprasad (approved):
approved (e.g. conflict was resolved from earlier)

PR #1345, opened 2024-03-01
add randomgen to dependency list

chikeabuah:


mccalluc:
- I just filed #1347 to describe the problem we're trying to fix.
- Switched this to draft to avoid confusion.
- Chiké ran pip-compile under python 3.8... which ends up introducing distutils... which isn't supported in the latest python versions --> The fix might just be to run pip-compile in 3.12 instead: Hopefully the dependency set that produces will be backward compatible to 3.8.

Issue #1347, opened 2024-03-02
Python dev environment optional dependencies + CI


mccalluc:
There are a few points in the code where we have special behavior if numpy is installed. We want want to:
- [ ] Add numpy to requirements-dev ... although maybe indicate that's its actually optional?
- [ ] Make sure tests work with or without numpy ... but probably not worth permanently expanding the test matrix?
- [ ] And do the same for randomgen?

Maybe try all combinations, once, but not something to check on every commit? There is some fuzziness here... but I had just gotten used to three failing tests in my local environment, and that's not good either. Sorry!

Background:
- #1037 was closed as won't-fix, then reopened, then closed by #1045, but that may have been a mistake.

This problem has been around for a while, but the desired behavior hasn't been clear.

PR #1348, opened 2024-03-04
Address numpy test failures

mccalluc:
- Towards #1347 

For now, draft: I want to check different patterns of optional installs with an enlarged test matrix, but I don't think we actually want to check all `2^(optional installs)` in CI on every commit.


mccalluc:
Update: I was actually hoping a test would fail -- Should have pushed the change to the test matrix before the change to the tests. The caching may be getting in the way of demoing the behavior: both the [no-extra-installs](https://github.com/opendp/opendp/actions/runs/8142711453/job/22253151020?pr=1348#step:9:277) and [two-extra-installs](https://github.com/opendp/opendp/actions/runs/8142711453/job/22253150108?pr=1348#step:9:277) have one test skipped.

Next: Remove CI caching, and add a pip-freeze to help debug.

mccalluc:
Test failures, yay! I think we want all of these combinations to pass.

mccalluc:
So when the tests pass for all combinations of `numpy`, `sklearn`, and `randomgen`, I'm not sure that we're done -- If someone had one of these partial install states, we want to be sure that opendp does not act up. Perhaps the end state we want is no skipif guards... Where we have them now, wrap the content of the test with a try-except, where the `except` makes sure that the error is something that explains the missing dependency. Perhaps we make sure there is an explicit `import randomgen` for example, unless we want an even friendlier exception?

mccalluc:
The last commit demonstrates the direction we might want to go: Where there are dependencies on optional libraries in the source code (excluding the tests), we can throw an exception explaining that an extra install is required. Is this a good direction?

mccalluc:
lean away from skip-ifs: instead try catch inside the test, looking for ImportException with a friendly message explaining optional install. Most important when randomgen is not installed, because the error is really cryptic.

mccalluc:
Keeping as draft to avoid premature merge, but would appreciate feedback: Tests are passing with all combinations of optional dependencies, and when a dependency is not present, we check that a user-friendly exception is raised.

The most interesting part is [`optional_dependency`](https://github.com/opendp/opendp/pull/1348/files#diff-daab583dc7c020875621ff8e37f4a2e5e9a4d18e6cb03cd4ce864b47069acdf2): This is used in tests to wrap lines which we expect to have a particular optional dependency -- If it's installed, test proceeds as usual; if it's not installed, we confirm that we get the expected user-friendly error message, and then the rest of the test is skipped.

Along with this, instead of wrapped imports in the tests, I'm using [`pytest.importorskip`](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-importorskip) extensively, which works well when the dependency is in the test itself.

Questions:
- What do we want for the test matrix? Maybe all the combinations are run nightly? Though since they're all parallel, it's really not that slow, although it is more total compute time.
- Do we want to change how the `np_csprng` works? Maybe change it to a singleton, raising an exception if dependencies aren't present?
- Is there too much `try-import-except-raise` boilerplate? The tests are also looking for a particular error message, so this is also kind of fragile. Could make a `import_optional_dependency` (or error) utility function.

mccalluc:
> Do we want to change how the np_csprng works? Maybe change it to a singleton, raising an exception if dependencies aren't present?

yeah: Making it a singleton makes sense.

.github/workflows/smoke-test.yml
```diff
@@ -90,7 +90,12 @@ jobs:
         working-directory: python
     strategy:
       matrix:
-        python-version: [3.8, 3.12]
+        python-version:
+          # - 3.8
+          - 3.12
+        optional-numpy: ['numpy', '']
+        optional-randomgen: ['randomgen', '']
+        optional-scikit-learn: ['scikit-learn', '']
```
Shoeboxam, line 98: We only need to test the extras we make available on the package. Instead of testing 2^3 combinations, we should only test the base install, install with the numpy extra, and install with the sklearn-extra (which is a superset of the numpy extra). 

When polars comes along, we can then also test with just the polars extra. I don't think we need to test combinations of polars with other features.

python/src/opendp/_extrinsics/_make_np_eigenvalues/__init__.py
```diff
@@ -10,7 +10,12 @@ def make_np_eigenvalues(input_domain: Domain, input_metric: Metric) -> Transform
     :param input_domain: instance of `_np_sscp_domain(size=_, num_columns=_)`
     :param input_metric: instance of `symmetric_distance()`
     """
-    import numpy as np  # type: ignore[import]
+    try:
+        import numpy as np # type: ignore[import-not-found]
+    except ImportError:
+        raise ImportError(
+            "The optional install numpy is required for this functionality"
+        )
```
Shoeboxam, line 18: Can you abstract this with a helper function? For reference, see how pandas does it:
https://github.com/pandas-dev/pandas/blob/main/pandas/compat/_optional.py

Shoeboxam (changes requested):
Nice adjustments!

python/src/opendp/_extrinsics/_make_np_eigenvalues/__init__.py
mccalluc: Will do: Don't think we need all the options in that example, but good to know this is a pattern.

.github/workflows/smoke-test.yml
mccalluc: Will do, but I'll save this for last to make sure I don't break an edge case in the course of the other work.

python/src/opendp/_extrinsics/_make_np_eigenvalues/__init__.py
Shoeboxam: Agreed, Pandas' approach is a bit more beefy than we need.

mccalluc:
The PRNG refactor was getting complicated, so I started a separate draft PR for that:
- #1433

So there are a couple `skipif`s left to handle that, but I think it would be good to review and merge this much.

Shoeboxam (approved):
Very cool! I like the shorter syntaxes throughout, and more principled approach towards extras. Looking forward, my guess is we'll probably have another optional extra `polars`, as well as `all`.

PR #1353, opened 2024-03-06
remove putting-it-together.rst, and move its diagram

### LINKED ISSUE #1293: Is putting-it-together still needed?
I think the new real-world example covers this. It's also a little out of place -- The API docs are really a reference, it muddles things to have a narrative example in here.

The diagram might be pulled out and reused on real-world.

mccalluc:
- Fix #1293

In some cases, like this, I think it's possible to have too much documentation. The typical-workflow page is new, and I think it's a good replacement for putting-it-together. I don't think many people will be reading this straight through... but some people may just want to find examples, so we should make sure they can be found.

This is subjective.


raprasad (approved):
approved

PR #1354, opened 2024-03-06
Update 404 template

### LINKED ISSUE #986: Friendlier 404s

mccalluc:
- Fix #986?
- Change the CI process: Just use plain HTML, so missing assets won't be a problem. (A different approach would be to copy up all the assets we might need.)
- add a list of substrings that should be replaced.


raprasad (approved):
approved

PR #1355, opened 2024-03-06
Runtime error if non-function is passed to `new_function`

### LINKED ISSUE #1190: Remove redundant function check is mod.py
I think that explains the `if not isinstance(other, Function)` above but I'm curious about `if isinstance(other, Function)` -- It seems like it would always pass now? Can't use GH suggestion because of scope, but would this work?

```diff
-        if isinstance(other, Function):
-            from opendp.combinators import make_chain_pm
-            return make_chain_pm(other, self)
-
-        raise ValueError(f"rshift expected a postprocessing transformation, got {other}")
+        from opendp.combinators import make_chain_pm
+        return make_chain_pm(other, self)
```

_Originally posted by @mccalluc in https://github.com/opendp/opendp/pull/1147#discussion_r1471596307_

### LINKED ISSUE #1339: Check that TransitionFn and CallbackFn are functions
_Originally posted by @Shoeboxam in https://github.com/opendp/opendp/issues/1190#issuecomment-1970137089_

mccalluc:
- Fix #1190
- Fix #1339
- Test the failing behavior
- Update the typing
- Remove coverage skip that isn't needed.

I think this does what we need, but I could be missing something.

### DIFF
--- a/python/src/opendp/core.py
+++ b/python/src/opendp/core.py
@@ -453,7 +453,7 @@ def measurement_output_measure(
 
 
 def new_function(
-    function: Callable,
+    function,
     TO: RuntimeTypeDescriptor
 ) -> Function:
     r"""Construct a Function from a user-defined callback.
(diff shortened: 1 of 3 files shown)


mccalluc:
Generally like to have checks at a low level, but that is generated code, and only called from one place.

Shoeboxam (approved):
Thanks for this!

Issue #1338, opened 2024-02-28
cargo build error, trying to use OpenDP 0.9.2 and 0.8.0


mccalluc:
From a fresh directory, ran `cargo init`, and updated `Cargo.toml`:
```
$ cat Cargo.toml
[package]
name = "docs"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
opendp = { version = "0.9.2", features = ["contrib"] }
```

`cargo build` fails:
```
error[E0432]: unresolved import `crate::combinators::IsSizedDomain`
 --> /Users/chuckmccallum/.cargo/registry/src/index.crates.io-6f17d22bba15001f/opendp-0.9.2/src/transformations/cast_metric/mod.rs:4:5
  |
4 |     combinators::IsSizedDomain,
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^ no `IsSizedDomain` in `combinators`

For more information about this error, try `rustc --explain E0432`.
error: could not compile `opendp` (lib) due to 1 previous error
```

If I change to 0.8.0 I get the same error. Is another step needed to init a project, or is something wrong with our build?

Shoeboxam:
This specific issue is that the amplify module is behind "honest-but-curious", and the trait is defined in that module.
To fix, move the trait out of that module.

The broader issue is that we should `cargo check` more combinations of features in CI. Should only add a few seconds to the CI.

mccalluc:
- Retitling to clarify scope
- Filed #1343

mccalluc:
@Shoeboxam -- organizationally, where does it make sense for `IsSizedDomain` to go?
- Up to `combinators/mod.rs`?
- Collected with other `Is*` traits in a new file?
- Split out on its own under `combinators`?
- Something else?

Shoeboxam:
Good question, I think it should go under traits. Possibly a new file, use your best judgement ;)

PR #1356, opened 2024-03-08
1338 test minimal cargo

mccalluc:
- Towards #1338

We want to make sure Rust OpenDP works with a minimal set of features. A place to start is a test of the current behavior, and it would also be nice to integrate it with the docs.

Down the road we probably want to do something like rewrite the `Cargo.toml` during test so it uses the local source code... but for documentation for external users, we need something that pulls from crates.io, rather than the local checkout.

I'm not sure if there's a more elegant way to organize this than to running `cargo init` on the directory: It would be nice to have the rust code sample next to the other code samples, but cargo works well because of conventions and I'm not sure we should fight that.

Some other edits to the RST, and move closing tags to the next line: the snippets were being truncated.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -65,6 +65,20 @@ jobs:
           name: vendor
           path: R/opendp/src/vendor.tar.xz
 
+  rust-examples:
+    runs-on: ubuntu-22.04
+    defaults:
+      run:
+        working-directory: docs/source/getting-started
+    env:
+      CARGO_TERM_COLOR: always
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Cargo run
+        run: cargo run
+
   python-lint:
     runs-on: ubuntu-22.04
     steps:
--- a/docs/source/getting-started/Cargo.toml
+++ b/docs/source/getting-started/Cargo.toml
@@ -1,0 +1,9 @@
+[package]
+name = "getting-started"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+opendp = { version = "0.9.2", features = ["contrib", "honest-but-curious"] }
--- a/docs/source/getting-started/quickstart.R
+++ b/docs/source/getting-started/quickstart.R
@@ -1,8 +1,10 @@
 # init
 library(opendp)
-enable_features("contrib") # /init
+enable_features("contrib")
+# /init
 
 # demo
 space <- c(atom_domain(.T = "f64"), absolute_distance(.T = "f64"))
 base_laplace <- space |> then_base_laplace(1.)
-dp_value <- base_laplace(arg = 123.0) # /demo
+dp_value <- base_laplace(arg = 123.0)
+# /demo
--- a/docs/source/getting-started/quickstart.rst
+++ b/docs/source/getting-started/quickstart.rst
@@ -1,7 +1,7 @@
 Quickstart
 ==========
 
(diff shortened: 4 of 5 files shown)


mccalluc:
Look at single file crates: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#single-file-packages

We'd also have some script to scan the docs for rust files and run them

and we'd also need to reference open dp by path

and we'd want to find a way to reference the built artifact, rather than rebuilding for every example.

mccalluc:
- Wait for #1371, and then move the rust file into the new subdirectory.

mccalluc:
Trying to keep the scope here small... but I may have made it too small. I'm thinking these tasks could be left for the future:
- In CI, compile the rust script against the opendp version on the branch rather than a pinned version. (while making sure the docs still reference the pinned version)
- Discover and check all rust files in docs. Probably just a `for` loop over `find`. Wait until there's at least one more example.
- Keep an eye on https://github.com/rust-lang/cargo/issues/12207 and remove reference to nightly when we can.
- See if there's any way to reference a cached build, rather than doing it over for each script we run. See https://github.com/rust-lang/cargo/issues/12207#issuecomment-1776089794
- I don't think this needs to wait until we have fixed #1338, but maybe there's a reluctance to put up docs that need to include a workaround for a bug.
- Make sure the examples are formatted.

docs/source/getting-started/quickstart.rs
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env -S cargo +nightly -Zscript
+```cargo
+# init
+[dependencies]
+opendp = { version = "0.9.2", features = ["contrib", "honest-but-curious"] }
```
Shoeboxam, line 5: I think this should be a relative path:
* We'd rather not need to update these version numbers throughout the scripts
* We'd like to be able to test things that haven't been released yet

docs/source/getting-started/quickstart.rs
```diff
@@ -0,0 +1,22 @@
...
+    measurements::then_base_laplace,
+    metrics::AbsoluteDistance,
+};
+
+fn main() {
+    let space = (AtomDomain::default(), AbsoluteDistance::default());
+    let base_laplace = space >> then_base_laplace(1.0, None);
+    let dp_value = base_laplace.expect("unexpected error").invoke(&123.0);
+    println!("DP value: {}", dp_value.expect("unexpected error"));
+}
```
Shoeboxam, line 21:
```suggestion
fn main() -> Fallible<()> {
    let space = (AtomDomain::default(), AbsoluteDistance::default());
    let laplace_mechanism = (space >> then_laplace(1.0, None))?;
    let dp_value = laplace_mechanism.invoke(&123.0)?;
    println!("DP value: {}", dp_value);
    Ok(())
}
```

A bit more rust-y. I think you'll also need to use `opendp::error::Fallible`.

docs/source/getting-started/quickstart.rs
Shoeboxam: The downside here is that we'll need a separate init block. Think we could have one init block in the quickstart, and omit init blocks elsewhere?

Shoeboxam (changes requested):
Getting there!

docs/source/getting-started/quickstart.rs
mccalluc: Updated: Now there's the real cargo block that uses the path, and suggested one that will be included in the docs.

raprasad (approved):
Reviewed with @Shoeboxam and @mccalluc

Shoeboxam (commented):
Approved, noting the new rust-examples task doesn't impact compile times:
* only takes one minute
* runs in parallel

PR #1359, opened 2024-03-11
Call mypy and flake8 as subprocesses from pytest

### LINKED ISSUE #1357: Wrap linting and mypy as subprocess tests
We have an increasing number of CI checks -- This is a good thing! -- but it also means:
- we have so many green checkmarks that it's hard to see everything at a glance
- these are all run as separate jobs, and so they all have their own start-up cost on CI
- there's no easy way to run them all locally.

Proposal: Move at least mypy and flake8 into a [`subprocess.run`](https://docs.python.org/3/library/subprocess.html#subprocess.run) call, wrapped by a test. This way, just running `pytest` will cover more ground. We could also look at pytest-parallel or pytest-xdist to so we still have the parallelism that CI jobs give us.

mccalluc:
- Fix #1357
- `vendor` is a new directory to ignore
- Simpler CI config
- For developers, just running `pytest` locally will catch more of the simple mistakes

`python-notebooks` might be added... but it is fairly slow, and also has an additional dependency.

Also considered adding the check for local modifications (`git diff --exit-code`) but local modifications are absolutely routine during development, so this would often give false positives if run locally.


mccalluc:
Fixing Numpy should fix CI...

mccalluc:
draft until CI is happy

raprasad (approved):
approved

PR #1360, opened 2024-03-11
More R linting

mccalluc:
Apply fixes and ignores so that that fewer linting rules need to be ignored. All the changes in this set seem reasonable to me. Looking ahead, I'm not sure how much farther to push linting:

    # If "linters_with_defaults" is used then only these problems are reported:
    indentation_linter = NULL, # Indentation should be ... spaces but is ... spaces.

Requires changes to code gen:
- #1368 is a first step,

and then need to regen the r code, and make hand edits in the other files.

    line_length_linter = NULL, # Lines should not be more than 80 characters.

Do we care about long lines?

    object_length_linter = NULL, # Variable and function names should not be longer than 30 characters.
    object_name_linter = NULL, # Variable and function name style should match snake_case or symbols.

Fixing these could require bigger changes to our conventions.

    # With "all_linters", these problems are also reported:
    expect_identical_linter = NULL, # Use expect_identical(x, y) by default; resort to expect_equal() only when needed, e.g. when setting ignore_attr= or tolerance=.
    expect_s3_class_linter = NULL, # expect_s3_class(x, k) is better than expect_equal(class(x), k). Note also expect_s4_class() available for testing S4 objects.

Seem ok, but the second should probably be tackled first, since it's a smaller set.

    extraction_operator_linter = NULL, # Use `[[` instead of `$` to extract an element.

Not sure the suggested change is always the best idiom

    implicit_integer_linter = NULL, # Integers should not be implicit. Use the form 1L for integers or 1.0 for doubles.

Seems excessive to me?

    routine_registration_linter = NULL, # Register your native code routines with useDynLib and R_registerRoutines().

Feeling out of my depth on this one, and what the implications might be.

    todo_comment_linter = NULL, # TODO comments should be removed.

I think TODOs can have a place, in conjunction with issues to track and surface issues.

    undesirable_function_linter = NULL, # Function "library" is undesirable. As an alternative, use roxygen2's @importFrom statement in packages and `::` in scripts, instead of modifying the global search path.
    undesirable_operator_linter = NULL, # Operator `<<-` is undesirable. It assigns outside the current environment in a way that can be hard to reason about. Prefer fully-encapsulated functions wherever possible, or, if necessary, assign to a specific environment with assign(). Recall that you can create an environment at the desired scope with new.env().

Global side effects aren't good... but it's also possible that we're relying on these side effects now, so we'd want to be cautious here. 

    unused_import_linter = NULL # Package 'testthat' is attached but never used.

I think this must be doing something useful for us, or it wouldn't have been included?


mccalluc:
Back out of draft: I'm going to mark this as
- fixes #1342

The lint rules we're still ignoring are either
- hard to fix (hanging indents, naming conventions)
- questionable utility (`L` suffix on integers, `[[` instead of `$`, disallow TODO)
- a little out of my depth (use useDynLib? avoid `<<-`?)

Shoeboxam (dismissed):
Approved as-is.

* A lot of the long lines come from links. Not sure how to get around those long lines. Ok with not linting these.
* I had no idea there were so many expect variants. Not in a rush to specialize these.
* Agreed on `$` vs `[[`. `$` is a narrower version of `[[`, but we don't need the more general abilities anyways, and `$` is much more succinct.
* We do use `L` for integers, but don't always use `.0` to distinguish non-ints. I'll try to further disambiguate with `.0` going forward.
* `routine_registration_linter` surprises me, because we do already use `useDynLib` and `R_registerRoutines`. Maybe the issue is that R is detecting the functions in the header files that aren't registered? But I include functions in the header file that I need to reuse in other C files, internally.
* Agreed on TODOs.
* Surprised to see the library lint apply to tests and packaging code.
* The side-effects are limited to function scope, I think global variables have immutable binding. This reduces the complexity of fixing these, of which there are only 9 instances from a pattern I was using for hashmaps.
* If it's possible, testthat should be relegated to a dev dependency, not a suggests (in DESCRIPTION)

mccalluc:
Thought about adding `implicit_integer_linter` to this, but I think that's going to take a little bit of work, so I'd prefer just to get this much merged.

raprasad:
Action item: Rebase the branch on the latest main 
cc: @mccalluc

mccalluc:
Rebased, and started a new PR:
- #1408

Move this back to draft, and delete if that can be merge successfully.

mccalluc:
- The new rebased PR looks good: #1408
- Issue filed for follow-ups: #1384

Closing.

PR #1361, opened 2024-03-11
LaTeX cache, temp, and output files

Shoeboxam:
I noticed no PR was made from this branch. I hope you don't mind I made one @abigail-gentle!


mccalluc (approved):
Approved... but I do feel a little weird about having so many extension based ignores: It's not feasible to have a git-ignored output directory? I'm thinking it'd be nice to have a more concise list, and also good to keep a distinction between source code and output. But don't let me block if that doesn't resonate with you.

Issue #1364, opened 2024-03-11
2 or 4 spaces for R indentation?


mccalluc:
For 2 spaces:
- It's the lintr default
- It's the [tidyverse style](https://style.tidyverse.org/syntax.html#indenting)

For 4 spaces:
- Would match the rest of our code in python and rust.
- Consistent with most of our current R code, the exceptions being `mod.R`, `configure.R`, and `update_notes.R`.

Clarification needed for:
- #1360

mccalluc:
Talked with Mike: 2 spaces for all R, both source code and examples.

mccalluc:
- fixed by #1408

closing

PR #1368, opened 2024-03-12
Add `number_of_spaces` param to `indent`

mccalluc:
Prereq for 
- #1364 / #1342

steps:
- added param to function
- search and replace usages
- cargo fmt, which introduces most of the changes, considered line by line.

Thought about introducing wrapper functions `indent_2` and `indent_4`, but this seems simpler, but no strong feelings.

Implicitly tested by the check for local changes to the generated code after the cargo build.

### DIFF
--- a/rust/opendp_tooling/src/codegen/mod.rs
+++ b/rust/opendp_tooling/src/codegen/mod.rs
@@ -18,8 +18,7 @@ pub fn write_bindings(base_dir: PathBuf, files: HashMap<PathBuf, String>) {
     }
 }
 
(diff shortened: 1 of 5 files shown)


Shoeboxam (commented):
How do you feel about `tab_r` and `tab_py`?

mccalluc:
Also removed some dead code warnings... maybe they date to before the functions were used? Or maybe CI will surface the problem that they prevent.

Shoeboxam (approved):
What a clever naming scheme!

PR #1350, opened 2024-03-04
Fix nightly

mccalluc:
- Fix #1349

envvar usage prevented CI from starting. See https://github.com/opendp/opendp/actions/runs/8135561781/workflow. The problem is really with `release.yml`, not `nightly.yml`. Simplify, by just repeating `inputs.dry_run || inputs.fake`. In some cases, only check `fake`, because not every `dry_run` is also `fake`

- Does the usage of `dry_run` vs `fake` now represent the behavior we want?
- Do we really need to support `fake`? I guess the idea is to speed up the build when we're exercising other parts of this... but I think it has been confusing.
- If we do want to support it, would it be more clear to have a select with three options, rather than than two checkboxes, one of which implies the other?

Need to manually run nightly on this branch and confirm that it works.

PR #1378, opened 2024-03-13
Upgrade github actions

### LINKED ISSUE #1372: Upgrade from node 16 to 20 in CI
Seeing this warning from action runs:
> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.

mccalluc:
- Fix #1372

There are backward incompatible changes to artifacts: Formerly, they were mutable, and you could add files to an artifact in multiple steps. On a quick skim of the code, I don't think we were using that functionality, but I could be wrong. Steps from here:
- Confirm that smoke-test passes
  - and that there are no Node 16 warnings
- Run the build manually, and confirm that it passes. (Wait until #1350 is merged to main, and then merge main to this branch, before trying to run manually.)
  - and that there are no Node 16 warnings
- Do a more thorough inspection to see if there are any workflows the steps above didn't cover, or if there are places where we're using the mutability of artifacts, but the testing wasn't strong enough.

Draft for now.


mccalluc:
smoke-test passed; doing a dry-run now: https://github.com/opendp/opendp/actions/runs/8271743407

mccalluc:
Several failures on dry-run, all with the same message:
- [libs-windows](https://github.com/opendp/opendp/actions/runs/8271743407/job/22632204279): `Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run`

Looks like we were depending on the mutability of artifacts, so this upgrade will be a little harder. [Migration guide](https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md).

mccalluc:
- Filed #1400 -- Don't think it has anything to do with this PR in particular.
- macos and linux artifacts need to be further qualified by machine type.

mccalluc:
nightly sync v1 dry-run: https://github.com/opendp/opendp/actions/runs/8329607066

mccalluc:
only failing on docs build now:
```
 cp: cannot stat '404.html': No such file or directory
```

mccalluc:
Nightly sync v3 dry-run: https://github.com/opendp/opendp/actions/runs/8330766088

mccalluc:
[`publish-python` fails](https://github.com/opendp/opendp/actions/runs/8330766088/job/22796974464#step:5:156):
> 400 File already exists

The version number that we supply in the github UI isn't passed to this job, so the names collide. I'll see if there's a light-weight way to override this.

mccalluc:
Looked into passing the version number specified in the UI and adding it as a suffix to the version number, but that adds a lot of complexity. Maybe this is sufficient?

raprasad (approved):
Reviewed at standup with @mccalluc and @Shoeboxam

raprasad (approved):
approved

PR #1380, opened 2024-03-13
Remove dead_code markers that do not cause warnings in IDE

### LINKED ISSUE #1374: Clean up `#[allow(dead_code)]`
In this PR:
- #1368

I removed a `#[allow(dead_code)]`, with no ill effect. My guess is that the function was checked in before we had a use for it, and the flag was necessary. There are ten other places where this is used -- It would be good to see if they are still necessary, and if they are, confirm that we still think that the code will be used at some point. Might add a comment in the source code giving the issue for the work to be done.

mccalluc:
- Fix #1374

These don't seem to be necessary.

### DIFF
--- a/rust/opendp_tooling/src/codegen/mod.rs
+++ b/rust/opendp_tooling/src/codegen/mod.rs
@@ -8,7 +8,6 @@ use crate::{Argument, TypeRecipe};
 pub mod python;
 pub mod r;
 
-#[allow(dead_code)]
 pub fn write_bindings(base_dir: PathBuf, files: HashMap<PathBuf, String>) {
     for (file_path, file_contents) in files {
         File::create(base_dir.join(file_path))
@@ -44,7 +43,6 @@ pub(crate) fn tab_c(text: String) -> String {
 }
 
 /// resolve references to derived types
-#[allow(dead_code)]
 fn flatten_type_recipe(type_recipe: &TypeRecipe, derived_types: &Vec<Argument>) -> TypeRecipe {
     let resolve_name = |name: &String| {
         derived_types
--- a/rust/src/core/mod.rs
+++ b/rust/src/core/mod.rs
@@ -281,7 +281,6 @@ where
         })
     }
 
-    #[allow(dead_code)]
     pub(crate) fn with_map<MI2: Metric, MO2: Measure>(
         &self,
         input_metric: MI2,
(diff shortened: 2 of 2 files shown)


rust/opendp_tooling/src/codegen/mod.rs
```diff
@@ -8,7 +8,6 @@ use crate::{Argument, TypeRecipe};
 pub mod python;
 pub mod r;
 
-#[allow(dead_code)]
 pub fn write_bindings(base_dir: PathBuf, files: HashMap<PathBuf, String>) {
```
Shoeboxam, line 11: This one _is_ dead code if the bindings feature is disabled. Could instead do:

```suggestion
#[cfg(feature = "bindings")]
pub fn write_bindings(base_dir: PathBuf, files: HashMap<PathBuf, String>) {
```

Shoeboxam (changes requested):
Nice catches.

rust/opendp_tooling/src/codegen/mod.rs
mccalluc: I might be misunderstanding your request, but that doesn't work for me. This change:
```diff
 pub mod python;
 pub mod r;

+#[cfg(feature = "bindings")]
 pub fn write_bindings(base_dir: PathBuf, files: HashMap<PathBuf, String>) {
     for (file_path, file_contents) in files {
         File::create(base_dir.join(file_path))
```
produces two build errors like this:
```
$ cargo build --features untrusted,bindings
...
error[E0425]: cannot find function `write_bindings` in module `codegen`
  --> build/derive.rs:39:22
   |
39 |             codegen::write_bindings(base_dir, codegen::python::generate_bindings(&_modules));
   |                      ^^^^^^^^^^^^^^ not found in `codegen`
   |
note: found an item that was configured out
  --> /Users/chuckmccallum/github/opendp/opendp/rust/opendp_tooling/src/codegen/mod.rs:12:8
   |
12 | pub fn write_bindings(base_dir: PathBuf, files: HashMap<PathBuf, String>) {
   |        ^^^^^^^^^^^^^^
   = note: the item is gated behind the `bindings` feature
```

My first guess was that downstream usages would also need to be gated, but adding `#[cfg(feature = "bindings")]` in `derive.rs` doesn't change anything. Maybe also need a change in `build/main.rs`?

Shoeboxam (approved):

Issue #1393, opened 2024-03-15
Drop support for Python 3.8 in October 2024?


mccalluc:
3.8 reaches EOL in October: https://devguide.python.org/versions/

When we drop 3.8 we can simply our typing, and just use `list` instead of `Typing.List`. Mypy may also work better with 3.9 than 3.8.

Before October we can upgrade our github actions to 3.9.

PR #1398, opened 2024-03-18
3.8 -> 3.12 in CI (except for smoke-test)

mccalluc:
- Towards #1393

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -256,7 +256,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v3
         with:
-          python-version: 3.8
+          python-version: 3.12
           cache: pip
 
       - name: Download libs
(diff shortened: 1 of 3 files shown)


mccalluc:
[r linting failures](https://github.com/opendp/opendp/actions/runs/8327661390/job/22785859499?pr=1398), all of the form `[object_usage_linter] no visible global function definition for ...`. Will try rerunning, and then see if other pushes are also failing. Possibilities:
- Erratic failure: we won't see it again?
- Un-pinned dependency issue?
- Actually some dependency between python version and R?

raprasad:
Note, earlier versions of Python are still running in the test matrix. This is for dev tooling

Shoeboxam (approved):
Rationale makes sense- dev tooling doesn't need to be minimum Python.

PR #1386, opened 2024-03-14
Enhance context docs

mccalluc:
- Towards #1375

To work on this issue, and others, I need to understand context better. Since the docs are thin, filling them out is one way to do this. Very much a draft!

Issue #1400, opened 2024-03-18
Erratic r-linting failures


mccalluc:
- Working on #1398, hit this failure: https://github.com/opendp/opendp/actions/runs/8327661390/job/22785859499?pr=1398 ... resolved on re-run
- Working on #1378, hit the same failure: https://github.com/opendp/opendp/actions/runs/8328906500/job/22789976615

```
Warning: file=R/accuracy.R,line=27,col=11,[object_usage_linter] no visible global function definition for 'parse_or_infer'
```
repeated 1250 times, for different functions.

- Maybe a race condition? Is something only partially installed before we continue?
- Since it only seems to be one linter that complains, just disable that one?
- Google the error message?
- Are there any errors earlier in the log that are the real root cause?

mccalluc:
Perhaps this is a fix? https://github.com/r-lib/lintr/issues/1137

There are a number of reports of errors with this linter, so it may be worth doing from more research.

PR #1401, opened 2024-03-18
Measure `__str__` -> `__repr__`

### LINKED ISSUE #1390: Add `repr` for Measure, and maybe others
```
>>> dp.loss_of(epsilon=1.0)
(<opendp.mod.Measure object at 0x1034c4bd0>, 1.0)
>>> m, x = dp.loss_of(epsilon=1.0)
>>> m
<opendp.mod.Measure object at 0x1034c4dd0>
>>> str(m)
'MaxDivergence(f64)'
```
I think the later is more useful, particularly writing doctests: I think we just want to define `__repr__` instead of `__str__`.

mccalluc:
- Fix #1390
- Will cause a small conflict with #1386, but didn't want to expand the scope of that PR unnecessarily.

Am I correct in thinking that `loss_of` is the most idiomatic way to construct a `Measure`? Change the wording here if not.

### DIFF
--- a/python/src/opendp/context.py
+++ b/python/src/opendp/context.py
@@ -180,10 +180,15 @@ def loss_of(*, epsilon=None, delta=None, rho=None, U=None) -> Tuple[Measure, flo
 
     :param U: The type of the privacy parameter.
 
-    >>> from opendp.context import loss_of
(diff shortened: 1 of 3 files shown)


mccalluc:
r-lint failure. 😢 
- #1400

PR #1415, opened 2024-03-21
Add setuptools to requirements

### LINKED ISSUE #1413: Python job fails on nightly: Missing `setuptools`
Nightly is failing. [setuptools is missing](https://github.com/opendp/opendp/actions/runs/8369403837/job/22915415668#step:6:38):
```
Build bdist
$ (cd python && python setup.py bdist_wheel -d wheelhouse)
  File "/home/runner/work/opendp/opendp/python/setup.py", line 1, in <module>
    from setuptools import setup  # type: ignore
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ModuleNotFoundError: No module named 'setuptools'
Error: Process completed with exit code 1.
```

Easy fix?

Also, consider adding a nightly status badge on the README so this will be a higher priority for us going forward.

mccalluc:
- Fix #1413

Draft for now: Want to try a dry-run.

Still curious what changed that caused it to break.


Shoeboxam:
I think this is because the Python version was switched to 3.12, and then we got affected by:
https://github.com/python/cpython/issues/95299

Shoeboxam (approved):

PR #1417, opened 2024-03-21
remove check from nightly, so it only calls release

### LINKED ISSUE #1416: Inaccurate "success" responses from nightly
If I look at the latest nightly builds, I see a number of green checkmarks:
- https://github.com/opendp/opendp/actions/workflows/nightly.yml

But on closer inspection, these runs were all very quick: If there have been no changes, the `check` job exits and reports success. This is problematic:
- It makes it harder to spot when a problem began, if every weekend the build is green.
- It gives us a false sense of security.
- Even if we added logic to only skip if the last build was good, builds aren't really deterministic: We depend on outside services, and if one of those services is flaky, we want to know.

We don't want to waste CI compute time, but I think doing a real build every night would make more sense.

mccalluc:
- Fix #1416

This makes sense to me, but there may be other views.


Shoeboxam (approved):
Considered asking to also condition on last run's failure status, but why keep the complexity? APIs will change, no reason to maintain this check to possibly save a small amount of compute on weekends.

PR #1423, opened 2024-03-21
Explain that R install does not require pre-compiled code

### LINKED ISSUE #1399: `r` job under `build.yml` doesn't seem to download `lib` artifact
- Working on #1378

The comments at the top explain (with the addition of `-*` in my local branch):
```
# * Downloads the lib-* and python_bindings artifacts, creates a Python wheel, uploads to the python_wheel artifact
# * Downloads the lib-* and r_bindings artifacts, creates an R package, uploads to the r_package artifact
```

and while I do see the `lib` download for python:

```
      - name: Download libs
        uses: actions/download-artifact@v4
        with:
          pattern: libs-*
          merge-multiple: true
          path: rust/

      - name: Download Python bindings
        uses: actions/download-artifact@v4
        with:
          name: python_bindings
          path: python/src/opendp/
```

I only see one download for R:
```
    steps:
      - name: Checkout repository
      - name: Set up R
      - uses: r-lib/actions/setup-r-dependencies@v2
      - name: Download R bindings
      - name: Source tar
      - name: Notes
      - name: Upload R package
```
Is the lib download implicit in another step? Have we actually seen a developer without a checkout of opendp install the opendp R package and use it successfully? For python we have the sanity test that can confirm that package install works in a fresh environment -- I don't know that we do anything like that for R.

### LINKED ISSUE #1407: In quickstart, explain Rust install process for R users
The R package will not work on its own: Users need to also install the Rust package. Explain this in Quickstart.

from https://github.com/opendp/opendp/issues/1399#issuecomment-2007559064

mccalluc:
- fix #1399
- fix #1407


Shoeboxam (approved):
Thanks for the clarification!

PR #1424, opened 2024-03-21
add a type variable, and remove mypy ignore

### LINKED ISSUE #1411: Correct return annotation on `loss_of`
Currently:
```python
def loss_of(epsilon=None, delta=None, rho=None, U=None) -> Tuple[Measure, float]:
    ...
        ...
        return max_divergence(T=U), epsilon
        ...
        return fixed_smoothed_max_divergence(T=U), (epsilon, delta) # type: ignore[return-value]
```
The last return is inconsistent with the signature. options:
- Change to `Tuple[Measure, Union[float, Tuple[float, float]]]`?
- ... and define a type variable for `Union[float, Tuple[float, float]]`, which could also be used in the signature for `Query`?
- Make it an `Array[float]` instead, and if there are the wrong number of elements, it's a runtime error?
- status quo?

mccalluc:
- Fix #1411

FYI: The rendered docs do not show the type variable:

> Return type:
> 
>     Tuple[Measure, float | Tuple[float, float]]


python/src/opendp/context.py
```diff
@@ -212,7 +212,10 @@ def metric_of(M) -> Metric:
     raise TypeError(f"unrecognized metric: {M}")
 
 
-def loss_of(epsilon=None, delta=None, rho=None, U=None) -> Tuple[Measure, float]:
+LossParameter = float | Tuple[float, float]
```
Shoeboxam, line 215: I dont think this will work before 3.10 without `from __future__ import annotations`?

Shoeboxam (changes requested):

mccalluc:
Lots of failures: All look like:
```
ValueError: predicate always fails. An example traceback is shown above at 0.0.
```

This PR has drifted: Maybe refocus on just getting the type annotation right, before trying to make it elegant.

mccalluc:
- Close in favor of #1524

PR #1430, opened 2024-03-25
gitignore generated bindings

### LINKED ISSUE #1425: gitignore Python and R bindings
Previously, they were needed, but after Andy's rewrite of the release CI last fall, they shouldn't need to be checked in. At the time we weren't sure if the first step was going to raise more issues, but the dust has settled on that and we should be able to move forward.

For example:
```
      - name: Checkout repository
        uses: actions/checkout@v4
        with:
          ref: ${{ inputs.channel }}
...
      - name: Download Python bindings
        uses: actions/download-artifact@v4
        with:
          name: python_bindings
          path: python/src/opendp/
```
right now, since the bindings are also checked in, the artifact download is redundant, but we'll rely on that going forward.

mccalluc:
- Fix #1425

Not sure if listing the files individually is the right call, but particularly for the Python, generated and hand-crafted files are mixed, (and I don't think it makes sense to try to separate them right now,) but if there's a new generated file, checking it in by mistake seems less annoying than the reverse.

Draft for now.


mccalluc:
- Notebook tests failing, so download the artifacts there, too.
- Was going to say we could get rid of the check for modifications to the checkout, but worth keeping. If we hit it, it's still a problem, though the root cause would be different.

Shoeboxam (changes requested):
There's another hidden piece to this, unfortunately. The generated bindings should get checked in to the release tag. The generated bindings should be committed into release tags because sphinx multi-version checks out to each of the tags to build the docs on that version. We don't want to have to run the rust build to generate these files in each tagged version.

mccalluc:
@Shoeboxam -- The bindings are probably also needed for the publish to pypi? If this isn't what you had in mind, let me know.

What do we need to test to be confident that this is working?
- install from test-pypi?
- docs build? but I'm not sure if there's a mechanism for pushing a dev docs release?

Shoeboxam (changes requested):
I think the key thing to verify is that, when you cut a dev build, the dev branch contains generated code. If this is the case, then the dev-build from test-pypi should also work, and I think the tags will also point to SHA's that contain the codegen. 

At this point, I'm thinking the dev branch probably won't have these files (on a new build), even though you've pulled the artifact down, because the files are still .gitignored.

mccalluc:
> At this point, I'm thinking the dev branch probably won't have these files (on a new build), even though you've pulled the artifact down, because the files are still .gitignored.

That would make sense. `git add --force` seems simplest. Other possibilities:
- Before making the commit to the channel, strip this section from the gitignores: `perl -ne 'print unless /Auto-generated files/../END/' -i .gitignore`
- Add new gitignores lower in the directory tree that override the top level.

Shoeboxam:
Agree, my money is on something like `git add --force` for all .py, .R and .c files.

mccalluc:
Running prepare.yml: dev / sync / v0 / not-dry-run: https://github.com/opendp/opendp/actions/runs/8526561113

mccalluc:
Assuming the `git add --force` works, is there the possibility that this is committing too much? Do we need something more selective than `.`? It would be great if that sufficed.

mccalluc:
FWIW: A couple times now since I filed this PR, I've found it useful to make some edits to the generated Python or R and submitted a draft PR -- It obviously fails, but it's an easy way to sketch out proposed changes before investing the time on `python.rs` or `r.rs`.

Shoeboxam:
* **#1430** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1430?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Shoeboxam (dismissed):
I rebased this PR and re-ignored the files. I would like to merge this, as I think it will significantly reduce the noise in PRs, and will help concentrate attention onto more meaningful files.

Shoeboxam (approved):
Thanks! I'll go through and rebase my PRs to avoid re-introducing these files.

PR #1433, opened 2024-03-26
randomgen guards

mccalluc:
- Extends #1348 
 
Moves PRNG into a singleton, and removes the last skipifs. 

Filed a new PR because this is failing locally for some combinations of installs, and introduces some significant changes, and I wanted to wrap up the parts that do work.


mccalluc:
replaced by #1562

Issue #1418, opened 2024-03-21
Usability: API Function Documentation


chikeabuah:
<img width="641" alt="Screen Shot 2024-03-21 at 9 36 15 AM" src="https://github.com/opendp/opendp/assets/2280440/5dfa4448-6295-4479-ae63-18a0a33a2dc9">

Consider, documenting each API function on a separate page, along with examples and DP-specific advice.

From study "Evaluating the Usability of Differential Privacy Tools with Data Practitioners" by Ngong et al.

mccalluc:
There are a couple separate issues here:
- Docs are searchable now, and hopefully it's easy to find the search? But I think they've been searchable for a while, so I'm not sure I understand the user's original problem.
- More examples in the API reference: This is hard, because the framework docs come from Rust. Could imagine adding logic to the docs-build to insert examples in the appropriate language, but it would add some complexity. Alternatively, could link to appropriate sections of the user guide from the API reference.
- Separately, we could have smaller pages, though I'm not sure this would be a big win. [Some options on SO](https://stackoverflow.com/questions/43357912/using-sphinx-to-automatically-generate-a-separate-document-for-each-function). FWIW, one function per page is the default for R docs.

Shoeboxam:
* Agreed, we've always had search.
* This is why there's such a harsh division between the user guide and API docs. The user guide is hand-written and should contain examples. Given that API docs are mostly generated, it's easy to go from the user guide to the API docs, but difficult to go from the API docs to the user guide.
* Separate pages for each constructor make sense, as the API docs are not meant to be read top-to-bottom.
* I would love to have more guidance on DP and I think the OpenDP Project is well-suited to do this. This was the reason for writing the first look at DP notebook, programming framework notebook and that more mathematical getting started notebook. I've always hoped that the educational working group might help produce some materials for this, but in the meantime I've gradually worked on adding guidance into the OpenDP docs.

Shoeboxam:
There's a fair amount we can do to reduce the surface area to document, as well. We could consider removing transformation constructors like casting, imputation, b_ary_tree, quantile scoring, lipschitz mul, and all the transformation aggregators in favor of rolled-up all-in-one measurement constructors. While on the one hand this would reduce the flexibility of the library, it would make it easier to see the forest from the trees.

mccalluc:
There will always be more work here, but at a high level things have improved a lot. 
- Search is working.
- we have some python-specific examples in the API reference, and want to add more for the most highly visible ones: #1545
[...]

PR #1439, opened 2024-03-27
Python example in API docs proposal

mccalluc:
- Towards #1418
- `.rst` extension because it works well in our IDE, but could be anything.
- Something similar could be done for R.
- Doctesting works. (It runs on the generated python code, not the `rust/**/*.rst`)
- Following the `opendp.prelude` pattern, though it's not actually required here.

Thinking about:
- #1430 

In this case, it was nice to be able to see the consequences of changes to the code generation, but this is probably an unusual case. We'll also want to make sure in that PR that the doctests run on the gitignored, generated code.


rust/opendp_tooling/src/codegen/python.rs
```diff
@@ -252,10 +257,16 @@ fn generate_docstring(func: &Function, hierarchy: &HashMap<String, Vec<String>>)
         }
     );
 
+    let example_path = format!("src/{}/{}.rst", &module_name, &func.name);
```
Shoeboxam, line 260: I think this would be better under `examples/`. I think the misc examples files could really clutter up a folder that should primarily have an implementation and the associated proofs.

Shoeboxam (changes requested):
Sorry to be a stickler on extra folders.

mccalluc:
Rather than `examples`, how about `code`? That would match the convention we have in docs. Admittedly, in the rust source directory everthing is code, so it's not that descriptive.

I don't feel strongly.

Shoeboxam (approved):
Looking good! Glad to have these under a folder now.

PR #1442, opened 2024-03-27
Use `import opendp.prelude as dp` in docs

### LINKED ISSUE #1409: Make sure that doctests use `dp` namepace
We want to encourage consistent patterns, so we should review the module doctests and make sure they are always referencing members via `dp`. (The members in the module are in scope during the doctest run, but we don't want to do something the user shouldn't do.)

(Derived from #1391 -- Originally thought this should go the opposite direction.)

mccalluc:
- Fix #1409

Right now, if you search for `>>>.*opendp`, you won't find the older idioms. Yay!

It occurs to me that a different case are module doctests which use may use unqualified names. By default, doctest provides these globals, but [maybe we can override](https://docs.python.org/3.5/library/doctest.html#what-s-the-execution-context)?

Also, the notebooks should be updated.

Filing as draft.


mccalluc:
Thought about overriding globals, but although it's something that doctest itself supports, it's not an option pytest exposes. Rather than making the test code more complicated, it seems better to just make this our responsibility.

Shoeboxam (approved):
Long overdue, thanks for handling this.

PR #1445, opened 2024-03-28
prelude note on each module's API docs

### LINKED ISSUE #1388: Mention the `import opendp.prelude as dp` idiom at the top of every module doc
In the body of the documents, it will still reference the full paths, but this would at least be a reminder.

mccalluc:
- Fix #1388

Sorry for the frankenmarkdown, but this seems to work.


rust/opendp_tooling/src/codegen/python.rs
```diff
@@ -69,16 +69,34 @@ from opendp.measures import *"#
...
-        "measurements" => "The ``measurements`` module provides functions that apply calibrated noise to data to ensure differential privacy.",
-        "measures" => "The ``measures`` modules provides functions that measure the distance between probability distributions.",
-        "metrics" => "The ``metrics`` module provides fuctions that measure the distance between two elements of a domain.",
-        "transformations" => "The ``transformations`` module provides functions that deterministicly transform datasets.",
-        _ => "TODO!"
+        "accuracy" => format!("{}\n\n{}",
+            "The ``accuracy`` module provides functions for converting between accuracy and scale parameters.", prelude_note),
+        "combinators" => format!("{}\n\n{}{}",
+            "The ``combinators`` module provides functions for combining transformations and measurements.", prelude_note,
+            "    >>> dp.c.make_basic_composition # doctest: +ELLIPSIS\n    <function make_basic_composition at ...>"),
```
Shoeboxam, line 84: This is getting a little scary. Doctests embedded inside rst inside Rust strings. I don't have a better recommendation though. No actionable feedback here.

Shoeboxam (approved):
Up to you, but it might be nice to explain that certain modules are accessed via `.c`, `.m`, `.t` so as to help distinguish between combinators, measurements and transformations.

mccalluc:
Maybe drop the doctest weirdness (instead leave that for future python examples below to make clear), and for now just mention the one-letter-namespace? I'll switch this back to draft until fixed.

mccalluc:
- Close in favor of #1458

PR #1446, opened 2024-03-29
Update introductory paragraph, remove last outdated section

### LINKED ISSUE #1295: In new docs branch, clean up theory/getting-started.html
At the top it says:
> This notebook is an introduction to programming in the OpenDP Library

which is out of place for something nearly at the end of the docs, as is the name of this page

Revise to say that this is an Intro to OpenDP, with an emphsis on connecting to the underlying theory, or something like that?

... but also double check that this is still needed, and that it's not covered elsewhere.

mccalluc:
- Fix #1295


raprasad (commented):
approved

raprasad (approved):
Approved

PR #1447, opened 2024-03-29
API ToC just down to modules

### LINKED ISSUE #1294: In new docs branch, api/python/index.html, full tree shown
Probably better just to show the next level down, if we have control over this? Not something to block PR over.

mccalluc:
With this change, the ToC looks like:
Screenshot 2024-03-29 at 10 40 15 AM

I think it's an improvement, but it is subjective.
- Fix #1294

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -28,7 +28,7 @@ clean:
 
 html:
 	@set -e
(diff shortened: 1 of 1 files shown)


raprasad (approved):
approved (rebuilt locally)

PR #1448, opened 2024-03-29
Explain extra installs for R / Add "is" on homepage

### LINKED ISSUE #1238: Explain the extra brew installs for R on macOS
```
brew install harfbuzz fribidi libgit2 xquartz
```

Suggest adding a `<details>` section that explains these: Are they needed for R in general, or is it something particular about opendp? I know that xquartz is needed with the standard R install, because that install also includes tcltk stuff, so X11 compatibility is not surprising, but not sure about the others.

mccalluc:
- Fix #1238


raprasad (commented):
Approved

raprasad (approved):
Approved!

PR #1450, opened 2024-03-29
Remove sphinx doctest tags

### LINKED ISSUE #1284: Turn off `sphinx.ext.doctest`
We're no longer using sphinx to run doctests: Instead, all the python tests are handled by pytest. (The relevant bit in `.pytest.ini` is `--doctest-glob *.rst`)

So to avoid confusion we should remove `sphinx.ext.doctest` from `conf.py`. The RST will also need updates.

Wait until this PR is merged:
- #1177

We don't want to put any roadblocks in the way of updating the docs.

mccalluc:
- Fix #1284
- and remove some other unused sphinx options while I'm here.


raprasad (approved):
Approved.

PR #1452, opened 2024-03-29
Rename `_extrinsics` to `extras`

### LINKED ISSUE #1432: document `pip install opendp[numpy]` and `pip install opendp[scikit-learn]`?
I could be overlooking it, but I don't think we mention these extra requires anywhere in our docs.

I think the right place is in the API reference, and we'd probably just add a note for each method that needs extra installs, either:
> This method requires extra installs: `pip install opendp[numpy]`

or 
> This method requires extra installs: `pip install opendp[scikit-learn]`

I also worry that people might be tempted to just `pip install numpy` -- Would it make any sense to rename our extras to reflect functionality, rather than libraries? something like: `pip install opendp[pca]`

mccalluc:
Planning to:
- rename the module
- make it public, and confirm that it shows up in API docs
- hide members that we don't want public
- add documentation about `pip install opendp[numpy]` and `pip install opendp[scikit-learn]`
- change `register_measurement`: instead of putting the argument into a different module, it will update the `constructors` dict in `context.py`
- Fix #1432

Draft for now.


mccalluc:
@Shoeboxam - Keeping as draft to prevent premature merge, but would appreciate feedback on this much.

- Introduced `dp.x` after we spoke: I think it might make sense to not simply import these into the `dp` namespace, but instead flag them so usages will stand out.
- Still need to update notebooks.
- `register_measurement`: It doesn't seem to be causing any errors right now: Before just jumping into that code, I'd like to have a failing test first, to be more confident that the eventual fix does the right thing.
- Sphinx is having trouble with the dynamically defined SKLPCA class. I'm not sure exactly what the fix will be there, so I'd suggest not holding up this PR for that.
- The one-function-per-module arrangement makes the doc structure strange. Any interest in reducing it to two public submodules under `extras`, one for numpy-related tools, and one for sk-learn?

Shoeboxam:
Disposition to merge after 0.10!

python/src/opendp/_lib.py
```diff
@@ -287,16 +287,16 @@ def proven(function): # pragma: no cover
         source_dir = os.path.dirname(inspect.getfile(function))
         absolute_proof_path = os.path.abspath(os.path.join(source_dir, matched_path))
 
-
-        # split the path at the extrinsics directory
-        extrinsics_path = os.path.join(os.path.dirname(__file__), "_extrinsics")
-        relative_proof_path = os.path.relpath(absolute_proof_path, extrinsics_path)
+        # TODO: Is special handling for extras still needed?
```
Shoeboxam, line 290: Yes, I believe this is still necessary- it has to do with links being relative from the repository root.

Shoeboxam (changes requested):
I'm thinking that the `.x` is redundant in cases where the end API is clearly Python: in particular, `dp.x.sklearn` is a bit strange. Can you add sklearn to the prelude directly? 

Then we could also move polars.py into extras. The prelude would then have `dp.polars` and `dp.sklearn`. Essentially, bespoke Python APIs get modules.

mccalluc:
> in particular, `dp.x.sklearn` is a bit strange. Can you add sklearn to the prelude directly?

I'll try. There are also enough changes here that it might be easier to start from a clean branch rather than resolving conflicts. TBD.

mccalluc:
One goal here is also to resolve
- #1798

but this PR doesn't do that... In fact, it makes the docs somewhat messier:

Screenshot 2024-07-17 at 2 17 33 PM

 and PCA is still not documented, because of:
```
class PCA(_SKLPCA):
    # TODO: If I add a docstring here, it also tries to locate _SKLPCA, and fails
```

@Shoeboxam
- I'll try working on this, but if we can't also fix #1798, is that a blocker on this PR? 
- I'll also try moving polars... but I'm not sure that needs to be in this PR, or could it be handled separately?

mccalluc:
I think these would be the next changes, but I'm not super confident here:
```
	modified:   python/src/opendp/_convert.py
	modified:   python/src/opendp/context.py
	renamed:    python/src/opendp/polars.py -> python/src/opendp/extras/polars.py
	modified:   python/src/opendp/extras/sklearn/__init__.py
	modified:   python/src/opendp/prelude.py
```

<details>

```diff
diff --git a/python/src/opendp/_convert.py b/python/src/opendp/_convert.py
index d1a649e24..d233c87b3 100644
--- a/python/src/opendp/_convert.py
+++ b/python/src/opendp/_convert.py
@@ -150,7 +150,7 @@ def c_to_py(value: Any) -> Any:
             query_type = RuntimeType.parse(queryable_query_type(value))

             if query_type == "OnceFrameQuery":
-                from opendp.polars import OnceFrame
+                from opendp.extras.polars import OnceFrame
                 return OnceFrame(value)

             return Queryable(value, query_type)
diff --git a/python/src/opendp/context.py b/python/src/opendp/context.py
index 272cb1af8..3e4126355 100644
--- a/python/src/opendp/context.py
+++ b/python/src/opendp/context.py
@@ -51,7 +51,7 @@ from opendp.mod import (
 )
 from opendp.typing import RuntimeType
 from opendp._lib import indent, import_optional_dependency
-from opendp.polars import LazyFrameQuery, Margin
+from opendp.extras.polars import LazyFrameQuery, Margin
 from dataclasses import asdict


@@ -176,7 +176,7 @@ def domain_of(T, infer: bool = False) -> Domain:
     if infer:
         pl = import_optional_dependency("polars", raise_error=False)
         if pl is not None and isinstance(T, pl.LazyFrame):
-            from opendp.polars import _lazyframe_domain_from_schema
+            from opendp.extras.polars import _lazyframe_domain_from_schema

             return _lazyframe_domain_from_schema(T.collect_schema())

diff --git a/python/src/opendp/polars.py b/python/src/opendp/extras/polars.py
(venv) (1432-extrinsics-to-extras) ~/github/opendp/opendp $ git diff --cached | cat
diff --git a/python/src/opendp/_convert.py b/python/src/opendp/_convert.py
index d1a649e24..d233c87b3 100644
--- a/python/src/opendp/_convert.py
+++ b/python/src/opendp/_convert.py
@@ -150,7 +150,7 @@ def c_to_py(value: Any) -> Any:
             query_type = RuntimeType.parse(queryable_query_type(value))

             if query_type == "OnceFrameQuery":
-                from opendp.polars import OnceFrame
+                from opendp.extras.polars import OnceFrame
                 return OnceFrame(value)

             return Queryable(value, query_type)
diff --git a/python/src/opendp/context.py b/python/src/opendp/context.py
index 272cb1af8..3e4126355 100644
--- a/python/src/opendp/context.py
+++ b/python/src/opendp/context.py
@@ -51,7 +51,7 @@ from opendp.mod import (
 )
 from opendp.typing import RuntimeType
 from opendp._lib import indent, import_optional_dependency
-from opendp.polars import LazyFrameQuery, Margin
+from opendp.extras.polars import LazyFrameQuery, Margin
 from dataclasses import asdict


@@ -176,7 +176,7 @@ def domain_of(T, infer: bool = False) -> Domain:
     if infer:
         pl = import_optional_dependency("polars", raise_error=False)
         if pl is not None and isinstance(T, pl.LazyFrame):
-            from opendp.polars import _lazyframe_domain_from_schema
+            from opendp.extras.polars import _lazyframe_domain_from_schema

             return _lazyframe_domain_from_schema(T.collect_schema())

diff --git a/python/src/opendp/polars.py b/python/src/opendp/extras/polars.py
similarity index 99%
rename from python/src/opendp/polars.py
rename to python/src/opendp/extras/polars.py
index 786b6426d..8a820e6bc 100644
--- a/python/src/opendp/polars.py
+++ b/python/src/opendp/extras/polars.py
@@ -1,5 +1,5 @@
 '''
-The ``opendp.polars`` module adds differential privacy to the
+The ``polars`` module adds differential privacy to the
 `Polars DataFrame library <https://docs.pola.rs>`_.

 For convenience, all the members of this module are also available from :py:mod:`opendp.prelude`.
@@ -8,6 +8,8 @@ We suggest importing under the conventional name ``dp``:
 .. code:: python

     >>> import opendp.prelude as dp
+
+The methods of this module will then be accessible at ``dp.polars``.
 '''
 from __future__ import annotations
 from dataclasses import dataclass
@@ -39,7 +41,7 @@ class DPExpr(object):

     >>> import polars as pl
     >>> pl.len().dp
-    <opendp.polars.DPExpr object at ...>
+    <opendp.extras.polars.DPExpr object at ...>
     '''
     def __init__(self, expr):
         """Apply a differentially private plugin to a Polars expression."""
diff --git a/python/src/opendp/extras/sklearn/__init__.py b/python/src/opendp/extras/sklearn/__init__.py
index 7e8862583..3078c5f12 100644
--- a/python/src/opendp/extras/sklearn/__init__.py
+++ b/python/src/opendp/extras/sklearn/__init__.py
@@ -1 +1,14 @@
+'''
+The ``opendp.sklearn`` module offers differentially private analogs for some of the tools of
+`scikit-learn <https://scikit-learn.org>`_.
+
+For convenience, all the members of this module are also available from :py:mod:`opendp.prelude`.
+We suggest importing under the conventional name ``dp``:
+
+.. code:: python
+
+    >>> import opendp.prelude as dp
+
+The methods of this module will then be accessible at ``dp.sklearn``.
+'''
 from .pca import PCA
\ No newline at end of file
diff --git a/python/src/opendp/prelude.py b/python/src/opendp/prelude.py
index 6a0763218..aff5102f4 100644
--- a/python/src/opendp/prelude.py
+++ b/python/src/opendp/prelude.py
@@ -36,7 +36,7 @@ For example:
 '''

 from opendp.mod import *
-from opendp.extras import sklearn
+from opendp.extras import sklearn, polars
 import opendp.extras as x
 import opendp.transformations as t
 import opendp.measurements as m
@@ -48,6 +48,5 @@ from opendp.typing import *
 from opendp.accuracy import *
 from opendp.core import new_function, new_queryable
 from opendp.context import *
-from opendp.polars import *

-__all__ = ["sklearn", "x", "t", "m", "c"]
+__all__ = ["sklearn", "polars", "x", "t", "m", "c"]
```

</details>

Shoeboxam:
That change seems right to me. This only partly resolves #1798.

mccalluc:
If this much seems coherent, great! Moved out of draft, and requested re-review.

Shoeboxam (changes requested):

python/src/opendp/prelude.py
```diff
@@ -49,4 +50,4 @@
 from opendp.context import *
 from opendp.polars import *
 
-__all__ = ["t", "m", "c"]
+__all__ = ["sklearn", "x", "t", "m", "c"]
```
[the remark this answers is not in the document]
mccalluc: I've been a bit uncomfortable with our existing short namespaces, and would prefer to spell things out. `np` and `pl` are the conventional import names for their respective libraries, but they are just conventions. 

Otherwise, having them accessible from the prelude seems fine to me:

```
dp.numpy.array2_domain
dp.sklearn.PCA
dp.polars.DPExpr
```

Thoughts?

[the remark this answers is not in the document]
Shoeboxam: This makes sense. Let's do this for extras.

python/src/opendp/extras/__init__.py
```diff
@@ -0,0 +1,7 @@
+'''
+The ``extras`` module is a namespace for functionality that requires extra, optional Python dependencies.
+'''
+from .domains import np_array2_domain
+from .make_np_clamp import make_np_clamp
+from .make_np_pca import make_private_np_pca, make_private_np_mean
```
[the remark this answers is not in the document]
mccalluc: I'm going to try just deleting this `__init__.py`, and moving everything down into the appropriate subdirectory: `sklearn/`, `numpy/`, or `polars/`.

mccalluc:
@Shoeboxam : Thanks for the nudge: This is more coherent, and I think could be merged now, though there are rough edges.
- PCA still doesn't have any real API docs, but at least now there's a link from where the API docs would be to the tutorial that was merged earlier.
- For simplicity's sake, I just moved all the `make_np_*` methods  to the `numpy/` folder... though many of them might better belong to `sklearn/`? Would it be confusing to have a `np` function outside the `numpy` directory?
- Moving Polars: I think this much makes sense, more would be moved in here, and will be done as a follow-up, in any event.

mccalluc:
@Shoeboxam , I think this much is looking good. Skimmed the docs and fixed a couple typos, too. I'm going to create a branch from this for polars.

python/src/opendp/extras/numpy/make_np_clamp/__init__.py
```diff
@@ -49,7 +62,7 @@ def function(arg):
     return dp.t.make_user_transformation(
         input_domain,
         input_metric,
-        dp.np_array2_domain(**kwargs),
+        dp.numpy.np_array2_domain(**kwargs),
```
Shoeboxam, line 65: If we change the path to introduce `numpy`, then `np_` in APIs is now redundant. I'd rather we drop `np_` from any APIs under `numpy` now, instead of changing APIs twice.

python/src/opendp/extras/sklearn/pca.py
```diff
@@ -1,22 +1,35 @@
...
+        '''
         def __init__(*args, **kwargs):
             raise ImportError(
                 "The optional install scikit-learn is required for this functionality"
             )
 
 
-class PCA(SKLPCA):
+class PCA(_SKLPCA):
+    # TODO: If I add a docstring here, it also tries to locate _SKLPCA, and fails
```
Shoeboxam, line 32: This is the same issue as in LazyFrameQuery from Polars, where we have an API inheriting from an optional dependency. It's not clear to me which approach is better, but it would be best if both places ultimately solve the problem the same way.

Doesn't have to be resolved at this point.

Shoeboxam (changes requested):
A few comments on APIs.

mccalluc:
From Mike on Slack:

> I thought about it for a bit and I think the only API changes would be:
> - np_array2_domain --> array2_domain
> - _np_sscp_domain --> _sscp_domain
> 
> No matter what we do, we should make sure to not have any flavor of dp.array_domain at this time (at root of prelude), but having a dp.numpy.array2_domain would be good.

Shoeboxam:
Do the API docs render more nicely if you leave the clamp module private (`_make_np_clamp/__init__.py`) and add `from _make_np_clamp import make_np_clamp` to `extras/numpy/__init__.py`?

python/src/opendp/extras/__init__.py
```diff
@@ -0,0 +1,3 @@
+'''
+The ``extras`` module is a namespace for functionality that requires extra, optional Python dependencies.
```
mccalluc, line 2: Looking at #1782, I'm reminded that what defined `_extrinsics` was not the dependence on an extra install, it was that it was pure-python code: It just happened that the examples we had depended on outside libraries.

I'd suggest that the module description just be updated:
```suggestion
The ``extras`` module contains pure-python DP tools.
These are not yet supported by the Rust or R bindings.
 Some tools require extra, optional Python dependencies, and are grouped under corresponding namespaces.
```

mccalluc (commented):
Reviewed
- #1782 

and I realize that I hadn't scoped `extras` this correctly. New text proposed.

mccalluc:
> Do the API docs render more nicely if you leave the clamp module private (`_make_np_clamp/__init__.py`) and add `from _make_np_clamp import make_np_clamp` to `extras/numpy/__init__.py`?

@Shoeboxam : no, that won't work. Sphinx docs aren't showing all the names available in a module, only the ones defined there. In a way, this makes sense: Python doesn't have any special syntax for re-exporting names, so if a file has `import numpy as np` you could import `np` from that file, though you wouldn't want to do that!

One thing we could do is move any public functions out of their standalone files and into the `__init__.py`, and if there are eventually any proof documents for public functions, they can also be moved up a level. I'll make a commit to try that out, but if the status quo is referable, that's fine by me: Could revert back to this state.

mccalluc:
make_np_clamp is up a level now: I think this makes for a more idiomatic API, but I don't feel strongly about it. If this is the right direction, should something similar be done for PCA? If it's not the right direction, I'll revert.

mccalluc:
What the docs look like with this change:
Screenshot 2024-07-30 at 3 32 50 PM

Shoeboxam (approved):
Thanks for the iterations. Looks much better now.

PR #1453, opened 2024-04-01
Check function signature documentation for consistency

### LINKED ISSUE #1395: Check that docstrings agree with signatures
Looking at `compositor` now, and the docs are out of sync with the signature.

It looks like there is an option in [pylint](https://pylint.pycqa.org/en/latest/user_guide/checkers/extensions.html#parameter-documentation-checker) that can check this. I'm not seeing other options, but do a bit more research.

I'm also not sure about interactions between flake8 and pyflake. Is this something useful enough that we would run it in CI? Off hand, I think pyflake advertises itself as slower but more scrupulous.

Another option is not to run it in CI on every commit, but only nightly, or something like that.

mccalluc:
- Fix #1395

draft until the new check passes.

### DIFF
--- a/python/.pylintrc
+++ b/python/.pylintrc
@@ -1,0 +1,685 @@
+[MAIN]
+
+# Analyse import fallback blocks. This can be used to support both Python 2 and
+# 3 compatible code, which means that the block might have code that exists
+# only in one or another interpreter, leading to false positives when analysed.
+analyse-fallback-blocks=no
+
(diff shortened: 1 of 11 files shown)


mccalluc:
I don't think this is the right direction: Adding empty `:type:` tags causes the typing information we do have in the docs to be hidden. Also better not to add another library if we can avoid it.

PR #1457, opened 2024-04-01
Warn if large priv loss

### LINKED ISSUE #1363: Context API: Raise a warning if privacy loss is large
I think the best place to raise the warning is in `dp.loss_of`, and add another argument to hide warnings (`warn=False`)?

Threshold based on rule of thumb:
* epsilon no greater than 1.
* delta no greater than 1e-6
* rho no greater than 0.25

Threshold based on large values:
* epsilon no greater than 5.
* delta no greater than 1e-6
* rho no greater than 0.5

I think we should only warn based on the thresholds for large values.

A scrappy script that I haven't prettified:

```python
import opendp.prelude as dp
dp.enable_features("contrib")

    data=[1] * 100,
    privacy_loss=dp.loss_of(epsilon=5., delta=1e-6),
    split_evenly_over=1
)
subcontext = context.query().compositor(split_evenly_over=1, output_measure=dp.zero_concentrated_divergence(T=float)).release()

query = subcontext.query().clamp((0, 1)).sum().gaussian()
m_gauss = dp.m.make_gaussian(dp.atom_domain(T=float), dp.absolute_distance(T=float), query.param())
print(m_gauss.map(1.))
print(dp.c.make_zCDP_to_approxDP(m_gauss).map(1.).epsilon(delta=1e-6))
```

mccalluc:
- Fix #1363

### DIFF
--- a/python/src/opendp/context.py
+++ b/python/src/opendp/context.py
@@ -242,7 +242,11 @@ def loss_of(epsilon=None, delta=None, rho=None, U=None) -> Tuple[Measure, float]
     def range_warning(name, value, info_level, warn_level):
         if value > warn_level:
             if info_level == warn_level:
-                logger.warning(f'{name} should be less than {warn_level}')
+                if name == 'delta':
+                    logger.warning(f'{name} should be equal to or less than {warn_level}')
+                else:
+                    logger.warning(f'{name} should be less than {warn_level}')
+
             else:
                 logger.warning(f'{name} should be less than {warn_level}, and is typically less than {info_level}')
         elif value > info_level:
--- a/python/src/opendp/context.py
+++ b/python/src/opendp/context.py
@@ -2,6 +2,7 @@
 The ``context`` module provides :py:class:`opendp.context.Context` and supporting utilities.
 '''
 
+import logging
 from typing import Any, Callable, List, Optional, Tuple, Union
 import importlib
 from inspect import signature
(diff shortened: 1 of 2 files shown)


python/src/opendp/context.py
```diff
@@ -235,15 +239,25 @@ def loss_of(epsilon=None, delta=None, rho=None, U=None) -> Tuple[Measure, float]
     if epsilon is None and delta is not None:
         raise ValueError("Epsilon must be specified if delta is given.")
 
+    def range_warning(name, value, info_level, warn_level):
+        if info_level < value <= warn_level:
+            logger.info(f'{name} is typically less than {info_level}')
+        if warn_level < value:
+            logger.warning(f'{name} should be less than {warn_level}, and is typically less than {info_level}')
```
raprasad, line 246: For delta, the info_level and warn_level are the same which makes the phrasing slightly awkward.
```
delta should be less than 1e-06, and is typically less than 1e-06
``

Add an additional message if the info_level and warn_level are the same

**Review by raprasad — DISMISSED** [review]
Approving, can address comment if desired

**Comment by mccalluc** [comment]
Last change introduced special handling for delta:
```
                if name == 'delta':
                    logger.warning(f'{name} should be equal to or less than {warn_level}')
                else:
                    logger.warning(f'{name} should be less than {warn_level}')
```
CI fails because of coverage. I'll fix.

raprasad (approved):
approved

PR #1458, opened 2024-04-01
links between user guide and api reference

### LINKED ISSUE #973: Links from examples to API
Problem: Examples shouldn’t be derailed by a complete explanation of the API… but the user might want to learn more. For example, the Polars documentation offers links to the API from the Examples – Not sure if these are automatically generated, or encoded in the documentation.

Fix: Not sure. Would be best to find a tool that already exists that can extract the information and generate the correct paths. This might be even harder if we’re trying to do it inside a notebook renderer.

See also:
- Links in the other direction – #448

### LINKED ISSUE #1388: Mention the `import opendp.prelude as dp` idiom at the top of every module doc

mccalluc:
- Fix #1388 (Replaces #1445)
- Fix #973 (could do more, but enough for now)
- Link from every module to a corresponding section of the user guide
- Link from many sections of the user guide to corresponding modules
- And special boilerplate for our three special modules.
- Make the anchors more consistent.

### DIFF
--- a/docs/source/api/user-guide/context/index.rst
+++ b/docs/source/api/user-guide/context/index.rst
@@ -20,7 +20,6 @@ it can raise an exception and prevent you from running more queries than your bu
     .. code:: python
 
       >>> import opendp.prelude as dp
-      >>> dp.enable_features("contrib")
       >>> context = dp.Context.compositor(
       ...     data=[5.0] * 100,
       ...     privacy_unit=dp.unit_of(contributions=1),
--- a/docs/source/api/user-guide/context/index.rst
+++ b/docs/source/api/user-guide/context/index.rst
@@ -20,12 +20,19 @@ it can raise an exception and prevent you from running more queries than your bu
 
     .. code:: python
 
+      >>> from typing import List
       >>> import opendp.prelude as dp
       >>> dp.enable_features("contrib")
       >>> context = dp.Context.compositor(
-      ...     data=[1, 2, 3],
-      ...     privacy_unit=dp.unit_of(contributions=3),
-      ...     privacy_loss=dp.loss_of(epsilon=3.0),
+      ...     data=[1.0, 2.0, 3.0, 4.0, 5.0],
+      ...     privacy_unit=dp.unit_of(contributions=1),
+      ...     privacy_loss=dp.loss_of(epsilon=1.0),
       ...     split_evenly_over=1,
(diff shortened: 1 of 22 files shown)


mccalluc:
We want something more than just TODO in the context user guide. Back to draft.

mccalluc:
Trying to get this example working I hit a number of usability problems. I'm going to make a small PR that just focuses on usability issues in this API.

docs/source/api/user-guide/programming-framework/supporting-elements.rst
```diff
@@ -287,6 +297,6 @@ Practically speaking, the smaller the ``d_out``, the tighter your analysis will
 You might find it surprising that metrics and measures are never actually evaluated!
 The framework does not evaluate these because it only needs to relate a user-provided input distance to another user-provided output distance.
 Even the user should not directly compute input and output distances:
-they are :ref:`solved-for <determining-accuracy>`, :ref:`bisected <parameter-search>`, or provided by the Context API.
+they are :ref:`solved-for <accuracy-user-guide>`, :ref:`bisected <parameter-search>`, or provided by the Context API.
```
raprasad, line 300: @mccalluc : does it make sense to link the "Context API" at the end of the sentence

raprasad (approved):
Nice improvements. (Added one optional-to-fix comment)

docs/source/api/user-guide/programming-framework/supporting-elements.rst
mccalluc: Sure: will do.

docs/source/api/user-guide/context/index.rst
```diff
@@ -20,6 +20,7 @@ it can raise an exception and prevent you from running more queries than your bu
     .. code:: python
 
       >>> import opendp.prelude as dp
+      >>> dp.enable_features("contrib")
```
mccalluc, line 23: @raprasad : If our testing is working correctly, this code sample ran without `dp.enable_features("contrib")` -- do we want to include it for the sake of consistency? Or do you think this example just should not work without `contrib` enabled? 

If the latter case, either the test framework itself is broken, or else the context code should add a check for `contrib`.

raprasad (approved):
approved

Issue #1463, opened 2024-04-02
Checklist for april release


mccalluc:
- [x] #1460 
- [x] fix nightly build #1464
~~- [ ] polars API (no public documentation)~~
- [x] #1462

PR #1464, opened 2024-04-02
fix nightly docs build

mccalluc:
- I believe it was broken by changes in #1450
- Bring back all the extensions that that commit removed.
- Towards #1463


mccalluc:
- Just started a dry run of the docs workflow: https://github.com/opendp/opendp/actions/runs/8554825873

Shoeboxam (approved):
Tricky. I guess if we do get into a situation where an extension is not maintained/working, we can always condition the extensions list based on the version number.

PR #1466, opened 2024-04-02
Add R example to "typical workflow"

### LINKED ISSUE #1287: Factor R example runner out into standalone script

### LINKED ISSUE #1460: R tabs on typical-workflow.html
Python framework examples are there, so this should be possible.

mccalluc:
- fix #1460
- fix #1287

Draft for now.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -235,8 +235,11 @@ jobs:
         working-directory: R/opendp/
         run: Rscript -e 'devtools::test(reporter = c("summary", "fail"))'
 
-      - name: Documentation examples
+      - name: Run sphinx examples
         run: python tools/r_doc_examples.py
+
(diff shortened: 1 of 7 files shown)


mccalluc:
Works locally.

The number of differences between the APIs make me hesitant to suggest to anyone that it's easy to translate code between the languages. At the same time,  I also don't think it does anyone much good to try to list all the differences, and expect the user to keep them straight.

raprasad (approved):
approved

PR #1472, opened 2024-04-04
More typing on context

mccalluc:
I had started this a while ago and then forgot about it. Just some more type annotations.


Shoeboxam (approved):

mccalluc:
Seemed like it should be easy, but there is a mypy error. Will investigate:
```
src/opendp/context.py:273: error: Incompatible return value type (got "tuple[Measure, float | None]", expected "tuple[Measure, float | tuple[float, float]]")  [return-value]
src/opendp/context.py:277: error: Incompatible return value type (got "tuple[Measure, tuple[float | None, float]]", expected "tuple[Measure, float | tuple[float, float]]")  [return-value]
Found 2 errors in 1 file (checked 75 source files)
```

python/src/opendp/context.py
```diff
@@ -243,11 +247,6 @@ def loss_of(epsilon=None, delta=None, rho=None, U=None) -> Tuple[Measure, Union[
     :param U: The type of the privacy parameter; Inferred if not provided.
 
     """
-    if epsilon is None and rho is None:
-        raise ValueError("Either epsilon or rho must be specified.")
-    if epsilon is None and delta is not None:
-        raise ValueError("Epsilon must be specified if delta is given.")
-
```
Shoeboxam, line 250: Why remove these checks? It is valid for `rho` to be the only argument.

python/src/opendp/context.py
```diff
@@ -262,6 +261,9 @@ def range_warning(name, value, info_level, warn_level):
         U = RuntimeType.parse_or_infer(U, rho)
         return zero_concentrated_divergence(T=U), rho
 
+    if epsilon is None:
+        raise ValueError("Either epsilon or rho must be specified.")
```
Shoeboxam, line 265: Won't this raise if I call it like `loss_of(rho=.5)`?

Shoeboxam (dismissed):

python/src/opendp/context.py
mccalluc: Unless I'm confused, no: If `rho` is set we'll return zero_concentrated_divergence just above.

python/src/opendp/context.py
mccalluc: The prompt for doing this is that the annotation additions in the earlier commit in the PR were causing mypy failures later in this function. The checks here are a little too complicated, and mypy was thinking that `None`s could come through. After replacing these with just one guard below, mypy is happy.

raprasad (approved):
Approved.

raprasad (approved):
approved

PR #1480, opened 2024-04-08
r-docs artifact: We were uploading r-docs with v2, downloading with v4

mccalluc:
I didn't file a new issue for currently nightly failure.

Haven't dug into the git history to understand what went wrong, but the fix seems simple. Confirmed that with this commit every `actions/upload-artifact` is v4.


raprasad (approved):
approved

Issue #1476, opened 2024-04-05
Investigate converting RST to notebooks


mccalluc:
We're planning for more of our examples and documentation to be tabs in RST, rather than notebooks:
- The source code is is easier to read and maintain.
- We have a good pattern for multi-language examples.
- At least in Python, the output checks provided by doctest are stronger than what we have in notebooks.

At the same time, if users want to run our code, notebooks are better than copy and pasting code, particularly if there's an option to open with one click in Binder, or perhaps Colab or Kaggle in the future.

We should investigate tools that can generate ipynb from rst, and see if they can be incorporated into our doc build process. One challenge is that we don't want to just extract all the codeblocks: For python, we'd like to just capture the Context API examples, and the Framework API example should be a parallel notebook, if included at all.

It looks like there are two options ([SO summary](https://stackoverflow.com/questions/76151653/convert-restructured-text-to-jupyter-notebook)):
- https://github.com/QuantEcon/sphinxcontrib-jupyter/ (More popular, but last commit 4 years ago)
- https://github.com/nthiery/rst-to-ipynb (older, but there was a commit six months ago?)

mccalluc:
Was hoping there was a third option: Just using pandoc.

Without any changes, the includes are not handled. [Fix is to use RST built-ins instead of Sphinx add-ons.](https://github.com/jgm/pandoc/issues/6631)

Change
```
        .. literalinclude:: code/typical-workflow-context.rst
            :language: python
```
to
```
        .. include:: code/typical-workflow-context.rst
            :code: python
```

but this still isn't everything we need: the include is processed, but it shows up as a triple-backtick block inside a markdown cell, instead of being a separate code cell.

mccalluc:
Investigate is done. Not sure if we'll use it, but there is a helper script that encapsulates what I learned. Closing.

PR #1483, opened 2024-04-10
util script for RST to NB

mccalluc:
- Towards #1476

Not sure this is the right course, but wanted to put this much out.

One possibility is to just let this PR wait, and even if it is merged, only use it to manually create notebooks, rather than incorporating it into the build.

Or we could work with this as a skeleton, and if tweaks are needed in the future, do that with pandoc filters operating on the AST.


raprasad (approved):
approved

PR #1484, opened 2024-04-10
Just add a note to explain duplication

### LINKED ISSUE #1477: Duplicate sphinx apidoc config
In the Makefile:
```
	    $(SPHINXAPIDOC) -f -F -e -d 3 -H "OpenDP" -A "The OpenDP Project" -V $(VERSION) \
  -o source/api/python ../python/src/opendp --templatedir source/_templates
```
In conf.py:
```
sphinx_apidoc_cmd = 'sphinx-apidoc -f -F -e -H "OpenDP" -A "The OpenDP Project" -V $VERSION \
  -o source/api/python ../python/src/opendp --templatedir source/_templates'
```

I think I added the `-d 3` recently to control of the depth of the ToC: should it be in conf.py too? Is there some way we can avoid this duplication? Is one of these used in CI and the other locally?

mccalluc:
- Fix #1477 

Pulling this out to a standalone script that they both reference seems like overkill, but this doesn't seem great either.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -28,6 +28,7 @@ clean:
 
 html:
 	@set -e
+	# If sphinx-apidoc options change, also update conf.py!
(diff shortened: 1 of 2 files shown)


raprasad (approved):
approved

PR #1486, opened 2024-04-10
Minor typing edit: float implies int

### LINKED ISSUE #1485: Simplify annotations for numbers: float implies int
https://peps.python.org/pep-0484/#the-numeric-tower:

> when an argument is annotated as having type float, an argument of type int is acceptable

So annotations like this can be simplified:

```diff
- predicate: Callable[[Union[float, int]], bool]
+ predicate: Callable[[float], bool]
```

I feel like this came up before, but not seeing the issue.

mccalluc:
- Fix #1485

### DIFF
--- a/python/src/opendp/mod.py
+++ b/python/src/opendp/mod.py
@@ -962,8 +962,8 @@ def binary_search(
 
 
 def exponential_bounds_search(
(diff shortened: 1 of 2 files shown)


raprasad (approved):
approved

PR #1494, opened 2024-04-12
R examples in docs via stand-alone files

### LINKED ISSUE #1490: R documentation examples
- To complement #1439

Figure out something analogous for R.

mccalluc:
- fixes #1490

Had some problems at first: notes below.

<details>

The machinery essentially works, but the example itself fails, at least as it is being run in my environment.

https://rstudio.github.io/r-manuals/r-exts/Writing-R-documentation-files.html explains that examples are run by default:
> `\examples{...}`
>
> Examples of how to use the function. Code in this section is set in typewriter font without reformatting and is run by example() unless marked otherwise (see below).

And what we see in the generated documentation isn't pretty:

`opendp/R/opendp/docs/reference/atom_domain.html`:
> Examples
> ```
> atom_domain(.T = "i32")
> #> Error in .Call("domains__atom_domain", bounds, nullable, .T, rt_parse(.T.bounds),     log, PACKAGE = "opendp"): "domains__atom_domain" not available for .Call() for package "opendp"
> ```

https://r-pkgs.org/man.html#sec-man-examples has a lot of information on writing examples, and I haven't read it yet.

Filing as draft.
- The example needs to work.
- If it doesn't, there should be a CI failure, and I'm not sure that there is one right now.

</details>


mccalluc:
In CI, `devtools::check("R/opendp")` should run examples, and I see this in the log:
```
✔  checking examples (552ms)
```
Debug strategy:
- Add an example that is definitely an error, and re-run check: Maybe it's missing the example?
- If it catches other errors, then there must be something different about the docs build environment.

raprasad (approved):
approved

PR #1499, opened 2024-04-16
explain docs build in each target language

mccalluc:
- Realized this needed to be cleaned up while working on #1490
- Add a section for R docs
- Explain that Rust stands outside the CI process.


raprasad (approved):
approved

Issue #943, opened 2023-10-02
Investigate when Python constructors can do dtype coercion


Shoeboxam:
Now that dtypes are typically filled in via partial chaining, we may be able to relax the checks on certain int vs float arguments.

Integers can be coerced to floats, and sometimes floats can be coerced into integers.

mccalluc:
@Shoeboxam -- Can you help me map out what this work entails? Are you envisioning more conversion happening in `py_to_c`, with a corresponding loosening of the checks in `check_similar_scalar`? Or would the work be done somewhere else?

Shoeboxam:
@mccalluc Exactly. More work in py_to_c, as that function already has the expected dtype given to it.

Shoeboxam:
Closed in #1509

PR #1509, opened 2024-04-18
cast int to float if needed in `py_to_c`

### LINKED ISSUE #1468: Context API Improvement: Laplace throws unclear error with integer data
from @mccalluc 

## Problem

Reference:
  - https://github.com/opendp/opendp/pull/1465/files#diff-d4d89fbd8d4c766e30502bed20068b70616b587d73d472f6eabb6da6967bfb03R22
  
```Python
@pytest.mark.xfail(raises=TypeError)
def test_int_data_laplace_param():
    # Currently fails with:
    #   TypeError: inferred type is i32, expected f64. See https://github.com/opendp/opendp/discussions/298
    # Possible resolution:
    #   Explicit parameter on laplace works with int data, or the error message should suggest the fix.
        data=[1, 2, 3, 4, 5],
        split_evenly_over=1,
    )
    sum_query = context.query().clamp((1, 10)).sum()
    sum_query.laplace(100)
```
  
## Possible resolution:
- [ ] Explicit parameter on laplace works with int data, or the error message should suggest the fix.
- [ ] Update test referenced above

mccalluc:
- Towards #943 (not sure if there are other coercions we'd also like to see?)
- Fix #1468

Not super confident about this approach, but `check_similar_scalar` really seems to be the center of things: This gets called from many different points, so doing it outside this function seemed like a piecemeal solution.

This does fix `test_int_data_laplace_param` and I've added a lower level conversion test, but there is a change in behavior in binary search, but maybe that's ok?

Draft for now, but feedback welcome.

### DIFF
--- a/python/src/opendp/_convert.py
+++ b/python/src/opendp/_convert.py
@@ -39,13 +39,33 @@ _ERROR_URL_298 = "https://github.com/opendp/opendp/discussions/298"
 
 def _check_and_cast_scalar(expected, value):
     '''
-    
-    If expected is not one of a small number of Rust types:
+    Any number is cast to float, if f32 or f64 is expected:
     >>> _check_and_cast_scalar('f32', 1.0)
-    ???
-    >>> _check_and_cast_scalar('f32', 1)
-    ???
+    1.0
+    >>> _check_and_cast_scalar('f64', 1)
+    1.0
 
+    Ints cast to ints, unsurprisingly:
+    >>> _check_and_cast_scalar('u8', 1)
+    1
+
+    ... but there are range checks:
+    >>> _check_and_cast_scalar('u8', 256)
+    Traceback (most recent call last):
+    ...
+    ValueError: 256 is not representable by u8
+
+    Floats cannot be cast to ints:
+    >>> _check_and_cast_scalar('i32', 1.0)
+    Traceback (most recent call last):
+    ...
+    TypeError: inferred type is f64, expected i32. See https://github.com/opendp/opendp/discussions/298
+
+    Unrecognized types will fail, but note that the error message refers to "f64":
+    >>> _check_and_cast_scalar('fake', 1)
+    Traceback (most recent call last):
+    ...
+    TypeError: inferred type is f64, expected fake. See https://github.com/opendp/opendp/discussions/298    
     '''
     inferred = RuntimeType.infer(value)
     
--- a/python/src/opendp/_convert.py
+++ b/python/src/opendp/_convert.py
@@ -37,7 +37,16 @@ INT_SIZES = {
 _ERROR_URL_298 = "https://github.com/opendp/opendp/discussions/298"
 
 
-def check_and_cast_scalar(expected, value):
+def _check_and_cast_scalar(expected, value):
+    '''
+    
+    If expected is not one of a small number of Rust types:
+    >>> _check_and_cast_scalar('f32', 1.0)
+    ???
+    >>> _check_and_cast_scalar('f32', 1)
+    ???
+
+    '''
     inferred = RuntimeType.infer(value)
     
     if inferred in ATOM_EQUIVALENCE_CLASSES:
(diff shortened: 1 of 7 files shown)


mccalluc:
Error is coming from [R build](https://github.com/opendp/opendp/actions/runs/8765450324/job/24056618201#step:5:5217)... probably not from this PR?

<details>

```
  Error: 
  ! error in pak subprocess
  Caused by error: 
  ! `path` must exist
  ---
  Backtrace:
  1. pak::lockfile_install(".github/pkg.lock")
  2. pak:::remote(function(...) { …
  3. err$throw(res$error)
  ---
  Subprocess backtrace:
   1. base::withCallingHandlers(cli_message = function(msg) { …
   2. get("lockfile_install_internal", asNamespace("pak"))(...)
   3. plan$install()
   4. pkgdepends::install_package_plan(plan, lib = private$library, num_workers = nw, …
   5. base::withCallingHandlers({ …
   6. pkgdepends:::start_task(state, task)
   7. pkgdepends:::start_task_build(state, task)
   8. pkgdepends:::make_build_process(path, pkg, tmp_dir, lib, vignettes, needscompilation, …
   9. pkgdepends:::withr_with_libpaths(c(tmplib, lib), action = "prefix", pkgbuild::pkgbuild_…
  10. base::force(code)
  11. pkgbuild::pkgbuild_process$new(path, tmp_dir, binary = binary, …
  12. local initialize(...)
  13. pkgbuild:::rcb_init(self, private, super, path, dest_path, binary, vignettes, …
  14. pkgbuild:::build_setup(path, dest_path, binary, vignettes, manual, clean_doc, …
  15. base::stop("`path` must exist", call. = FALSE)
  16. | base::.handleSimpleError(function (e) …
  17. global h(simpleError(msg, call))
  Execution halted
```

</details>

Can file an issue if we see this again. Will update and re-run CI.

raprasad:
(leaving open until @Shoeboxam has chance to review)

Shoeboxam:
The reason why it's taken so long to review this PR is because I find the new code to be very complicated, and I feel that it could be simpler, but I don't know what to recommend off-hand. I've tinkered with it some myself, but haven't had uninterrupted time to take a good look.

mccalluc:
@Shoeboxam - Can you say if the complexity that is most concerning is in `check_and_cast_scalar` or in the usages, or both? If it is `check_and_cast_scalar`, I could make the diff simpler by restoring the original if/then, though there was some redundancy there that I was also trying to clean up.

Another thing I could do would be to add a comment explaining the logic of the code before this change, and that could also help explain the new flow.

python/src/opendp/_convert.py
```diff
@@ -36,18 +36,24 @@ def c_int_limits(type_name):
 }
 _ERROR_URL_298 = "https://github.com/opendp/opendp/discussions/298"
 
-def check_similar_scalar(expected, value):
+
+def check_and_cast_scalar(expected, value):
```
raprasad, line 40: @mccalluc : based on your last comment, can add a docstring to this function. Thanks

python/src/opendp/_convert.py
```diff
@@ -36,18 +36,24 @@ def c_int_limits(type_name):
...
+        (not inferred_in_eq_class and expected != inferred)
+    ):
+        if not isinstance(value, int):
             raise TypeError(f"inferred type is {inferred}, expected {expected}. See {_ERROR_URL_298}")
-    else:
-        if expected != inferred:
-            raise TypeError(f"inferred type is {inferred}, expected {expected}. See {_ERROR_URL_298}") 
-
+        value = float(value)
+        value = check_and_cast_scalar(expected, value)
```
raprasad, line 51: Is this a "double check" after the cast?

raprasad (dismissed):
I think this is fine if comments are added

python/src/opendp/_convert.py
mccalluc: will do.

Shoeboxam:
* **#1509** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1509?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Shoeboxam:
@mccalluc I've approved this, but I've made changes myself. If you also approve, then you can trigger the merge.

Shoeboxam (approved):

mccalluc:
Thanks for cleaning this up: Much clearer!

PR #1511, opened 2024-04-18
provide specialized message if we can say what the fix is

### LINKED ISSUE #1470: Context API Improvement: Better message if domain is incorrect
from @mccalluc 

## Problem

Reference:
  - https://github.com/opendp/opendp/pull/1465/files#diff-d4d89fbd8d4c766e30502bed20068b70616b587d73d472f6eabb6da6967bfb03R72
  
  ```Python
@pytest.mark.xfail(raises=TypeError)
def test_scalar_instead_of_vector():
    # Currently fails with:
    #   TypeError: inferred type is Vec<i32>, expected i32. See https://github.com/opendp/opendp/discussions/298
    # Possible resolution:
    #   Would the domain ever not be a Vector? Maybe we check first for that, and give a more precise error.
        data=[1, 2, 3, 4, 5],
        split_evenly_over=1,
        domain=dp.domain_of(int),
    )
    sum_query = context.query().clamp((1, 10)).sum()
    sum_query.laplace()
  ```
  
## Possible resolution:
- [ ] Better error handling
- [ ] Update test referenced above

mccalluc:
- Fix #1470

We can't imagine everything the user could do wrong, but when we can anticipate problems, and give feedback without adding too much cruft to our code, seems worth considering.

Maybe the error message should just be replaced instead of being appended?


raprasad (approved):
approved

PR #1512, opened 2024-04-18
Steer users in the right direction if they try to call a domain descriptor

### LINKED ISSUE #1510: Usability: mismatch between reprs and utility method constructors
Ideally, a repr is close to an actual python expression that can be evaluated to obtain the object. A user might think that an object represented as:
```
VectorDomain(AtomDomain(T=i32))
```
could be constructed with something like:
```
dp.VectorDomain(dp.AtomDomain(T=dp.i32))
```
but they'd be wrong!
```
*** TypeError: 'DomainDescriptor' object is not callable
```
They actually want 
```
dp.vector_domain(dp.atom_domain(T=dp.i32))
```
Is there a way to steer them onto the right course?

This is probably overthinking it, but would it be helpful to define a `__call__` method, just to raise an exception?

----

- Alternatively, could `DomainDescriptors` actually be used as the constructors, and we'd just get rid of `dp.vector_domain` etc.? Probably not feasible, but it might be the least surprising arrangement.
- There's also something slightly non-standard about the way we use the term "constructor" -- I think a stricter definition would distinguish between constructors (called using the class name) and factory methods (which create new objects, but may hide from the caller the details of exactly what class is being instantiated.)

mccalluc:
- Fix #1510

A proposal: It does hide a problem that could be detected by mypy and instead waits till runtime, and I feel like I nixed something like that earlier when you proposed it.


raprasad (approved):
approved

PR #1521, opened 2024-04-22
add context, remove raise_error kwarg

### LINKED ISSUE #1520: optional dependency context
This might be a little cleaner:
```diff
- pl = import_optional_dependency("polars", raise_error=False)
- if pl is not None:
+ with optional_dependency("polars") as pl:
```

mccalluc:
- Fix #1520

If the optional install isn't so optional, use `import_optional_dependency` and get a good error message; If it really is optional, use `with optional_dependency`, and it will just skip the body if not found.


mccalluc:
context managers can't skip code. Closing.

PR #1523, opened 2024-04-23
add polars feature to instructions

mccalluc:
- Without this, python tests fail at the collection phase
- If we wanted to be sure these didn't fall out of sync again, I could imagine:
  - pulling the `cargo build` in the RST out into an include.
  - treating that include as a shell script in CI, so docs and CI would reflect the same features.
  - ... but probably overkill?

### DIFF
--- a/docs/source/contributing/development-environment.rst
+++ b/docs/source/contributing/development-environment.rst
@@ -48,7 +48,7 @@ Now run ``cargo build`` in the ``rust`` subdirectory of the repo:
 .. code-block:: bash
 
     cd rust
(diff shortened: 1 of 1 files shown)


Shoeboxam (changes requested):
Add to vscode-suggested as well?

mccalluc:
Change of direction: We want to use `--all-features` pretty much everywhere and explain the exceptions. But that is failing right now on `sample_discrete_laplace_Z2k`
- See #1525

Closing in favor of that issue.

PR #1524, opened 2024-04-23
fix return signature on loss_of

### LINKED ISSUE #1411: Correct return annotation on `loss_of`

mccalluc:
- Fix #1411
- Replace #1424: It got too complicated


raprasad (approved):
approved

PR #1526, opened 2024-04-24
use `--all-features`

### LINKED ISSUE #1525: Should `cargo test --all-features` work? Should we use it?
Right now we reference different subsets of features in CI and in developer docs:
```
      - name: Test
        run: cargo test --verbose --features untrusted,ffi,polars
```
```
    cargo build --features untrusted,bindings
...
Substitute ``cargo build`` with ``cargo test`` to test
```
Should we be able to use `--all-features` instead? Currently failing locally:
```
error[E0599]: no function or associated item named `sample_discrete_laplace_Z2k` found for type `f64` in the current scope
   --> src/traits/samplers/discretize/mod.rs:292:27
    |
292 |             .map(|_| f64::sample_discrete_laplace_Z2k(shift, scale, -1074))
    |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^ function or associated item not found in `f64`

error[E0599]: no function or associated item named `sample_discrete_gaussian_Z2k` found for type `f64` in the current scope
   --> src/traits/samplers/discretize/mod.rs:306:27
    |
306 |             .map(|_| f64::sample_discrete_gaussian_Z2k(shift, scale, -1074))
    |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ function or associated item not found in `f64`
```
It looks like we could also define a `default`, if there's a reason not to use all.

mccalluc:
- will fix #1525

WIP:
- test runs now produce vega graphs -- If we don't want these as a part of every run, maybe use skip instead of feature gating?
- Still need to handle `tasks.json`
- `Makevars*` and `rust_build.sh` parameterize `--features` ... Maybe we still want the flexibility there?


mccalluc:
I'm going to say this much is ready for review. There are more references to `--features` but I'll address that in a separate PR, to keep this much from getting muddled.

```
R/opendp/src/Makevars.in:
  30: 		@BEFORE_CARGO_BUILD@ cargo build --manifest-path rust/Cargo.toml --color always --release --jobs 2 --offline --features @RUST_FEATURES@; \

R/opendp/src/Makevars.win.in:
  34: 		@BEFORE_CARGO_BUILD@ cargo build --manifest-path rust/Cargo.toml --target=$(TARGET) --color always --release --jobs 2 --offline --features @RUST_FEATURES@; \

tools/rust_build.sh:
  86:   CMD+=(--manifest-path=rust/Cargo.toml --features="$FEATURES")
```

mccalluc:
[CI failure in r-tests](https://github.com/opendp/opendp/actions/runs/8833489592/job/24253800196#step:10:153): Also saw the same problem on `main` and filed an issue:
- #1531

The last commit here only touched `tasks.json`, so the r-tests problem is non-deterministic, even though it seems to be related to the build. Exciting.

Shoeboxam (approved):
I was wondering about the plots, glad they're ignored. Thanks, one less thing to maintain.

PR #1528, opened 2024-04-24
udf -> plugin

### LINKED ISSUE #1527: Rename "UDF" to "plugins"
... although, it looks like we don't actually say "UDF" or "user defined function" in the docs. Would these renamings make sense?
```diff
- user_domain
+ plugin_domain
- user_divergence
+ plugin_divergence
- user_transformation
+ plugin_transformation
```
etc.

- Prompted by #1508

mccalluc:
- Fix #1527

Should be sufficient.

### DIFF
--- a/python/test/test_interactive.py
+++ b/python/test/test_interactive.py
@@ -74,7 +74,7 @@ def test_sequential_composition_approxdp():
     sc_qbl(sum_meas)
 
 
-def test_udf_queryable_int():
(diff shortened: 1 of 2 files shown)


raprasad:
(Adding to Polars milestone in that the term "plugin" is aligned with Polars lib)

raprasad (approved):
approved

PR #1532, opened 2024-04-25
use `--all-features` in build

mccalluc:
- Complements #1526

My guess is that this isn't quite what we want (for example, I think this would include the `test` feature), but the solution may be to define a default set that is defined in one place rather than trying to keep sets in sync across the codebase


mccalluc:
Not needed: R will not have polars features immediately, so we still need `--features` there. Closing.

PR #1533, opened 2024-04-25
Move rust tests to standalone files

### LINKED ISSUE #1513: Move inline Rust test modules to separate files
If unit tests are located in a separate, adjacent test.rs file, then changes to the tests won't trigger a re-compilation of the crate, making test recompilation faster.

mccalluc:
With tests in separate files, they code they test won't need to be rebuilt as often.

- Fix #1513 
- Branched from #1526: Review that first

Draft


mccalluc:
This is most of it:
```
$ egrep 'cfg.*test' -r src -A 1 | grep 'mod.rs' | grep -v cfg | grep -v 'mod test;' | wc -l
       9
$ git checkout main
Switched to branch 'main'
Your branch is up to date with 'origin/main'.
$ egrep 'cfg.*test' -r src -A 1 | grep 'mod.rs' | grep -v cfg | grep -v 'mod test;' | wc -l
      63
```

- I narrowed the scope to only do this refactor on `mod.rs` files (with the exception of `error.rs` which I happened to do first). If we carry it through to all files, then it's effectively saying that every file should be `mod.rs` in its own directory, and that might be too much of a good thing.
- Along the way, I renamed to `test`, although that causes some problems: there are some public modules with utilities, that are then `*` imported, and we would now have name conflicts.
- In some cases there are multiple `[cfg(test)]` blocks with different sets of extra features. For now, I've just moved one over to `test.rs`.
- An easy mistake to make would be to clear out the test code in file, but not move it all to the new file, but I think the test count is the same here and in main: 216.

Since this touches so many files, it's likely to pick up merge conflicts ... but conversely, it might introduce merge conflicts in other PRs which are higher priority.

Shoeboxam (approved):
This is a pretty mechanical change. Might as well get it in before conflicts arise!

PR #1548, opened 2024-04-26
Move a few more rust tests to standalone files

mccalluc:
- Follow-up to #1533
- At some point might distinguish public `test-helper.rs` from private `test.rs`, but this is sufficient for now.
- The are two impls guarded by `#[cfg(test)]`, but otherwise all of the test code has been moved out of the `mod.rs` files.


Shoeboxam (approved):
Looks good. At some point I was going to pull the test data set for polars out into somewhere shared, but that doesn't need to be in this PR.

Issue #1544, opened 2024-04-26
Python API reference: Examples for measurements


mccalluc:
- Enabled by #1439
- Prioritized by #1489

mccalluc:
Reopening:
- We want want longer examples (like #1557)
- And since it's all on the same page (unlike R), we will drop the `then_` examples, and users can follow the see-also.

mccalluc:
Filed lots of follow-ups. This can be closed.

PR #1550, opened 2024-04-26
Python measurement examples

mccalluc:
- Some python measurement examples.
- Needed to tweak the code gen for `then_` functions
- Add `dp` to the global scope for doctests, so we don't need to repeat it everywhere.
- Towards #1544: A couple remaining functions are more complicated.


raprasad (approved):
approved

PR #1554, opened 2024-04-29
`make_base_laplace_threshold` -> `make_laplace_threshold`

### LINKED ISSUE #1551: `make_base_laplace_threshold` -> `make_laplace_threshold`?
- follow-up to #1005

My sense is that the `base` is no longer needed. Was this function just overlooked?

mccalluc:
Seems like it could have been shortened at the same time as the others, but if this one really is different, never mind.

Some larger changes in generated code because of change in alphabetical order.

- Fix #1551


Shoeboxam (approved):
Was originally holding off because I thought I'd have time to do an iteration on this mechanism. Thanks for going ahead and making the naming change now for consistency.

PR #1558, opened 2024-04-30
Use features of Python 3.9

### LINKED ISSUE #1393: Drop support for Python 3.8 in October 2024?

mccalluc:
- Fix #1393


python/test/test_space_of.py
```diff
@@ -1,13 +1,12 @@
 import opendp.prelude as dp
-from typing import Dict, List
 
 def test_typed_space_of():
     # metric defaults to symmetric_distance on vector_domain
-    space = dp.space_of(List[int])  # can also do list[int] if python 3.8+
+    space = dp.space_of(list[int])  # can also do list[int] if python 3.8+
```
Shoeboxam, line 5: Comment is no longer necessary!

Shoeboxam (changes requested):
Approved, except for one small nit- a comment is no longer necessary.

Shoeboxam (approved):

PR #1559, opened 2024-04-30
Error if missing `arg` param on measurement in R

### LINKED ISSUE #1555: R usability: Measurements w/o `arg` should error, instead of just returning NULL
```
> gaussian <- make_gaussian(atom_domain(.T = f64), absolute_distance(.T = f64), scale = 1.0)
> gaussian(100.0)
> print(gaussian(100.0))
NULL
> gaussian(arg = 123.0)
[1] 123.2615
```
Translating from python to R I was stuck for a while because nothing was coming back.

Mike says:
> I think it would be better if an error were raised if the argument is not named. The reason why args are necessary is because they distinguish whether you're calling the function or map. I got feedback quite a few R experts that a more object-oriented API is not ideal

mccalluc:
- Fix #1555
- Started off with a more general solution that actually checked the parameters from the call frames... but that was complicated, and caused failures, I think because `|>` fills in `attr` as a positional parameter: ie, sometimes calling with positional parameters is ok.
- So instead just check if `attr` is numeric... This assumes that mistakenly unnamed `arg` will always be numeric... but I'm not confident that that's true?
- Does not address "Should also verify similar checks exist on other library primitives"... not sure that there is a general solution.

For reference, for checking whether params are named, I started from https://stackoverflow.com/questions/69644911/how-to-ensure-that-parameters-have-been-called-by-name-and-not-by-position and stripped it down to just `length(names(as.list(sys.call()))) != length(as.list(sys.call()))`... and that highlighted that we actually do need to support calling with positional params.


R/opendp/R/mod.R
```diff
@@ -158,6 +158,9 @@ new_measurement <- function(ptr, log) {
     if (!missing(arg)) {
       return(measurement_invoke(ptr, arg))
     }
+    if (is.numeric(attr)) {
+      stop("numeric attr not allowed; Did you mean 'arg='?")
+    }
```
Shoeboxam, line 163: This is nice QOL.

Shoeboxam (approved):
Thanks!

Shoeboxam (changes requested):
On further consideration, I think we should improve the error message `stop("unrecognized attribute")` to direct the user to specify an attr, and which attrs they may pass. In the case where `attr` is not a string, suggest the user specify either `arg` or `d_in`.

Similar adjustments can be made to transformation, as well as more helpful messages when an attr is not recognized for supporting elements.

I don't see much utility in the repeated `expect_error(meas(), "expected exactly one of attr, arg or d_in")`, I feel like that will just calcify the code.

mccalluc:
>  I think we should improve the error message stop("unrecognized attribute") to direct the user to specify an attr, and which attrs they may pass. In the case where attr is not a string, suggest the user specify either arg or d_in.

Your suggestion is that the error handling can be done there instead? I don't think that will work, because the `switch` statement is [weird](https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/switch):

> If the value of EXPR is not a character string it is coerced to integer. ... If the integer is between 1 and nargs()-1 then the corresponding element of … is evaluated and the result returned

```
> attr = 1.0; switch(attr, a = "apple", b = "banana", stop("unrecognized"))
[1] "apple"
> attr = 1L; switch(attr, a = "apple", b = "banana", stop("unrecognized"))
[1] "apple"
```

So an input validation needs to happen before the `switch`.

Shoeboxam (approved):
Just looked again. I'm happy with it now! Thanks for your patience.

raprasad (approved):
approved

PR #1563, opened 2024-05-01
add check for null measurement

### LINKED ISSUE #1556: R usability: Friendlier error message if objects from previous session reused
In an earlier session:
```
gaussian_mechanism <- make_gaussian(atom_domain(.T = f64), absolute_distance(.T = f64), 1.0)
```
exit, and start a new session:
```
> gaussian_mechanism(arg=123.0)
Error in measurement_invoke(ptr, arg) :
  could not find function "measurement_invoke"
```
after `devtools::load_all("R/opendp/")`
```
> gaussian_mechanism(arg=123.0)
Error in measurement_input_carrier_type(this) :
  [FFI] : null pointer: this
```
Mike says:
> interesting, hadn't thought of this. old objects should not work at this point. We could consider restoring them via the logger info if we wanted to support this, but I think a better error message would be sufficient for now

mccalluc:
- Fix #1556
```
Browse[1]> gaussian_mechanism(arg=123.0)
Error in measurement_invoke(ptr, arg) :
  Measurement is a null pointer. Reusing measurements from previous sessions is unsupported.
```
... but maybe this should be at a higher level? I'm also not sure about the best way of testing this: just calling `measurement_invoke` with a null pointer is contrived, and if the implementation changed, the test wouldn't necessarily keep up.

This approach to testing for null pointer is from https://stackoverflow.com/a/27350487/10727889


mccalluc:
@Shoeboxam -- This would be a change to the generated code, and would be applied to all the functions. Before I invest time in code generation, can you give your thoughts on this much?

Shoeboxam:
Do you think you could avoid the need for codegen changes but get the same errors by adjusting a lower-level helper function?

I would think the code paths would all pass through the sexp_to_* functions here: 
https://github.com/opendp/opendp/blob/main/R/opendp/src/convert_elements.c

mccalluc:
I myself feel less confident in the C code than in the R, but that doesn't mean the R is the best place. We can figure out the best direction when you're back.

mccalluc:
But if this could be addressed at a lower level, maybe that also means there's a way for this problem in Python? That would actually be a good thing, because I'm more confident about testing and debugging the Python API.

mccalluc:
Since the time this was originally filed, we've git ignored the generated files, so I'm much less concerned about keeping everything in sync. If I follow the steps in the original reproducer, I now see:
```
> gaussian_mechanism(arg=123.0)
Error in measurement_invoke(ptr, arg) :
  Got null pointer. Reusing objects from previous sessions is unsupported.
```

Shoeboxam:
* **#1563** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1563?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Shoeboxam:
Updated code now fails with the following message:
```
> meas("input_domain")
Error in measurement_input_domain(ptr) :
  Got null pointer. Reusing a measurement from a previous session is not supported.
```

mccalluc (commented):
Looks good to me. I'm the original author so I can't approve.

Shoeboxam (approved):

Issue #1553, opened 2024-04-27
alp_queryable freezes


mccalluc:
I'm trying code very similar to test_alp_histogram:
```
>>> import opendp.prelude as dp
>>> dp.enable_features("contrib")
>>> counter = dp.t.make_count_by(
...     dp.vector_domain(dp.atom_domain(T=str)),
...     dp.symmetric_distance(),
...     MO=dp.L1Distance[int])
>>> alp_queryable = dp.m.make_alp_queryable(
...     *counter,
...     scale=1.,
...     total_limit=24,
...     value_limit=24)
```
but for me, at that point it freezes and is unresponsive. I've waited a couple minutes.

I'm probably doing something wrong, but freezing is unusual.

mccalluc:
No, shouldn't freeze. Chuck can look into it more, or at least see if it reproduces.

mccalluc:
Still freezing for me. For context, this is based on `test_alp_histogram`, except that uses `then`, and I've confirmed that it works.
```
>>> import opendp.prelude as dp
>>> dp.enable_features('contrib')
>>> counter = dp.t.make_count_by(
...     dp.vector_domain(dp.atom_domain(T=str)),
...     dp.symmetric_distance(),
...     MO=dp.L1Distance[int])
>>>
>>> alp_meas = counter >> dp.m.then_alp_queryable(
...     scale=1.,
...     total_limit=24,
...     value_limit=24,
... )
```

Shouldn't make a difference / maybe there's another typo / but even if there is, freezing is not the expected behavior. I'll try to dig into this.

mccalluc:
This is weird: If I redefine the function, I can hit the breakpoint:
```
def make_alp_queryable(
    *args, **kwargs
) -> Measurement:
    breakpoint()
```
but a breakpoint at the top of the function doesn't work. (Although the `then_` construct calls it indirectly, and that works.)

I've pushed to github, and we'll see if the freeze is replicated there.

mccalluc:
Decided we should see if defining `__iter__` helps.

Mike says:
> Here's a smaller repro that might help: [*counter]. Is it safe to override __iter__?

PR #1564, opened 2024-05-01
Trying to diagnose why make_alp freezes

mccalluc:
Does the behavior reproduce on github CI?
- see #1553


Shoeboxam:
Here's a smaller repro that might help: `print([*counter])`

Shoeboxam:
Closing in favor of #1823

PR #1566, opened 2024-05-01
cancel old smoke-test CI runs

Shoeboxam:
If I ever make a mistake and submit my stack twice, the old batch of CI runs don't get canceled, meaning the CI runners are consumed on useless tasks for an extended time.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -1,5 +1,9 @@
 name: Smoke Test
 
+concurrency: 
+  group: ${{ github.workflow }}-${{ github.ref_name }}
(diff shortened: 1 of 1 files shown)


Shoeboxam:
* **#1566** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1566?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

mccalluc (approved):
Makes sense, but should it also be applied to the rust build, at least?

Shoeboxam:
@mccalluc Since it is located at the root level, I think this applies to all tasks in the workflow

PR #1575, opened 2024-05-02
Merge "then" examples into "make" and regenerate

### LINKED ISSUE #1570: For python examples, drop the `then_` examples
users can follow the see-also.

mccalluc:
- Fix #1570
- I'll file a separate PR that extends this by using the same snippet for both "then" and "make" docs: I'm not sure that either of us are sure about the best course.


Shoeboxam (approved):
Great! Thanks.

raprasad (approved):
approved

Issue #1340, opened 2024-02-29
Sporadic R null pointer errors in tests


mccalluc:
[R tests failed](https://github.com/opendp/opendp/actions/runs/8085390731/job/22093043633?pr=1177) with the following error, but no error when re-running. Tests should be deterministic.
```
══ Failed ══════════════════════════════════════════════════════════════════════
── 1. Error ('test-mod.R:8:34'): binary search ─────────────────────────────────
Error: Error in `make_chain_mt(rhs, lhs)`: [FFI] : null pointer: measurement1
Error: Error: Failures detected.
Execution halted
disabled backtrace
Backtrace:
     ▆
  1. ├─opendp::binary_search_chain(...) at test-mod.R:8:5
  2. │ ├─opendp (local) make_chain(...) at opendp/R/mod.R:507:3
  3. │ │ └─opendp::then_laplace(t_sum, s) at test-mod.R:8:34
  4. │ │   ├─opendp:::new_constructor_log(...) at opendp/R/measurements.R:1032:5
  5. │ │   │ └─opendp::new_hashtab(...) at opendp/R/mod.R:431:3
  6. │ │   └─opendp::new_hashtab(list("scale", "QO"), list(scale, .QO)) at opendp/R/mod.R:405:3
  7. │ └─opendp::binary_search_param(...) at opendp/R/mod.R:405:3
  8. │   └─opendp::binary_search(...) at opendp/R/mod.R:525:3
  9. │     └─opendp (local) predicate(mid) at opendp/R/mod.R:584:5
 10. │       └─make_chain(param)(d_in = d_in, d_out = d_out) at opendp/R/mod.R:526:5
 11. │         └─opendp::measurement_check(ptr, d_in, d_out) at opendp/R/mod.R:155:9
 12. │           └─opendp::measurement_input_distance_type(measurement) at opendp/R/core.R:54:5
 13. │             ├─opendp:::new_constructor_log(...) at opendp/R/core.R:133:5
 14. │             │ └─opendp::new_hashtab(...) at opendp/R/mod.R:431:3
 15. │             └─opendp::new_hashtab(list("this"), list(this)) at opendp/R/mod.R:405:3
 16. └─opendp::make_chain_mt(rhs, lhs) at opendp/R/mod.R:405:3

══ DONE ════════════════════════════════════════════════════════════════════════
Error: Process completed with exit code 1.
```

mccalluc:
Another occurrence:

```
[...]

PR #1576, opened 2024-05-02
Use the same examples for `make` and `then`

mccalluc:
- Extends #1575
- I think it could be helpful to some users to have the examples in both places, but either is fine: Your call.


mccalluc:
Test failure in R:
```
── 1. Error ('test-measurements.R:73:5'): test_gaussian_search ─────────────────
Error in `make_fix_delta(make_zCDP_to_approxDP(then_gaussian(input_space, 
    scale)), delta)`: [FFI] : null pointer: measurement
disabled backtrace
Backtrace:
    ▆
 1. └─opendp::binary_search_param(...) at test-measurements.R:79:3
 2.   └─opendp::binary_search(...) at opendp/R/mod.R:525:3
 3.     └─opendp (local) predicate(mid) at opendp/R/mod.R:584:5
 4.       └─opendp (local) make_chain(param) at opendp/R/mod.R:526:5
 5.         └─opendp (local) make_smd_gauss(s, 1e-05) at test-measurements.R:80:5
 6.           └─opendp::make_fix_delta(...) at test-measurements.R:73:5
```
Seems to work after restart. 
- Might be the same as #1340

raprasad (approved):
approved

PR #1579, opened 2024-05-02
speed up smoke-test

Shoeboxam:
* -3 Don't run rust tests twice (these got re-introduced in the rust-build task from the --all-features pr)
* -12 Don't run the disk cleaner (not actually seeing CI fail now, and it didn't used to take this long. leaving commented)
* -3 Don't vendor dependencies (they're not used by the r tests)
* -5 Don't run rust doctests (just the 31 doctests, many of which are internal, take 5 minutes on their own)
    * Perhaps we can move rust doctests out to unit tests in future work

This reduces smoke-test from 31 minutes to 7 minutes.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -21,6 +21,7 @@ jobs:
       CARGO_TERM_COLOR: always
     steps:
 
+      # We're no longer running out of space, but freeing disk space uses time, so disable. Re-enable if needed.
       # - name: Free disk space
       #   uses: jlumbroso/free-disk-space@v1.3.1
       #   with:
@@ -89,6 +90,7 @@ jobs:
       CARGO_TERM_COLOR: always
     steps:
       
+      # We're no longer running out of space, but freeing disk space uses time, so disable. Re-enable if needed.
       # - name: Free disk space
       #   uses: jlumbroso/free-disk-space@v1.3.1
       #   with:
@@ -114,7 +116,7 @@ jobs:
           restore-keys: ${{ runner.os }}-cargo-
 
       - name: Test
-        run: cargo test --verbose --all-features --lib
+        run: cargo test --verbose --all-features
         
 
   rust-examples:
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -21,16 +21,16 @@ jobs:
       CARGO_TERM_COLOR: always
     steps:
 
-      - name: Free disk space
-        uses: jlumbroso/free-disk-space@v1.3.1
-        with:
-          tool-cache: false     # no: will remove rust
-          android: true         # 8.7GiB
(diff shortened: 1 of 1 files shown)


Shoeboxam:
* **#1579** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1579?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

.github/workflows/smoke-test.yml
```diff
@@ -21,16 +21,16 @@ jobs:
...
-        uses: jlumbroso/free-disk-space@v1.3.1
-        with:
-          tool-cache: false     # no: will remove rust
-          android: true         # 8.7GiB
-          dotnet: true          # 1.6GiB
-          docker-images: true   # 3.2GiB
-          haskell: false        # no: little/no benefit
-          large-packages: false # no: slow
-          swap-storage: false   # no: having swap is useful
+      # - name: Free disk space
```
mccalluc, line 24:
```suggestion
      # We're no longer running out of space, but freeing disk space uses time, so disable. Re-enable if needed.
      # - name: Free disk space
```

.github/workflows/smoke-test.yml
```diff
@@ -101,27 +89,16 @@ jobs:
...
```
mccalluc, line 92:
```suggestion
      # We're no longer running out of space, but freeing disk space uses time, so disable. Re-enable if needed.
      # - name: Free disk space
```

.github/workflows/smoke-test.yml
```diff
@@ -137,7 +114,7 @@ jobs:
           restore-keys: ${{ runner.os }}-cargo-
 
       - name: Test
-        run: cargo test --verbose --features untrusted,ffi,polars
+        run: cargo test --verbose --all-features --lib
```
mccalluc, line 117:
```suggestion
        run: cargo test --verbose --all-features
```
I'm uncomfortable disabling doctests entirely right now. Maybe hold off on this part?
- Are there particular doctests that are using a lot of time, or is it simply that doctests in general are slow to run? Sometimes just a couple tests dominate.
- If they are dropped from smoke test, could we make sure they are run nightly? Might already be the case.
- Do we want to discourage doctests going forward? Or only use them on high visibility functions? Or make sure that single doctests cover a lot of ground, to make up for the compilation cost?

mccalluc (changes requested):
Really uncomfortable dropping the tests... Maybe something else covers this logic, but I don't think we know.

(But overall, thanks for looking into this!)

Shoeboxam:
Maybe things will speed up in the future with https://github.com/rust-lang/rust/issues/75341.

In the meantime I might open a PR that moves some doctests to unit tests.

mccalluc (approved):
Thanks!

PR #1581, opened 2024-05-02
`make_private_lazyframe` example

mccalluc:
Started with the example in the tests, and then
- inlined the utility functions
- gave the columns meaningful names
- added some narrative
- changed details so we don't reference the private data till the very end,
- and formatted the private data for readability

Expecting most of the test matrix to fail because polars is not available; need to think about how to make doctests conditional.


mccalluc:
Possible fix: At the end of the module, add something like:
```
try:
    import polars
except ImportError:
    make_private_lazyframe.__doc__ = ''
```
I've done this locally, and it seems to work.
- (I think) sphinx parses the file without executing it, so it sees the docstring and renders the example.
- (I think) pytest executes the file, so the docstring is cleared if polars is not installed

This would need to be part of the code generation: It feels weird, but relatively simple.

Alternatively, less weird, but more complex, register a custom flag for doctest (via pytest) to let it skip tests that depend on polars. More code, and harder to trace the path, but less hacky?

raprasad (approved):
approved

Issue #1502, opened 2024-04-16
Sphinx fails on type annotations for optional dependencies


Shoeboxam:
I noticed that even if I guard the Polars import...

```python
if TYPE_CHECKING:
    import polars
```

...Sphinx still fails to document.

I've continued to leave types off for now.

_Originally posted by @Shoeboxam in https://github.com/opendp/opendp/pull/767#discussion_r1559973331_

mccalluc:
This is in generated code, and is also touched by:
- #1507

Once that's merged, and there's hopefully no longer an `Any`, see about adding a more precise type instead in the code generation. Oh hold for now.

mccalluc:
Looking back at the original PR, this was the context:
```
def new_arrow_array(
    name: str
):
    r"""Allocate an empty ArrowArray and ArrowSchema that Rust owns the memory for.
    The ArrowArray and ArrowSchema are initialized empty, and are populated by the bindings language.
```
Originally, the return type was `Any`, which didn't seem useful: I was expecting the return to be something like `ArrowArray`... but that's not what this function does: It's actually returning `opendp._lib.FfiSlicePtr`... which is an implementation detail that users shouldn't need to see, and this is an internal function in any case.

But I think there's a different concern, which is documenting the return types on public polars extensions, and that's probably still useful.

mccalluc:
Locally, my sphinx build works, and then I got off the rails, but learned some things.

I made this change:
```diff
-    lazyframe,
+    lazyframe: polars.LazyFrame,
```
and the docs rendered locally, though `LazyFrame` is just italicized.

Would it be possible to link to the Polars docs? In principle, yes: [intersphinx](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html) can be configured with URLs for 3rd party library documentation sites, but it relies on a generated index file, and the index for polars is [incomplete](https://github.com/pola-rs/polars/issues/7027). We could reference a local file, but it's a [binary](https://sphobjinv.readthedocs.io/en/stable/cli/index.html) and I can't imagine adding that complexity to our build process is worth it right now: If someone wants to know more about polars they'll just have to navigate over their on their own, instead of having a generated link from our docs.

PR #1587, opened 2024-05-03
add annotations for polars objects

mccalluc:
- Towards #1502
- Expect to fail at least because because of change to generated code, but interested in whether other errors will arise as well.

Draft for now.


mccalluc:
Closing old half-baked PR to clear the slate.

PR #1600, opened 2024-05-09
Polars: boolean binary operations

Shoeboxam:

PR #1597, opened 2024-05-08
Polars: separate public module

Shoeboxam:

PR #1596, opened 2024-05-07
Polars: add Gaussian, generalize Laplace

Shoeboxam:

PR #1592, opened 2024-05-04
Polars: horizontal stack

Shoeboxam:

PR #1602, opened 2024-05-10
Polars: OnceFrame

Shoeboxam:
Currently in Polars, Measurements emit LazyFrames. LazyFrames can be repeatedly executed, and each time it is executed, the privacy loss increases (but is not tracked anywhere). 

This PR instead returns a queryable with the LazyFrame in its internal state. When the user wants to execute the LazyFrame, the queryable executes the LazyFrame and then frees it, making it impossible to execute again.

A queryable is used so that the user can control how the LazyFrame is executed: either collect into memory, sink to a CSV, sink to Parquet, to IPC or to JSON. Queryables may also share the query plan, or allow the user to configure query plan optimizations.

In both the Python and Rust APIs, the Queryable emitted mimics a (very) limited subset of the LazyFrame API. We can expand this API in the future as needed.


Shoeboxam:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/opendp/opendp/1602?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1602** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1602?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1597** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1597?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1596** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1596?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1592** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1592?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1591** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1591?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1534** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1534?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1600** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1600?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1539** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1539?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1540** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1540?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1586** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1586?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1590** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1590?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1584** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1584?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1542** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1542?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1498** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1498?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1503](https://github.com/opendp/opendp/pull/1503) <a href="https://app.graphite.dev/github/pr/opendp/opendp/1503?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1479** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1479?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1496](https://github.com/opendp/opendp/pull/1496) <a href="https://app.graphite.dev/github/pr/opendp/opendp/1496?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#908** <a href="https://app.graphite.dev/github/pr/opendp/opendp/908?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#913](https://github.com/opendp/opendp/pull/913) <a href="https://app.graphite.dev/github/pr/opendp/opendp/913?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#890** <a href="https://app.graphite.dev/github/pr/opendp/opendp/890?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#856](https://github.com/opendp/opendp/pull/856) <a href="https://app.graphite.dev/github/pr/opendp/opendp/856?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#829** <a href="https://app.graphite.dev/github/pr/opendp/opendp/829?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2 other dependent PRs ([#854](https://github.com/opendp/opendp/pull/854) <a href="https://app.graphite.dev/github/pr/opendp/opendp/854?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#888](https://github.com/opendp/opendp/pull/888) <a href="https://app.graphite.dev/github/pr/opendp/opendp/888?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1519** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1519?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1506](https://github.com/opendp/opendp/pull/1506) <a href="https://app.graphite.dev/github/pr/opendp/opendp/1506?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1566** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1566?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

mccalluc:
I'll start a more thorough code review, but just skimming it, it's hard for me to see exactly what this means for the user. Would it be possible to update or copy
- #1632

and integrate that example into this PR to demonstrate the interface?

I also have some small PRs out that improve test coverage, particularly on generated code. Having those checks in place would make it easier for me to see what code here is actually being exercised.

Finally, I'd worry about holding up the release for this. When we do make a tagged release with this, I think we also need at least orphan docs for the new functionality.

For the code review, I'll try putting in more a lot more comments along the way. I'm not sure how this will work, but it might be good for me to be really transparent about what I understand and what I don't understand.

python/src/opendp/_data.py
```diff
@@ -25,6 +29,31 @@
 ]
 
 
+def _onceframe_extract_lazyframe(
```
mccalluc, line 32: This is used in turn by `_lazyframe`, but that function doesn't seem to have any usages. Do we either/both of these?

python/src/opendp/_data.py
```diff
@@ -252,6 +281,86 @@ def object_type(
...
+    lib_function = lib.opendp_data__onceframe_collect
+    lib_function.argtypes = [AnyObjectPtr]
+    lib_function.restype = FfiResult
+
+    output = c_to_py(unwrap(lib_function(c_onceframe), AnyObjectPtr))
+
+    return output
+
+
+def onceframe_sink_csv(
```
mccalluc, line 308: If we have a method to return a DataFrame, should that be enough? Is there a concern that going through an intermediate data structure will be too slow or inefficient for large datasets and we need this method and `onceframe_sink_parquet` to make it more direct? If that is the motivation, worry about premature optimization. What determines that these two formats are the ones we support?

python/src/opendp/mod.py
```diff
@@ -425,12 +426,6 @@ def eval(self, query):
         from opendp.core import queryable_eval # pragma: no cover
         return queryable_eval(self.value, query) # pragma: no cover
 
-    @property
-    def query_type(self):
-        from opendp.core import queryable_query_type
-        from opendp.typing import RuntimeType
-        return RuntimeType.parse(queryable_query_type(self.value))
```
mccalluc, line 432: Oh! It looks like you had a wrapper around `RuntimeType.parse` before, and are moving away from that. There's probably a good reason, but I don't understand the motivation.

python/src/opendp/polars.py
```diff
@@ -133,3 +133,24 @@ def median(self, candidates, scale=None):
...
+
+    def sink_csv(self, path):
+        from opendp._data import onceframe_sink_csv
+        onceframe_sink_csv(self.value, path)
+
+    def sink_parquet(self, path):
+        from opendp._data import onceframe_sink_parquet
+        onceframe_sink_parquet(self.value, path)
+
+    def _lazyframe(self):
```
mccalluc, line 154: Noted above, but I think this is unused?

python/src/opendp/prelude.py
```diff
@@ -47,5 +47,6 @@
 from opendp.accuracy import *
 from opendp.core import new_function, new_queryable
 from opendp.context import *
+import opendp.polars
```
mccalluc, line 50: I don't think any of the python code uses this, right now, so could leave it out of this PR? At one point, I think we were thinking of including this method and others under a new `dp.x` namespace, which would have all the methods which require optional installs?

python/test/test_polars.py
```diff
@@ -118,7 +118,7 @@ def test_private_lazyframe_explicit_sum(measure):
             pl.Series("A", [10.0] * 5),
         ]
     )
-    df_act = m_lf(lf).sort("B").collect()
+    df_act = m_lf(lf).collect()
```
mccalluc, line 121: Is it wrong to `sort` as a post processing step? Or maybe it just exercises more of the machinery to do it here? If it's not a good thing to do, that should get mentioned in the docs. (Though maybe it's already there?)

rust/src/data/ffi/mod.rs
```diff
@@ -8,14 +8,17 @@ use std::os::raw::c_char;
...
 
 #[cfg(feature = "polars")]
-use arrow::ffi::{ArrowArray, ArrowSchema};
+use ::polars::export::arrow;
 #[cfg(feature = "polars")]
-use polars::export::arrow;
+use ::polars::prelude::*;
 #[cfg(feature = "polars")]
-use polars::prelude::*;
+use arrow::ffi::{ArrowArray, ArrowSchema};
```
mccalluc, line 15: Reminder to self: `::polars` resolves ambiguity -- It specifies the external crate instead of the internal module.

mccalluc (commented):
Just some comments on the Python code so far. Not sure that all the functions that get exposed need to be exist, but you know the big picture better than me.

Later this afternoon I'll have another batch of comments on the Rust code.

Shoeboxam:
This PR makes the user experience worse-- users can only conduct a limited set of actions on a OnceFrame, instead of all possible actions with LazyFrame. However it prevents users from violating privacy guarantees by executing the plan multiple times. This should be merged before releasing 0.10 to ensure that the privacy guarantees we give are valid.

python/src/opendp/_convert.py
```diff
@@ -141,10 +141,20 @@ def c_to_py(value: Any) -> Any:
...
         from opendp._data import object_type, object_as_slice, slice_free
         obj_type = object_type(value)
+        
         if "SMDCurve" in obj_type:
             return SMDCurve(value)
+        
         if "Queryable" in obj_type:
-            return Queryable(value)
+            from opendp.core import queryable_query_type
+            query_type = RuntimeType.parse(queryable_query_type(value))
```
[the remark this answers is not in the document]
mccalluc: (I now see the change in `mod.py` below, so it looks like it used to be like that, so it's probably changing for a good reason?)

R/opendp/src/opendp.h
```diff
@@ -731,6 +731,41 @@ const struct FfiResult______AnyObject *ffiresult_err(char *message, char *backtr
  */
 bool opendp_data__fill_bytes(uint8_t *ptr, uintptr_t len);
 
+/**
+ * Internal function. Collects a DataFrame from a OnceFrame, exhausting the OnceFrame.
+ *
+ * # Arguments
+ * * `onceframe` - The queryable holding a LazyFrame.
+ */
+struct FfiResult_____AnyObject opendp_data__onceframe_collect(struct AnyObject *onceframe);
```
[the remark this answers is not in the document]
Shoeboxam: Yes, this should be blacklisted.

python/src/opendp/mod.py
Shoeboxam: The query type of a queryable will never change (it is static), so there's no need to repeatedly retrieve the query type at every call. Since I already need to know the query type to convey it into the correct Python type (Queryable or OnceFrame), I might as well simply the property to just an attribute.

python/src/opendp/polars.py
Shoeboxam: I adjusted this to make it public, and to assert "honest-but-curious".

python/src/opendp/_data.py
Shoeboxam: This is a great comment. That is a motivation, yes, but I also had another hope for OnceFrames that are not yet used by this PR. I think `trait PrivateLogicalPlan` shouldn't emit measurements with type `LogicalPlan`, as this also has the same problem as the `LazyFrame` type. `OnceFrame` better suits this. I was thinking this would be in a future PR.

Arguably we could just make `OnceFrame` internal, and only release data frames from `make_private_lazyframe`. However, grouping key sets can get larger than memory, so permitting these use-cases would be useful. One minor simplification here is that I could remove the auxiliary `sink_*` constructors, and just keep `collect` and the honest-but-curious `lazyframe`.

python/test/test_polars.py
Shoeboxam: The output of the measurement is a OnceFrame now, which has a more limited API. One of the additions in this PR is to allow sorting within `make_private_lazyframe`. Operations that are on what are now OnceFrames have been moved into the constructor.

mccalluc (commented):
Overall looks good. Many of the questions here parallel questions from the python review. Talk and decide on next steps?

rust/src/measurements/make_private_expr/mod.rs
```diff
@@ -33,7 +33,7 @@ mod expr_postprocess;
 pub(crate) mod expr_report_noisy_max_gumbel;
 
 #[bootstrap(
-    features("contrib"),
+    features("contrib", "honest-but-curious"),
```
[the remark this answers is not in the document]
Shoeboxam: Note to self: should document why the constructor is honest-but-curious.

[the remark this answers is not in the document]
Shoeboxam: Yes, file an issue from this comment-- maybe this should be a hand-written doc with an automated check?

python/src/opendp/polars.py
```diff
@@ -133,3 +133,24 @@ def median(self, candidates, scale=None):
...
+        from opendp._data import onceframe_collect
+        return onceframe_collect(self.value)
+
+    def sink_csv(self, path):
+        from opendp._data import onceframe_sink_csv
+        onceframe_sink_csv(self.value, path)
+
+    def sink_parquet(self, path):
+        from opendp._data import onceframe_sink_parquet
+        onceframe_sink_parquet(self.value, path)
```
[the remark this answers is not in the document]
Shoeboxam: I've removed the `sink_*` methods and associated machinery. For the time being, if someone wants it, they should call `.lazy()` with honest-but-curious first.

rust/src/measurements/make_private_lazyframe/mod.rs
```diff
@@ -98,6 +103,17 @@ where
         output_measure: MO,
         global_scale: Option<f64>,
     ) -> Fallible<Measurement<LogicalPlanDomain, LogicalPlan, MS, MO>> {
+        #[cfg(feature = "contrib")]
```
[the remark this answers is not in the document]
Shoeboxam: Maybe I'm being too forward-thinking, but I anticipate the removal of "contrib" to happen piecemeal as each individual PDF proof gets a green light.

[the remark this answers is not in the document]
Shoeboxam: I think this matcher doesn't need its own contrib flag, though. Admittedly somewhat arbitrary.

rust/src/data/ffi/polars.rs
```diff
@@ -0,0 +1,116 @@
...
+        onceframe(c_type = "AnyObject *", rust_type = "AnyQueryable"),
+        path(c_type = "const char *", rust_type = "String")
+    )
+)]
+/// Internal function. Sinks the data from a OnceFrame into a CSV file, exhausting the OnceFrame.
+///
+/// # Arguments
+/// * `onceframe` - The queryable holding a LazyFrame.
+#[no_mangle]
+pub extern "C" fn opendp_data__onceframe_sink_csv(
```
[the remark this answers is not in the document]
Shoeboxam: I narrowed down to just `collect` and `lazy`.

python/src/opendp/prelude.py
Shoeboxam: On second thought, I think this does make at least some sense. I added this because otherwise the plugin namespace won't get registered. You still can't access any imports from that module in the prelude.

rust/src/measurements/make_private_expr/mod.rs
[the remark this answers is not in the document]
mccalluc: - Filed #1662

mccalluc (commented):
Just a couple questions. After the tests pass, maybe just grab me and see if there are any last questions, and then I'll approve?

Shoeboxam:
I think it's ok that diffs for blacklisted functions are showing up in the R header file, because the generated header file is meant generally for any language for bind with, and not all the functions in a header file need to be used.

python/src/opendp/prelude.py
```diff
@@ -48,4 +48,8 @@
 from opendp.core import new_function, new_queryable
 from opendp.context import *
 
+# we don't want to make this module public, but do want the module setup to run
+import opendp.polars as _x_polars
+del _x_polars
```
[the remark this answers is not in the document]
Shoeboxam: This is no longer a problem after the PR restack.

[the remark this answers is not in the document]
Shoeboxam: The module should be public, it just shouldn't be accessible via `dp.opendp.polars`. That would be such a weird import.

mccalluc (approved):
Spot checking the earlier threads, looks good. Let me know if there's anything you'd like me to look at more carefully.

Make sure CI passes before merge? Not sure why it wasn't run on last commits.

PR #1612, opened 2024-05-15
For rust example in docs, use normal `cargo run`

### LINKED ISSUE #1601: CI fails because -Zscript now needs a hashbang
I think the nightly chains now require the following to run?
```
#!/usr/bin/env -S cargo +nightly -Zscript
```

Not sure, but CI is failing:
https://github.com/opendp/opendp/actions/runs/9020660022/job/24786298591?pr=1597

### LINKED ISSUE #1609: Unexpected drop in test count
https://github.com/opendp/opendp/actions/runs/9086573056/job/24972742426
```
    def test_number_of_tests_found(request):
        tests_found = len(request.session.items)
>       assert tests_found >= 308
E       assert 304 >= 308
```
Should review the recent commits to main: I remember recently raising the lower bound, but it's possible that a separate PR consolidated tests in a way that is fine, but should look through the history to confirm before lowering the bound again.

(And if this becomes a nuisance, we could drop it.)

mccalluc:
There's probably a way to get the `-Zscript` working, but nightly is inherently unstable. This adds a bit of clutter to the docs source, but it builds the example in a more standard way.

- Fix #1601 
- Fix #1609

(Draft until CI passes.)


raprasad (approved):
approved

PR #1613, opened 2024-05-15
1605 fix r failure in nightly

### LINKED ISSUE #1605: Nightly r-docs failing: maybe because OPENDP_LIB_DIR is undefined?
**Describe the bug**
Nightly is failing, and has been since last month, though it looks like the failure has changed over time, and is now near the end of the CI chain.

**To Reproduce**
[ci logs](https://github.com/opendp/opendp/actions/runs/9056941387/job/24881540577#step:5:30):
```
using C compiler: ‘gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0’
expected either a src/binary.tar.xz, src/source.tar.xz (run bash tools/r_stage.sh) or OPENDP_LIB_DIR to be set
make: *** [Makevars:16: rust_build] Error 1
ERROR: compilation failed for package ‘opendp’
* removing ‘/tmp/RtmpaKQVmk/devtools_install_17a51673c2d6/opendp’
Error in `(function (command = NULL, args = character(), error_on_status = TRUE, …`:
! System command 'R' failed
---
Exit status: 1
stdout & stderr: <printed>
---
Backtrace:
 1. devtools::document("R/opendp")
...
```

**More broadly...**

Do we need to move more of Nightly into the normal PR CI so that we don't find ourselves surprised by failures like this?

mccalluc:
- Fix #1605

Draft: Not entirely confident about this, but using the same  script in smoketest as we do in nightly seems like a good direction. (`r_stage.sh` also calls `pkgdown::build_site`, so there is a difference.)

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -29,6 +29,8 @@ jobs:
   r-docs:
     needs: credential-check
     runs-on: ubuntu-22.04
+    env:
+      OPENDP_LIB_DIR: ${{ github.workspace }}/libs
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -274,6 +274,9 @@ jobs:
           working-directory: R/opendp
           extra-packages: devtools, RcppTOML, lintr
 
+      - name: Build R docs (and run API examples)
+        run: bash tools/r_stage.sh -d
+
       # running the OpenDP install in a separate command (instead of local::. above)
       #   provides significantly more debug info if it fails
       - name: Install OpenDP
@@ -289,9 +292,6 @@ jobs:
 
       - name: Lint sphinx examples
         run: Rscript -e 'lintr::lint_dir("docs/source")'
-
-      - name: Document
-        run: cd R/opendp/ && Rscript -e 'devtools::document()'
 
       - name: Check
         run: cd R/opendp/ && Rscript -e 'devtools::check(error_on="warning")'


mccalluc:
Just started a manual run of doc.yml on this branch: https://github.com/opendp/opendp/actions/runs/9209139022

mccalluc:
Next step: What was the change when these failures began?

mccalluc:
Next steps:
- Make a clean checkout, and try the steps in `r-docs` locally: Does it reproduce the problem?
- Try running `docs.yml` against different versions. Although the `docs` job requires artifacts from earlier steps in `release.yml`, `r-docs` doesn't. 
  - One problem: Github only lets us run workflows on branches or tags, not arbitrary hashes... so we'd need to make branches or tags at points in history, and then run.

mccalluc:
R docs build does not work locally, either, though it's a different error. This is on `main`:
```
cd rust
rm -rf target
cargo build --all-features
cd ..
export OPENDP_LIB_DIR=`realpath rust/target/debug`
tools/r_stage.sh -d
```
```
   clang -arch arm64 -dynamiclib -Wl,-headerpad_max_install_names -undefined dynamic_lookup -L/Library/Frameworks/R.framework/Resources/lib -L/opt/R/arm64/lib -o opendp.so accuracy.o combinators.o convert.o convert_elements.o core.o data.o domains.o lib.o measurements.o measures.o metrics.o transformations.o libopendp.a -F/Library/Frameworks/R.framework/.. -framework R -Wl,-framework -Wl,CoreFoundation
   ld: warning: __eh_frame section too large (max 16MB) to encode dwarf unwind offsets in compact unwind table
   installing to /private/var/folders/jm/40wztt291610f8pk5w0sw20r0000gq/T/RtmpO6wO4M/devtools_install_d2de225f6311/00LOCK-opendp/00new/opendp/libs
   ** checking absolute paths in shared objects and dynamic libraries
─  DONE (opendp)
Registered S3 method overwritten by 'opendp':
  method        from
  print.hashtab utils
Error in dyn.load(dll_copy_file) :
  unable to load shared object '/var/folders/jm/40wztt291610f8pk5w0sw20r0000gq/T//RtmpO6wO4M/pkgloadd2de24de5b9f/opendp.so':
  dlopen(/var/folders/jm/40wztt291610f8pk5w0sw20r0000gq/T//RtmpO6wO4M/pkgloadd2de24de5b9f/opendp.so, 0x0006): symbol not found in flat namespace '_PyBaseObject_Type'
Calls: <Anonymous> ... <Anonymous> -> load_dll -> library.dynam2 -> dyn.load
Execution halted
```

mccalluc:
```
git checkout 6712fcd29 # BAD!
git checkout 97a5472f8 # GOOD!
git checkout c65042c2d # GOOD!
```
ok, double checking, but I think this is where it goes bad:
```
git checkout 6712fcd29 # still BAD!
```
That corresponds to 
- #829 

... and there's a lot in there, but no R changes.

mccalluc:
This has gone entirely off the rails as a PR, although the notes could be useful. Closing.

PR #1614, opened 2024-05-15
Chucks branch from polars tip plus updates from main

mccalluc:
I think Mike's usual pattern is to rebase the PR stack, but since I should get started with the documentation for features at the tip, and I don't think I should rebase his PRs, I've made this branch which just merges in updates from main. There shouldn't be any work here that doesn't originate elsewhere.

I'll make branches from this for my work, and probably take this as the base of any PRs, but that may change.


mccalluc:
Closing: Development has caught up with this. The two PRs that got merged here instead of against main have been redone:
- #1727
- #1695 

closing.

PR #1615, opened 2024-05-15
regenerate python to include example

mccalluc:
This PR:
- #1576

clobbered the API example from
- #1581 

and caused the tests to fail when merged. 

The longterm fix to avoid this problem is
- #1430 
 
though I have some ambivalence there.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -130,8 +130,8 @@ jobs:
     strategy:
       matrix:
         python-version:
-          - 3.9
-          - 3.12
+          - '3.9'
+          - '3.12'
         extra-install:
           - '.'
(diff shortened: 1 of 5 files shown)


mccalluc:
CI config to skip polars unless installed not working...

mccalluc:
huh: 3.12 on CI is also missing some lines -- not sure why I get more coverage locally.
```
Name                     Stmts   Miss  Cover   Missing
------------------------------------------------------
src/opendp/_convert.py     348      2    99%   197, 247
src/opendp/_lib.py         143      5    97%   108, 117-118, 134, 146
```

For now, I'll just add ignores so we can get CI passing.

raprasad (approved):
Approved.

Issue #1410, opened 2024-03-20
Polars MVP documentation


raprasad:
Include examples of using Polars for sums, means, groupby, and quantiles

raprasad:
(depending on time, this may move to the next, smaller release)

mccalluc:
In the API docs, `rust/src/measurements/code/make_private_lazyframe.rst` already has sums and groupby, adding mean wouldn't be hard, and with a data set this trivial, median might be sufficient to demonstrate quantiles?

raprasad:
closing in favor of multiple notebook issues

PR #1627, opened 2024-05-16
polars typical-workflow example

mccalluc:
I'm not sure right now how to translate some things like unit of privacy to the new context, and I have a local environment problem, but wanted to save this much.

- Towards #1410


mccalluc:
We decided typical-workflow should be non-polars, and instead emphasize parallels with R, and only use vector data, instead of the old dataframe methods.

PR #1632, opened 2024-05-20
examples of mean and median for Polars

### LINKED ISSUE #1410: Polars MVP documentation

mccalluc:
- Fix #1410: Add demo of mean and median
- Since it's getting repetitive, I've added some helper functions, though that is not usually my preference in doctests.
- I'm also not 100% confident on the math -- the means seem high -- I found one bad variable, but the numbers could be a fluke.


mccalluc:
This has been handled elsewhere. Closing.

PR #1634, opened 2024-05-20
Clean up network dependency in doctest

### LINKED ISSUE #1441: Remove network dependency from `typical-workflow-context.rst`
I've noticed this doctest being flakey, and it's probably because of:
```
>>> import urllib.request
>>> data_url = "https://raw.githubusercontent.com/opendp/opendp/sydney/teacher_survey.csv"
```
We should check if the file exists locally, and download if not available. (ie, someone has copy and pasted the code and tried to run it.)

In notebooks we have this idiom:
```
![ -e data.csv ] || wget ...
```
adapt?

mccalluc:
- Fix #1441

... but the symptom for which I thought the network request was the root cause is still there: On some local runs it just stalls on
```
getting-started/code/typical-workflow-context.rst
```

Also noticed a new problem: I think our CSV parser may not see the last line if it is not terminated with `\n`. If I make this replacement:
```diff
- data = 'age\n42\n'
+ data = 'age\n42'
```
it fails with this error:
```
opendp.mod.OpenDPException:
  MakeTransformation("row size must be greater than zero")
```
- Filed #1635

Since this PR doesn't actually fix the problem I was aiming it at, maybe it shouldn't go in... does make the docs a little kludgy. Your call.

### DIFF
--- a/docs/source/getting-started/code/typical-workflow-context.rst
+++ b/docs/source/getting-started/code/typical-workflow-context.rst
@@ -28,8 +28,9 @@
 # mediate
 >>> import urllib.request
 >>> data_url = "https://raw.githubusercontent.com/opendp/opendp/sydney/teacher_survey.csv"
->>> with urllib.request.urlopen(data_url) as data_req:
-...     data = data_req.read().decode('utf-8')
+>>> if data is None:
+...     with urllib.request.urlopen(data_url) as data_req:
+...         data = data_req.read().decode('utf-8')
 
 >>> context = dp.Context.compositor(
 ...     data=data,
--- a/docs/source/getting-started/code/typical-workflow-context.rst
+++ b/docs/source/getting-started/code/typical-workflow-context.rst
@@ -25,7 +25,7 @@
 # /public-info
 
 
->>> data = 'age\n42\n' # Minimal data so doctest can run without hitting network.
+>>> data = 'John Smith,M,42,,,,,,,,\n' # Minimal data so doctest can run without hitting network.
 
 # mediate
 >>> import urllib.request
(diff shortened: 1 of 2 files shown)


mccalluc:
CI failed in an [interesting way](https://github.com/opendp/opendp/actions/runs/9165218380/job/25198456656#step:7:318):
```
getting-started/code/quickstart-context.rst::quickstart-context.rst PASSED [ 97%]
getting-started/code/typical-workflow-context.rst::typical-workflow-context.rst PASSED [ 97%]
Fatal Python error: Aborted

Current thread 0x00007f7d47a27b80 (most recent call first):
  File "/home/runner/work/opendp/opendp/python/src/opendp/core.py", line 548 in queryable_eval
  File "/home/runner/work/opendp/opendp/python/src/opendp/mod.py", line 422 in __call__
  File "<doctest typical-workflow-framework.rst[20]>", line 1 in <module>
  File "/opt/hostedtoolcache/Python/3.9.19/x64/lib/python3.9/doctest.py", line 1334 in __run
  File "/opt/hostedtoolcache/Python/3.9.19/x64/lib/python3.9/doctest.py", line 1481 in run
```
Line 20 is here, so I'm not sure if that's the real problem:
```
# public-info
>>> col_names = [
```

mccalluc:
On [this run](https://github.com/opendp/opendp/actions/runs/9165218380/job/25199098347?pr=1634#step:7:340):
```
070 # mean
071 >>> mean_query = (
UNEXPECTED EXCEPTION: OpenDPException('MakeTransformation', 'row size must be greater than zero'
```
This error happened in `3.9, .[scikit-learn]`, but not in `3.12, .`: I don't think the optional dependencies should make any difference here, so is the problem non-deterministic?

I also realize that my fake data doesn't really represent the original structure, so I'll try adjusting that, though it's been good to exercise some edge cases.

mccalluc:
This is moot: typical-workflow-context now just generates some random numbers. Will also close the issue.

PR #1636, opened 2024-05-21
Python polars version: Extra info on error if mismatch with Rust

### LINKED ISSUE #1631: More helpful message if there's a polars version mismatch
If the Python and Rust polars versions have diverged far enough, the user may get an error like:
```
  File "/Users/chuckmccallum/github/opendp/opendp/python/src/opendp/_data.py", line 283, in slice_as_object
    output = unwrap(lib_function(c_raw, c_T), AnyObjectPtr)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/chuckmccallum/github/opendp/opendp/python/src/opendp/_lib.py", line 292, in unwrap
    raise OpenDPException(variant, message, backtrace)
opendp.mod.OpenDPException:
  FFI("Error when deserializing LazyFrame. This may be due to mismatched polars versions. Semantic(None, "unknown variant `GroupBy`, expected one of `Selection`, `Cache`, `Scan`, `DataFrameScan`, `Projection`, `Aggregate`, `Join`, `HStack`, `Distinct`, `Sort`, `Slice`, `MapFunction`, `Union`, `HConcat`, `ExtContext`, `Sink`")")
```

We know that the fix is to install from extras to prevent this, but the end user may not: They may be wondering what the Rust and Python Polars versions are, and which is the right one to change. In the future, this could also be a transitive dependency which would make it ever harder to track down.

I don't know if there's an easy way for Rust code to inform Python of the Polars version it has, but an alternative would be:
- Record the target Polars version as a constant in Python.
- Have a test that checks that constant against the version installed in CI (so if we bump the specified version in the extra, we'll also get a CI error.)
- Use this constant in the above error message, so we can point the user in the right direction.

Feedback welcome.

mccalluc:
- Fix #1631

Lots of "keep X in sync with Y" messages across the codebase are generally not a good thing, but maybe the clarity for the user in this case would be worth it.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -171,7 +171,7 @@ jobs:
         run: python -m pip install -e ${{ matrix.extra-install }}
 
       - name: Test
-        run: coverage run -m pytest -v ${{ contains(matrix.extra-install, 'polars') && '' || '-k "not lazyframe"'}}
+        run: coverage run -m pytest -v ${{ ! contains(matrix.extra-install, 'polars') && '-k "not lazyframe and not polars"' || ''}}
 
       - name: Test coverage
         # We only expect test coverage with all extra dependencies.
--- a/R/opendp/R/combinators.R
+++ b/R/opendp/R/combinators.R
@@ -336,8 +336,8 @@ then_fix_delta <- function(
 #'
 #' The DIA, DO, MI and MO between the input measurement and amplified output measurement all match.
 #'
-#' Protected by the "honest-but-curious" feature flag
-#' because a dishonest adversary could set the population size to be arbitrarily large.
+#' Requires `honest-but-curious`:
+#' a dishonest adversary could set the population size to be arbitrarily large.
 #'
 #' [make_population_amplification in Rust documentation.](https://docs.rs/opendp/latest/opendp/combinators/fn.make_population_amplification.html)
 #'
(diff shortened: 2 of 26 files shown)


raprasad (approved):
approved

Shoeboxam:
Why do we need to check strict version number adherence? Isn't the error when serialization fails enough? Two downsides to this approach:

1. There should be a range of acceptable Polars python point releases that maintain binary compatibility, and if the user happens to have any of them it should be fine. We already have a good place to raise an error where binary compatibility fails, we could just make that error richer.
2. This would mean we'd have to maintain polars version numbers in sync in another place.

mccalluc:
For users, this doesn't enforce strict version matches... but if there is an error, and there is a mismatch, it includes more information, saying exactly what version it saw and what version it expected. With the current state, I think it could be frustrating to be a downstream user, particularly if some other library includes opendp as a dependency, but forgot to reference our `[polars]` extra. The user is going to have some trouble figuring out what version would fix the error.

Also, it's also possible that version mismatch is not the problem... I think I hit this at one point, though I don't remember the details, and I may be misremembering. But if anyone else were to hit it, it would be good not to point them in the wrong direction.

Going to move this back to draft:
- Shouldn't merge without buy-in.
- Also want to edit the lower-level message: `This may be due to mismatched polars versions` is superseded.

mccalluc:
Ack: Forgot to switch branches. To keep the git history simple, I'm going to close this PR, and open two new ones: One with the the original content of this PR, and one for the new work relating to #1662.

PR #1639, opened 2024-05-21
Fix polars override namespace warning

### LINKED ISSUE #1619: Fix `UserWarning: Overriding existing custom namespace 'dp' (on 'Expr')`
- On #1614:
```
src/opendp/polars.py:7
  .../opendp/python/src/opendp/polars.py:7: UserWarning: Overriding existing custom namespace 'dp' (on 'Expr')
    @pl.api.register_expr_namespace("dp")
```
- Fix the warning.
- Configure pytest to fail on warnings.

mccalluc:
- Fix #1619 by removing duplicate code
- Pytest will now treat warnings as errors
- ... which flagged an open filehandle warning, so fix that, too

### DIFF
--- a/python/.pytest.ini
+++ b/python/.pytest.ini
@@ -5,6 +5,9 @@ testpaths =
 	../docs/source
 	../docs/tools
 
+# Treat warnings as errors:
+filterwarnings = error
+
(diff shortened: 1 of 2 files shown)


raprasad (approved):
approved

PR #1640, opened 2024-05-21
Upgrade pydata theme for search snippets

### LINKED ISSUE #1517: Show context snippet for search results
When I search on docs.opendp.org, all I see is the document title. Sphinx should able to show us snippets, giving us our search term in context. For example: https://www.sphinx-doc.org/en/master/search.html?q=context

There are a couple layers to the problem: The JS-based search only has an index of the documents. For snippets, it actually needs to load the document and parse its contents, but if we're just using `file:///` that's not possible: The page does not have permission to read arbitrary files from the file system -- For good reason! see https://github.com/sphinx-doc/sphinx/issues/9156

But we don't see snippets on docs.opendp.org either. If we look in the debugger, we see lots of errors:
```
Uncaught (in promise) TypeError: docContent is null
```
coming from the second line here:
```javascript
    const docContent = htmlElement.querySelector('[role="main"]');
    if (docContent !== undefined) return docContent.textContent;
```
`htmlElement.documentElement.innerHTML` looks like it is the full html, but `docContent` gets set to `null`. Looking at the source html, we are missing the `role="main"` attribute.

Looking at the [docs](https://www.sphinx-doc.org/en/master/development/templating.html):
> In order for the built-in JavaScript search to show a page preview on the results page, the document or body content should be wrapped in an HTML element containing the role="main" attribute. 

I'll investigate...

mccalluc:
- Fix #1517 

Confirmed locally that search now shows snippets. The only other difference I see with this release is the styling of tabs: They seem clunkier to me, but I don't think it's something we should fight.

Screenshot 2024-05-21 at 5 43 41 PM

When there is a final release, we should certainly use it, but I'm not seeing any problems with the RC, and the search snippets are a big win.

Took me longer than I'd like to admit to confirm that it works locally. The upstream fix is a change to the document template (to restore a `role="main"` attribute) so there are a number a cache layers to clear:
- Checkout the branch and `pip install -r requirements.txt`
- Clear the sphinx build cache with `rm -r build/html`
- Because of browser security rules, it doesn't work with just `file:///`: Instead you'll need `python -m http.server` or similar.
- My browser was also caching old versions of page, so clear the browser cache.

### DIFF
--- a/docs/requirements.in
+++ b/docs/requirements.in
@@ -9,7 +9,7 @@ sphinx
 sphinx-prompt
 sphinx-design
 git+https://github.com/samtygier-stfc/sphinx-multiversion.git@prebuild_command
-pydata-sphinx-theme==0.15.3rc1
+pydata-sphinx-theme==0.15.3rc1 # TODO: Remove version pin when RC is released.
 nbsphinx
 ipython  # for notebook syntax highlighting in docs
 pandoc
--- a/docs/requirements.in
+++ b/docs/requirements.in
@@ -9,7 +9,7 @@ sphinx
 sphinx-prompt
 sphinx-design
 git+https://github.com/samtygier-stfc/sphinx-multiversion.git@prebuild_command
-pydata-sphinx-theme
+pydata-sphinx-theme==0.15.3rc1
 nbsphinx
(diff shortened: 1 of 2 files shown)


raprasad (approved):
approved

Issue #542, opened 2022-09-02
Add a code of conduct page


raprasad:
Use the GitHub code of conduct template/instructions
  - Github instructions: https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/adding-a-code-of-conduct-to-your-project
  - OpenDP, current code: https://opendp.org/code-conduct

---

- github default for contributors: https://www.contributor-covenant.org/version/2/0/code_of_conduct.html

PR #1641, opened 2024-05-22
Create CODE_OF_CONDUCT.md

raprasad:
Code of conduct for:
- #542

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,128 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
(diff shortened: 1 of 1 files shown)


mccalluc (commented):
- Would it be worth setting up an email account separate from info@opendp.org to handle this?
- Do we know who the "community leaders" are? Who responds if we do receive something? If it's not one of the groups here https://opendp.org/our-people, could we add a section for the "community leaders"?

As much as this Code of Conduct sets expectations for contributors, I think it also commits us to respond clearly, quickly, and consistently. Are we set up for that?

raprasad (commented):
- Adapt the Django conduct enforcement manual
- Need people in place, exec committee approval

mccalluc:
Also, might be obvious, but any reports should be treated with care: I'll not sure who gets cc'd on the `info` emails, but it could be easy to forget that it also is used for this.

And, as long as we're under Harvard, should the Title IX Office be mentioned as another route for addressing complaints?

mccalluc:
- close in favor of #1905

PR #1646, opened 2024-05-23
Improved test coverage in _lib.py

mccalluc:
- Towards #1035 
- add a couple lines of doc tests
- Remove ignores that do not seem to be needed
- add notes in a couple places where we have branches that are not exercised... before just adding a test to exercise them, are we sure we need this behavior?


raprasad (approved):
approved

PR #1648, opened 2024-05-28
Improve test coverage in `context.py`

mccalluc:
- Towards #1035
- Add some tests where the behavior seems clear enough.
- Where there are still gaps, add an explicit TODO.
- On `context.__call__` suggest deleting unused/untested branch, particularly since it simplifies the typing.
- Add typing, and a couple run-time checks.

There is the risk here of just hardening behavior that isn't what we really want, but I'm hoping that this at least helps to make gaps visible.


raprasad (approved):
nice improvements

PR #1656, opened 2024-05-30
Improve test coverage in generated code

mccalluc:
- `.coveragerc` has been ignoring generated files... but many are tested, so we can give ourselves credit.
- Add some bare-minimum tests.
- Remove one untested method in generated code: Seemed to be unused, and described as private?


rust/src/data/ffi.rs
```diff
@@ -993,11 +993,6 @@ extern "C" fn opendp_data__new_arrow_array(name: *const c_char) -> FfiResult<*mu
     }
 }
 
-#[bootstrap(
-    name = "to_string",
-    arguments(this(rust_type = b"null")),
-    returns(c_type = "FfiResult<char *>")
-)]
```
Shoeboxam, line 1000: Can the Rust code corresponding to this also be removed?

python/test/test_tests.py
```diff
@@ -1,3 +1,29 @@
+import opendp.prelude as dp
+
+
 def test_number_of_tests_found(request):
     tests_found = len(request.session.items)
     assert tests_found >= 304
+
+
+def test_barely():
```
Shoeboxam, line 9: If we're not really testing them, we shouldn't take credit! Can these be distributed into the corresponding tests? Is it possible to tweak any of those tests to get full coverage?

Shoeboxam (changes requested):

rust/src/data/ffi.rs
mccalluc: done!

python/test/test_tests.py
mccalluc: Where I feel confident about the behavior I'll try to add stronger assertions and move to a more appropriate file. I'm not sure I'll be able to do that for all.

In the background, there is the question of what "test coverage" means -- **Real** 100% coverage is something to strive for ... but what we've been doing is requiring "100%", but actually having exceptions scattered all over the place. Rather than sweeping problems under the rug, I was thinking of this as a centralized TODO list, and if new untested generated code is added, CI would fail unless we have a TODO test here, or a real test elsewhere.

There are projects that have a "100% test coverage" badge, but I certainly wouldn't advertise that for opendp.

mccalluc:
Moved the new accuracy tests to `test_accuracy.py`... but looking at that file, my preference would be to have stronger assertions in the existing tests as well. (There are lots of points where we have print statements in the tests, and I have never really been sure I understood the model.) Not sure it should be a high priority.

This is hardly the end of the road, but it does:
- Confirm that everything at least runs
- In the future, it will nudge us if there is absolutely no testing on a new mechanism.

Shoeboxam (approved):
Thanks, in retrospect the tests out here in Python don't need to be very heavyweight, that's more a job for Rust.

PR #1659, opened 2024-06-03
Misc. copy-edits on the API User Guide

mccalluc:
Just re-reading the User Guide and making some edits along the way. I'll start a new PR for the next batch. Comments give some motivation, but this is all subjective, and I don't feel too strongly about any of it.


raprasad (approved):
approved

PR #1663, opened 2024-06-06
Python polars version: Extra info on error if mismatch with Rust

mccalluc:
- Replaces #1636
 
Mistakenly added new commits to that branch, and a new PR seems like the least confusing way to go forward.

Last comment from that PR:

> For users, this doesn't enforce strict version matches... but if there is an error, and there is a mismatch, it includes more information, saying exactly what version it saw and what version it expected. With the current state, I think it could be frustrating to be a downstream user, particularly if some other library includes opendp as a dependency, but forgot to reference our `[polars]` extra. The user is going to have some trouble figuring out what version _would_ fix the error.
> 
> ... it's also possible that version mismatch is not the problem... I think I hit this at one point, though I don't remember the details, and I may be misremembering. But if anyone else were to hit it, it would be good not to point them in the wrong direction.
> 
> This does require an exact version match for devs (at least for passing tests), and it is an additional thing to track -- though I think we'll be bumping this version pretty infrequently. It is a tradeoff, but I think it will be relatively painless for devs.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -171,7 +171,7 @@ jobs:
         run: python -m pip install -e ${{ matrix.extra-install }}
 
       - name: Test
-        run: coverage run -m pytest -v ${{ contains(matrix.extra-install, 'polars') && '' || '-k "not lazyframe"'}}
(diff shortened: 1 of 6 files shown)


Shoeboxam (approved):
Being told the expected version is a nice change. My initial thoughts were that this would be un-necessary: if the serialization parses, then it's probably good. But with an exact version match we'll at least know ser/de will pass our tests.

PR #1664, opened 2024-06-06
Require honest-but-curious note

### LINKED ISSUE #1662: Flag (and explain?) `honest-but-curious` (and `contrib`?) in API docs

mccalluc:
- Eventually, fix #1662
- For now, draft: Just want to get feedback on this approach.
- If good, next step is to actually fill in short explanations for each method.

### DIFF
--- a/rust/opendp_tooling/src/bootstrap/docstring.rs
+++ b/rust/opendp_tooling/src/bootstrap/docstring.rs
@@ -312,7 +312,7 @@ pub fn get_proof_path(
     })
 }
 
-/// confirm prescence of note for honest-but-curious 
+/// confirm prescence of note for honest-but-curious
 pub fn confirm_note_for_honest_but_curious(name: String, attributes: &mut Vec<Attribute>) -> () {
     let doc_comment = get_doc_comment(attributes.to_vec());
     let requires = "Requires `honest-but-curious`";
@@ -322,13 +322,18 @@ pub fn confirm_note_for_honest_but_curious(name: String, attributes: &mut Vec<At
 }
 
 fn get_doc_comment(attributes: Vec<Attribute>) -> String {
-    attributes.iter()
+    attributes
+        .iter()
         .filter(|attr| attr.path.get_ident().map(ToString::to_string).as_deref() == Some("doc"))
-        .map(|attr| if let Ok(comment) = parse_doc_attribute(attr.clone()) {
-            comment
-        } else {
-            "".to_string()
-        }).collect::<Vec<_>>().join("\n")
+        .map(|attr| {
+            if let Ok(comment) = parse_doc_attribute(attr.clone()) {
+                comment
+            } else {
+                "".to_string()
+            }
+        })
+        .collect::<Vec<_>>()
+        .join("\n")
 }
 
 /// add attributes containing the proof link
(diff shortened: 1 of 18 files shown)


rust/src/transformations/make_user_transformation/mod.rs
```diff
@@ -26,6 +26,7 @@ use crate::{
     )
 )]
 /// Construct a Transformation from user-defined callbacks.
+/// Requires `honest-but-curious`: TODO
```
Shoeboxam, line 29:
```suggestion
/// Requires `honest-but-curious`: The OpenDP Library cannot verify that the stability of the resulting transformation is governed by the provided stability map.
```

Shoeboxam: For measurements:

```suggestion
/// Requires `honest-but-curious`: The OpenDP Library cannot verify that the privacy of the resulting measurement is governed by the provided privacy map.
```

rust/src/transformations/manipulation/mod.rs
```diff
@@ -135,6 +135,7 @@ where
     generics(D(suppress), M(suppress))
 )]
 /// Make a Transformation representing the identity function.
+/// Requires `honest-but-curious`: TODO
```
Shoeboxam, line 138:
```suggestion
/// Requires `honest-but-curious`: The resulting transformation is only valid if the input domain and input metric form a metric space.
```

Shoeboxam (changes requested):
This looks great, I left a few comments, and filled some TODOs.

Shoeboxam (changes requested):
From Salil, would you also mind updating the text for honest-but-curious? 

> Enable to include constructors whose differential privacy (or stability) properties rely on the constructor arguments being correct.  That is, if a user/adversary is 'honest' in specifying the constructor arguments, then even if they later become 'curious' and try to learn something from the measurement outputs, they will not be able to violate the differential privacy promises of the measurement.

mccalluc:
Salil's revised text made it in #1810. I'll try to address the other points here.

rust/opendp_tooling/src/bootstrap/docstring.rs
```diff
@@ -312,6 +312,30 @@ pub fn get_proof_path(
     })
 }
 
+/// confirm prescence of note for honest-but-curious
+pub fn confirm_note_for_honest_but_curious(name: String, attributes: &mut Vec<Attribute>) -> () {
+    let doc_comment = get_doc_comment(attributes.to_vec());
+    let requires = "Requires `honest-but-curious`";
+    if !doc_comment.contains(requires) {
+        panic!("In {name} doc comment, add '{requires} because ...'")
```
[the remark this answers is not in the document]
mccalluc: Will do. Is the idea that even if I think an error is unrecoverable, we don't want to make it impossible for it to be handled by the caller?

mccalluc:
Having a build error, but feel out of my depth:
```
error: unexpected end of macro invocation
  --> opendp_derive/src/full.rs:62:85
   |
15 | macro_rules! try_ {
   | ----------------- when calling this macro
...
62 |         try_!(confirm_note_for_honest_but_curious(function.name, &mut item_fn.attrs));
   |                                                                                     ^ missing tokens in macro arguments
   |
note: while trying to match `,`
  --> opendp_derive/src/full.rs:16:16
   |
16 |     ($expr:expr, $original:expr) => {
   |                ^
```

mccalluc:
Closing this PR.
- #2014 

seems sufficient to me, and it is less of a maintenance burden. If anyone strongly believes that we should also require an explanation of why a particular feature is required for a particular function, file a new issue.

PR #1666, opened 2024-06-11
mental health tutorial w polars and context

### LINKED ISSUE #380: Mental Health Notebook Port
- depends on #366 

---

Port the mental health notebook to the opendp library:

https://github.com/opendp/smartnoise-samples/blob/master/analysis/tutorial_mental_health_in_tech_survey.ipynb

(Note: this notebook was originally created by @MayanaPereira)

GKD-stack:
Mental health tutorial with polars and the Context API

- Fix #380


mccalluc:
(Pinged Mike on slack: Expect that we'll get CI passing on the base before going farther here.)

mccalluc:
There's a link to a pre-processing notebook. 
- Link to it as well?
- Integrate it into this notebook?
- Handwave the details?

Shoeboxam:
The CI failure on `--no-default-features` is fixed in main (from the OnceFrame PR).

mccalluc:
```
#this was borrowed from github
# https://github.com/pola-rs/polars/issues/4289
df = (
    df
    .with_columns([
        pl.col(pl.Int64).cast(pl.Int32).keep_name(),
        pl.col(pl.Float64).cast(pl.Float32).keep_name(),
    ])
)
```
(File is too large to open as diff in github, so I'll add comments here instead of taking the usual review process.)

- Need to explain what's going on here.
- Is there a usability issue filed for this?

```
df = (df.select(vars))
```
I'd move this up, maybe immediately after the definition of `vars`, just to keep the reference spans shorter.

mccalluc:
Talked with Mike: more will happen on main and then that can be merged or rebased here. Nothing needs to be done here right now.

mccalluc:
(Should try rebasing this to main now and make sure the code works with the latest release.)

GKD-stack:
Got it, thanks Chuck!

I made all the edits you suggested and have some clarification questions on the following: 

1. I actually don't see the keep_name deprecation warning, I don't use that function either
2. I didn't change this portion, it was previously already there.
`Due to sequential composition, given $q_i$ queries. if we assume that i=n then I think it's what you're saying `. If we assume that i=n then I think it's what you're saying 
3. Bar graphs instead of pie charts. I think the pie charts actually showcase more variety in visualization. The bar charts are consistently used throughout other examples. 

Should I make a new branch that follows the naming convention?

mccalluc:
> I actually don't see the keep_name deprecation warning, I don't use that function either

Huh, that's strange. (This is one reason it's nice to keep files smaller so we see the same diffs in the web interface.) Maybe we're looking at different things?

```
$ git checkout notebook
$ grep -n keep_name docs/tutorial_mental_health.ipynb
102:      "/var/folders/dg/nlwnz_jj7rsblr2tzwj962b40000gn/T/ipykernel_64512/2319299402.py:16: DeprecationWarning: `keep_name` is deprecated. It has been moved to `name.keep`.\n",
103:      "  pl.col(pl.Int64).cast(pl.Int32).keep_name(),\n",
104:      "/var/folders/dg/nlwnz_jj7rsblr2tzwj962b40000gn/T/ipykernel_64512/2319299402.py:17: DeprecationWarning: `keep_name` is deprecated. It has been moved to `name.keep`.\n",
105:      "  pl.col(pl.Float64).cast(pl.Float32).keep_name(),\n"
125:    "        pl.col(pl.Int64).cast(pl.Int32).keep_name(),\n",
126:    "        pl.col(pl.Float64).cast(pl.Float32).keep_name(),\n",
```

[Deprecation note in Polars docs](https://docs.pola.rs/api/python/version/0.19/reference/expressions/api/polars.Expr.keep_name.html)

mccalluc (commented):
I don't think there have been changes since the last review? I'd suggest either rebasing this on `main`, or what might be simpler, starting a new branch and new PR, referencing this old PR, and just copying the files over there, and closing this.

mccalluc:
I'll confirm on Slack, but I'll close this as well as #380. We learned things by exercising Polars, but the tutorials are being built around the employment dataset, and we understand how each of those parts will fit into the larger structure.

PR #1668, opened 2024-06-11
 nb to rst

### LINKED ISSUE #1667: notebook to rst where it's easy
Started looking at
- #1654

and was going to make a copy edit on proof-initiation, but realized it was a notebook. There are no graphics, nor any python code here, so we're not gaining anything with the notebook format. `nbconvert` should make this straightforward. See if there are any other candidates for easy conversion.

mccalluc:
- Fix #1667
- Automatically convert low-python notebooks to rst
- For some medium-python notebooks, convert, and then edit to change to doctest

Benefits:
- Simpler source code
- Stronger, more uniform tests. (Notebooks are run, but we don't check the output.)


Shoeboxam (changes requested):
Double-check the HTML and transition ipython3 -> doctest?

mccalluc:
- Close in favor of #1700

PR #1669, opened 2024-06-12
patch R docs build failure on CI

### LINKED ISSUE #1605: Nightly r-docs failing: maybe because OPENDP_LIB_DIR is undefined?

Shoeboxam:
Closes #1605

Provides an alternative docs build that does not rely on the existence of a library binary. Behaves similarly as OPENDP_HEADLESS for Python.

CI Run:
https://github.com/opendp/opendp/actions/runs/9490435694


mccalluc:
I'll come back to this, but I was hoping the manual run might have pushed r-docs to nightly, but https://docs.opendp.org/en/nightly/api/r/reference/accuracy_to_discrete_gaussian_scale.html is 404. 

My sense is that this is hacky, and ideally it would work differently, but I'm not sure what that ideal looks like. Should there be an issue for that?

(After looking at other PRs, which I expect to be easier, I'll come back to this and try to understand better how it works.)

tools/r_stage.sh
```diff
@@ -94,6 +94,37 @@ function docs() {
   log "R package docs are ready in R/opendp/docs/index.html"
 }
 
+function dummydocs() {
```
mccalluc, line 97:
```suggestion
function dummy_docs() {
```

tools/r_stage.sh
```diff
@@ -108,14 +139,15 @@ if (($# == 0)); then
...
-while getopts ":cbsvnd" OPT; do
+while getopts ":cbsvndu" OPT; do
   case "$OPT" in
     c) clean ;;
     b) binary_tar ;;
     s) source_tar ;;
     v) vendor_tar ;;
     n) notes ;;
     d) docs ;;
+    u) dummydocs ;;
```
mccalluc, line 150:
```suggestion
    u) dummy_docs ;;
```

mccalluc (approved):
- Filed #1675

It's going to take me some time to understand this, and even then, the best solution probably isn't going to be obvious. Approving to get nightly fixed.

PR #1672, opened 2024-06-12
Add a script to help create changelog

### LINKED ISSUE #1629: Document process for creating changelog / Stronger conventions with PR names?
The CHANGELOG has a consistent structure, but is it mostly hand made? The one instruction I see is

> Update ``CHANGELOG.md`` on ``main`` (based on `Keep a Changelog <https://keepachangelog.com/en/1.0.0/>`_) .

For #1620, I'm starting from the git log, and using some adhoc perl, but is there a better way to do this?

```
$ git log | perl -ne 'print if //../#1218/' | perl -ne 'if (/^\s+\w.*\(#\d+\)/) {s/^\s+/- /; s{\(#(\d+)\)}{[#\1](https://github.com/opendp/opendp/pull/\1)}; print}' | head
- regenerate python to include example [#1615](https://github.com/opendp/opendp/pull/1615)
- More typing on context [#1472](https://github.com/opendp/opendp/pull/1472)
- Use the same examples for `make` and `then` [#1576](https://github.com/opendp/opendp/pull/1576)
- For rust example in docs, use normal `cargo run` [#1612](https://github.com/opendp/opendp/pull/1612)
- Error if missing `arg` param on measurement in R [#1559](https://github.com/opendp/opendp/pull/1559)
- randomgen guards, again [#1562](https://github.com/opendp/opendp/pull/1562)
- Merge "then" examples into "make" and regenerate [#1575](https://github.com/opendp/opendp/pull/1575)
- speed up smoke-test [#1579](https://github.com/opendp/opendp/pull/1579)
- Polars: add `make_private_quantile_expr` [#908](https://github.com/opendp/opendp/pull/908)
- Polars: bounded-DP mean via postprocessing [#890](https://github.com/opendp/opendp/pull/890)
```

One idea might be to raise the bar on our PR names, and add a category there. We might still want to rewrite and reorganize, but we'd have a good head start.

(Or `git log --oneline` might be a better place to start.)

mccalluc:
- Fix #1629


Shoeboxam (changes requested):
This is a good start, but if we want to go this route I think it should be more closely tied in with channel_tool.py, and it would be great to adopt conventions on PR names first.

Shoeboxam (approved):
I found this useful when updating and I appreciate the docs improvements.

PR #1678, opened 2024-06-14
feat: mechanism: add `make_permute_and_flip`

tudorcebere:
Adding support for Permute&Flip (https://arxiv.org/abs/2010.12603), floating-point stable.

TODOs:
* bindings

### DIFF
--- a/rust/src/measurements/permute_and_flip/mod.rs
+++ b/rust/src/measurements/permute_and_flip/mod.rs
@@ -1,37 +1,27 @@
-use crate::error::Fallible;
-use crate::traits::samplers::sample_bernoulli_exp;
-use crate::traits::samplers::SampleUniformIntBelow;
-use super::Optimize;
-use dashu::rational::RBig;
-use dashu::base::Sign;
-
 use crate::{
     core::{Function, Measurement, PrivacyMap},
     domains::{AtomDomain, VectorDomain},
+    error::Fallible,
     measures::MaxDivergence,
     metrics::LInfDistance,
-    traits::{Float, Number},
+    traits::{Float, Number, DistanceConstant},
+    traits::samplers::{sample_bernoulli_exp, SampleUniformIntBelow, CastInternalRational},
+    measurements::Optimize
 };
-
-use crate::traits::{
-    samplers::CastInternalRational,
-    DistanceConstant,
-};
-
-use crate::error::Error;
+use dashu::base::Sign;
+use dashu::rational::RBig;
 
 
 fn exact_fisher_yates(n: usize, trials_per_coin: Option<usize>) -> Fallible<Vec<usize>> {
     let mut permutation: Vec<usize> = (0..n).collect();
 
     for i in (1..n).rev() {
-        let j =  usize::sample_uniform_int_below(i, trials_per_coin)?;
+        let j = usize::sample_uniform_int_below(i, trials_per_coin)?;
         permutation.swap(i, j);
     }
 
     Ok(permutation)
 }
-
 
 /// Make a Measurement that takes a vector of scores and privately selects the index of the highest score via permute and flip.
 ///
@@ -62,7 +52,7 @@ where
         return fallible!(MakeMeasurement, "scale must not be negative");
     }
 
-    let scale_frac: RBig = scale.clone().into_rational()?;
+    let scale_frac: RBig = scale.into_rational()?;
     let sgn: Sign = match optimize {
         Optimize::Max => Sign::Positive,
         Optimize::Min => Sign::Negative,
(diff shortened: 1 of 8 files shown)


rust/src/measurements/permute_and_flip/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+    measurements::Optimize,
+    measures::MaxDivergence,
+    metrics::LInfDistance,
+    traits::samplers::{sample_bernoulli_exp, CastInternalRational, SampleUniformIntBelow},
+    traits::{DistanceConstant, Float, Number},
+};
+use dashu::base::Sign;
+use dashu::rational::RBig;
+
+fn exact_fisher_yates(n: usize, trials_per_coin: Option<usize>) -> Fallible<Vec<usize>> {
```
Shoeboxam, line 14: * I think it would be better to write this as a function that shuffles a slice
* I don't think you need to plumb the number of trials up, unless you have a good sense for how you might handle the error and still keep this algorithm timing-safe.

I think this signature then becomes:
`fn exact_fisher_yates<T>(&mut [T] slice) -> Fallible<()>`

Calling like:
```
let mut permutation = (0..arg.len()).collect();
exact_fisher_yates(permutation.as_mut_slice())?;
```

rust/src/measurements/permute_and_flip/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+            let argmax = rational_elements
+                .iter()
+                .enumerate()
+                .max_by_key(|&(_, value)| value)
+                .map(|(index, _)| index)
+                .ok_or(err!(FailedFunction, "there must be at least one candidate"))?;
+
+            let max_element = rational_elements
+                .get(argmax)
+                .ok_or(err!(FailedFunction, "there must be at least one candidate"))?;
```
Shoeboxam, line 79: I don't think you ever need the index of the max score, so the map could just return the value.

rust/src/measurements/permute_and_flip/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+
+            // iterate over the permutations and throw a coin for each
+            for i in permutation {
+                let candidate = rational_elements
+                    .get(i)
+                    .ok_or(err!(FailedFunction, "Wrong indexing"))?;
```
Shoeboxam, line 85: I would just use `let candidate = &rational_elements[i]` here, as we know the index will be in-bounds.

rust/src/measurements/permute_and_flip/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+
+                let coin_bias_scale = max_element - candidate;
+                let coin_flip = sample_bernoulli_exp(coin_bias_scale / &scale_frac)?;
+                if coin_flip {
+                    return Ok(i);
+                }
+            }
+            Ok(argmax)
```
Shoeboxam, line 93: What should happen here? Should the argmax really be returned?

rust/src/measurements/permute_and_flip/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+            if d_in.is_sign_negative() {
+                return fallible!(InvalidDistance, "sensitivity must be non-negative");
+            }
+
+            if scale.is_zero() {
+                return Ok(QO::infinity());
+            }
+
+            let rational_two =
+                QO::inf_cast(TIA::from(2).ok_or(err!(FailedCast, "Failed to cast constant."))?)?;
```
Shoeboxam, line 113: This isn't actually rational, and so the `d_in * rational_two` below is not exact. 

However, with a proof, I expect this two to be redundant with the range distance conversion above.

rust/src/measurements/permute_and_flip/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+
+            // d_out >= (2 * d_in) / scale
+            (d_in * rational_two).inf_div(&scale)
+        }),
+    )
+}
+
+#[cfg(feature = "floating-point")]
+#[cfg(test)]
+mod test;
```
Shoeboxam, line 123: We only require floating-point in cases where the guarantees do not hold due to floating-point imprecision. This algorithm should be exact, so it doesn't need it, even though it deals with floating-point numbers.

rust/src/measurements/permute_and_flip/test.rs
```diff
@@ -0,0 +1,40 @@
...
+    assert_eq!(out.len(), 10);
+    // check that out contains all elements from 0 to 9
+    for i in 0..10 {
+        assert!(out.contains(&i));
+    }
+
+    // check that the elements are shuffled
+    let mut sorted = out.clone();
+    sorted.sort();
+    assert_ne!(out, sorted);
```
Shoeboxam, line 17: Unlikely, but will fail with probability 1/10!

Shoeboxam (changes requested):
Looks great, thanks for the implementation! Left a few comments.

Shoeboxam:
The latex-build CI here fails because this PR originated from a fork (https://stackoverflow.com/a/75746222). 

It runs properly in a draft PR (https://github.com/opendp/opendp/pull/1726)

I'm fine with ignoring the failure. Will keep this gotcha in mind for the future. :P

Shoeboxam (changes requested):
Whoops! Not yet! My bad.

Shoeboxam:
Closing in favor of #2220, specifically [this implementation](https://github.com/opendp/opendp/pull/2220/files#diff-c3b525d9f6d97f06c857cb9e0eaf036a14559c56d8e999d6aaf21eae6c45f52d) and surrounding proofs/tests.

PR #1621, opened 2024-05-16
Polars documentation

mccalluc:
- Fix #1618

PR #1688, opened 2024-06-18
Polars: expand API docs

### LINKED ISSUE #1618: Fix references in polars docs
`make html` gives this error:
```
polars.py:docstring of opendp.polars.DPExpr.gaussian:3: WARNING: py:func reference target not found: opendp.measurement.make_private_lazyframe
polars.py:docstring of opendp.polars.DPExpr.laplace:3: WARNING: py:func reference target not found: opendp.measurement.make_private_lazyframe
polars.py:docstring of opendp.polars.DPExpr.mean:3: WARNING: py:func reference target not found: opendp.measurement.make_private_lazyframe
polars.py:docstring of opendp.polars.DPExpr.noise:3: WARNING: py:func reference target not found: opendp.measurement.make_private_lazyframe
polars.py:docstring of opendp.polars.DPExpr.sum:3: WARNING: py:func reference target not found: opendp.measurement.make_private_lazyframe
```
- Should be an easy fix.
- I though bad doc references would be caught by CI?

mccalluc:
- Replaces #1621: `polars.py` has changed a lot, and it was easier to start fresh than sort out the conflicts.
- Fix #1618


mccalluc:
https://github.com/opendp/opendp/actions/runs/9586399684/job/26434669648
```
── 1. Error ('test-measurements.R:64:3'): test_gaussian_curve ──────────────────
Error: Error in `make_zCDP_to_approxDP(then_gaussian(input_space, 4))`: [FFI] : null pointer: measurement
disabled backtrace
Backtrace:
    ▆
 1. └─opendp::make_zCDP_to_approxDP(then_gaussian(input_space, 4)) at test-measurements.R:64:3
```
restarting...

python/src/opendp/polars.py
```diff
@@ -11,6 +20,28 @@
...
+    Both of these expressions return a ``DPExpr``:
+
+    >>> import polars as pl
+    >>> pl.len().dp
+    <opendp.polars.DPExpr object at ...>
+    >>> DPExpr(pl.len())
+    <opendp.polars.DPExpr object at ...>
+
+    However, a static analysis tool like `mypy <https://mypy.readthedocs.io/en/stable/>`_
+    will see the type of the first expression as ``Any``, while the later is ``DPExpr``.
```
mccalluc, line 43: I think you'd like us to say something like this? I feel like it might be introducing unnecessary complexity: Libraries that aren't very strictly typed are not a surprise, and if the user wanted stricter typing they could do this:

```python
dp_expr: dp.DPExpr = pl.len().dp
```

Shoeboxam: I'm fine with it not being mentioned if you prefer! Your judgement. If you are familiar with mypy typing then this may be apparent. The most important thing to communicate in this docstring is that these methods are accessible under .dp, which you've already handled.

python/src/opendp/polars.py
```diff
@@ -43,39 +74,73 @@ def noise(
...
-        If scale is None it is filled by `global_scale` in :py:func:`opendp.measurements.make_private_lazyframe`.
+        If scale is None it is filled by ``global_scale`` in :py:func:`opendp.measurements.make_private_lazyframe`.
 
-        :param scale: Noise scale parameter for the Laplace distribution. `scale` == standard_deviation / sqrt(2).
+        :param scale: Noise scale parameter for the Laplace distribution. ``scale == standard_deviation / sqrt(2)``
+
+        :example:
+
+        >>> import polars as pl
+        >>> expression = pl.col('numbers').dp.laplace()
```
Shoeboxam, line 84: While this makes an expression, the expression would not be accepted by the lib. I'd go with `pl.len().dp.laplace()` instead. Similarly below for `gaussian` and above for `noise` (which doesn't have an example yet).

python/src/opendp/polars.py
```diff
@@ -357,19 +446,26 @@ def release(self) -> OnceFrame:
 
 @dataclass
 class Margin(object):
+    '''
+    Instances of ``Margin`` are used to collect DP statistics on
+    `marginal distributions <https://en.wikipedia.org/wiki/Marginal_distribution>`_.
```
Shoeboxam, line 451:
```suggestion
    The ``Margin`` class is used to describe what information is known publicly about a grouped dataset:
    like the values you might expect to find in the margins of a table.
    
    Be aware that aspects of your data marked as "public information" are not subject to privacy protections,
    so it is important that public descriptors about the margin should be set conservatively, or not set at all.
```

Shoeboxam (changes requested):
Would've been nice in 0.10, but it'll be in before long. ;)

Shoeboxam (approved):
Thanks for the edits/additions!

PR #1694, opened 2024-06-20
chore: remove `UnknownType` and instead immediately raise exception

mccalluc:
Prompted by https://github.com/opendp/opendp/pull/1652/files#r1642781252 

I don't understand the data load process in detail, but at least on the evidence of our tests it doesn't seem to be needed now.

### DIFF
--- a/python/src/opendp/typing.py
+++ b/python/src/opendp/typing.py
@@ -279,7 +279,7 @@ class RuntimeType(object):
         >>> dp.RuntimeType.infer([])
         Traceback (most recent call last):
         ...
-        opendp.mod.UnknownTypeException: cannot infer atomic type when empty
+        opendp.mod.UnknownTypeException: Cannot infer atomic type when empty
         """
         if type(public_example) in ELEMENTARY_TYPES:
             return ELEMENTARY_TYPES[type(public_example)]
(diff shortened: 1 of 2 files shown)


python/src/opendp/typing.py
```diff
@@ -330,8 +336,14 @@ def infer_homogeneous(value):
...
-        if public_example is None:
-            return RuntimeType('Option', [UnknownType("Constructed Option from a None variant")])
+        if isinstance(public_example, Measurement): # pragma: no cover
+            return "AnyMeasurementPtr"
+
+        if isinstance(public_example, Transformation): # pragma: no cover
+            return "AnyTransformationPtr"
+
+        if public_example is None: # pragma: no cover
+            raise UnknownTypeException("Constructed Option from a None variant")
```
Shoeboxam, line 346: This error could be clearer... It made sense when the code would make an `UnknownType`, but this raises now.

```suggestion
        if public_example is None: # pragma: no cover
            raise UnknownTypeException("Type of Option cannot be inferred from None")
```

Shoeboxam (changes requested):

Shoeboxam (approved):
Thanks for updating this!

PR #1695, opened 2024-06-20
Add more informative error message if Python and Rust polars versions don't match

### LINKED ISSUE #1631: More helpful message if there's a polars version mismatch

mccalluc:
- Fix #1631 
- Redo of #1663, which I merged onto the wrong branch.
- (And that PR replaced #1636.)
- This just rebases it onto main.

> For users, this doesn't enforce strict version matches... but if there is an error, and there is a mismatch, it includes more information, saying exactly what version it saw and what version it expected. With the current state, I think it could be frustrating to be a downstream user, particularly if some other library includes opendp as a dependency, but forgot to reference our `[polars]` extra. The user is going to have some trouble figuring out what version _would_ fix the error.
> 
> ... it's also possible that version mismatch is not the problem... I think I hit this at one point, though I don't remember the details, and I may be misremembering. But if anyone else were to hit it, it would be good not to point them in the wrong direction.
> 
> This does require an exact version match for devs (at least for passing tests), and it is an additional thing to track -- though I think we'll be bumping this version pretty infrequently. It is a tradeoff, but I think it will be relatively painless for devs.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -171,7 +171,7 @@ jobs:
         run: python -m pip install -e ${{ matrix.extra-install }}
 
       - name: Test
-        run: coverage run -m pytest -v ${{ contains(matrix.extra-install, 'polars') && '' || '-k "not lazyframe"'}}
+        run: coverage run -m pytest -v ${{ ! contains(matrix.extra-install, 'polars') && '-k "not lazyframe and not polars"' || ''}}
 
       - name: Test coverage
         # We only expect test coverage with all extra dependencies.
--- a/python/setup.cfg
+++ b/python/setup.cfg
@@ -45,7 +45,8 @@ scikit-learn =
 polars =
 	# A strict version requirement is necessary to keep the serialization format of LazyFrames stable.
 	# If you change the version of Polars, 
-	# be sure to also change the Polars version in rust/Cargo.toml and test binary compatibility.
+	# be sure to also change the Polars version in rust/Cargo.toml and test binary compatibility,
+	# and update _EXPECTED_POLARS_VERSION in mod.py as well.
 	polars==0.20.16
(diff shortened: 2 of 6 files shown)


Shoeboxam (approved):
Same approval.

PR #1715, opened 2024-06-24
Polars: add cast transformation

Shoeboxam:
closes #1710

PR #1716, opened 2024-06-24
Polars: laplace thresholding for private key-sets

Shoeboxam:
In this setting, the key set is not public:

```python
    data=pl.LazyFrame(schema={"A": pl.Int32, "B": pl.String}),
    privacy_loss=dp.loss_of(epsilon=1.0, delta=1e-7),
    split_evenly_over=2,
    margins={
        ("B",): dp.Margin(max_partition_length=5),
    },
)
```

However, you can still release grouped queries on it:
```python
query = (
    context.query()
    .group_by("B")
    .agg(pl.len().dp.noise(), pl.col("A").fill_null(2).dp.sum((0, 3)))
)
```

This is because the library adds thresholding to the query based on the noisy `pl.len()` to satisfy approximate-DP.

## ProductOrd change
This PR also changes the behavior of `ProductOrd` on tuples:
A tuple is only considered greater than another if all elements are greater than the corresponding element in the other tuple. The library previously used a lexicographic ordering, which, while this makes tuples totally ordered, effectively only guarantees that the epsilon in the epsilon/delta tuple is solved for. This is not really what we want from the check in our maps: the check should only succeed if both epsilon and delta are satisfied.

While the new product ordering makes check perform as expected, it also means tuples no longer have a total ordering, so binary searches over approx-DP mechanisms now fail. I've adjusted all of these call locations to first solve for epsilon, and then delta.

### DIFF
--- a/docs/source/getting-started/examples/histograms.ipynb
+++ b/docs/source/getting-started/examples/histograms.ipynb
@@ -282,7 +282,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.11.6"
+   "version": "3.12.3"
   }
  },
  "nbformat": 4,
--- a/docs/source/getting-started/examples/histograms.ipynb
+++ b/docs/source/getting-started/examples/histograms.ipynb
@@ -191,8 +191,8 @@
     "    def privatize(s, t=1e8):\n",
     "        return preprocess >> dp.m.then_laplace_threshold(scale=s, threshold=t)\n",
     "    \n",
-    "    s = dp.binary_search_param(lambda s: privatize(s=s), d_in, d_out)\n",
-    "    t = dp.binary_search_param(lambda t: privatize(s=s, t=t), d_in, d_out)\n",
+    "    s = dp.binary_search(lambda s: privatize(s=s).map(d_in)[0] <= d_out[0])\n",
+    "    t = dp.binary_search(lambda t: privatize(s=s, t=t).map(d_in)[1] <= d_out[1])\n",
     "\n",
     "    return privatize(s=s, t=t)"
    ]
--- a/python/src/opendp/context.py
+++ b/python/src/opendp/context.py
@@ -702,12 +702,18 @@ class PartialChain(object):
 
         The discovered parameter is assigned to the param attribute of the returned transformation or measurement.
         """
-        param = binary_search(
-            lambda x: _cast_measure(self.partial(x), output_measure, d_out).check(
-                d_in, d_out
-            ),
-            T=T,
-        )
+        # When the output measure corresponds to approx-DP, only optimize the epsilon parameter.
+        # The delta parameter should be fixed in _cast_measure, and if not, then the search will be impossible here anyways.
+        if output_measure == fixed_smoothed_max_divergence(T=float):
+            def predicate(param):
+                meas = _cast_measure(self.partial(param), output_measure, d_out)
+                return meas.map(d_in)[0] <= d_out[0] # type: ignore[index] 
+        else:
+            def predicate(param):
+                meas = _cast_measure(self.partial(param), output_measure, d_out)
+                return meas.check(d_in, d_out)
+        
+        param = binary_search(predicate, T=T)
         chain = self.partial(param)
         chain.param = param
         return chain
--- a/python/src/opendp/measurements.py
+++ b/python/src/opendp/measurements.py
@@ -663,7 +663,8 @@ def make_private_lazyframe(
     input_metric: Metric,
     output_measure: Measure,
     lazyframe,
-    global_scale = None
+    global_scale = None,
+    threshold = None
 ) -> Measurement:
     r"""Create a differentially private measurement from a [`LazyFrame`].
 
@@ -686,7 +687,8 @@ def make_private_lazyframe(
     :param output_measure: How to measure privacy loss.
     :type output_measure: Measure
     :param lazyframe: A description of the computations to be run, in the form of a [`LazyFrame`].
-    :param global_scale: A tune-able parameter that affects the privacy-utility tradeoff.
+    :param global_scale: Optional. A tune-able parameter that affects the privacy-utility tradeoff.
+    :param threshold: Optional. Minimum number of rows in each released partition.
     :rtype: Measurement
     :raises TypeError: if an argument's type differs from the expected type
     :raises UnknownTypeException: if a type argument fails to parse
@@ -768,20 +770,22 @@ def make_private_lazyframe(
     c_output_measure = py_to_c(output_measure, c_type=Measure, type_name=None)
     c_lazyframe = py_to_c(lazyframe, c_type=AnyObjectPtr, type_name=LazyFrame)
     c_global_scale = py_to_c(global_scale, c_type=AnyObjectPtr, type_name=RuntimeType(origin='Option', args=[f64]))
+    c_threshold = py_to_c(threshold, c_type=AnyObjectPtr, type_name=RuntimeType(origin='Option', args=[u32]))
 
     # Call library function.
     lib_function = lib.opendp_measurements__make_private_lazyframe
-    lib_function.argtypes = [Domain, Metric, Measure, AnyObjectPtr, AnyObjectPtr]
+    lib_function.argtypes = [Domain, Metric, Measure, AnyObjectPtr, AnyObjectPtr, AnyObjectPtr]
     lib_function.restype = FfiResult
 
-    output = c_to_py(unwrap(lib_function(c_input_domain, c_input_metric, c_output_measure, c_lazyframe, c_global_scale), Measurement))
+    output = c_to_py(unwrap(lib_function(c_input_domain, c_input_metric, c_output_measure, c_lazyframe, c_global_scale, c_threshold), Measurement))
 
     return output
 
 def then_private_lazyframe(
     output_measure: Measure,
     lazyframe,
-    global_scale = None
+    global_scale = None,
+    threshold = None
 ):  
     r"""partial constructor of make_private_lazyframe
 
@@ -791,7 +795,8 @@ def then_private_lazyframe(
     :param output_measure: How to measure privacy loss.
     :type output_measure: Measure
     :param lazyframe: A description of the computations to be run, in the form of a [`LazyFrame`].
-    :param global_scale: A tune-able parameter that affects the privacy-utility tradeoff.
+    :param global_scale: Optional. A tune-able parameter that affects the privacy-utility tradeoff.
+    :param threshold: Optional. Minimum number of rows in each released partition.
 
     :example:
 
@@ -865,7 +870,8 @@ def then_private_lazyframe(
         input_metric=input_metric,
         output_measure=output_measure,
         lazyframe=lazyframe,
-        global_scale=global_scale))
+        global_scale=global_scale,
+        threshold=threshold))
 
 
 
--- a/python/src/opendp/polars.py
+++ b/python/src/opendp/polars.py
@@ -423,6 +423,7 @@ try:
                     threshold=threshold,
                 )
             
+            # when the output measure is δ-approximate, then there are two free parameters to tune
             if query._output_measure.type.origin == "FixedSmoothedMaxDivergence":  # type: ignore[union-attr]
 
                 # search for a scale parameter. Solve for epsilon first, 
@@ -434,7 +435,7 @@ try:
 
                 # attempt to return without setting a threshold
                 try:
-                    return make(scale)
+                    return make(scale, threshold=None)
                 except OpenDPException:
                     pass
                 
@@ -444,8 +445,11 @@ try:
                     T=int,
                 )
                 
+                # return a measurement with the discovered scale and threshold
                 return make(scale, threshold)
 
+            # when no delta parameter is involved, 
+            # finding a suitable measurement just comes down to finding scale
             return binary_search_chain(make, d_in, d_out, T=float)
 
         def release(self) -> OnceFrame:
@@ -458,6 +462,10 @@ try:
             """Retrieve noise scale parameters and accuracy estimates for each output.
 
             If ``alpha`` is passed, the resulting data frame includes an ``accuracy`` column.
+
+            If a threshold is configured for censoring small/sensitive partitions, 
+            a threshold column will be included,
+            containing the cutoff for the respective count query being thresholded.
             
             :example:
 
--- a/python/src/opendp/polars.py
+++ b/python/src/opendp/polars.py
@@ -14,7 +14,14 @@ from dataclasses import dataclass
 import os
 from typing import Literal, Tuple
 from opendp._lib import lib_path
-from opendp.mod import Domain, Measurement, assert_features
+from opendp.mod import (
+    Domain,
+    Measurement,
+    OpenDPException,
+    assert_features,
+    binary_search,
+    binary_search_chain,
+)
 from opendp.domains import series_domain, lazyframe_domain, option_domain, atom_domain
 from opendp.measurements import make_private_lazyframe
 
@@ -142,6 +149,7 @@ class DPExpr(object):
         [(col("numbers").clip([...]).sum()...:noise()) / (len())]
         """
         import polars as pl  # type: ignore[import-not-found]
+
         return self.expr.dp.sum(bounds, scale) / pl.len()
 
     def _discrete_quantile_score(self, alpha: float, candidates: list[float]):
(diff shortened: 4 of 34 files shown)


Shoeboxam:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/opendp/opendp/1716?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1716** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1716?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1703** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1703?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1715](https://github.com/opendp/opendp/pull/1715) <a href="https://app.graphite.dev/github/pr/opendp/opendp/1715?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`

<h2></h2>

mccalluc:
Notes on how to review from Mike:

> I'll ask Vikrant to take a look at the more math-y parts. 
> 
> I'd be sure to understand how the logic for parsing the DSL works.
> The general idea for this algorithm is that, you only want to release statistics on a partition if there is no neighboring dataset that doesn't have that partition. If removing any one person causes that partition to go away, then it shouldn't be part of the release. The way this is done privately is, you compute a DP count of the number of records in each partition, and if the count is too small (less than or equal to a threshold), then the whole partition is filtered away.
> 
> Unfortunately, there's still a chance that a partition that really only has one person in it is still included in the release, because the filtering is based on a noisy count estimate. This can happen if the noise added to that small partition's count was large and positive. The probability of this happening is reflected in δ, which is tuned to be cryptographically small.

mccalluc (commented):
Is my understanding of the goal of this PR correct?

> In this setting, the key set is not public:
> 
> ```
> context = dp.Context.compositor(
>     # code that currently on main would require explicit public_info in the margin?
> )
> ```
> 
> However, you can still release grouped queries on it:
> 
> ```
> query = (
>     # code that currently on main would fail without public_info?
> )
> ```

If that's right, could the existing `test_polars_context` (the only test I think we have of `margins=`) be tweaked, to demonstrate what's now possible?

I've also looked at the issue this fixes 
- #1714

... but I don't get a clear sense from that of what the current behavior on main is, and what we'd like it to be.

I'll start on a more thorough review... but understanding what we're trying to do will be useful.

docs/source/getting-started/examples/histograms.ipynb
```diff
@@ -191,8 +191,8 @@
     "    def privatize(s, t=1e8):\n",
     "        return preprocess >> dp.m.then_laplace_threshold(scale=s, threshold=t)\n",
     "    \n",
-    "    s = dp.binary_search_param(lambda s: privatize(s=s), d_in, d_out)\n",
-    "    t = dp.binary_search_param(lambda t: privatize(s=s, t=t), d_in, d_out)\n",
+    "    s = dp.binary_search(lambda s: privatize(s=s).map(d_in)[0] <= d_out[0])\n",
+    "    t = dp.binary_search(lambda t: privatize(s=s, t=t).map(d_in)[1] <= d_out[1])\n",
```
mccalluc, line 195: The new version seems to be at a slightly lower level, and is slightly longer. Would the previous way still work?

python/src/opendp/context.py
```diff
@@ -702,12 +702,18 @@ def fix(self, d_in: float, d_out: float, output_measure: Optional[Measure] = Non
...
         """
-        param = binary_search(
-            lambda x: _cast_measure(self.partial(x), output_measure, d_out).check(
-                d_in, d_out
-            ),
-            T=T,
-        )
+        # When the output measure corresponds to approx-DP, only optimize the epsilon parameter.
+        # The delta parameter should be fixed in _cast_measure, and if not, then the search will be impossible here anyways.
+        if output_measure == fixed_smoothed_max_divergence(T=float):
```
mccalluc, line 707: Just glancing at the docs for `fixed_smoothed_max_divergence`, it doesn't mention approx-DP. Should it? (Not a blocker for this PR, but maybe file an issue from this comment.)

python/src/opendp/measurements.py
```diff
@@ -686,7 +687,8 @@ def make_private_lazyframe(
     :param output_measure: How to measure privacy loss.
     :type output_measure: Measure
     :param lazyframe: A description of the computations to be run, in the form of a [`LazyFrame`].
-    :param global_scale: A tune-able parameter that affects the privacy-utility tradeoff.
+    :param global_scale: Optional. A tune-able parameter that affects the privacy-utility tradeoff.
+    :param threshold: Optional. A tune-able parameter that filters smaller partitions.
```
mccalluc, line 691: I'd assume that the threshold is something like the minimum number of elements in the partition? Wouldn't hurt to be more explicit about what it does?

Shoeboxam:
@mccalluc: Code that is currently on main should require explicit public info in the margin-- but it currently only requires the margin exists.

python/src/opendp/polars.py
```diff
@@ -398,25 +406,46 @@ def wrap(*args, **kwargs):
...
+                    lazyframe=lf_plan,
+                    global_scale=scale,
+                    threshold=threshold,
+                )
+            
+            if query._output_measure.type.origin == "FixedSmoothedMaxDivergence":  # type: ignore[union-attr]
+
+                # search for a scale parameter. If threshold unknown, set it to u32::MAX so as not to interfere
+                scale = binary_search(
+                    lambda s: make(s, 2**32 - 1).map(d_in)[0] < d_out[0],  # type: ignore[index]
```
mccalluc, line 430: In the line above, `threshold` doesn't mean the parameter here... I think you mean farther down? Consider rephrasing or moving comment for clarity.

mccalluc (commented):
A few more comments. Should pick up here next week. Take care.

docs/source/getting-started/examples/histograms.ipynb
Shoeboxam: No, the previous way will throw an error about the tuples being incomparable. When searching, you're likely to find situations where epsilon is greater but delta is smaller, or vice versa, resulting in a failure to check.

python/src/opendp/context.py
Shoeboxam: I have another PR that refactors the measures: #1768. Will follow-up in work related to that.

rust/src/measurements/make_private_lazyframe/group_by/mod.rs
```diff
@@ -21,136 +33,184 @@ use polars_plan::prelude::GroupbyOptions;
...
+    let threshold = if margin.public_info.is_some() {
+        None
+    } else if let Some((name, threshold_value)) = expr_threshold {
+        let noise = find_len_expr(&dp_exprs, Some(name.as_str()))?.1;
+        Some((name, noise, threshold_value))
+    } else if let Some(threshold_value) = threshold {
+        let (name, noise) = find_len_expr(&dp_exprs, None)?;
+        Some((name, noise, threshold_value))
+    } else {
+        // TODO: link to documentation page showing how to filter.
```
mccalluc, line 116: I like to include an issue URL on TODO notes, but if you've got this in your head, not critical.

rust/src/measurements/make_private_lazyframe/group_by/mod.rs
```diff
@@ -21,136 +33,184 @@ use polars_plan::prelude::GroupbyOptions;
...
 
-#[cfg(test)]
-mod test;
+macro_rules! impl_measure_non_catastrophic {
+    ($ty:ty) => {
+        impl ApproximateMeasure for $ty {
+            fn add_delta(_d_out: Self::Distance, _delta_p: f64) -> Fallible<Self::Distance> {
+                return fallible!(
+                    MakeMeasurement,
+                    "key-sets cannot be privatized under {:?}. Approximate-DP is necessary.",
```
mccalluc, line 187:
```suggestion
                    "key-sets cannot be privatized under {:?}. FixedSmoothedMaxDivergence is necessary.",
```

Not 100% sure that the suggestion is correct, but the idea that the suggested alternative is also a `*Divergence`... unless you think that will be obvious?

rust/src/measurements/make_private_lazyframe/group_by/mod.rs
```diff
@@ -21,136 +33,184 @@ use polars_plan::prelude::GroupbyOptions;
...
+fn integrate_discrete_noise_tail(
+    distribution: Distribution,
+    scale: f64,
+    tail_bound: u32,
+) -> Fallible<f64> {
+    match distribution {
+        Distribution::Laplace => integrate_discrete_laplacian_tail(scale, tail_bound),
+        Distribution::Gaussian => fallible!(
+            MakeMeasurement,
+            "gaussian tail bounds are not currently implemented"
```
mccalluc, line 213: Perhaps include an issue URL here, so if someone hits this they can add their usecase where we'll see it?

rust/src/measurements/make_private_lazyframe/group_by/pseudocode/make_private_group_by.py
```diff
@@ -0,0 +1,101 @@
...
+    input_, keys, aggs, expr_threshold = match_group_by(plan)
+
+    t_prior = input_.make_stable(input_domain, input_metric)
+    middle_domain, middle_metric = t_prior.output_space()
+
+    grouping_columns = match_grouping_columns(keys)
+
+    margin = middle_domain \
+        .margins \
+        .get(grouping_columns, Margin.default())
```
mccalluc, line 19:
```suggestion
    margin = (middle_domain
        .margins
        .get(grouping_columns, Margin.default()))
```

mccalluc (commented):
A few more little comments. Only big concern is, if this is an API change that users need to worry about, how we communicate that to them, either in the changelog, or in an error message.

rust/src/measurements/make_private_lazyframe/group_by/pseudocode/make_private_group_by.py
```diff
@@ -0,0 +1,101 @@
...
+
+    if margin.public_info is not None:
+        threshold = None
+    elif expr_threshold is not None:
+        name, threshold = expr_threshold
+        plugin = find_len_expr(dp_exprs, name.as_str())[1]
+        threshold = name, plugin, threshold
+    elif threshold is not None:
+        name, plugin = find_len_expr(dp_exprs, None)
+        threshold = name, plugin, threshold
```
mccalluc, line 51: I find the redefinition of variables, particularly `threshold`, hard to follow here, particularly with the same variables used in the conditionals. Lower down it looks like there's a distinction between `threshold_value` and `threshold` and maybe that is applicable here?

rust/src/measurements/make_private_lazyframe/group_by/make_private_group_by.tex
```diff
@@ -0,0 +1,125 @@
...
+
+Start by establishing proof properties of the following three variable.
+
+\begin{itemize}
+    \item By the postcondition of \texttt{StableDslPlan.make\_stable}, \texttt{t\_prior} is a valid transformation.
+    \item By the postcondition of \texttt{match\_grouping\_columns}, \texttt{grouping\_columns} holds the names of the grouping columns.
+        \texttt{margin} denotes what is considered public information about the key set, 
+        pulled from descriptors in the input domain.
+        By the postcondition of \texttt{make\_basic\_composition}, 
+    \item \texttt{m\_exprs} is a valid measurement that prepares a batch of expressions that,
```
mccalluc, line 49:
```suggestion
    \item By the postcondition of \texttt{make\_basic\_composition}, 
        \texttt{m\_exprs} is a valid measurement that prepares a batch of expressions that, 
```

rust/src/measurements/make_private_lazyframe/group_by/make_private_group_by.tex
```diff
@@ -0,0 +1,125 @@
...
+In the case where grouping keys are considered public,
+no thresholding is performed.
+In the setting when grouping keys are considered private information,
+it is necessary to know if a suitable release for the data set length is made
+that can be used to filter sensitive keys.
+
+The function returns a DslPlan that applies each expression from \texttt{m\_exprs}
+to \texttt{arg} grouped by \texttt{keys}.
+
+The measurement for each expression expects data set distances in terms of a triple:
```
mccalluc, line 61: These are rendered indented: Not sure if that's right?

rust/src/measurements/make_private_lazyframe/group_by/make_private_group_by.bib
```diff
@@ -0,0 +1,9 @@
+@misc{rogers2023unifyingprivacyanalysisframework,
+    title={A Unifying Privacy Analysis Framework for Unknown Domain Algorithms in Differential Privacy}, 
+    author={Ryan Rogers},
+    year={2023},
+    eprint={2309.09170},
+    archivePrefix={arXiv},
+    primaryClass={cs.CR},
+    url={https://arxiv.org/abs/2309.09170},
```
mccalluc, line 8: Should there be a link rendered in the PDF? It just seems to be plain text.

mccalluc (commented):
Possible typos in tex? I know less about how this should work.

python/src/opendp/polars.py
```diff
@@ -398,25 +406,47 @@ def wrap(*args, **kwargs):
...
             """Resolve the query into a measurement."""
-            from opendp.context import PartialChain
 
             # access attributes of self without getting intercepted by Self.__getattribute__
             lf_plan = object.__getattribute__(self, "_lf_plan")
             query = object.__getattribute__(self, "_query")
             input_domain, input_metric = query._chain
+            d_in, d_out = query._d_in, query._d_out
+
+            def make(scale, threshold=None):
```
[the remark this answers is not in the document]
Shoeboxam: The default parameter makes this a suitable constructor for binary_search_chain, and is used for the non-approx-dp case.

python/test/test_polars.py
```diff
@@ -388,6 +388,41 @@ def test_polars_describe():
...
+        },
+    )
+
+    expected = pl.DataFrame(
+        {
+            "column": ["len", "A"],
+            "aggregate": ["Len", "Sum"],
+            "distribution": ["Integer Laplace", "Integer Laplace"],
+            "scale": [8.0, 8.0],
+            "threshold": [130, None]
```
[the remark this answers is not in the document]
Shoeboxam: It's a bit different: only rows for which the "len" column is greater than 130 will be retained in the output dataframe.

It is possible for the user to configure this via `.filter(pl.col("len") > my_threshold)`, but the tooling will automatically find the smallest suitable threshold based on the structure of the query. When scale is fixed, tuning the threshold only affects the delta parameter.

python/test/test_polars.py
```diff
@@ -422,3 +457,38 @@ def test_polars_collect_early():
...
+        .collect()
+    )
+
+    print(
+        context.query()
+        .group_by("B")
+        .agg(pl.len().dp.noise())
+        .release()  # type: ignore[union-attr]
+        .collect()
+    )
```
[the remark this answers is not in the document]
Shoeboxam: I like it, updated. Also strengthened the test to check the query accuracy.

rust/src/accuracy/polars/mod.rs
```diff
@@ -75,21 +77,29 @@ struct UtilitySummary {
...
+    let mut utility = logical_plan_utility(&lazyframe.logical_plan, alpha, None)?;
 
     // only include the accuracy column if alpha is passed
     if alpha.is_none() {
         utility = utility.drop("accuracy")?;
     }
+    // only include the threshold column if a threshold is set
+    if utility.column("threshold")?.is_null().all() {
+        utility = utility.drop("threshold")?;
+    }
```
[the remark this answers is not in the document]
Shoeboxam: I kind of like keeping the description thin so that it is presentation-ready. As a user I'd also prefer things to error if I tried to access accuracies without setting alpha.

rust/src/accuracy/test.rs
```diff
@@ -312,6 +313,32 @@ pub fn test_empirical_discrete_gaussian_accuracy() -> Fallible<()> {
...
+    let n = 50_000;
+    let empirical_alpha = (0..n)
+        .filter(|_| m_dlap.invoke(&0).unwrap().clamp(-127, 127) > tail as i8)
+        .count() as f64
+        / n as f64;
+
+    println!("Discrete Laplace tail significance levels/alpha");
+    println!("Theoretical: {:?}", theoretical_alpha);
+    println!("Empirical:   {:?}", empirical_alpha);
+    assert!((empirical_alpha - theoretical_alpha).abs() < 1e-2);
```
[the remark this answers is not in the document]
Shoeboxam: I added notes to all of these non-deterministic tests.

rust/src/measurements/make_private_lazyframe/group_by/make_private_group_by.bib
Shoeboxam: Annoying that latex doesn't handle this automatically. I don't immediately see a setting to enable it.

python/src/opendp/polars.py
```diff
@@ -398,25 +406,47 @@ def wrap(*args, **kwargs):
...
+                # now that scale has been solved, find a suitable threshold
+                threshold = binary_search(
+                    lambda t: make(scale, t).map(d_in)[1] < d_out[1],  # type: ignore[index]
+                    T=int,
                 )
-            ).resolve()
+                
+                return make(scale, threshold)
+
+            return binary_search_chain(make, d_in, d_out, T=float)
```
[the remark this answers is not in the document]
Shoeboxam: I decorated this with more comments. It's a bit of a tricky spot in the code.

rust/src/measurements/make_private_lazyframe/group_by/pseudocode/make_private_group_by.py
```diff
@@ -0,0 +1,101 @@
...
+             expr,
+             global_scale,
+        ) for expr in aggs
+    ])
+
+    # reconcile information about the threshold |\label{reconcile-threshold}|
+    dp_exprs = m_exprs.invoke((input_, all()))
+
+    if threshold is not None and expr_threshold is not None:
+        raise "two thresholds set"
```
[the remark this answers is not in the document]
Shoeboxam: I dropped this validation, so that people can still filter outputs through the context API, even when not necessary for privacy. This can give some nice computational benefits.

rust/src/measurements/make_private_lazyframe/group_by/pseudocode/make_private_group_by.py
Shoeboxam: Refactored for clarity.

mccalluc:
CI:
```
diff --git a/python/src/opendp/accuracy.py b/python/src/opendp/accuracy.py
index b43eb47..42c11cd 100644
--- a/python/src/opendp/accuracy.py
+++ b/python/src/opendp/accuracy.py
@@ -184,7 +184,7 @@ def describe_polars_measurement_accuracy(
 ):
     r"""Get noise scale parameters from a measurement that returns a OnceFrame.
 
-    If a threshold is configured for censoring small/sensitive partitions, 
+    If a threshold is configured for censoring small/sensitive partitions,
     a threshold column will be included,
     containing the cutoff for the respective count query being thresholded.
```
(There's whitespace at the end of the line which is making a difference.)

Thanks for the explanations: I'll re-review and approve.

python/test/test_polars.py
```diff
@@ -477,6 +478,25 @@ def test_polars_threshold():
...
+        "aggregate": ["Len"],
+        "distribution": ["Integer Laplace"],
+        "scale": [2.0]
+    })
+
+    # check that no threshold is set when keys are known
+    pl_testing.assert_frame_equal(actual, expected)
+
+    # check that query runs.
+    # output should be two columns ("A" and "len") with one row (1, ~1000)
```
mccalluc, line 499: Thanks for adding this. Not something to block this PR, but how would you feel about including this in another `print` statement? Particularly if a test has multiple outputs, could make it easier to align the comment with the appropriate output.

mccalluc (approved):
Thanks! Always more I could poke at, but if Gurman's work depends on this, probably makes sense to merge it to `main`, and recruit Vikrant for another look at the proofs when he's available?

Shoeboxam:
I opened an issue for the migration change.

PR #1727, opened 2024-06-25
Change test warnings to errors

mccalluc:
Most of the lines in
- #1639

were about fixing
- #1619

but that got fixed elsewhere, which is good, since #1639 never made it all the way to main... but making warnings into errors is useful on its own, so I'm filing a separate PR.


Shoeboxam (approved):
Stronger checks!

PR #1728, opened 2024-06-25
Move plugins from user-guide to contributing

### LINKED ISSUE #1725: Plugins should only be in Contributors section
They really shouldn't be touched by most end users -- Instead, they are on the path the leads to a library contribution.

There is currently a reference in the User Guide that should be cleaned up.

mccalluc:
- Fix #1725


Shoeboxam:
Upon more thinking, I believe I was wrong in stating that the plugins should only be in the contributor documentation.

We have four levels of documentation:

1. Getting Started: suitable for non-technical users
2. User Guide: assumes the user is somewhat technical
3. Theory: assumes the user is comfortable with reading mathematics
4. Contributor Guide: assumes the user is wanting to make a library contribution

I could certainly see the plugin APIs being useful for a technical user who wants to plug a private or stable primitive in, but not make a contribution. So the current location of the plugin documentation is correctly placed. Sorry for the mis-direction!

mccalluc:
No worries! It was good to clarify this.

PR #1730, opened 2024-06-25
Add a little more explanation for honest-but-curious

### LINKED ISSUE #1227: Explain `honest-but-curious`
On https://docs.opendp.org/en/stable/contributor/development-environment.html:

> `honest-but-curious`: Enable to include constructors that are only private if the constructor arguments are honest.

In this context, it's a little mysterious. Would giving examples help? Perhaps:

> For example, user defined functions require this flag because privacy depends on the stability specified by the user. Similarly, population amplification relies on an honest population size.

mccalluc:
- Fix #1227

### DIFF
--- a/docs/source/contributing/development-environment.rst
+++ b/docs/source/contributing/development-environment.rst
@@ -74,8 +74,8 @@ Setting a feature changes how the crate compiles:
           - Enable to include constructors that have not passed the vetting process.
         * - ``honest-but-curious``
           - Enable to include constructors that are only private if the constructor arguments are honest.
-            For example, user defined functions require this flag because privacy depends on the stability specified by the user.
-            Similarly, population amplification relies on an honest population size.
+            For example, population amplification relies on the assumption that the user is honest (and correct) in saying that the dataset under observation is a simple sample.
+            Similarly, library plugins rely on the assumption that the privacy or stability of a user-defined function is governed by a user-defined map.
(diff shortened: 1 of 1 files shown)


docs/source/contributing/development-environment.rst
```diff
@@ -74,8 +74,10 @@ Setting a feature changes how the crate compiles:
           - Enable to include constructors that have not passed the vetting process.
         * - ``honest-but-curious``
           - Enable to include constructors that are only private if the constructor arguments are honest.
+            For example, user defined functions require this flag because privacy depends on the stability specified by the user.
+            Similarly, population amplification relies on an honest population size.
```
Shoeboxam, line 78:
```suggestion
            For example, population amplification must assume the user is honest and correct in their assertion that the dataset under observation is a simple sample.
            Similarly, library plugins must assume that the privacy or stability of a given function is governed by the given map.
```

Shoeboxam (changes requested):
I moved pop amplification first, as plugins are not quite as user-facing. Nitted wording of amplification because it mainly depends on the sample being a simple sample.

Thanks!

docs/source/contributing/development-environment.rst
mccalluc: Sounds good. Thanks!

Shoeboxam (approved):
Great!

PR #1735, opened 2024-06-26
docs: Noise comparison

GKD-stack:
This walks the user through the different types of noise, explores use cases where one may be better suited, and then demonstrates how to specify them in the context API. 

This should be merged first: https://github.com/opendp/opendp/pull/1812
https://github.com/opendp/opendp/pull/1834
https://github.com/opendp/opendp/pull/1836

this is a rough draft and I did have these questions: 
1. is rho generally 0.5?

these are polars questions mostly
1. why does adding a column with with_columns and then casting it to a string cause a dtype error with defining the margin? 
2. trying to do pl.sum also causes a margin error
`(context.query().fill_nan(40.0).fill_null(40.0).group_by("ILOSTAT").agg(pl.sum("HWUSUAL").dp.laplace().alias("Laplace Noise"))).release().collect()`

`MakeMeasurement("Failed to find margin for {"ILOSTAT"}")`

but the margin is defined here

`context = dp.Context.compositor(
    data=df,
    split_evenly_over=10,
    margins={
        ("ILOSTAT", ): dp.Margin(public_info="keys", max_partition_length=60_000_000),
        ("HWUSUAL", ): dp.Margin(public_info="keys", max_partition_length=60_000_000),
        (): dp.Margin(max_partition_length=60_000_000),
    },
)`

(The HWUSUAL margin may not be included in the version uploaded, but it doesn't work either way)

Thanks!


docs/Noise.ipynb
```diff
@@ -0,0 +1,467 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 10,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "/var/folders/dg/nlwnz_jj7rsblr2tzwj962b40000gn/T/ipykernel_68933/709051719.py:11: DeprecationWarning: `apply` is deprecated. It has been renamed to `map_elements`.\n",
```
mccalluc, line 330: Repeating a comment made elsewhere, we want notebooks to run without warnings... and if that's not possible, we should comment to explain why, and have an issue.

docs/Noise.ipynb
```diff
@@ -0,0 +1,467 @@
...
+   "source": [
+    "noise_values = laplace_values.join(gaussian_values, on=[\"ILOSTAT\"])\n",
+    "all_values = actual_values.join(noise_values, on=[\"ILOSTAT\"])\n",
+    "\n",
+    "labor_status_codes = {\n",
+    "    9: 'NA',\n",
+    "    2: 'Not Working But Employed',\n",
+    "    1: 'Working for Pay', \n",
+    "    3: 'Laid Off'\n",
+    "}\n",
```
mccalluc, line 374: Great! But where does this come from? Source URL in comment.

mccalluc (commented):
I think Mike also would have comments about the presentation of this information. I would defer to his judgement, but my comments might have some useful points on writing documentation more generally, even if you don't end up using them here.

mccalluc:
Also, note
- #1774

We'll want this in a new directory in the docs. I'll make a separate PR for that, just to get the ball rolling.

docs/Noise.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "## Introduction"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In the examples up to now, we've been using `.noise()` to make our queries differentially private. However, there are different types of noise available including Laplacian noise and Gaussian noise. In this notebook, we will describe how the two noise distributions differ and how can one choose which is more appropriate for your particular analysis. \n",
+    "\n",
+    "The default `.noise()` by OpenDP chooses one of these distributions for you based on the limited information you've provided. However, with a better understanding of both distributions, you'll be able to specify the type of noise and its parameters for higher utility!"
```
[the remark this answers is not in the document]
Shoeboxam: More specifically, it chooses based on the privacy measure. If pure-DP, then laplace, if zCDP, then gaussian.

docs/Noise.ipynb
```diff
@@ -0,0 +1,467 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The Laplace and Gaussian distributions are the two primary mechanisms to add noise to ensure differential privacy and preserve relationships for statistical analysis. The two distributions may appear similar at a first glance, but there are key differences in their use in differential privacy techniques. \n",
+    "\n",
+    "**1. Definition of Privacy:**\n",
+    " Differential privacy mechanisms using the Laplace distribution satisfy the pure ε-dp, whereas differentially private outputs using the Gaussian mechanism satisfy a more relaxed definition of differential privacy, ε-δ dp. Therefore, Laplace provides stronger privacy guarantees. \n",
```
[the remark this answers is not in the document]
Shoeboxam: There is actually a notion of "strength" of privacy measures. I would put this more along the lines that pure DP is a stronger definition of privacy than approximate-DP, and since the gaussian mechanism cannot satisfy pure-DP, it's categorically different which can make it difficult to compare the two mechanisms fairly.

Not so relevant to the notebook, but more broadly, you can think of the strength of a privacy measure as a kind of partial ordering. All pure-DP guarantees can be converted into approx-DP guarantees, but not vice-versa. This is why pure-DP is "stronger" than approx-DP. [The following graph is not complete](https://mermaid.live/edit#pako:eNp1kLsKgzAUhl8lHOhWKWR0KEg7dgjpZuIQ9NgKaiQk9CK-e62JthaaKf8lfD_pIdcFQgxlrW_5VRlLTly2ZDxMMGcwI1G0J6l4Ho4s80E6WVxwbB_Vyku84F4IZnRZ1bgro-XtZkPYXA0YKpKuM_oeSLOMvonUI5dshaYr9k_n_xRGP0OCMxmMLlPTMGkxeNgBW2jQNKoqxq_r37EEe8UGJcTjtcBSudpKkO0wVpWz-vxoc4itcbgF1xXK4rFSF6Mabw4vpWVwFw), but gives a sense of the relationships between some of the common privacy measures.

docs/Noise.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "\n",
+    "**2. Accuracy:**\n",
+    "Differentially private outputs using the Gaussian mechanism are often further from the true answer compared to outputs using the Laplace mechanism.\n",
+    "\n",
+    "This is supported by the following plot which graphs the emperical probability density function of the Laplace and Gaussian mechanisms with ε = 1 and δ = 10^-5. \n",
+    "\n",
+    "![Histogram of Laplace and Gaussian](https://programming-dp.com/_images/3022199fd49e470e2f86adb6c633df9f586459106067366915cd578cd7e05a5d.png)\n",
```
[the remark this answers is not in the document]
Shoeboxam: I agree, we definitely shouldn't be pulling assets from other resources.

```python
import matplotlib.pyplot as plt
import numpy as np

def laplace_pdf(b, mu=0):
    return lambda x: np.exp(-abs(x - mu) / b) / (2 * b)

def gaussian_pdf(b, mu=0):
    return lambda x: np.exp(-((x - mu) / b) ** 2 / 2) / (np.sqrt(2 * np.pi) * b)


scale = 1.
cont_support = np.linspace(-5, 5, num=1000)
plt.plot(cont_support, laplace_pdf(scale)(cont_support), label="M(\cdot) ~ Laplace")
plt.plot(cont_support, gaussian_pdf(scale)(cont_support), label="M(\cdot) ~ Gaussian")
plt.xlabel("support ($y$)")
plt.ylabel("$\Pr[M(0) = y]$")
plt.title("Laplace vs. Gaussian Distributions")
plt.legend()
plt.show()
```

docs/Noise.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "\n",
+    "\n",
+    "3. Type of Output: \n",
+    "Gaussian mechanisms perform better when the output is a vector of real numbers rather than a single real number. This is because the Laplace mechanism requires use of the L1 norm (sum of absolute values) and results in a sensitivity of k (length of output vector), whereas the Gaussian mechanism allows the use of the L1 norm or the L2 norm (Euclidean distance). With the Gaussian mechanism and L2 norm, the sensitivty can be $\\sqrt{k}$ In such cases, the L2 sensitivity is significantly lower than the L1 sensitivity. Therefore, the Gaussian mechanism allows for less noise to be added and results in higher utility. \n",
+    "\n",
+    "To learn more about the methodology behind the sensitivity calculation, read [chapter 6](https://programming-dp.com/ch6.html)!\n",
```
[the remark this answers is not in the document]
Shoeboxam: Agree with Chuck about using newlines, it makes it much easier to review.

The gaussian mechanism doesn't necessarily perform better than the laplace mechanism for a vector-- in the case where the L1 sensitivity is equal to the L2 sensitivity, the laplace mechanism provides overall lower variance.

Gaussian utility is better particularly in the cases where:
1. the L2 sensitivity is smaller than L1 (and discussing when this might be the case would be interesting)
2. there are many compositions

docs/Noise.ipynb
```diff
@@ -0,0 +1,467 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "context = dp.Context.compositor(\n",
+    "    data=df,\n",
+    "    privacy_unit=dp.unit_of(contributions=36),\n",
```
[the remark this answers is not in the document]
GKD-stack: im leaving the adding a link to the introduction page HTML comment in markdown as with the other notebooks

docs/Noise.ipynb
```diff
@@ -0,0 +1,467 @@
...
+   "metadata": {},
+   "source": [
+    "### Using Gaussian"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "To use the Gaussian mechanism for adding noise, we specify a different parameter, `rho`, instead of `epsilon`. Similar to how `epsilon` is usually set to one, `rho` is usually set to 0.5. "
```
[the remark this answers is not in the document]
GKD-stack: ok, im leaving the todo in the markdown to insert a link

docs/Noise.ipynb
```diff
@@ -0,0 +1,467 @@
...
+   ],
+   "source": [
+    "context = dp.Context.compositor(\n",
+    "    data=df,\n",
+    "    privacy_unit=dp.unit_of(contributions=36),\n",
+    "    privacy_loss=dp.loss_of(rho=0.5),\n",
+    "    split_evenly_over=10,\n",
+    "    margins={\n",
+    "        (\"ILOSTAT\", ): dp.Margin(public_info=\"keys\", max_partition_length=60_000_000),\n",
+    "        (\"HWUSUAL\", ): dp.Margin(public_info=\"keys\", max_partition_length=60_000_000),\n",
```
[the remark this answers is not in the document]
GKD-stack: nope, took it out for hours usual

mccalluc (approved):
Tests pass, and the explanation seems good. There might be more edits before publication, but I'll go ahead with merge.

docs/source/getting-started/assessing-utility/noise.ipynb
```diff
@@ -0,0 +1,527 @@
...
+   "metadata": {},
+   "source": [
+    "## Introduction"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Thus far, we've been using `.noise()` to make our queries differentially private. However, there are different types of noise available including Laplacian noise and Gaussian noise. In this notebook, we will describe how the two noise distributions differ and how you can choose one which is most appropriate for your particular analysis. \n",
```
vikrantsinghal, line 21: "...different types of distributions available to sample the noise from, including the Laplacian and the Gaussian distributions."

docs/source/getting-started/assessing-utility/noise.ipynb
```diff
@@ -0,0 +1,527 @@
...
+    "\n",
+    "The default `.noise()` method chooses one of these distributions for you based on the limited information you've provided, specifically the privacy measure. If the privacy measure requires pure-DP to be satisfied, then the Laplace distribution will be used for noise. Otherwise, the Gaussian distribution may be used to satisfy zero concentrated DP. However, with a better understanding of both distributions, you can specify the type of noise and its parameters for higher utility!"
+   ]
+  },
+  {
+   "cell_type": "code",
```
vikrantsinghal, line 27: Mention just before this code snippet what this is doing. Could be as simple as, "As always we do the blah blah thing in the following."

docs/source/getting-started/assessing-utility/noise.ipynb
```diff
@@ -0,0 +1,527 @@
...
+   "metadata": {},
+   "source": [
+    "## Laplace vs. Gaussian"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The Laplace and Gaussian distributions are the two primary mechanisms to add noise to ensure differential privacy and preserve relationships for statistical analysis. The two distributions may appear similar at first glance, but there are key differences in their use in differential privacy techniques. \n",
```
vikrantsinghal, line 53: They don't really appear similar, to be honest. They have pretty different tail properties, so can't really say that.

docs/source/getting-started/assessing-utility/noise.ipynb
```diff
@@ -0,0 +1,527 @@
...
+    "\n",
+    "**1. Definition of Privacy:**\n",
+    " Differential privacy mechanisms using the Laplace distribution satisfy the pure ε-dp, whereas differentially private outputs using the Gaussian mechanism satisfy a more relaxed definition of differential privacy, approximate DP. Therefore, Laplace provides more precise privacy guarantees.\n",
```
vikrantsinghal, line 56: - "satisfy (pure) ε-DP, whereas..."
- What does precise guarantee mean? I think approx DP is also a "precise" guarantee. Might just want to say that Laplace provides a "stronger" privacy guarantee.

docs/source/getting-started/assessing-utility/noise.ipynb
```diff
@@ -0,0 +1,527 @@
...
+    "plt.show()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "**3. Type of Output:**\n",
+    "\n",
+    "Gaussian mechanisms perform better than Laplace mechanisms particularly when \n",
```
vikrantsinghal, line 105: "when:"

docs/source/getting-started/assessing-utility/noise.ipynb
```diff
@@ -0,0 +1,527 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current\n",
```
vikrantsinghal, line 158: Again, this looks untidy. Enclose it in a box or something because this is either code or some output.

docs/source/getting-started/assessing-utility/noise.ipynb
```diff
@@ -0,0 +1,527 @@
...
+    ")\n",
+    "\n",
+    "laplace_values"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "To use the Gaussian mechanism for adding noise, we specify a different parameter, `rho`, instead of `epsilon`; `rho` is set to 1 for this tutorial. See `TODO` for a fuller discussion of setting a privacy budget. "
```
vikrantsinghal, line 322: One issue here is that we're talking about ρ-zCDP in this line without giving any context on that. Previously, we only talk about approx DP, but here in the tutorial, we don't really do that, but use zCDP, instead.

GKD-stack: I took out this line from the introduction: Otherwise, the Gaussian distribution may be used to satisfy zero concentrated DP...

is that what you meant?

PR #1763, opened 2024-07-03
docs: Composition

### LINKED ISSUE #1762: Notebook: Composition
go over sequential and parallel comp, interactivity, privacy interps

GKD-stack:
Hey! I had a question on the parameters, privacy budget. I noticed when using 'make_sequential_comp' the budget had to be a lot higher for the same scale noise values. so I would end up increasing the noise values and the budget. is there another way to fix this?

- fix #1762


docs/composition.ipynb
```diff
@@ -0,0 +1,739 @@
...
+   "source": [
+    "#construct measurement for median\n",
+    "discrete_scores = dp.t.make_quantile_score_candidates(dp.vector_domain(dp.atom_domain(T=float)), \n",
+    "                                    dp.symmetric_distance(), \n",
+    "                                    np.array(candidates).astype(float), \n",
+    "                                    0.5)\n",
+    "input_space = dp.vector_domain(dp.atom_domain(T=dp.usize), 31), dp.linf_distance(T=dp.usize)\n",
+    "select_index_measurement = dp.m.make_report_noisy_max_gumbel(*input_space, scale=1.0, optimize='min')\n",
+    "\n",
+    "median_meas = discrete_scores >> select_index_measurement >> (lambda index: candidates[index])\n",
```
mccalluc, line 372: Getting down to the noisy_max_gumbel distribution, etc. At other points, this is an implementation detail the user doesn't need to see. It's not clear to me here if you're using it to demonstrate how things work at a lower level... or if this is something that actually would be the best solution for some usecase.

docs/composition.ipynb
```diff
@@ -0,0 +1,739 @@
...
+    }
+   ],
+   "source": [
+    "#get the domain of the datafarame\n",
+    "df_domain = dp.domain_of(df, infer=True)\n",
+    "#specify the parameters when grouped by Working cateory\n",
+    "df_domain = dp.with_margin(df_domain, by=[\"ILOSTAT\"], \n",
+    "                           public_info=\"lengths\", \n",
+    "                           max_partition_length=50, \n",
+    "                           max_num_partitions=3)\n",
```
mccalluc, line 417: Comment / justify parameters

docs/composition.ipynb
```diff
@@ -0,0 +1,739 @@
...
+    "print(\"DF Domain: \",df_domain)\n"
```
mccalluc, line 418: Generally, I do like to show the intermediate object... but here it's so big that I worry it would be a distraction to the user.

docs/composition.ipynb
```diff
@@ -0,0 +1,739 @@
...
+     ]
+    }
+   ],
+   "source": [
+    "#get the domain of the datafarame\n",
+    "df_domain = dp.domain_of(df, infer=True)\n",
```
mccalluc, line 412: is infer required?

docs/composition.ipynb
```diff
@@ -0,0 +1,739 @@
...
+    "#specify the parameters when grouped by Working cateory\n",
+    "df_domain = dp.with_margin(df_domain, by=[\"ILOSTAT\"], \n",
```
mccalluc, line 414: Generally prefer giving it a new name if it's a new thing... but if it feels to you like we're modifying the preceding instance, it's fine this way, too.

docs/composition.ipynb
```diff
@@ -0,0 +1,739 @@
...
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "#define polars queries\n",
+    "proposed_plan = context.query().groupby(\"ILOSTAT\").agg(\n",
+    "    pl.col(\"HWUSUAL\").fill_null(40.0).dp.mean((1,98), scale = 8).alias(\"Mean Hours Worked\"), \n",
+    "    pl.col(\"HWUSUAL\").fill_null(40.0).dp.median(candidates, 8).alias(\"Median Hours Worked\"),\n",
+    ")\n",
+    "proposed_plan"
```
mccalluc, line 455: Again, this one is hardy readable, so might leave it out.

docs/composition.ipynb
```diff
@@ -0,0 +1,739 @@
...
+       "└─────────┴───────────────────┴─────────────────────┴──────────────┴───────────────────────────┘"
+      ]
+     },
+     "execution_count": 5,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "candidates = list(range(25,56))\n",
```
mccalluc, line 219: Maybe not on this page, but where we first introduce candidates for quantiles, do we explain that they don't need to be every integer in a range, and that that could be a better tradeoff in some cases?

docs/composition.ipynb
```diff
@@ -0,0 +1,739 @@
...
+      "│ i64     ┆ f64               ┆ i64                 │\n",
+      "╞═════════╪═══════════════════╪═════════════════════╡\n",
+      "│ 1       ┆ 37.671921         ┆ 29                  │\n",
+      "└─────────┴───────────────────┴─────────────────────┘\n",
+      "<opendp.polars.OnceFrame object at 0x12ea6f740>\n"
+     ]
+    }
+   ],
+   "source": [
+    "# IN SERVER\n",
```
mccalluc, line 487: Are we shifting from explorative code running locally to code that could run in the secure environment?

docs/composition.ipynb
```diff
@@ -0,0 +1,739 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Parallel composition can be viewed as an alternative to sequential composition since it's another way to calculate a bound on the total privacy cost of multiple queries. \n",
+    "\n",
+    "Parallel composition involves partitioning the input dataset into disjoint partitions and applying a differentially private mechanism on each chunk independent of all other chunks. Assuming each individual is represented once in the dataset, their data will appear in exactly one chunk. If there are $k$ chunks, there will be $k$ runs of the differentially private mechanisms.\n",
+    "\n",
+    "To apply parallel composition, we need to set up our margin or define our query to allow each individual to influence at most 1 partition, or appear at most in 1 chunk. \n"
```
mccalluc, line 528: "... and this is different from earlier in this notebook, where we assumed 36 contributions" ... or something like that.

docs/composition.ipynb
```diff
@@ -0,0 +1,739 @@
...
+   "metadata": {},
+   "source": [
+    "### Noninteractive Parallel Composition"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In the following example, each individual is represented just one each year, quarter, and labor status. Moreover, this example also satisfies the properties of noninteractive sequential composition since all the queries are executed simultaneously. "
```
mccalluc, line 542:
```suggestion
    "In the following example, each individual is represented just once for each combination of year, quarter, and labor status. Moreover, this example also satisfies the properties of noninteractive sequential composition since all the queries are executed simultaneously. "
```

docs/composition.ipynb
```diff
@@ -0,0 +1,739 @@
...
+       "│ 2004 ┆ 2       ┆ 1       ┆ 41.136433         ┆ 55                  │\n",
+       "│ 2004 ┆ 3       ┆ 1       ┆ 37.531945         ┆ 55                  │\n",
+       "│ 2004 ┆ 4       ┆ 1       ┆ 37.375663         ┆ 55                  │\n",
+       "│ 2005 ┆ 1       ┆ 1       ┆ 35.841911         ┆ 55                  │\n",
+       "│ …    ┆ …       ┆ …       ┆ …                 ┆ …                   │\n",
+       "│ 2012 ┆ 4       ┆ 1       ┆ 38.014647         ┆ 25                  │\n",
+       "│ 2013 ┆ 1       ┆ 1       ┆ 37.630491         ┆ 25                  │\n",
+       "│ 2013 ┆ 2       ┆ 1       ┆ 35.786751         ┆ 55                  │\n",
+       "│ 2013 ┆ 3       ┆ 1       ┆ 38.219257         ┆ 55                  │\n",
+       "│ 2013 ┆ 4       ┆ 1       ┆ 37.850275         ┆ 25                  │\n",
```
mccalluc, line 579: Any chance you have a sign problem? Instead of picking the median, it seems to be picking one of the extremes.

mccalluc (commented):
Thanks! As usual, lots of little nitpicks. At the very end of the notebook, the medians don't look right... I'm guessing there's a sign problem?

docs/composition.ipynb
```diff
@@ -0,0 +1,739 @@
...
+    "    privacy_unit=dp.unit_of(contributions=36),\n",
+    "    privacy_loss=dp.loss_of(epsilon=1.0),\n",
+    "    split_evenly_over=10,\n",
+    "    margins={\n",
+    "        (\"ILOSTAT\", ): dp.Margin(public_info=\"lengths\", max_partition_length=60_000_000, max_num_partitions=3),\n",
+    "        (\"YEAR\", \"QUARTER\", \"ILOSTAT\",): dp.Margin(public_info=\"lengths\", max_partition_length=60_000_000, max_partition_contributions=1, max_num_partitions=1),\n",
+    "        (): dp.Margin(public_info=\"lengths\", max_partition_length=60_000_000, max_num_partitions=1),\n",
+    "    },\n",
+    ")"
+   ]
```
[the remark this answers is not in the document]
GKD-stack: add a link to the introduction in tabular data that talks about context

docs/composition.ipynb
```diff
@@ -0,0 +1,739 @@
...
+    "\n",
+    "1. **Noninteractive:** \n",
+    "\n",
+    "All queries are composed and executed at once. Specifically, take $F$ as a noninteractive algorithm composed of individual queries $F = \\text{Comp}(F_1, F_2, \\ldots, F_{k-1})$ and dataset x as the input dataset. $F(x)$ is our output, an algorithm that executes all the queries. \n",
+    "\n",
+    "2. **Interactive:** \n",
+    "\n",
+    "In an interactive algorithm, the queries are not defined in advance, and they can be determined based on the results from previous queries. This is an **adaptive** scenario. \n",
+    "\n",
+    "A non-adaptive scenario is where you simply want to compute statistics such as the mean, median, and sum of particular columns in a dataset. An adaptive scenario is where you start computing particular statistics about a column, such as \"income,\" and then change your queries depending on the results you find. For example, if you notice that the mean and median are extremely different, you may explore the distribution of the data. Otherwise, you may choose to focus your queries on other, more interesting aspects of your data. \n",
```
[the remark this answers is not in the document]
GKD-stack: i believe its a interactivity is a superset, made the distinction more clear

docs/composition.ipynb
```diff
@@ -0,0 +1,739 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We can execute multiple queries at once using the `select` or `agg` methods. To start off, let's calculate the actual and differentially private mean and median for the variable `HWUSUAL`, which is the number of hours per week usually worked in the main job by respondents that were working for pay at the time of the survey.\n",
+    "\n",
+    "Notes: \n",
+    "- Remember to reference the [EU Labour Force Survey User Guide](https://ec.europa.eu/eurostat/documents/1978984/6037342/EULFS-Database-UserGuide.pdf) for any questions about variables. \n",
+    "- Do not calculate the actual mean to impute null values. We shouldn't reference our data after creating the compositor. "
```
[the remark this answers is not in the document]
GKD-stack: notes for user!

docs/composition.ipynb
```diff
@@ -0,0 +1,739 @@
...
+    "    pl.col(\"HWUSUAL\").fill_null(40.0).mean().alias(\"Mean Hours Worked\"), \n",
+    "    pl.col(\"HWUSUAL\").fill_null(40.0).median().alias(\"Median Hours Worked\"),\n",
+    ").collect()\n",
+    "\n",
+    "dp_hours = context.query().groupby(\"ILOSTAT\").agg(\n",
+    "    pl.col(\"HWUSUAL\").fill_null(40.0).dp.mean((1,98), scale = 1/97).alias(\"Mean Hours Worked\"), \n",
+    "    pl.col(\"HWUSUAL\").fill_null(40.0).dp.median(candidates, 1/97).alias(\"Median Hours Worked\"),\n",
+    ").release().collect()\n",
+    "\n",
+    "hours_cat_df = actual_hours.join(dp_hours, on=[\"ILOSTAT\"]).filter(pl.col(\"ILOSTAT\")==1)\n",
```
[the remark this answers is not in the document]
GKD-stack: yep, i those other labor codes are usually people not working so they have a lot of 99 null values

docs/composition.ipynb
```diff
@@ -0,0 +1,739 @@
...
+   "source": [
+    "#define the compositor by specifying parameters\n",
+    "sc_meas = dp.c.make_sequential_composition(\n",
+    "    input_domain=dp.vector_domain(dp.atom_domain(T=float)),\n",
+    "    input_metric=dp.symmetric_distance(),\n",
+    "    output_measure=dp.max_divergence(T=float),\n",
+    "    #d_in is the upper bound on distances\n",
+    "    d_in=1,\n",
+    "    #d_mids is the privacy consumption for each query\n",
+    "    d_mids=[1., 1., 2.]\n",
```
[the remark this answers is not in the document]
GKD-stack: first bullet was mixed a prior PR. and I clarified the second in the introduction

docs/composition.ipynb
```diff
@@ -0,0 +1,739 @@
...
+    ")\n",
+    "\n",
+    "#this is the privacy consumption of the entire composition\n",
+    "print(\"Privacy Consumption of Entire Composition: \" + str(sc_meas.map(1)))\n",
+    "\n",
+    "#clean and compile data\n",
+    "hours_list = list(np.array(df.select(\n",
+    "                                    pl.col(\"HWUSUAL\").\n",
+    "                                    fill_null(40.0).\n",
+    "                                    fill_nan(40.0).\n",
```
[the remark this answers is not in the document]
GKD-stack: it would skew the results

docs/composition.ipynb
GKD-stack: it's a part of the measurement to create a DP median, so across the different types of compositors dp mean and dp median are reported, just through different methods.

docs/composition.ipynb
GKD-stack: yep, although unsure how to interpret why. when its not included, I get this: 
`TypeError: "'=='" comparison not supported for LazyFrame objects`

docs/composition.ipynb
GKD-stack: don't think so

GKD-stack:
need to delete the data file here

docs/source/getting-started/assessing-utility/composition.ipynb
```diff
@@ -0,0 +1,742 @@
...
+    "vscode": {
+     "languageId": "shellscript"
+    }
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current\n",
```
vikrantsinghal, line 101: Same comment about enclosing this in a box.

docs/source/getting-started/assessing-utility/composition.ipynb
```diff
@@ -0,0 +1,742 @@
...
+   "metadata": {},
+   "source": [
+    "## Interactive Sequential Composition"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In cases where you want to apply interactive sequential composition, you can use the `dp.c.make_sequential_composition` method. There are two ways to implement sequential composition: using transformations and measurements directly and passing in polars queries. \n",
```
vikrantsinghal, line 278: "(1) using transformations and measurements directly, and (2) passing in polars queries."

docs/source/getting-started/assessing-utility/composition.ipynb
```diff
@@ -0,0 +1,742 @@
+{
```
vikrantsinghal, line 1: One thing I will say about this. It's not fully clear through this doc what the differences are among interactive, sequential, and adaptive. Can these three things be true about a set of queries simultaneously? Can a subset of these three be true for a set of queries? Which combinations are valid. It looks a bit mixed up right now, unfortunately.

docs/source/getting-started/assessing-utility/composition.ipynb
```diff
@@ -0,0 +1,742 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 8,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Privacy Consumption of Entire Composition: 4.0\n"
```
vikrantsinghal, line 308: Enclose these kinds of outputs in a box or something again for more clarity and separation from the main text.

vikrantsinghal (changes requested):

docs/source/getting-started/assessing-utility/composition.ipynb
GKD-stack: yeah, currently i go into all of them a bit in this introduction suggestion. what combinations are valid and invalid? I could add that in

mccalluc: My sense is that 
- *mechanisms* can be described as interactive or non-interactive, 
-  but *compositions* can be described as adaptive or non-adaptive.

(Basing this on pages 104 and 151 in _Hands-on DP_.)

So, in the discussion here, maybe we only mention adaptive/non-adaptive compositions, and not mention the parallel distinction between  interactive/non-interactive mechanisms?

mccalluc: From Mike:

> Mechanisms can be interactive or non-interactive. The difference between them is that interactivity allows an adversary to gain more information via communication with the output. We technically always consider measurements in the library to be interactive mechanisms, and in the case where the mechanism would typically be considered non-interactive, the adversary is trivial (there are zero rounds of communication). Informally, a mechanism is non-interactive if there is no queryable in the output (nothing to communicate with).
>
> Compositors are a kind of mechanism. Compositors can also be either interactive or non-interactive. Some examples:
> - non-interactive: `make_basic_composition` (for each privacy measure), advanced composition (for pure/approx/BR-DP)
> - interactive: `make_sequential_composition` (likewise), advanced composition (likewise)
> You can even have non-interactive composition of interactive mechanisms (for example `make_basic_composition` over a bunch of interactive mechanisms).
>
> If the composition is interactive, then it may either be adaptive or concurrent.
> - If the compositor is adaptive, then there is a sequentiality constraint on queries to interactive mechanisms spawned by the compositor: only the most recent queryable spawned may be interacted with.
> - If the compositor is concurrent, then the adversary may submit multiple interactive mechanisms to the compositor, thus spawning multiple queryables, and interleave queries amongst them.
> 
> ### Book vs. Library
> - In the book, I put these three cases along one axis of "query interactivity": a compositor is either non-adaptive (corresponding to non-interactive), adaptive, or concurrent
> - In the library, I've hidden the distinction between adaptive and concurrent composition. Concurrent composition allows for everything adaptive composition does and more. So if the user wants an interactive compositor, they simply use `make_sequential_composition`, and the library allows concurrency if it has been proven. If the privacy measure doesn't support concurrent composition, then the implementation imposes a sequentiality constraint to get "adaptive" composition. If we later prove that a definition of privacy allows for concurrent composition, we can then lift the sequentiality constraint for that privacy measure without requiring any changes from the user.

mccalluc:
CI fails in the "Noninteractive Sequential Composition" section:
```
dp_hours = context.query().select(
    pl.col("HWUSUAL").fill_null(40.0).dp.mean((1,45)).alias("DP Mean Hours Worked"), 
    pl.col("HWUSUAL").fill_null(40.0).dp.median(candidates).alias("DP Median Hours Worked"),
).release().collect()

dp_hours
```

<details>

```
OpenDPException                           Traceback (most recent call last)
File ~/work/opendp/opendp/python/src/opendp/mod.py:1093, in exponential_bounds_search(predicate, T)
   1092 try:
-> 1093     predicate(center)
   1094 except Exception:

File ~/work/opendp/opendp/python/src/opendp/mod.py:873, in binary_search_param.<locals>.<lambda>(param)
    869 # one might think running scipy.optimize.brent* would be better, but 
    870 # 1. benchmarking showed no difference or minor regressions
    871 # 2. brentq is more complicated
--> 873 return binary_search(lambda param: make_chain(param).check(d_in, d_out), bounds, T)

File ~/work/opendp/opendp/python/src/opendp/mod.py:110, in Measurement.check(self, d_in, d_out, debug)
    109 try:
--> 110     return measurement_check(self, d_in, d_out)
    111 except OpenDPException as err:

File ~/work/opendp/opendp/python/src/opendp/core.py:212, in measurement_check(measurement, distance_in, distance_out)
    210 lib_function.restype = FfiResult
--> 212 output = c_to_py(unwrap(lib_function(c_measurement, c_distance_in, c_distance_out), BoolPtr))
    214 return output

File ~/work/opendp/opendp/python/src/opendp/_lib.py:254, in unwrap(result, type_)
    253         message = f'Installed python polars version ({pl.__version__}) != expected version ({_EXPECTED_POLARS_VERSION}). {message}' # pragma: no cover
--> 254 raise OpenDPException(variant, message, backtrace)

OpenDPException: 
  Overflow("(0).inf_add(inf) is not finite. Consider tightening your parameters.")
```

</details>

PR #1760, opened 2024-07-01
Polars: utility estimation

Shoeboxam:
Resolves #1759

PR #1732, opened 2024-06-25
Polars: simplify code by reducing ownership

Shoeboxam:
closes #1731

PR #1767, opened 2024-07-08
Polars: no kwargs in public APIs

### LINKED ISSUE #1761: Polars: don't use keyword arguments in public APIs
Going forward, we should stop using Polars keyword arguments in DP plugins:
* Keyword arguments need to be un-pickled, which presents a security issue for Python
* SQL has no ability to distinguish between keyword and data arguments, so keyword arguments are not supported at all
* R does not support keyword arguments, due to the pickling scheme
  - ^ **Needed for using Polars with R**

The alternative approach is to pass all arguments as data arguments, parse the data arguments into keyword arguments within the constructor, and embed the parsed info into the output lazyframe as keyword arguments

Shoeboxam:
Closes #1761

The design here is that each operation we want to add to Polars is implemented with two plugins:
* a shim, that takes no keyword arguments. What would typically be keyword arguments are passed as expressions.
* the actual plugin, which uses keyword args (and pickling). 

Users always write queries in terms of the shims. When OpenDP emits a private lazyframe from the function, it swaps the shims out with the real plugins.


Shoeboxam:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/opendp/opendp/1767?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1767** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1767?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1716** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1716?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1760** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1760?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1703** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1703?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2 other dependent PRs ([#1715](https://github.com/opendp/opendp/pull/1715) <a href="https://app.graphite.dev/github/pr/opendp/opendp/1715?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1732](https://github.com/opendp/opendp/pull/1732) <a href="https://app.graphite.dev/github/pr/opendp/opendp/1732?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`

<h2></h2>

docs/source/getting-started/examples/histograms.ipynb
```diff
@@ -191,8 +191,8 @@
     "    def privatize(s, t=1e8):\n",
     "        return preprocess >> dp.m.then_laplace_threshold(scale=s, threshold=t)\n",
     "    \n",
-    "    s = dp.binary_search_param(lambda s: privatize(s=s), d_in, d_out)\n",
-    "    t = dp.binary_search_param(lambda t: privatize(s=s, t=t), d_in, d_out)\n",
+    "    s = dp.binary_search(lambda s: privatize(s=s).map(d_in)[0] <= d_out[0])\n",
```
mccalluc, line 194: I think I've commented on this change elsewhere, but off-hand it looks like things are becoming more complicated for the user, but that might be a trade-off for something else that we need?

mccalluc (changes requested):
I haven't completed the review here because many of the changes are also in
- #1760

I'm guessing that you only want these changes in one PR or the other? Or let me know if there are some things I can safely ignore here, and instead only focus on other parts of the diff?

mccalluc:
#1340 strikes again! re-running test...

rust/src/measurements/gumbel_max/mod.rs
```diff
@@ -20,22 +18,20 @@ use crate::traits::{
...
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(
-            f,
-            "{}",
-            match *self {
-                Optimize::Min => "min",
-                Optimize::Max => "max",
-            }
-        )
+impl std::fmt::Debug for Optimize {
```
mccalluc, line 29: Is it always better to define `Debug` rather than `Display`? Or is the change particular to what's going on in this PR?

rust/src/measurements/make_private_expr/expr_index_candidates/mod.rs
```diff
@@ -72,117 +73,150 @@ static INDEX_CANDIDATES_PLUGIN: &str = "index_candidates";
...
-                        .into_iter()
-                        .map(|x| match x {
-                            Value::$dtype(v) => Ok(v),
-                            _ => Err(serde::de::Error::custom("Expected homogenous candidates")),
-                        })
-                        .collect::<Result<Vec<_>, D::Error>>()?,
-                )
-            };
-        }
+//         for value in self.0.iter() {
```
mccalluc, line 130: Generally want to avoid commented out code blocks.

rust/src/transformations/make_stable_expr/expr_discrete_quantile_score/mod.rs
```diff
@@ -68,6 +59,8 @@ where
             "Quantile estimation requires non-null inputs"
         );
     }
+    println!("{:?}", candidates.dtype());
```
mccalluc, line 62: This was probably just for debugging?
```suggestion
```

mccalluc (changes requested):
Looks like there is some code to clean up here? printlns, commented out blocks.

I see a lot of parallel changes, but I'm not really understanding at a basic level what this does. It's not essential that I understand, but if you'd like to point me at a small section to read more closely, let me know.

rust/src/measurements/gumbel_max/mod.rs
Shoeboxam: I've updated to the approach that Debug is as automatically derived by the macro (uppercase), and Display is lowercase.

rust/src/measurements/make_private_expr/expr_index_candidates/mod.rs
Shoeboxam: Agree-- I wasn't sure if it would work without the code and forgot to return to it.

Shoeboxam:
@mccalluc I updated the description of this PR to explain the general pattern.

mccalluc (commented):
One of the original motivations was:
> R does not support keyword arguments, due to the pickling scheme

but it sounds like with this new scheme, at a low-level, the pickling is still there?
> the actual plugin, which uses keyword args (and pickling).

Do these changes get us on the path where R will be more tractable? Or is this still a useful change, even if we can't check that box?

(Doing a full review now, but wanted to give a high-level observation.)

mccalluc (dismissed):
Latest changes all look good. One question about how the two-level scheme relates to a potential R implementation, but not something that should block.

Shoeboxam:
> One of the original motivations was:
> 
> > R does not support keyword arguments, due to the pickling scheme

The public API doesn't have any pickling, so the API is now compatible with R (and SQL). 

> but it sounds like with this new scheme, at a low-level, the pickling is still there?
> 
> > the actual plugin, which uses keyword args (and pickling).
The only way you could encounter the pickled data from R is if you were to extract the lazyframe from a onceframe, then serialize it into json, and then traverse the lazyframe plan down into the plugins. But R can still execute this plan that has pickled data inside of it, because it passes the entire plan to Polars to execute, and Polars can work with the pickled data.

> Do these changes get us on the path where R will be more tractable? Or is this still a useful change, even if we can't check that box?

Yes. I'm not sure if the Polars R package supports defining custom plugins, but even if they don't, I can write a helper function to do this, or open a PR on Polars R.

PR #1777, opened 2024-07-10
Polars: Teacher Workflow Update

GKD-stack:
The notebook has the changes to make. I didn't know what rst file to update so I didn't make the changes directly. 

The main questions are: 
- What are the alternatives to the .params() since its not an attribute 
- should I use the new accuracy method here?
- why does the scan_csv not read in a link to lazyframe and it has to converted from dataframe to lazyframe?


mccalluc (commented):
Assuming it's ok with you, I'll make the tweak to index.rst to fix the `.../docs/source/getting-started/teacher_draft.ipynb: WARNING: document isn't included in any toctree`? I'll also rename the file -- Maybe this work is a draft, and I'll mark it as such in github if you want, but that doesn't need to be in the filename (particularly since I'll be adding a reference.)

For context, does this replace an earlier PR, or does it fix an issue? I think the answer may be "yes" to both, but I think you have a better handle on it. Could you add references to the PR description?

I'll come back to this soon and actually look at the content of the notebook...

mccalluc:
I'm sorry, I hadn't understood what this is about: You're previewing changes that would be made, eventually, to typical-workflow, but in a notebook, since that's easier to work with. This is valuable feedback on using the API, but we should learn from this, but it's not on a path where it would be merged any time soon. Think I've got it now.

docs/source/getting-started/teacher_draft.ipynb
```diff
@@ -0,0 +1,185 @@
...
+    "        (): dp.Margin(public_info=\"lengths\", max_partition_length=8000, max_num_partitions=1),\n",
+    "    },\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We set the `max_partition_length` to be 8000 since we know our data has less than 8000 entries. The `max_num_partitions` is set to 1 since the order that our methods are using on the backgrounds requires `max_num_partitions` to be specified and it's 1 since one of our queries require only 1 partition. "
```
chikeabuah, line 99: i can't parse the phrase "the order that our methods are using on the backgrounds" what does this mean?

docs/source/getting-started/teacher_draft.ipynb
```diff
@@ -0,0 +1,185 @@
...
+   "metadata": {},
+   "source": [
+    "Based on previous examples, one of the initial ways you may try to find the count of age directly could be \n",
+    "\n",
+    "```\n",
+    "context.query().select(pl.col(\"age\").len().dp.noise()\n",
+    ").release().collect() \n",
+    "```\n",
+    "\n",
+    "However, this method won't work yet since in this case we're using `expr.len`, which isn't supported yet by OpenDP and is different than `pl.len` in polars. Therefore, we'll directly get the length of the dataframe by doing `pl.len().dp.noise()`. However, we also need to impose some type of bound, which we can do by filtering the variable of interest. "
```
chikeabuah, line 120: this seems like TMI. why are we referencing unsupported expressions in a beginner tutorial?

mccalluc:
Going back to an earlier comment:
> You're previewing changes that would be made, eventually, to typical-workflow, but in a notebook, since that's easier to work with

We actually decided to simplify typical-workflow, and that PR is ready to go: 
- #1909

so we learned from this work, but the PR can be closed.

Issue #1781, opened 2024-07-11
Add private selection algorithm with known threshold


shlomihod:
The private selection algorithm is a differentially private mechanism that repeatedly executes another DP mechanism until it achieves satisfactory performance according to a given quality score and threshold, incurring only a constant factor on the privacy loss budget. See Algorithm 1 in [Liu and Talwar (STOC, 2021)](https://arxiv.org/abs/1811.07971).

PR #1782, opened 2024-07-11
Add measurement for private selection with known threshold

shlomihod:
Implement #1781.


Shoeboxam (commented):
Thanks for opening this! I have a few higher-level questions and comments, from tinkering with this in what time I had available today.

What is the purpose of defaulting T to min_steps? Why not default to some arbitrarily large number (effectively infinity)? It doesn't seem to hurt privacy, and the utility is always better when T is larger, as far as I can tell.

In fact, if the API just takes T instead of epsilon_selection, then I think you could solve for epsilon_selection like so:
```python
    min_steps = 1 + 1 / (np.e * stop_probability)
    if steps < min_steps:
        raise ValueError(
            f"steps must be greater than or equal to {int(np.ceil(min_steps))}"
        )

    epsilon_selection = 2 / np.exp(steps * stop_probability)
    if not 0 <= epsilon_selection <= 1:
        raise ValueError(
            f"steps * stop_probability must be greater than or equal to {np.log(2)}"
        )
```

This way you'd always get a point on the pareto-curve between T and epsilon_selection, because epsilon_selection would always take full advantage of T. Dropping epsilon_selection from the API may also make it possible to generalize this API to other privacy measures, like RDP, as in https://arxiv.org/abs/2110.03620.

The three remaining arguments would then correspond to the three stopping criteria. The algorithms in the paper seem to use some combination of these three stopping criteria:
1. exceeds threshold: 3.1, 3.4
2. stopping probability: 3.1, 3.2
3. step limit: 3.1

It looks like your implementation corresponds to 3.1 if epsilon selection and stop probability are set. 
If they're not set the implementation is similar to 3.4, but the algorithm still has stopping condition 3, in that it will still terminate after `steps` runs. Does this still satisfy the privacy guarantee in 3.4?

A lot of what I'm thinking about here is trying to find a good general API for private selection, but it may be a bit tricky to find it. Just so we don't get stuck, I'm totally ok with moving forward with APIs that specifically correspond to algorithms in the paper (or modifications with proofs).

python/src/opendp/_extrinsics/_make_private_selection/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+                                     epsilon_selection: float,
+                                     steps: Optional[int] = None):
+
+    """Measurement for private selection with known threshold.
+
+    This executes an ε-DP mechanism M that returns a tuple (q, x) repeatedly
+    until the score q is at least T, a predefined threshold.
+    Only the last output of M is released.
+
+    Algorithm 1: https://arxiv.org/pdf/1811.07971.pdf#page=7
```
mccalluc, line 20: Make this an RST link with the title as the clickable text?

```suggestion
    Algorithm 1 in `Private selection from private candidates <https://arxiv.org/pdf/1811.07971.pdf#page=7>`_
```

python/src/opendp/_extrinsics/_make_private_selection/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+
+    :param meas: A measurement function that returns a 2-tuple when invoked,
+                 where the first element is a score.
```
mccalluc, line 23:
```suggestion
                 where the first element is a score, and the second is the actual output.
```
Or something to explain the second element.

python/src/opendp/_extrinsics/_make_private_selection/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+    :param threshold: The threshold score. Return immediately if the score is above this threshold.
+    :param stop_probability: The probability of stopping early at any iteration.
+    :param epsilon_selection: The epsilon allocated to the private selection.
+    :param steps: Optional. The number of steps to run. If not specified, will run the minimum number of steps.
```
mccalluc, line 27: Are `stop_probability` and `steps` about exiting early, before a solution is found? Or something else? Might also mention the particular exception is that is raised on an early exit.

python/src/opendp/_extrinsics/_make_private_selection/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+
+    dp.assert_features("contrib", "floating-point")
+
+    if not 0 <= stop_probability <= 1:
+        raise ValueError("stop_probability must be between 0 and 1")
+
+    if not 0 <= epsilon_selection <= 1:
+        raise ValueError("epsilon must be between 0 and 1")
+
+    assert (stop_probability == 0) == (epsilon_selection == 0), "either both stop_probability and epsilon_selection should be 0, or neither should be 0."
```
mccalluc, line 41: `assert` statements aren't guaranteed to run: My sense is that they are good for stating the assumptions of code, but aren't really for input validation. Another if-ValueError would be fine.

python/src/opendp/_extrinsics/_make_private_selection/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+
+            if score >= threshold:
+                return score, output
+
+            biased_coin = np.random.binomial(n=1, p=stop_probability)
+            if biased_coin:
+                return
+
+            if max_steps is not None and step >= max_steps:
+                return
```
mccalluc, line 77: Oh: I assumed an exception would be thrown, not a `return None`. I'm not sure if one or the other makes more sense... I should see how this works in context.

python/src/opendp/_extrinsics/_make_private_selection/__init__.py
```diff
@@ -0,0 +1,86 @@
+from typing import Optional
```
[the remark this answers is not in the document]
mccalluc: On second, second thought, this does depend on numpy, so it would belong in `extras.numpy`, if that PR gets merged first.

python/test/extrinsics/test_make_private_selection.py
```diff
@@ -0,0 +1,45 @@
...
+
+from ..helpers import optional_dependency
+
+
+def test_private_selection_threshold():
+    from opendp._extrinsics._make_private_selection import make_private_selection_threshold
+
+    dp.enable_features("contrib", "floating-point")
+
+    np = pytest.importorskip("numpy")
```
mccalluc, line 13: You could move more of the test inside `with optional_dependency("numpy"):`, as needed, and just have a normal `import numpy as np` inside.... Or you could just drop the `with optional_dependency`.

`optional_dependency`, unlike `pytest.importorskip`, checks that the friendly opendp specific errors are thrown if the dependency is not available: having both is a little redundant.

mccalluc (commented):
Mike has other comments, so I expect this will all change a bit. Some doc tweaks (include title of article, and explain what Exceptions could be thrown, depending on the params), and don't use `assert` for input validation. Thanks!

shlomihod:
My reflection on a conversation with @Shoeboxam:

There are two primary tasks regarding private selection:

1. Return a result above a predefined threshold (this contribution falls into this bucket)
2. Return the best output

The second most important aspect of a private selection algorithm is how many times to run the mechanism (in either one of the tasks). It seems that the most general presentation is as follow: sample an integer K from a predefine specific distribution, run the mechanism at most K times and return the first result the passes the threshold (for task 1) or run the mechanism for K times and return the best output (task 2).

For task 1, K should be sampled from the geometric distribution with $0 \leq \gamma < 1$ (see Algorithm 1 in [LT19] and Appendix C in [PS21]). Therefore, we will have a single parameter `stop_probability`.

For task 2, the distribution for sampling K is the truncate negative binomial [1], and it is a generalization of the distribution presented in [LT19] (where is the truncate geometric distribution [2]), that can provide different privacy-utility tradeoffs. There are two parameters here: `\gamma` and `\eta`.

Therefore, unfortunately it seems that is less user-friendly to try and unify the API for the two tasks.

We will drop the `epsilon_selection` and `steps` parameters because they used to control the maximal runtime of the algorithm. But this can be controlled via the `stop_probability` without spending extra privacy budget.

[1] Negative binomial distribution is "a discrete probability distribution that models the number of failures in a sequence of independent and identically distributed Bernoulli trials (of probability $\gamma$) before a specified (non-random) number of successes $\eta$ occurs.
Truncate negative binomial distribution: setting Pr[K = 0] = 0.

[2] With $\eta = 0$.
The algorithms in [LT1] uses more explicit representation of the geometric distribution. Instead of sampling the number of attempts K, they flips a $\gamma$-based coin every iteration that decide whether to stop the algorithm. This is equivalent to sampling K ahead of time.

[LT19] [Private Selection from Private Candidates](https://arxiv.org/pdf/1811.07971)

[PS21] [Hyperparameter Tuning with Renyi Differential Privacy](https://arxiv.org/abs/2110.03620)

python/src/opendp/_extrinsics/_make_private_selection/__init__.py
shlomihod: If a None is returned, the privacy budget is still consumed. That's why I believe an exception is inappropriate here.

python/test/extrinsics/test_make_private_selection.py
shlomihod: Oh, thank you for the this explanation - it is very helpful!

python/src/opendp/_extrinsics/_make_private_selection/__init__.py
shlomihod: Thanks! This line is now removed.

python/src/opendp/_extrinsics/_make_private_selection/__init__.py
shlomihod: (see reply below)

chikeabuah (approved):

python/src/opendp/_extrinsics/_make_private_selection/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def make_private_selection_threshold(meas: Measurement,
+                                     threshold: float,
+                                     stop_probability: float) -> Measurement:
+
+    """Measurement for private selection with known threshold.
+
+    This executes an ε-DP mechanism M that returns a tuple (q, x) repeatedly
+    until the score q is at least T, a predefined threshold.
```
Shoeboxam, line 13:
```suggestion
    until the score q is at least `threshold`.
```

python/src/opendp/_extrinsics/_make_private_selection/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+    Only the last output of M is released.
+
+    Algorithm 1 in `Private selection from private candidates <https://arxiv.org/pdf/1811.07971.pdf#page=7>`_ (Liu and Talwar, STOC 2019).
+
+    :param meas: A measurement function that returns a 2-tuple when invoked,
+                 where the first element is a score and the second element is the output of intreset.
```
Shoeboxam, line 19:
```suggestion
                 where the first element is a score and the second element is the output of interest.
```

python/src/opendp/_extrinsics/_make_private_selection/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+        raise ValueError("meas must be pure differential privacy (max_divergence(T=float))")
+
+    def function(data):
+
+        while True:
+
+            score, *output = meas(data)
+
+            if score >= threshold:
+                return score, output
```
Shoeboxam, line 44: I was going to suggest returning *output, but on further consideration, when this goes to Rust it will have to be a 2-tuple anyways.

python/src/opendp/_extrinsics/_make_private_selection/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+
+    np_csprng = get_np_csprng()
+
+    # If stop_probabily is 1, the measurement is executed only once with double of the privacy budget,
+    # so we prevent this inefficient case.
+    if not 0 <= stop_probability < 1:
+        raise ValueError("stop_probability must be between 0 and 1")
+
+    if meas.output_measure != dp.max_divergence(T=float):
+        raise ValueError("meas must be pure differential privacy (max_divergence(T=float))")
```
Shoeboxam, line 35:
```suggestion
        raise ValueError("meas must satisfy pure differential privacy (max_divergence(T=float))")
```

python/src/opendp/_extrinsics/_make_private_selection/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+    :param threshold: The threshold score. Return immediately if the score is above this threshold.
+    :param stop_probability: The probability of stopping early at any iteration.
+    :returns: A tuple from `meas` with the first element being the score.
+    """
+
+    dp.assert_features("contrib", "floating-point")
+
+    np_csprng = get_np_csprng()
+
+    # If stop_probabily is 1, the measurement is executed only once with double of the privacy budget,
```
Shoeboxam, line 29:
```suggestion
    # If stop_probability is 1, the measurement is executed only once with double of the privacy budget,
```

python/src/opendp/_extrinsics/_make_private_selection/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+                return score, output
+
+            biased_coin = np_csprng.binomial(n=1, p=stop_probability)
+            if biased_coin:
+                return
+
+    return dp.m.make_user_measurement(
+        input_domain=meas.input_domain,
+        input_metric=meas.input_metric,
+        output_measure=dp.max_divergence(T=float),
```
Shoeboxam, line 53:
```suggestion
        output_measure=meas.output_measure,
```

Shoeboxam (changes requested):
Minor nits at this point.

Shoeboxam:
* **#2126** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2126?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1782** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1782?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @shlomihod and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

Shoeboxam:
Sorry for the never-ending API picking, but I think the following API handles everything and is pretty accessible:

```python
def make_select_private_candidate(
	measurement,
    stop_probability,
	threshold=None,
    num_trials=1,
	max_candidates=None
)
```

* If you just specify measurement and stop_probability, you get algorithm 2/theorem 3.2 satisfying 3ε-DP. 
* If you then also specify threshold, you get algorithm 1/theorem 3.1.
* num_trials controls eta (for PS21) but defaults to geometric (for LT19).
* If max_candidates is set, you add an additional ε0 term.

Raise a construction-time error if an invalid combination of parameters is passed.

For now, the API would be this, following the same algorithm as in your existing PR:
```python
def make_select_private_candidate(
	measurement,
    stop_probability,
	threshold,
)
```

We can add other features gradually as-needed in a backwards-compatible way later. 

BTW the Rust impl in the PR linked above is working, but I still need to polish it and implement these changes.

shlomihod:
API discussions are fun!

For `max_iteration`, I'm uncertain about the added value it provides to the user. While it offers more deterministic runtime control, the user incurs a privacy budget cost, ε0. It seems that the deterministic aspect may not be essential, as users can adjust the stop_probability parameter to achieve an expected runtime without added costs. Additionally, how does this parameter interact with PS21?

Shoeboxam (approved):
Merging, but will be making changes in the follow-up #2126

Issue #1783, opened 2024-07-11
`with_privacy` should pass `args` and `kwargs`


shlomihod:
`with_privacy` does not pass `args` and `kwargs` to the transformation, similarly to how it is already implemented in `to_then`.

For example, this is currently impossible:
```python

from opendp._extrinsics._utilities import to_privacy

# ASSUME: make_my_transformation(input_domain, input_metric, param1, param2)
make_my_measurement = to_privacy(make_my_transformation)
make_my_measurement(input_domain, input_metric, param1, param2=param2)  # DOES NOT WORK
```

PR #1784, opened 2024-07-11
`with_privacy` should pass `args` and `kwargs`

shlomihod:
Implementing #1783.


Shoeboxam (approved):
Thanks for making this enhancement. The original function didn't have this kind of generality in mind.

PR #1785, opened 2024-07-12
Polars: rewrite binary paths in Polars plugins

### LINKED ISSUE #1697: Polars: replace binary paths in FFI plugins
When a Polars plugin is constructed, it includes a path to the binary file that contains the plugin logic. If the plan is constructed on one computer and then passed to another computer, the path to the binary file may no longer be valid. 

When a lazyframe is passed through a transformation or measurement function, it should emit a lazyframe where the binary paths within are correct for the current machine.

Shoeboxam:
Closes #1697


Shoeboxam:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/opendp/opendp/1785?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1785** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1785?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1767** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1767?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1716** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1716?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1768](https://github.com/opendp/opendp/pull/1768) <a href="https://app.graphite.dev/github/pr/opendp/opendp/1768?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1760** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1760?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1703** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1703?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2 other dependent PRs ([#1715](https://github.com/opendp/opendp/pull/1715) <a href="https://app.graphite.dev/github/pr/opendp/opendp/1715?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1732](https://github.com/opendp/opendp/pull/1732) <a href="https://app.graphite.dev/github/pr/opendp/opendp/1732?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`

<h2></h2>

python/src/opendp/polars.py
```diff
@@ -24,6 +24,10 @@
 )
 from opendp.domains import series_domain, lazyframe_domain, option_domain, atom_domain
 from opendp.measurements import make_private_lazyframe
+from opendp._data import set_opendp_lib_path
+
+if lib_path is not None:
+    set_opendp_lib_path(str(lib_path))
```
mccalluc, line 30:
```suggestion
if lib_path is not None: # Should only be None when OPENDP_HEADLESS is set.
    set_opendp_lib_path(str(lib_path))
```
Is my understanding correct?

rust/src/data/ffi/polars.rs
```diff
@@ -1,12 +1,38 @@
...
+/// # Arguments
+/// * `opendp_lib_path` - Absolute path to the OpenDP Library binary.
+#[no_mangle]
+pub extern "C" fn opendp_data__set_opendp_lib_path(
+    opendp_lib_path: *mut c_char,
+) -> FfiResult<*mut ()> {
+    let opendp_lib_path = Arc::from(try_!(to_str(opendp_lib_path)).to_string());
+    try_!(OPENDP_LIB_PATH
+        .try_lock()
+        .map_err(|_| err!(FFI, "failed to set OPENDP_LIB_PATH due to lock poisoning")))
```
mccalluc, line 31: Would it be worth interpolating the path here? It might not be. Whatever makes the most sense.

mccalluc (commented):
I assume this path is the one that shows up in the serializations? Just to test that this wiring does what we think it does, add a simple test that 
- calls `set_opendp_lib_path` and sets the path to `foo/bar`
- calls `str` on an `Expr` and makes sure it has `foo/bar`
- calls `set_opendp_lib_path` to restore the original.

mccalluc (commented):
I checked out this branch, and the only occurrence of `OPENDP_LIB_PATH` is on these lines, which seems odd. Any chance `OPENDP_LIB_DIR` is what you meant? And if this is a mistake, could there be a test that would catch it?

Shoeboxam:
@mccalluc Yes, `OPENDP_LIB_PATH` is what I meant. Perhaps we could create a documentation page that covers all of the library's features and environment variables.

mccalluc (dismissed):
If FSO is waiting on this I shouldn't block, but maybe these reservations should go in a follow-up issue?

Documentation. I don't know that there needs to be one reference for all envvars, but when a new feature is introduced I'd like to have some public documentation of it. A doctest based on this example might be good.

I put a breakpoint in this test, and I'm a little surprised about how it works. I was expecting all the serializations to change, but maybe this is only for measurements?

```
(Pdb) str(m_expr((pl.LazyFrame(dict()), pl.all())))
'len().testing!:noise_plugin()'
(Pdb) pl.len().dp.noise(scale=1.)
<Expr ['len()./Users/chuckmccallum/git…'] at 0x15B62B1D0>
(Pdb) str(pl.len().dp.noise(scale=1.))
'len()./Users/chuckmccallum/github/opendp/opendp/rust/target/debug/libopendp.dylib:noise([null, dyn float: 1.0])'
```

Shoeboxam:
I wasn't thinking it would be used for the client-side, as well, but I guess someone could set it to obscure local paths from the server?

I renamed the var to make it clearer that this is just about Polars, and opened a new PR with documentation.

mccalluc (approved):
Thanks!

Issue #1790, opened 2024-07-14
Microdata: Faithfulness


shlomihod:
Initiating the microdata subpackage with an implementation of the faithfulness criteria as described in the [Israeli Birth Dataset release](https://arxiv.org/pdf/2405.00267#page=13).

PR #1791, opened 2024-07-14
Microdata: Faithfulness

shlomihod:
Implementing #1790.

Few open questions:

1. How to enforce public knowledge of a LazyDataframe? Perhaps adding SizedLazyFrame? For example,
```python

def sized_lazyframe_domain(lf_domain, size):
    return user_domain(
        f"Sized{lf_domain}", 
        member=lambda x: lf_domain.member(x) and len(x.collect()) == size,
        descriptor={"domain": lf_domain, "size": size})
```

2. How to include `igraph` as optional dependency?

### DIFF
--- a/python/src/opendp/_extrinsics/microdata/faithfulness.py
+++ b/python/src/opendp/_extrinsics/microdata/faithfulness.py
@@ -1,0 +1,128 @@
+from typing import Callable
+
+import polars as pl
+from opendp.mod import Domain, Metric, Transformation
+import opendp.prelude as dp
+from opendp._lib import import_optional_dependency
+from opendp._extrinsics._utilities import to_then, with_privacy
+
+
+__all__ = ["make_faithfulness",
+           "then_faithfulness",
+           "make_private_faithfulness",
+           "then_private_faithfulness"]
+
+
+def make_faithfulness(
+        input_domain: Domain,
+        input_metric: Metric,
+        *,
+        reference_dataset: pl.LazyFrame,
+        similarity: Callable | dict[str, float]) -> Transformation:
+    r"""Construct a Transformation that returns the optimal (maximal) faithfulness of two datasets.
+    Faithfulness assesses whether the released dataset resemble the original dataset in a record-level granularity.
+    Namely, there is a 1-to-1 matching between the records of released dataset and the records of the original dataset: each released record is matched with an original record.
+    Two records are considered similar if their feature space distance, according to the similarity function, is less than or equal to 1.
+
+    Presented in [Israel's Birth Dataset release](https://arxiv.org/pdf/2405.00267#page=13).
+
+    :param input_domain: instance of `LazyFrameDomain`
+    :param input_metric: instance of `symmetric_distance()`
+    :param reference_dataset: public dataset to compare against, instance of `LazyFrame` or `DataFrame`
+    :param similarity: function measuing the similarity between two records
+            or a dictionary of weights for each column scaling the absolute difference
+
+    :returns a Transformation that computes a tuple of (mean, S, Vt)
+    """
+
+    np = import_optional_dependency("numpy")
+    scipy = import_optional_dependency("scipy")
+    sklearn_neighbors = import_optional_dependency("sklearn.neighbors")
+    igraph = import_optional_dependency("igraph")
+
+    dp.assert_features("contrib", "floating-point")
+
+    if "LazyFrameDomain" not in str(input_domain.type):
+        raise ValueError("input_domain must be a LazyFrame domain")
+
+    # TODO: how to enfornce public knowlege of dataset size, if at all?
+    # if (input_domain.descriptor.get("size")) is None:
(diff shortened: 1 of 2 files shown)


Shoeboxam:
For reference, I promised Shlomi I would add some FFI helpers to retrieve descriptors from the domain. It would help us avoid wrapping the domain in an user domain.

mccalluc:
> How to include igraph as optional dependency?

This would be something like how `numpy` or `sklearn` are now? You could add it in `setup.cfg` under `[options.extras_require]`... although this could support a DAG of optional dependencies, we're tried to keep it as a chain, so our test matrix in `smoke-test.yml` only needs to check `n` options, instead of `2^n`.

Is there a natural place that `igraph` would fit? The downsides of requiring it at a lower level are the potential for dependency version range conflicts, and just the install time... so if it doesn't have many strict transitive dependencies, and it's not huge, adding it to the `[numpy]` extra wouldn't be unreasonable.

python/src/opendp/_extrinsics/microdata/faithfulness.py
```diff
@@ -0,0 +1,127 @@
...
+    # assert input_domain.member(reference_dataset.lazy())
+
+    if input_metric != dp.symmetric_distance():
+        raise ValueError("input metric must be symmetric distance")
+
+    if callable(similarity):
+        similarity_metric = similarity
+
+    elif isinstance(similarity, dict):
+        assert reference_dataset.columns == list(similarity.keys())
```
mccalluc, line 60: Probably change to `if`, if this is input validation.

python/src/opendp/_extrinsics/microdata/faithfulness.py
```diff
@@ -0,0 +1,127 @@
...
+        metric_weights = np.array(list(similarity.values()))
+
+        def similarity_metric(xi, yj):
+            return np.sum(np.abs((xi - yj)) * metric_weights)
+
+    else:
+        raise ValueError("similarity must be a callable or a dictionary")
+
+    def similarity_fn(X, Y):
+        # TODO: consider using the underline scipy implementation directly
```
mccalluc, line 70: I'm really happy staying with public methods, unless there's something a lot better about the private underscore one?

python/src/opendp/_extrinsics/microdata/faithfulness.py
```diff
@@ -0,0 +1,127 @@
...
+        first_matched = matched_vertices[second_matched] - num_records
+
+        assert len(first_matched) == len(second_matched)
+        matching_cardinality = len(first_matched)
+
+        # matching = (first_matched, second_matched)
+        # vertex_indices = set(range(num_records))
+        # unmatched_indices = (vertex_indices - set(first_matched),
+        #                       vertex_indices - set(second_matched)
+        # )
```
mccalluc, line 111: Delete or explain commented-out code.

python/src/opendp/_extrinsics/microdata/faithfulness.py
```diff
@@ -0,0 +1,127 @@
...
+        bipartite_mask = [False] * num_records + [True] * num_records
+        g = igraph.Graph.Bipartite(bipartite_mask, edges, directed=False)
+
+        matched_vertices = np.array(g.maximum_bipartite_matching().matching)[:num_records]
+
+        second_matched_mask = matched_vertices[:num_records] != -1
+        second_matched = np.nonzero(second_matched_mask)[0]
+        first_matched = matched_vertices[second_matched] - num_records
+
+        assert len(first_matched) == len(second_matched)
```
mccalluc, line 104: If this is an algorithm check, and it would just surface a bug in our code if it failed, then this is ok. Still might add a second argument with a message to help the person who gets the bug report.

python/test/extrinsics/test_microdata_faithfulness.py
```diff
@@ -0,0 +1,51 @@
+import pytest
+
+import opendp.prelude as dp
+
+from ..helpers import optional_dependency
+
+dp.enable_features("contrib", "floating-point")
+
+
+def test_microdataa_faithfulness():
```
mccalluc, line 10: Slight preference for parameterizing the test. Only 2 conditions tested, so not a big win, but it does give us clearer feedback about any failure, and makes sure all tests are tried, even if one fails
```suggestion
@pytest.mark.parametrize("faithfulness,similarity", [
    (1, lambda x, y: 2*(x != y).any()),
    (2, {"A": 1/100, "B": 1/2})])
def test_microdataa_faithfulness(faithfulness, similarity):
```

See also slight preference for handling these as separate kwargs, so it might be more like:
```suggestion
@pytest.mark.parametrize("faithfulness,similarity_function,similarity_dict", [
    (1, lambda x, y: 2*(x != y).any(), None),
    (2, None, {"A": 1/100, "B": 1/2})])
def test_microdataa_faithfulness(faithfulness, similarity_function, similarity_dict):
```

mccalluc (commented):
Mostly just style comments.

Issue #1792, opened 2024-07-14
Microdata: Universal Microdata Scheme


shlomihod:
Expanding the microdata subpackage with a minimal implementation of the universal microdata scheme as described in the [Israeli Birth Dataset release](https://arxiv.org/pdf/2405.00267#page=13).

This issue depends on #1782 (private selection) #1790 (faithfulness).

PR #1793, opened 2024-07-14
Microdata: Universal Microdata Scheme

shlomihod:
Implementing #1792.

### DIFF
--- a/python/src/opendp/_extrinsics/microdata/universal_microdata_scheme.py
+++ b/python/src/opendp/_extrinsics/microdata/universal_microdata_scheme.py
@@ -1,0 +1,166 @@
+from typing import Callable, Union, Optional
+
+from opendp.mod import Domain, Metric, Measurement
+import opendp.prelude as dp
+from opendp._lib import get_np_csprng, import_optional_dependency
+from opendp._extrinsics._make_private_selection import make_private_selection_threshold
+from opendp._extrinsics.microdata.faithfulness import make_private_faithfulness
+from opendp._extrinsics.synth import make_private_synthesizer_trainer
+
+
+__all__ = ["make_private_universal_microdata_scheme"]
+
+
+def make_private_universal_microdata_scheme(input_domain: Domain,
+                                            input_metric: Metric,
+                                            synthesizer_epsilon: float,
+                                            configuration_candidates: Optional[list[dict]] = None,
+                                            quality_acceptance_criteria: Optional[list[tuple]] = None,
+                                            quality_epsilons: Optional[list[float]] = None,
+                                            faithfulness_simiarlity: Optional[Union[Callable, dict[str, float]]] = None,
+                                            faithfulness_threshold: Optional[float] = None,
+                                            faithfulness_epsilon: Optional[float] = None,
+                                            with_selection: bool = True,
+                                            selection_stop_probability: float = 0,
+                                            selection_epsilon: float = 0,
+                                            selection_steps: Optional[int] = None) -> Measurement:
+
+    np = import_optional_dependency("numpy")
+
+    if "LazyFrameDomain" not in str(input_domain.type):
+        raise ValueError("input_domain must be a LazyFrame domain")
+
+    # if (input_domain.descriptor.get("size")) is None:
+    #     raise ValueError("input_domain's size must be known")
+
+    if input_metric != dp.symmetric_distance():
+        raise ValueError("input metric must be symmetric distance")
+
+    if sum([quality_acceptance_criteria is None, quality_epsilons is None]) == 1:
+        raise ValueError("quality_acceptance_criteria and quality_epsilons must be both None or both not None")
+
+    if 0 < sum([faithfulness_simiarlity is None, faithfulness_threshold is None, faithfulness_epsilon is None]) < 3:
+        raise ValueError("faithfulness_simiarlity, faithfulness_threshold, faithfulness_epsilons must be all None or all not None")
+
+    if quality_epsilons is None and faithfulness_epsilon is None:
+        raise ValueError("At least one of quality_epsilons and faithfulness_epsilon must be not None")
+
+    output_measure = dp.max_divergence(T=float)
+
+    if configuration_candidates is None:
+        configuration_candidates = [{}]
+
+    acceptance_criteria_epsilons: list[float] = []
+
+    if quality_epsilons is not None:
+        acceptance_criteria_epsilons.extend(quality_epsilons)
+
+    if faithfulness_epsilon is not None:
+        acceptance_criteria_epsilons.append(faithfulness_epsilon)
+
+    epsilons = [synthesizer_epsilon, sum(acceptance_criteria_epsilons)]
+
+    # TODO: is it correct?
+    d_in = 1
+
+    overall_pipeline = dp.c.make_sequential_composition(input_domain,
+                                                        input_metric,
+                                                        output_measure,
+                                                        d_in,
+                                                        d_mids=epsilons)
+
+    np_csprng = get_np_csprng()
+
+    def function(dataset):
+
+        num_records = len(dataset.collect())
+        overall_pipline_comp = overall_pipeline(dataset)
+        del dataset
+
+        # Synthesizer
+
+        configuration = np_csprng.choice(configuration_candidates)
+
+        synthesizer_trainer = make_private_synthesizer_trainer(input_domain,
+                                                               input_metric,
+                                                               synthesizer_epsilon,
+                                                               **configuration["synthesizer"])
+
+        synthesizer = overall_pipline_comp(synthesizer_trainer)  
+
+        synth_dataset = synthesizer.sample(num_records)
+
+        # Acceptance Criteria
+        acceptance_criteria_pipeline = dp.c.make_sequential_composition(input_domain, input_metric, output_measure, d_in,
+                                                                        d_mids=acceptance_criteria_epsilons)
+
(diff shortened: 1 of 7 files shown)


mccalluc (commented):
Could you retarget this PR so it doesn't overlap with #1791, maybe by making `1790-microdata-faithfulness` the target, rather than `main`? #1782 is also out there, and I'm not sure if it overlaps with either of these, too.

The goal is for the reviewer not to have to sort out what they've seen in a previous PR and what's unique to this one.

Mike has a lot of experience stacking PRs, and uses graphite. For my PRs, that has seemed like overkill, and instead I'll just keep the later PRs as drafts until the base PRs are resolved. After the base PR is merged to main and auto-deleted, and the follow-up PR is auto-retargetted. Whatever works for you!

python/src/opendp/_extrinsics/microdata/universal_microdata_scheme.py
```diff
@@ -0,0 +1,166 @@
...
+
+__all__ = ["make_private_universal_microdata_scheme"]
+
+
+def make_private_universal_microdata_scheme(input_domain: Domain,
+                                            input_metric: Metric,
+                                            synthesizer_epsilon: float,
+                                            configuration_candidates: Optional[list[dict]] = None,
+                                            quality_acceptance_criteria: Optional[list[tuple]] = None,
+                                            quality_epsilons: Optional[list[float]] = None,
```
mccalluc, line 19: Sometimes for list args, rather than making it optional, it can be simpler to make the default `[]`, so you don't need to handle `None` as a special case... but if you actually want those to work differently, ignore me!

python/src/opendp/_extrinsics/microdata/universal_microdata_scheme.py
```diff
@@ -0,0 +1,166 @@
...
+import opendp.prelude as dp
+from opendp._lib import get_np_csprng, import_optional_dependency
+from opendp._extrinsics._make_private_selection import make_private_selection_threshold
+from opendp._extrinsics.microdata.faithfulness import make_private_faithfulness
+from opendp._extrinsics.synth import make_private_synthesizer_trainer
+
+
+__all__ = ["make_private_universal_microdata_scheme"]
+
+
```
mccalluc, line 13: Let's leave this off for now. We're still working out the conventions for the `extras`... might end up making this private and exporting from a public name from the top level? TBD.

python/src/opendp/_extrinsics/microdata/universal_microdata_scheme.py
```diff
@@ -0,0 +1,166 @@
...
+                                            selection_steps: Optional[int] = None) -> Measurement:
+
+    np = import_optional_dependency("numpy")
+
+    if "LazyFrameDomain" not in str(input_domain.type):
+        raise ValueError("input_domain must be a LazyFrame domain")
+
+    # if (input_domain.descriptor.get("size")) is None:
+    #     raise ValueError("input_domain's size must be known")
+
```
mccalluc, line 35:
```suggestion
```

Delete or explain.

python/src/opendp/_extrinsics/microdata/universal_microdata_scheme.py
```diff
@@ -0,0 +1,166 @@
...
+    if input_metric != dp.symmetric_distance():
+        raise ValueError("input metric must be symmetric distance")
```
mccalluc, line 37:
```suggestion
        raise ValueError("input_metric must be symmetric_distance")
```

python/src/opendp/_extrinsics/microdata/universal_microdata_scheme.py
```diff
@@ -0,0 +1,166 @@
...
+
+        # Gather Results and Check if all Acceptance Criteria Passed
+        acceptance_criteria_results = [acceptance_criteria_pipeline_comp(meas)
+                                       for meas in acceptance_criteria_meas]
+
+        assert len(acceptance_criteria_results) == len(acceptance_criteria_thresholds)
+        all_acceptance_criteria_passed = all(value <= threshold
+                                             for value, threshold in zip(acceptance_criteria_results, acceptance_criteria_thresholds))
+
+        return all_acceptance_criteria_passed, configuration, synthesizer, synth_dataset, acceptance_criteria_results
```
mccalluc, line 150: Return a namedtuple or something else that doesn't depend on the positions being stable.

python/src/opendp/_extrinsics/synth/__init__.py
```diff
@@ -0,0 +1,61 @@
+from opendp.mod import Domain, Metric
```
mccalluc, line 1: Not sure how this relates to the rest of it, but could this also live under `microdata`?

python/src/opendp/_extrinsics/synth/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+        self._is_fitted = False
+
+    def fit(self, data):
+        assert not self._is_fitted, "Synthesizer is already fitted"
+
+    def sample(self, num_records):
+        assert self._is_fitted, "Synthesizer must be fitted first"
+
+    def releasable(self):
+        assert self._is_fitted, "Synthesizer must be fitted first"
```
mccalluc, line 16: Ditto previous comments about `assert` for input validation. (I really wish there wasn't that caveat on how it works!)

python/src/opendp/_extrinsics/synth/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+        return self._data.sample(num_records)
+
+
+def make_private_synthesizer_trainer(input_domain: Domain,
+                                     input_metric: Metric,
+                                     epsilon: float,
+                                     name: str,
+                                     **kwargs):
+
+    print("THIS IS A PLACEHOLDER. NOT DIFFERENTIALLY PRIVATE!!!")
```
mccalluc, line 39: Placeholder in the sense that this is a TODO? Or is this code that is working correctly, but we want to be sure users don't do the wrong thing with it?

python/test/extrinsics/test_universal_microdata_scheme.py
```diff
@@ -0,0 +1,197 @@
...
+import opendp.prelude as dp
+from opendp.mod import Domain, Metric, Transformation
+from opendp._extrinsics._utilities import with_privacy
+
+from ..helpers import optional_dependency
+
+dp.enable_features("contrib", "floating-point")
+
+
+def test_microdataa_faithfulness():
```
mccalluc, line 15:
```suggestion
def test_microdata_faithfulness():
```

python/test/extrinsics/test_universal_microdata_scheme.py
```diff
@@ -0,0 +1,197 @@
...
+         for col in COLS])
+
+    lf_domain_with_margin = dp.with_margin(lf_domain,
+                                           by=COLS, public_info="keys")
+
+    def similarity_record_fn(x, y):
+        return np.sum(np.abs((x - y)))
+
+    with (optional_dependency("numpy"),
+          optional_dependency("polars"),
```
mccalluc, line 42:
```suggestion
```

Are all of these necessary? I just haven't seen where the code was using polars, but maybe it's indirect?

python/test/extrinsics/test_universal_microdata_scheme.py
```diff
@@ -0,0 +1,197 @@
...
+    np = pytest.importorskip("numpy")
+    pl = pytest.importorskip("polars")
+
+    dp.assert_features("contrib", "floating-point")
+
+    if "LazyFrameDomain" not in str(input_domain.type):
+        raise ValueError("input_domain must be a LazyFrame domain")
+
+    # TODO: assert margin with public_info "keys"
+    # TODO: extract cols (by) from the magin
```
mccalluc, line 145: If these wouldn't block merge, but they'd be nice to have some day, can you file corresponding issues and add the URLs here?

python/test/extrinsics/test_universal_microdata_scheme.py
```diff
@@ -0,0 +1,197 @@
...
+    cols = reference_dataset.columns
+
+    # assert input_domain.member(reference_dataset.lazy())
```
mccalluc, line 148: Delete commented out code.

mccalluc (commented):
Just a cursory review. Heading out now and back next week: feel free to request a re-review. Thanks!

python/src/opendp/_extrinsics/microdata/universal_microdata_scheme.py
Shoeboxam: (assuming you don't mutate them, as default args are shared objects across function invocations in Python)

python/src/opendp/_extrinsics/microdata/universal_microdata_scheme.py
```diff
@@ -0,0 +1,166 @@
...
+__all__ = ["make_private_universal_microdata_scheme"]
+
+
+def make_private_universal_microdata_scheme(input_domain: Domain,
+                                            input_metric: Metric,
+                                            synthesizer_epsilon: float,
+                                            configuration_candidates: Optional[list[dict]] = None,
+                                            quality_acceptance_criteria: Optional[list[tuple]] = None,
+                                            quality_epsilons: Optional[list[float]] = None,
+                                            faithfulness_simiarlity: Optional[Union[Callable, dict[str, float]]] = None,
```
Shoeboxam, line 20: *similarity

Shoeboxam (changes requested):
Some initial thoughts. Thanks for opening this PR!

PR #1794, opened 2024-07-15
docs: Polars Accuracy Method

GKD-stack:
I tried rebasing, but there are quite a few merge conflicts. i also tried forcing the main into my local branch but ran into issues with getting the environment to get accuracy run


mccalluc (changes requested):
I think you only want the accuracy notebook in this PR, and all the others should be in other PRs? We all find our own ways of working with git: For me, I almost always add the whole directory, rather than having multiple files out with modifications and trying to remember which get committed and which don't.

I'll look at `accuracy.ipynb` and give some more detailed feedback.

docs/source/getting-started/examples/accuracy.ipynb
```diff
@@ -0,0 +1,269 @@
...
+    "# Accuracy Method"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Many prior examples of queries used comparisons of the actual values to the computed differentially private values. Intuitively, we assessed the quality of our results by visually inspecting the proximity of the actual values to the differentially private outputs. \n",
+    "\n",
+    "However, this approach isn't concrete since interpretations can vary widely on scale, type of visualization, and other factors. A more objective way to evaluate differentially private outputs is by using the `accuracy` method. \n",
```
mccalluc, line 16: I'd emphasize that comparing one DP-release to the true value is a comparison with exactly one draw from a random variable: You might get "lucky" and get a release that is much more accurate than usual, or "unlucky" and get a release that is much less accurate than usual.

Also: with synthetic data a comparison with the "true" value is possible... but in the real world you can't do that and maintain privacy. Even if you don't release these intermediate results, if your analysis plan is influenced by what you see in the real data, that's a privacy leak.

docs/source/getting-started/examples/accuracy.ipynb
```diff
@@ -0,0 +1,269 @@
...
+    "\n",
+    "In this notebook ,we will: \n",
```
mccalluc, line 18:
```suggestion
    "In this notebook we will: \n",
```

docs/source/getting-started/examples/accuracy.ipynb
```diff
@@ -0,0 +1,269 @@
...
+    "        (): dp.Margin(max_partition_length=60_000_000),\n",
+    "    },\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We will use the same query we began with in [Part 1](Link Here): a count query for each year and quarter."
```
mccalluc, line 55: CI won't pass with this. Possible courses:
- Get part 1 merged in so you can point at something
- Rephrase so it doesn't depend on anything outside
- change to http://example.com/TODO or similar -- but also move this to the hidden toctree

mccalluc (commented):
A few more comments, and some questions for @Shoeboxam .

docs/source/getting-started/examples/accuracy.ipynb
Shoeboxam: When writing my book, before asking people to review, I used the VSCode Grammarly extension. I think if you used it before requesting reviews, it would help cut down on the noise, so that we can give deeper feedback.

docs/source/getting-started/examples/accuracy.ipynb
```diff
@@ -0,0 +1,269 @@
...
+      "text/plain": [
+       "<Figure size 1000x600 with 1 Axes>"
```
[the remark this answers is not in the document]
Shoeboxam: My intuition here is that if smoothing would help your utility, then it is likely you'd be better off with coarser bins. Both smoothing and coarser bins will reduce variance. This is another one of those bias/variance tradeoffs.

docs/source/getting-started/examples/accuracy.ipynb
```diff
@@ -0,0 +1,269 @@
...
+      ],
+      "text/plain": [
+       "shape: (1, 5)\n",
+       "┌────────┬───────────┬─────────────────┬───────┬─────────────┐\n",
+       "│ column ┆ aggregate ┆ distribution    ┆ scale ┆ accuracy    │\n",
+       "│ ---    ┆ ---       ┆ ---             ┆ ---   ┆ ---         │\n",
+       "│ str    ┆ str       ┆ str             ┆ f64   ┆ f64         │\n",
+       "╞════════╪═══════════╪═════════════════╪═══════╪═════════════╡\n",
+       "│ DP     ┆ Len       ┆ Integer Laplace ┆ 360.0 ┆ 1078.963271 │\n",
+       "└────────┴───────────┴─────────────────┴───────┴─────────────┘"
```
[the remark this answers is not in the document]
Shoeboxam: * if you are suggesting column names like `accuracy (α=.05)`, then it might be annoying to index
* if you are suggesting a new column for alpha, it would always be constant. This feels like an unnecessary piece of info to keep in the table, as the user is already aware-- they set alpha in the argument.

Shoeboxam:
Can you rebase onto main? Something like this:
```shell
git checkout main
git pull
git checkout 1789-accuracy
git rebase --onto main 4775062 1789-accuracy
```

Shoeboxam:
Can you commit the dataset you're using to this repo, and link to it from your notebooks?
https://github.com/opendp/dp-test-datasets/tree/master

docs/source/getting-started/examples/accuracy.ipynb
```diff
@@ -0,0 +1,269 @@
...
+    "context = dp.Context.compositor(\n",
+    "    data=df,\n",
+    "    privacy_unit=dp.unit_of(contributions=36),\n",
+    "    privacy_loss=dp.loss_of(epsilon=1.0),\n",
+    "    split_evenly_over=10,\n",
+    "    margins={\n",
+    "        (\"YEAR\", ): dp.Margin(public_info=\"keys\", max_partition_length=200_000, max_partition_contributions=4),\n",
+    "        (\"QUARTER\", ): dp.Margin(public_info=\"keys\", max_partition_length=200_000, max_partition_contributions=13),\n",
+    "        (\"YEAR\", \"QUARTER\",): dp.Margin(public_info=\"keys\", max_partition_length=200_000, max_partition_contributions=1),\n",
+    "        (): dp.Margin(max_partition_length=60_000_000),\n",
```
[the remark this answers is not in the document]
GKD-stack: leaving this for the same intro link insertion todo

mccalluc (commented):
If the upstream changes your work needs have been merged to `main`, I'd suggest just pulling out the files that you need from this PR, making a fresh branch from `main`, and closing this PR.

GKD-stack:
> If the upstream changes your work needs have been merged to `main`, I'd suggest just pulling out the files that you need from this PR, making a fresh branch from `main`, and closing this PR.

unsure of all changes this needs are merged into main, I see the accuracy method in the new polars.py in main but the following starter example is still failing: 

`import polars as pl
data = pl.LazyFrame([pl.Series("convicted", [0, 1, 1, 0, 1] * 50, dtype=pl.Int32)])

data=data,
split_evenly_over=1,
margins={(): dp.Margin(max_partition_length=1000)},
)

query = context.query().select(
pl.len().dp.noise(), 
pl.col("convicted").fill_null(0).dp.sum((0, 1))
)

query.accuracy(alpha=.05) `

mccalluc:
I've created a new PR with just the one file we're really interested in.

- #1857

Seems easier than sorting out the git history here.

PR #1797, opened 2024-07-16
Stub out getting-started/utility

### LINKED ISSUE #1774: New docs section: Getting Started > Assessing Utility
Assessing Utility section:
- Laplace vs. Gaussian noise (for both clipped-norms and grouped) #1720
- Bias vs. Variance tradeoff
- Accuracy (move here from User Guide)

mccalluc:
- Fix #1774
- After this is merged to main, #1735 should be updated to target this directory.
- The original issue mentions "Bias vs. Variance tradeoff" as one of the pages here. I don't think we have an issue just for that, and I'm not sure exactly what is imagined.
- Should there be any mention of the accuracy utility in the API User Guide? Maybe it wouldn't get its own page, but we'd incorporate it into other discussions?


mccalluc:
- #1340 strikes again!

Restarting...

GKD-stack (approved):

mccalluc:
- Filed #1801 as a separate issue.

Issue #1780, opened 2024-07-11
Simplify `typical-workflow` for now: no dataframe, just a vector


mccalluc:
In the long run we do want a parallel demonstration of polars in R and python
- ie #1491

but for now we just want to simplify `typical-workflow` to remove the old dataframe API.

Also, just do a sum instead of a mean. (Avoid resize transformations unless we have a good reason.)

mccalluc:
- Fixed by #1909

PR #1800, opened 2024-07-16
docs: document API layers

mccalluc:
- Preparing for #1780

It seemed that the discussion of the API layers needed one more layer, and that it might fit better on the previous page.

### DIFF
--- a/docs/source/getting-started/code/quickstart-context.rst
+++ b/docs/source/getting-started/code/quickstart-context.rst
@@ -7,7 +7,7 @@
 # /init
 
 # demo
->>> space = dp.space_of(float)
+>>> space = dp.atom_domain(T=float), dp.absolute_distance(T=float)
 >>> laplace_mechanism = space >> dp.m.then_laplace(scale=1.)
 >>> dp_value = laplace_mechanism(123.0)
 
--- a/docs/source/getting-started/code/quickstart-context.rst
+++ b/docs/source/getting-started/code/quickstart-context.rst
@@ -7,7 +7,7 @@
 # /init
 
 # demo
->>> space = dp.atom_domain(T=float), dp.absolute_distance(T=float)
+>>> space = (dp.atom_domain(T=float), dp.absolute_distance(T=float))
 >>> laplace_mechanism = space >> dp.m.then_laplace(scale=1.)
 >>> dp_value = laplace_mechanism(123.0)
 
--- a/docs/source/getting-started/code/quickstart-context.rst
+++ b/docs/source/getting-started/code/quickstart-context.rst
@@ -7,7 +7,8 @@
 # /init
 
 # demo
->>> laplace_mechanism = dp.space_of(float) >> dp.m.then_laplace(scale=1.)
+>>> space = dp.space_of(float)
+>>> laplace_mechanism = space >> dp.m.then_laplace(scale=1.)
 >>> dp_value = laplace_mechanism(123.0)
 
 # /demo
--- a/docs/source/getting-started/quickstart.rst
+++ b/docs/source/getting-started/quickstart.rst
@@ -100,16 +100,16 @@ This is obviously not the easiest way to add noise to a number,
 but it demonstrates a number of low-level OpenDP patterns:
 
 * First, define your metric space.
-* Then, chain operators together and end with a ``Measurement`` function.
+* Then, chain operators together to construct a ``Measurement`` (aka mechanism).
 * Invoke that measurement on a value to get a DP release.
 
 OpenDP has two APIs and we'll demonstrate how to use both:
 
 * The **Context API** is simpler and helps to enforce best practices. Currently available only for Python.
-* The **Framework API** is lower-level. Available for Python and R, it mirrors the underlying Rust framework.
+* The **Framework API** is lower-level. Available for Python, R and Rust, it directly implements the :ref:`OpenDP Programming Framework <programming-framework>`.
 
 Because the Context API is a wrapper around the Framework API, it is easier to use but less flexible:
 All calls ultimately pass through the Framework API.
 
-The next page will use the Framework and Context APIs to demonstrate the similarities between the Framework APIs in different languages.
+The next page will demonstrate usage of the Context API in Python, and Framework API in Python and R.
 After that, the remaining "Getting Started" documentation will focus just on Python.
--- a/docs/source/getting-started/quickstart.rst
+++ b/docs/source/getting-started/quickstart.rst
@@ -99,10 +99,10 @@ Let's apply Laplace noise to a value.
 This is obviously not the easiest way to add noise to a number,
 but it demonstrates a number of low-level OpenDP patterns:
 
-* Defining your metric space with ``space_of`` in Python's Context API, or a (domain, distance) tuple in any language..
-* Chaining operators together with ``>>`` in Python and Rust, or ``|>`` in R.
-* Constructing a ``Measurement`` function on your metric space with ``then_laplace``.
-* Invoking that measurement on a value to get a DP release.
+* First, define your metric space.
+* Then, chaining operators together.
+* Construct a ``Measurement`` function on your metric space with ``then_laplace``.
+* Invoke that measurement on a value to get a DP release.
 
 OpenDP has two APIs and we'll demonstrate how to use both:
 
--- a/docs/source/getting-started/quickstart.rst
+++ b/docs/source/getting-started/quickstart.rst
@@ -98,7 +98,7 @@ Let's apply Laplace noise to a value.
 
 This demonstrates a number of low-level OpenDP patterns:
 
-* First, define your metric space.
+* First, define your "metric space": a data domain and a definition of distance
 * Then, chain operators together to construct a ``Measurement`` (aka mechanism).
 * Invoke that measurement on a value to get a DP release.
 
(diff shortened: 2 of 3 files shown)


docs/source/getting-started/quickstart.rst
```diff
@@ -97,9 +97,20 @@ Let's apply Laplace noise to a value.
...
 * Defining your metric space with ``space_of`` in Python's Context API, or a (domain, distance) tuple in any language..
 * Chaining operators together with ``>>`` in Python and Rust, or ``|>`` in R.
 * Constructing a ``Measurement`` function on your metric space with ``then_laplace``.
 * Invoking that measurement on a value to get a DP release.
+
+OpenDP has layered APIs which provide increasing abstraction and usability:
+
+* The **Framework API** is low-level. Available for Python and R, it mirrors the underlying Rust framework.
+* The **Context API** introduces a ``Context`` class which ensures that queries do not exceed the privacy budget. Currently available only for Python.
+* The **Polars API** provides a DP extension to the `Polars <https://docs.pola.rs/>`_ dataframe library. Currently available only for Python.
```
Shoeboxam, line 111: I think it would be better to just state that we have two APIs: the Context API, and the Framework API. I'd list the Context API first. 

```
* The **Context API** is simpler to use and enforces best practices. The API is designed to ensure that mistakes that lead to under-estimated privacy guarantees cannot be made. Currently available only for Python.
* The **Framework API** is lower-level, and directly follows the OpenDP programming framework. Available for Python, R and Rust.
```

Polars and Sklearn cross-cut these: we aim to make either accessible from either, and we'll probably expand this set. I'd wait to introduce them until the tabular data and statistical modeling sections, respectively.

Shoeboxam (changes requested):
I agree that this is a better place for these comments, but I have some suggestions about the changes made to the text.

docs/source/getting-started/quickstart.rst
mccalluc: So, essentially restore the original text here, but it still makes sense to move it up to quickstart -- I feel like Polars at least should be mentioned here as well, but maybe it doesn't need to be elevated to a bullet point?

Shoeboxam: That works for me.

docs/source/getting-started/quickstart.rst
```diff
@@ -97,9 +97,20 @@ Let's apply Laplace noise to a value.
             :end-before: /demo
 
 This is obviously not the easiest way to add noise to a number,
-but it demonstrates a number of OpenDP patterns:
+but it demonstrates a number of low-level OpenDP patterns:
 
 * Defining your metric space with ``space_of`` in Python's Context API, or a (domain, distance) tuple in any language..
 * Chaining operators together with ``>>`` in Python and Rust, or ``|>`` in R.
 * Constructing a ``Measurement`` function on your metric space with ``then_laplace``.
 * Invoking that measurement on a value to get a DP release.
```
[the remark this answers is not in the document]
mccalluc: Since these are a sequence of steps, it seems weird to move some but not all of this? I think I like it outside because the main point is to emphasize the similarities across languages.

mccalluc:
I'm dropping any mention of Polars in this PR, and instead we'll update when the Polars docs are actually in place (#1862). We want the state of the docs to be internally consistent with every commit, even if we know the direction we're going.

docs/source/getting-started/quickstart.rst
```diff
@@ -97,9 +97,19 @@ Let's apply Laplace noise to a value.
...
-* Constructing a ``Measurement`` function on your metric space with ``then_laplace``.
-* Invoking that measurement on a value to get a DP release.
+* First, define your metric space.
+* Then, chain operators together and end with a ``Measurement`` function.
+* Invoke that measurement on a value to get a DP release.
+
+OpenDP has two APIs and we'll demonstrate how to use both:
+
+* The **Context API** is simpler and helps to enforce best practices. Currently available only for Python.
+* The **Framework API** is lower-level. Available for Python and R, it mirrors the underlying Rust framework.
```
Shoeboxam, line 109:
```suggestion
* The **Framework API** is lower-level. Available for Python, R and Rust, it directly implements the :ref:`OpenDP Programming Framework <programming-framework>`.
```

docs/source/getting-started/quickstart.rst
```diff
@@ -97,9 +97,19 @@ Let's apply Laplace noise to a value.
...
+
+Because the Context API is a wrapper around the Framework API, it is easier to use but less flexible:
+All calls ultimately pass through the Framework API.
+
+The next page will use the Framework and Context APIs to demonstrate the similarities between the Framework APIs in different languages.
```
Shoeboxam, line 114:
```suggestion
The next page will demonstrate usage of the Context API in Python, and Framework API in Python and R.
```

docs/source/getting-started/code/quickstart-context.rst
```diff
@@ -7,7 +7,8 @@
 # /init
 
 # demo
->>> laplace_mechanism = dp.space_of(float) >> dp.m.then_laplace(scale=1.)
+>>> space = dp.space_of(float)
```
Shoeboxam, line 10: For consistency with the other APIs, lets write this out:

```suggestion
>>> space = dp.atom_domain(T=float), dp.absolute_distance(T=float)
```

While `space_of` is cool when it works, it is a bit magical and isn't universally applicable. I don't think we should remove it from the public APIs, but it should be de-prioritized. The context API can be magical, lets keep examples in the framework API explicit.

docs/source/getting-started/quickstart.rst
```diff
@@ -97,9 +97,19 @@ Let's apply Laplace noise to a value.
             :end-before: /demo
 
 This is obviously not the easiest way to add noise to a number,
-but it demonstrates a number of OpenDP patterns:
+but it demonstrates a number of low-level OpenDP patterns:
```
Shoeboxam, line 100: I might drop "This is obviously not the easiest way to add noise to a number," unless you have an affinity to it.

docs/source/getting-started/quickstart.rst
```diff
@@ -97,9 +97,19 @@ Let's apply Laplace noise to a value.
...
 
-* Defining your metric space with ``space_of`` in Python's Context API, or a (domain, distance) tuple in any language..
-* Chaining operators together with ``>>`` in Python and Rust, or ``|>`` in R.
-* Constructing a ``Measurement`` function on your metric space with ``then_laplace``.
-* Invoking that measurement on a value to get a DP release.
+* First, define your metric space.
```
Shoeboxam, line 102: Your prose is better, as it talks about the concepts and not the syntax.

Maybe another tweak though:
```suggestion
* First, define your metric space (your data domain and definition of neighboring)
```

docs/source/getting-started/quickstart.rst
```diff
@@ -97,9 +97,19 @@ Let's apply Laplace noise to a value.
...
+* Then, chain operators together and end with a ``Measurement`` function.
```
Shoeboxam, line 103:
```suggestion
* Then, chain operators together to construct a ``Measurement`` (aka mechanism).
```

Shoeboxam (changes requested):
Good improvements Chuck! Some suggestions.

docs/source/getting-started/quickstart.rst
```diff
@@ -97,9 +97,19 @@ Let's apply Laplace noise to a value.
             :end-before: /demo
 
 This is obviously not the easiest way to add noise to a number,
```
mccalluc, line 99:
```suggestion
```

docs/source/getting-started/quickstart.rst
mccalluc: will do!

docs/source/getting-started/quickstart.rst
mccalluc: "definition of neighboring" is kind of clunk... Let me hold off on this and re-read the whole thing with updates.

docs/source/getting-started/quickstart.rst
mccalluc: Are we still planning to make a bulk change from "measurement" to "mechanism" at some point?

docs/source/getting-started/quickstart.rst
mccalluc: Ended up with 
> First, define your “metric space”: a data domain and a definition of distance

docs/source/getting-started/quickstart.rst
mccalluc: Now:
> This demonstrates a number of low-level OpenDP patterns:

docs/source/getting-started/quickstart.rst
Shoeboxam: Yes, at some point.

docs/source/getting-started/quickstart.rst
```diff
@@ -96,10 +96,19 @@ Let's apply Laplace noise to a value.
...
-but it demonstrates a number of OpenDP patterns:
+This demonstrates a number of low-level OpenDP patterns:
 
-* Defining your metric space with ``space_of`` in Python's Context API, or a (domain, distance) tuple in any language..
-* Chaining operators together with ``>>`` in Python and Rust, or ``|>`` in R.
-* Constructing a ``Measurement`` function on your metric space with ``then_laplace``.
-* Invoking that measurement on a value to get a DP release.
+* First, define your "metric space": a data domain and a definition of distance.
+* Then, chain operators together to construct a ``Measurement`` (aka mechanism).
+* Invoke that measurement on a value to get a DP release.
```
Shoeboxam, line 103: Not blocking, but for the future:
```suggestion
* Invoke that measurement on a dataset to get a DP release.
```

Shoeboxam (approved):
Good to go!

Issue #1808, opened 2024-07-16
Synthetic Data: Add MWEM Synthesizer


shlomihod:
Add [MWEM](https://www.cs.huji.ac.il/~katrina//papers/mwem-nips.pdf) as the first synthetic data generator in OpenDP. The implementation is based on a stripped-down version from [Smart Noise](https://docs.smartnoise.org/synth/synthesizers/mwem.html).

PR #1813, opened 2024-07-17
feat: Synthetic Data: Add MWEM Synthesizer #1808

shlomihod:
Implementing #1808.


Shoeboxam:
@shlomihod The AIM implementation in #2406 can be parameterized with a pure-DP output measure, and the mbi approach in that PR for postprocessing counts is a more capable algorithm for fitting the joint distribution. How do you feel about closing this as a subset of that work?

PR #1816, opened 2024-07-18
Conversion from privacy profile to tradeoff curve, posterior risk and relative risk

damienbfs:
relative risk.


Shoeboxam:
* **#1828** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1828?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1816** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1816?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @damienbfs and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

Shoeboxam:
Not a full review, but I noticed that creating a tradeoff curve with num_approximations=0 panics. In general, try to avoid unwrapping and panicking in Rust, because it will cause the Python interpreter to shut down.

In cases where you want to quickly return an error, and have an error, look into the try operator in rust (`?`).

Shoeboxam:
Also, given that this is another form of a privacy guarantee from the library, the final result should be computed with conservative arithmetic. If you don't want this to affect compute performance of the function, you can conduct the search in standard arithmetic and then only compute the conservative version once (as done in the δ(ε) rust conversion function I sent you).

rust/src/measures/f_dp/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+        if eps_left.is_nan() {
+            return fallible!(FailedMap, "epsilon(1-alpha) is NaN");
+        }
+        // Find an upper bound for valid epsilons.
+        // This avoids starting epsilon in a regime where the curve is not computable.
+        let mut delta_right = f64::MIN_POSITIVE;
+        while self.epsilon_unchecked(delta_right).is_err() {
+            delta_right *= 2.0;
+        }
+        let mut eps_right = self.epsilon_unchecked(delta_right)?.min(f64::MAX); // epsilon(delta) returns smallest epsilon st. delta(epsilon) ~= 0.0, fine since f(e1, d) > f(e2, d) if e1 < e2
```
damienbfs, line 62: Can we end up in a situation where `delta_right > delta_left = (1.0 - alpha)` ?

rust/src/measures/f_dp/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+            let eps_mid_left = eps_left + third;
+            let eps_mid_right = eps_right - third;
+
+            // Stopping criteria
+            if eps_left == eps_mid_left && eps_right == eps_mid_right {
+                let delta = self.delta(eps_right)?; // Arbitrary between left and right.
+
+                // This is the only place where conservative arithmetic is needed.
+                // If the search doesn't find the "optimal" curve due to numerical issues,
+                // then beta will be overestimated, which is fine.
```
damienbfs, line 75: underestimated?

rust/src/measures/f_dp/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+        let mut eps_right = self.epsilon_unchecked(delta_right)?.min(f64::MAX); // epsilon(delta) returns smallest epsilon st. delta(epsilon) ~= 0.0, fine since f(e1, d) > f(e2, d) if e1 < e2
+
+        loop {
+            let third = (eps_right - eps_left) / 3.0;
+            let eps_mid_left = eps_left + third;
+            let eps_mid_right = eps_right - third;
+
+            // Stopping criteria
+            if eps_left == eps_mid_left && eps_right == eps_mid_right {
+                let delta = self.delta(eps_right)?; // Arbitrary between left and right.
```
damienbfs, line 71: For my own understanding: do we arbitrarily choose because it is good enough, or is there another reason for not taking (eps, delta) pair that gives largest beta?

PR #1817, opened 2024-07-18
docs: release process

### LINKED ISSUE #1689: Document release process
Notes started in 1689-document-release-process

Shoeboxam:
Closes #1689

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -113,7 +113,18 @@ on github, or with the `gh` command line tool. Parameters:
 
 ## Making a release
 
-1. Check that changelog is up-to-date and the `VERSION` file on the `main` branch matches the version you want to release. If not, see step 3.
+1. Update `VERSION` file on the `main` branch to match the version you want to release.
+    - If it's a patch release, no change to the version number should be needed: The patch should have been incremented after the previous release.
+    - If it's a minor release, run `python tools/channel_tool.py bump_version --position minor`
+    - Similarly, if it's a major release, run `python tools/channel_tool.py bump_version --position major`
+
+1. Update changelog.
+
+    ```shell
+    python tools/channel_tool.py changelog
+    ```
+
+    Edit the contents manually. PR [#1672](https://github.com/opendp/opendp/pull/1672) provides a script to get you started (TODO: update these instructions once it is merged).
 
 1. We use a release train with nightly, beta and stable channels. Repeatedly run the Release workflow through each channel (see preceding section).
 
@@ -136,18 +147,19 @@ on github, or with the `gh` command line tool. Parameters:
     If beta release failed, don't forget to increment the beta version counter next time.
     If stable release failed, consider incrementing patch version and starting over.
 
-1. Update `main`.
+1. Increment patch version in `main`.
 
-    1. Update version numbers:
+    ```shell
+    python tools/channel_tool.py bump_version --position patch
+    ```
+    Add arguments as needed to bump or set version.
 
-        ```shell
-        python tools/channel_tool.py bump_version --position minor
-        ```
-        Add arguments as needed to bump or set version.
 
-    1. Create a new changelog:
-        
-        ```shell
-        python tools/channel_tool.py changelog
-        ```
-        Edit the contents manually.
+## Making a dev release
+
+It is also possible to cut a one-off release from a specific branch.
+* Use workflow from: `Branch: your-branch`
+* Target Channel: `dev`
+* Sync the channel from Upstream?: `yes`
+
+No updates to changelogs or versions are necessary.
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -106,7 +106,7 @@ One-off releases can be made with the
 on github, or with the `gh` command line tool. Parameters:
 
 - **Target channel** controls how the release is tagged, and what semantic version is given to the release. There is a git branch with the same name for each channel.
-- The **sync the Channel from upstream?** checkbox is for when you want to update the `nightly`, `beta` `stable` or `dev` branches. This replaces the contents of the respective branch with the branch the workflow is run from ("Use workflow from"). In practice, you'll almost always enable this.
+- The **sync the Channel from upstream?** checkbox is for when you want to update the `nightly`, `beta`, `stable`, or `dev` branches. This replaces the contents of the respective branch with the branch the workflow is run from ("Use workflow from"). In practice, you'll almost always enable this.
 - Update the **version counter** accordingly when you want to release multiple nightlies or betas in the same day.
 - **Dry runs** get sent to test-pypi, and don't update the docs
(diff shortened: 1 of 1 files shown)


Shoeboxam:
* **#1817** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1817?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

.github/workflows/README.md
```diff
@@ -99,14 +99,55 @@ graph TD
...
 - **Target channel** controls how the release is tagged, and what semantic version is given to the release. There is a git branch with the same name for each channel.
-- The **sync the Channel from upstream?** checkbox is for when you want to update the `nightly`, `beta` or `stable` branches.
+- The **sync the Channel from upstream?** checkbox is for when you want to update the `nightly`, `beta` `stable` or `dev` branches. This replaces the contents of the respective branch with the branch the workflow is run from ("Use workflow from"). In practice, you'll almost always enable this.
 - Update the **version counter** accordingly when you want to release multiple nightlies or betas in the same day.
 - **Dry runs** get sent to test-pypi, and don't update the docs
 - **Fake** is for developer convenience when debugging CI: it skips compilation and inserts dummy binaries instead
+
+## Making a release
+
+1. Ensure that changelog is up-to-date and version in `main` matches the version you want to release (otherwise see step 3).
```
mccalluc, line 116: >  1. Ensure that changelog is up-to-date

"Ensure" sounds like we're just checking, while we really need to author the changelog at the start of this process, I think?

```diff
- version in `main`
+ the `VERSION` file on the `main` branch
```

.github/workflows/README.md
```diff
@@ -99,14 +99,55 @@ graph TD
...
+
+2. We use a release train with nightly, beta and stable channels. Repeatedly run the Release workflow through each channel (see [Making one-off releases](#Making-one-off-releases)).
```
mccalluc, line 118:
```suggestion
2. We use a release train with nightly, beta and stable channels. Repeatedly run the Release workflow through each channel (see preceding section).
```

I find really short hop links a little confusing. If you do want to keep it, update the link to reflect the current title.

mccalluc (commented):
Biggest question is where the changelog edits come in. Thanks!

.github/workflows/README.md
```diff
@@ -113,21 +113,21 @@ on github, or with the `gh` command line tool. Parameters:
 
 ## Making a release
 
-1. Ensure that changelog is up-to-date and version in `main` matches the version you want to release (otherwise see step 3).
+1. Check that changelog is up-to-date and the `VERSION` file on the `main` branch matches the version you want to release. If not, see step 3.
```
mccalluc, line 116: Maybe talk this over in person? I might be thinking about this the wrong way, but unless the plan is to start updating the changelog incrementally (which isn't impossible, but in my experience leads to merge conflicts), updating the changelog needs to be at the start of the release process, not the end.

.github/workflows/README.md
```diff
@@ -99,14 +99,55 @@ graph TD
...
+    If stable release failed, consider incrementing patch version and starting over.
+
+1. Update `main`.
+
+    1. Update version numbers:
+
+        ```shell
+        python tools/channel_tool.py bump_version --position minor
+        ```
+        Add arguments as needed to bump or set version.
```
mccalluc, line 146: I would like this change, incrementing the version numbers right after release, but I feel like we might have talked about it earlier and I had thought you weren't sold on it? I may be misremembering.

Right now, if I look at [pypi](https://pypi.org/project/opendp/#history) or [crates.io](https://crates.io/crates/opendp/versions?sort=semver) (sorted by semver) the highest version is v0.10.0, and all the nightlies come below. With this change to our process, there will only be one v0.11 release, and the next nightly will be under v0.12... which makes me wonder if we actually want `--position patch` instead of `--position minor`? We might not actually make a v0.11.1 release and instead go to v0.12 in a few months, but we would have v0.11.1-nightlies.

.github/workflows/README.md
mccalluc: The auto-number is nice, but it does mean that if a step is inserted, "3" might be pointing at the right thing... but my preference is really to just get rid of this branch in the flow chart, and this would be moot.

mccalluc (commented):
Some more comments. I might be confused: Comment threads here might not be the most efficient way to resolve, but whatever makes sense to you? Thanks for working to clarify the process!

.github/workflows/README.md
mccalluc: Follow-up from discussion: 
- If it's a patch release, no change to the version number should be needed: We incremented the patch after the previous release.
- If it's a minor release, `python tools/channel_tool.py bump_version --position minor`
- Similarly, if it's a major release, `python tools/channel_tool.py bump_version --position major`

.github/workflows/README.md
Shoeboxam: I had lost this piece from my brain at some point, thanks for restoring it. Glad we have it documented now.

.github/workflows/README.md
```diff
@@ -99,14 +99,67 @@ graph TD
...
-## Making one-off releases
+## Knobs on the Release Workflow
 
 One-off releases can be made with the
 [`release.yml` workflow](https://github.com/opendp/opendp/actions/workflows/release.yml)
 on github, or with the `gh` command line tool. Parameters:
 
 - **Target channel** controls how the release is tagged, and what semantic version is given to the release. There is a git branch with the same name for each channel.
-- The **sync the Channel from upstream?** checkbox is for when you want to update the `nightly`, `beta` or `stable` branches.
+- The **sync the Channel from upstream?** checkbox is for when you want to update the `nightly`, `beta` `stable` or `dev` branches. This replaces the contents of the respective branch with the branch the workflow is run from ("Use workflow from"). In practice, you'll almost always enable this.
```
mccalluc, line 109:
```suggestion
- The **sync the Channel from upstream?** checkbox is for when you want to update the `nightly`, `beta`, `stable`, or `dev` branches. This replaces the contents of the respective branch with the branch the workflow is run from ("Use workflow from"). In practice, you'll almost always enable this.
```

.github/workflows/README.md
```diff
@@ -99,14 +99,67 @@ graph TD
...
+    If CI fails, debug on main and restart.
+    If beta release failed, don't forget to increment the beta version counter next time.
+    If stable release failed, consider incrementing patch version and starting over.
+
+1. Increment patch version in `main`.
+
+    ```shell
+    python tools/channel_tool.py bump_version --position patch
+    ```
+    Add arguments as needed to bump or set version.
```
mccalluc, line 155:
```suggestion
```

At this point in the process, aren't we always going to only bump the patch version? The comment makes it sound like there are cases where we might do something different.

mccalluc (approved):
A couple missing commas, but this looks great. Small question about the comment on the last increment step.

PR #1822, opened 2024-07-18
Remove epsilon/delta units in docs

### LINKED ISSUE #1802: ensure that ε, δ are not used as units in documentation
We should not use ε, δ as units. Minor edits necessary in compositors docs.

_Originally posted by @Shoeboxam in https://github.com/opendp/opendp/pull/1700#discussion_r1679961467_

mccalluc:
- Fix #1802

Is the right style just the numbers?


Shoeboxam (approved):
Thanks! These small details add up.

PR #1824, opened 2024-07-19
Polars: nit error messages

### LINKED ISSUE #1655: Polars: explain how failure to deserialize lazyframe can also be from missing features
- related to #1721
----
Trying to mock up an analysis for TPDP, I made a LazyFrame plan which included a `Categorical` type, because it seemed appropriate, but it errors:
```
...
group_by_role = pl.LazyFrame(schema={'role': pl.Categorical, 'expertise': pl.Int32}).group_by("role")
mean_expertise_plan = group_by_role.agg(
    pl.col("expertise").dp.mean((0, 5), scale=1.0))
def make_measurement(plan):
    return dp.m.make_private_lazyframe(
        input_domain=lf_domain_with_margin,
        input_metric=dp.symmetric_distance(),
        output_measure=dp.max_divergence(T=float),
        lazyframe=mean_expertise_plan,
        global_scale=1.0)
dp_mean_expertise_by_role = make_measurement(mean_expertise_plan)
```
```
OpenDPException: 
  FFI("Error when deserializing LazyFrame. This may be due to mismatched polars versions. Semantic(None, "unknown variant `Categorical`, expected one of `Boolean`, `UInt8`, `UInt16`, `UInt32`, `UInt64`, `Int8`, `Int16`, `Int32`, `Int64`, `Float32`, `Float64`, `String`, `Binary`, `Date`, `Datetime`, `Duration`, `Time`, `List`, `Array`, `Null`, `Struct`, `Unknown`")")
```

The "mismatched polars versions" is a red herring: we just don't support all the Polars types. This is probably not the best way to figure out the gap:
```
raw = '`Boolean`, `UInt8`, `UInt16`, `UInt32`, `UInt64`, `Int8`, `Int16`, `Int32`, `Int64`, `Float32`, `Float64`, `String`, `Binary`, `Date`, `Datetime`, `Duration`, `Time`, `List`, `Array`, `Null`, `Struct`, `Unknown`'
dp_polars_types = set(t.replace('`','') for t in raw.split(', '))
import re
polars_types = set(dt for dt in dir(pl.datatypes) if re.match(r'[A-Z].*[a-z]', dt))
polars_types - dp_polars_types
{'Categorical',
 'DataType',
 'DataTypeClass',
 'DataTypeGroup',
 'Decimal',
 'Enum',
 'Field',
 'IntegerType',
 'Object',
 'OneOrMoreDataTypes',
 'PolarsDataType',
 'PolarsTemporalType',
 'PythonDataType',
 'SchemaDefinition',
 'SchemaDict',
 'TemporalType',
 'Utf8'}
```
There is one PR out that already tweaks this error message:
- #1636

Add error message that explains that a given type is unsupported.

### LINKED ISSUE #1699: Error: Potential For Overflow
In the developer environment, I had to cast everything to Int32. With a really large max_partition_length was also really large, it resulted in a potential for overflow error. Working with Int64 might fix it. 

``````
OpenDPException: 
  MakeTransformation("potential for overflow when computing function")

During handling of the above exception, another exception occurred:
``````

Run this to reproduce the error
``````
import polars as pl 
import opendp.prelude as dp
dp.enable_features("contrib")

data = {"col1": [0, 2], "col2": [3, 7]}
df = pl.LazyFrame(data, schema={"col1": pl.Int32, "col2": pl.Int32})
    data=df,
    split_evenly_over=10,
    margins={
        ("col1", ): dp.Margin(public_info="keys", max_partition_length=60_000_000, max_partition_contributions=4),
    },
)

(context.query().
group_by("col1").
agg(pl.col("col2").dp.sum((1.0,1000.0)))).release().collect()
``````

### PR DESCRIPTION
**Pull request by Shoeboxam**
Closes #1699
Closes #1655

### DIFF
--- a/rust/src/data/ffi/mod.rs
+++ b/rust/src/data/ffi/mod.rs
@@ -214,7 +214,7 @@ pub extern "C" fn opendp_data__slice_as_object(
         // https://github.com/pola-rs/pyo3-polars/blob/5150d4ca27c287ff4be5cafef243d9a878a8879d/pyo3-polars/src/lib.rs#L147-L153
         // the slice is lf.__getstate__ from the python side and then deserialized here
         ciborium::de::from_reader(slice).map_err(
-            |e| err!(FFI, "Error when deserializing {}. {}", name, e)
+            |e| err!(FFI, "Error when deserializing {}. This may be because you're using features from Polars that are not currently supported. {}", name, e)
         )
     }
     #[cfg(feature = "polars")]
--- a/rust/src/transformations/make_stable_expr/expr_sum/mod.rs
+++ b/rust/src/transformations/make_stable_expr/expr_sum/mod.rs
@@ -185,7 +185,7 @@ macro_rules! impl_accumulator_for_float {
                 if Sequential::<$t>::float_sum_can_overflow(size_limit, (lower, upper))? {
                     return fallible!(
                         MakeTransformation,
-                        "potential for overflow when computing function"
+                        "potential for overflow when computing function. You could resolve this by choosing tighter clipping bounds."
                     );
                 }
                 Sequential::<$t>::relaxation(size_limit, lower, upper)
@@ -204,7 +204,7 @@ macro_rules! impl_accumulator_for_int {
                 if <$t>::int_sum_can_overflow(size_limit, (lower, upper))? {
                     return fallible!(
                         MakeTransformation,
-                        "potential for overflow when computing function"
+                        "potential for overflow when computing function. You could resolve this by choosing tighter clipping bounds or by using a data type with greater bit-depth."
                     );
                 }
                 Ok(0)
--- a/rust/src/transformations/sum/float/checked/mod.rs
+++ b/rust/src/transformations/sum/float/checked/mod.rs
@@ -68,7 +68,7 @@ where
     if S::float_sum_can_overflow(size_limit, bounds)? {
         return fallible!(
             MakeTransformation,
-            "potential for overflow when computing function"
+            "potential for overflow when computing function. You could resolve this by choosing tighter clipping bounds."
         );
     }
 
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

**Comment by Shoeboxam** [comment]
* **#1824** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1824?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

File: rust/src/data/ffi/mod.rs [hunk 1]
```diff
@@ -214,7 +214,7 @@ pub extern "C" fn opendp_data__slice_as_object(
         // https://github.com/pola-rs/pyo3-polars/blob/5150d4ca27c287ff4be5cafef243d9a878a8879d/pyo3-polars/src/lib.rs#L147-L153
         // the slice is lf.__getstate__ from the python side and then deserialized here
         ciborium::de::from_reader(slice).map_err(
-            |e| err!(FFI, "Error when deserializing {}. {}", name, e)
+            |e| err!(FFI, "Error when deserializing {}. This may be because you're using features from Polars that are not currently supported. {}", name, e)
```
mccalluc, line 217: With #1695 and #1663 in place, mismatched polars versions should be reported earlier. Is there anything _besides_ using unsupported features that we would expect to cause this? If so, do we have an issue to improve the reporting for those cases? If not, then could this message be more confident?

mccalluc (approved):
No concerns that definitely should block this, but:
- Do we think anything besides unsupported features could cause deserialization problems?
- Any interest in factoring out repeated strings? Analogous to localization.
- With #1830 and #1832 as goals, is this a place where we'd like to begin having more tests of the content of messages?

Shoeboxam:
* I don't know of other reasons that would cause deserialization problems, but I also don't know what I don't know. I guess the binary itself could be malformed.
* Broader changes to error.rs would facilitate this, I think the error enum could be updated and shared messages could be consolidated there.

PR #1825, opened 2024-07-19
Polars: evenly allocate budget amongst queries, by default

### LINKED ISSUE #1757: Polars: evenly allocate privacy budget amongst DP expressions
When two queries have significantly different sensitivities, the query with less sensitivity will get less noise relative to the other query.

In extreme cases, like when estimating quantiles where alpha breaks into a large fraction (1/3), the sensitivity is very large, and the scale parameter works out to be proportionately large, starving the privacy budget of any other queries.

Shoeboxam:
Closes #1757

### DIFF
--- a/python/test/test_polars.py
+++ b/python/test/test_polars.py
@@ -369,7 +369,7 @@ def test_polars_describe():
             "column": ["len", "A"],
             "aggregate": ["Len", "Sum"],
             "distribution": ["Integer Laplace", "Integer Laplace"],
-            "scale": [8.0, 8.0],
+            "scale": [4.0, 12.0],
         }
     )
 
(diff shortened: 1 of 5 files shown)


Shoeboxam:
* **#1825** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1825?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

python/test/test_polars.py
```diff
@@ -369,7 +369,7 @@ def test_polars_describe():
             "column": ["len", "A"],
             "aggregate": ["Len", "Sum"],
             "distribution": ["Integer Laplace", "Integer Laplace"],
-            "scale": [8.0, 8.0],
+            "scale": [4.0, 12.0],
```
mccalluc, line 372: Can you add a comment explain the 1/3 ratio? I see this in the original issue, and I can imagine that sum is more sensitive than len, but I don't immediately see where that ratio is coming from.

mccalluc (commented):
Add a comment to explain the ratio in the python test? Thanks.

python/test/test_polars.py
```diff
@@ -366,25 +366,32 @@ def test_polars_describe():
...
         }
     )
 
+    summer = pl.col("A").fill_null(2).dp.sum((0, 3))
+
     query = (
         context.query()
         .group_by("B")
-        .agg(pl.len().dp.noise(), pl.col("A").fill_null(2).dp.sum((0, 3)))
+        .agg(pl.len().dp.noise(), summer, summer.alias("B"))
```
mccalluc, line 385: I'm not sure how the introduction of the aliased column strengthens the test.

mccalluc (approved):
The explanation is helpful: Thanks! The alias column is just to make sure it still comes through when it's a little more complicated? Approved.

python/test/test_polars.py
Shoeboxam: It prevents the second sum from shadowing "A". Polars will otherwise reject it.

PR #1831, opened 2024-07-22
Replace ε(δ) curve with δ(ε) curve

damienbfs:


damienbfs:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/opendp/opendp/1831?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1831** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1831?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1828** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1828?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1816** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1816?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

 Join @damienbfs and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

R/opendp/tests/testthat/test-measurements.R
```diff
@@ -67,17 +67,25 @@ test_that("test_gaussian_curve", {
   expect_equal(curve(delta = 1e-3), 0.6880024554878086)
   expect_equal(curve(delta = 1.), 0.)
 
+  expect_equal(curve(epsilon = 0.), 0.1508457845622862)
+  expect_equal(curve(epsilon = 0.6880024554878086), 1e-3)
```
mccalluc, line 71: Prefer to have expressions that explain how floats like this are derived. Risk of just asserting the current behavior instead of confirming that the calculation is actually correct.

python/test/test_meas.py
```diff
@@ -7,21 +7,30 @@ def test_gaussian_curve():
     meas = dp.c.make_zCDP_to_approxDP(dp.m.make_gaussian(*input_space, 4.))
     profile = meas.map(d_in=1.)
     assert profile.epsilon(delta=0.) == float('inf')
-    assert profile.epsilon(delta=1e-3) == 0.6880024554878086
+    assert profile.epsilon(delta=1e-3) == 0.6880024554878085
```
mccalluc, line 10: Was the old value actually wrong? Should the test be less strict?

python/test/test_meas.py
```diff
@@ -7,21 +7,30 @@ def test_gaussian_curve():
...
     assert profile.epsilon(delta=1.) == 0.
+    assert profile.delta(epsilon=0.) == 0.1508457845622862
+    assert profile.delta(epsilon=0.6880024554878085) == 1e-3
```
mccalluc, line 13: ditto: explain floats.

python/test/test_meas.py
```diff
@@ -7,21 +7,30 @@ def test_gaussian_curve():
...
 
     profile = dp.c.make_zCDP_to_approxDP(dp.m.make_gaussian(*input_space, 4.)).map(d_in=0.0)
     assert profile.epsilon(0.0) == 0.0
     with pytest.raises(Exception):
         profile.epsilon(delta=-0.0)
-
+    with pytest.raises(Exception):
+        profile.delta(epsilon=-0.0)
```
mccalluc, line 20: I see that R has a negative zero that is not quite the same as zero, but does Python?

Also, preference for stronger checks with `pytest.raises`:
- It's probably a subclass of `Exception`?
- Usability: Is the message something an end user would understand?

rust/src/combinators/measure_cast/zCDP_to_approxDP/cdp_delta/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+        return fallible!(FailedMap, "epsilon must be non-negative");
+    }
+
+    if rho.is_zero() || eps.is_infinite() {
+        return Ok(0.0);
+    }
+
+    if rho.is_infinite() {
+        return Ok(1.0);
+    }
```
mccalluc, line 28: Is the special casing for infinity something that should also show up on the Python side? Should there be python tests with `float('inf')`?

rust/src/combinators/measure_cast/zCDP_to_approxDP/cdp_delta/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+
+        let a_mid = a_min + diff / 2.0;
+
+        if a_mid == a_max || a_mid == a_min {
+            break;
+        }
+
+        // calculate derivative
+        let deriv = (2.0 * a_mid - 1.0) * rho - eps + a_mid.recip().neg().ln_1p();
+        //        = (2α - 1)            ρ   - ε   + ln1p(-1/α)
```
mccalluc, line 68:
```suggestion
        //        = (2α - 1)            * ρ   - ε   + ln1p(-1/α)
```

rust/src/combinators/measure_cast/zCDP_to_approxDP/cdp_delta/test.rs
```diff
@@ -0,0 +1,15 @@
+use super::*;
+#[test]
+fn test_edge_cases() -> Fallible<()> {
+    // negativity checks
+    assert!(cdp_delta(-0., 0.).is_err());
+    assert!(cdp_delta(0., -0.).is_err());
+
+    assert_eq!(cdp_delta(0., 0.)?, 0.);
+    assert_eq!(cdp_delta(0.5, 0.)?, 0.5588356393474351);
```
mccalluc, line 9: ditto: replace with expression?

rust/src/measures/mod.rs
```diff
@@ -67,13 +67,83 @@ impl Clone for PrivacyProfile {
...
-        (self.0)(delta)
+    pub fn epsilon(&self, delta: f64) -> Fallible<f64> {
+        // reject negative zero
+        if delta.is_sign_negative() {
+            return fallible!(FailedMap, "delta must not be negative");
+        }
+
+        if !(0.0..=1.0).contains(&delta) {
+            return fallible!(FailedMap, "delta must be between zero and one");
+        }
```
mccalluc, line 82: The first check seems to be covered by the second. Also, do we want to include the actual value of delta in the message? Last week there was a PR that added a lot of values to messages to help with debugging, and I'd think that would be the default going forward.

rust/src/combinators/measure_cast/zCDP_to_approxDP/cdp_delta/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+mod test;
+
+pub(crate) fn cdp_delta(rho: f64, eps: f64) -> Fallible<f64> {
+    if rho.is_sign_negative() {
+        return fallible!(FailedMap, "rho must be non-negative");
+    }
+
+    if eps.is_sign_negative() {
+        return fallible!(FailedMap, "epsilon must be non-negative");
+    }
```
mccalluc, line 20: Include values in error messages?

rust/src/traits/arithmetic/mod.rs
```diff
@@ -525,11 +525,123 @@ macro_rules! impl_float_inf_uni {
...
+impl InfExp for f64 {
+    fn inf_exp(self) -> Fallible<Self> {
+        let not_finite = || {
+            err!(
+                Overflow,
+                concat!(
+                    "({}).",
+                    stringify!(inf_exp),
+                    "() is not finite. Consider tightening your parameters."
+                ),
```
mccalluc, line 543: Here and below, I think `concat!` just takes literals? Do you want a `format!` in here?

mccalluc (commented):
As usual, I think the comments I'm making are pretty superficial, but this is a case where I think I could understand the algorithm with a bit of study: Let me know if it's worth looking at anything more closely.

python/test/test_meas.py
Shoeboxam: They are both valid upper bounds. This calculation seems to be very slightly more numerically stable.

python/test/test_meas.py
Shoeboxam: Negative zero is equivalent to positive zero, but can be used by some software to represent underflow of a negative number. 
It's a part of the IEEE-754 standard, so it's present in R as well (and any software that adheres to that standard, which is basically all software).

rust/src/combinators/measure_cast/zCDP_to_approxDP/cdp_delta/mod.rs
Shoeboxam: It is tested, maybe surprisingly-- it's the case where scale is zero. Zero noise means the distributions are disjoint, resulting in infinite rho.

rust/src/measures/mod.rs
Shoeboxam: It does two things: rejects negative zero, and gives a better error message.

rust/src/combinators/measure_cast/zCDP_to_approxDP/cdp_delta/test.rs
Shoeboxam: The expression is pretty complex. I just copied the prior implementation and checked for equality. I think it would be good to keep the reverse map from bit rotting as well.

Shoeboxam: I have also checked the new implementation against the reference implementation.

R/opendp/tests/testthat/test-measurements.R
Shoeboxam: I updated to include code here that uses the reverse map. I think it makes more sense to have two implementations and check that they're matching, and point people to the docs of your dependency functions about this. The formula itself is not easy to explain or describe.

python/test/test_meas.py
```diff
@@ -7,10 +7,14 @@ def test_gaussian_curve():
...
     assert profile.epsilon(delta=0.) == float('inf')
-    # see cdp_delta for formula of 0.688...
+    # see cdp_delta for formula of 0.688 and 0.151
+
+    # cdp_epsilon(rho=(1/4)^2 / 2, delta=1e-3)
     assert profile.epsilon(delta=1e-3) == 0.6880024554878085
     assert profile.epsilon(delta=1.) == 0.
+    # cdp_delta(rho=(1/4)^2 / 2, epsilon=0.0)
     assert profile.delta(epsilon=0.) == 0.1508457845622862
+    # reuse the conent above
```
mccalluc, line 17:
```suggestion
    # reuse the content above
```

rust/src/combinators/measure_cast/zCDP_to_approxDP/cdp_delta/mod.rs
```diff
@@ -1,3 +1,47 @@
+// Copyright (c) 2022 President and Fellows of Harvard College
```
mccalluc, line 1:
```suggestion
// Copyright (c) 2024 President and Fellows of Harvard College
```
If the year even matters, and if we want to restate the license on this file?

mccalluc (approved):
Latest changes look good to me. Typos are inconsequential.

PR #1815, opened 2024-07-18
docs: features and env vars

Shoeboxam:
Closes #1814

PR #1839, opened 2024-07-23
Polars: OPENDP_LIB_PATH -> OPENDP_POLARS_LIB_PATH

### LINKED ISSUE #1838: Polars: OPENDP_LIB_PATH -> OPENDP_POLARS_LIB_PATH

Shoeboxam:
Closes #1838

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -295,6 +295,8 @@ jobs:
         run: |
           shopt -s globstar
           pytest --nbmake source/**/*.ipynb -n=auto
+          # Messes with envvars, so can't be run in parallel!
+          FORCE_TEST_REPLACE_BINARY_PATH=1 pytest -k 'test_replace_binary_path'
 
   r-test:
     needs: rust-build
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -193,6 +193,11 @@ jobs:
         # (Even in unittests we should use "importskip" rather than "skipif".)
         run: coverage run -m pytest -v ${{ ! contains(matrix.extra-install, 'polars') && '-k "not lazyframe and not polars and not pca"' || ''}}
 
+      - name: Test replace-binary-path
+        if: ${{ contains(matrix.extra-install, 'polars') }}
+        # Messes with envvars, so can't be run in parallel!
+        run: FORCE_TEST_REPLACE_BINARY_PATH=1 pytest -k 'test_replace_binary_path'
+
       - name: Test coverage
         # We only expect test coverage with all extra dependencies.
         # "polars" implies "scikit-learn": see setup.cfg.
@@ -295,8 +300,6 @@ jobs:
         run: |
           shopt -s globstar
           pytest --nbmake source/**/*.ipynb -n=auto
-          # Messes with envvars, so can't be run in parallel!
-          FORCE_TEST_REPLACE_BINARY_PATH=1 pytest -k 'test_replace_binary_path'
 
   r-test:
     needs: rust-build
--- a/docs/source/contributing/development-environment.rst
+++ b/docs/source/contributing/development-environment.rst
@@ -297,6 +297,7 @@ Environment Variables
        When OpenDP is used as a query server, library paths in queries submitted by clients are stale (local to the client).
        This environment variable overrides paths in new OpenDP Polars plugins and OnceFrames.
        For Python, you can read this value from ``opendp._lib.lib_path`` (read-only).
+       This is separate from ``OPENDP_LIB_DIR`` because we anticipate it diverging for R.
    * - ``OPENDP_HEADLESS``
      - Used by CI. When ``true``, The Python ``opendp`` package will import without the presence of the OpenDP Library binary.
    * - ``OPENDP_SPHINX_PORT`` and ``OPENDP_SPHINX_URI``
--- a/python/test/test_polars.py
+++ b/python/test/test_polars.py
@@ -549,7 +549,7 @@ def test_replace_binary_path():
     os.environ["OPENDP_POLARS_LIB_PATH"] = __file__
     pl = pytest.importorskip("polars")
 
-    with pytest.raises(dp.OpenDPException) as err:
+    with pytest.raises(dp.OpenDPException):
         dp.m.make_private_expr(
             dp.expr_domain(example_lf()[0], grouping_columns=[]),
             dp.partition_distance(dp.symmetric_distance()),
--- a/python/test/test_polars.py
+++ b/python/test/test_polars.py
@@ -544,20 +544,32 @@ def test_polars_threshold():
         .collect()
     )
 
-@pytest.mark.skipif(os.getenv('FORCE_TEST_REPLACE_BINARY_PATH') != "1", reason="setting OPENDP_POLARS_LIB_PATH interferes with the execution of other tests")
+@pytest.mark.skipif(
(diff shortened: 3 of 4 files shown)


Shoeboxam:
* **#1839** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1839?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

mccalluc (changes requested):
CI failing.

I was remembering PR #1815, but looking at it, I see that `OPENDP_POLARS_LIB_PATH` is used there, so missing `OPENDP_LIB_PATH` at that time was the oversight?

I also wonder, if the variables can't merged, whether we could rename `OPENDP_LIB_DIR` to `OPENDP_R_LIB_PATH` for consistency? ... although it looks like that one is also used by `OPENDP_TEST_RELEASE`?

Shoeboxam:
@mccalluc I introduced a second env var because:
* In Python: `OPENDP_POLARS_LIB_PATH` is in `OPENDP_LIB_DIR`
* In R (TBD): `OPENDP_POLARS_LIB_PATH` is separate from `OPENDP_LIB_DIR`, because R wraps the staticlib into a new dylib.

python/test/test_polars.py
```diff
@@ -532,18 +533,21 @@ def test_polars_threshold():
         .collect()
     )
 
-
+@pytest.mark.skip(reason="setting OPENDP_POLARS_LIB_PATH interferes with the execution of other tests")
```
mccalluc, line 536: It feels like one consequence of this PR is that a test that used to run will no longer be run? That doesn't seem like the right direction, but I might be missing something.

mccalluc (commented):
Not sure what the test is doing for us if we don't run it? Thank you for more explanation on the envvars: Maybe that can be added to the docs?

python/test/test_polars.py
Shoeboxam: There was one new test added in this PR, and it gets skipped by default. I don't have a good way to test this functionality within our current CI because it causes race issues/failures in other tests, when tests are run in parallel.

mccalluc: It's a bit kludgy, but in smoke-test.yml maybe run everything except this test, and then run just this test?
```
pytest --nbmake source/**/*.ipynb -n=auto -k 'not test_replace_binary_path'
pytest -k 'test_replace_binary_path'
```

Shoeboxam: I think pytest should run out-of-the box without special commands. Pushing a different approach.

.github/workflows/smoke-test.yml
```diff
@@ -295,6 +295,8 @@ jobs:
         run: |
           shopt -s globstar
           pytest --nbmake source/**/*.ipynb -n=auto
+          # Messes with envvars, so can't be run in parallel!
+          FORCE_TEST_REPLACE_BINARY_PATH=1 pytest -k 'test_replace_binary_path'
```
mccalluc, line 299: This is in the `python-notebooks` block: Should be under `python-test`? Unless it's here to avoid the test matrix?

mccalluc: ... but still, I think this should have located and run the test: Something else is wrong.

docs/source/contributing/development-environment.rst
```diff
@@ -297,6 +297,7 @@ Environment Variables
        When OpenDP is used as a query server, library paths in queries submitted by clients are stale (local to the client).
        This environment variable overrides paths in new OpenDP Polars plugins and OnceFrames.
        For Python, you can read this value from ``opendp._lib.lib_path`` (read-only).
+       This is separate from ``OPENDP_LIB_DIR`` because we anticipate it diverging for R.
```
mccalluc, line 300: Thanks for this explanation!

mccalluc (commented):
Test is failing right now. Not sure if it's being run from the right section of smoke-test.

.github/workflows/smoke-test.yml
Shoeboxam: Indeed something was wrong. Sorry for the iterations on this one.

python/test/test_polars.py
```diff
@@ -543,21 +544,33 @@ def test_polars_threshold():
         .collect()
     )
 
-
+@pytest.mark.skipif(
+    os.getenv('FORCE_TEST_REPLACE_BINARY_PATH') != "1", 
+    reason="setting OPENDP_POLARS_LIB_PATH interferes with the execution of other tests"
+)
 def test_replace_binary_path():
     import os
```
mccalluc, line 552:
```suggestion
def test_replace_binary_path():
```
Locally, I can remove the skipif, and the redundant `import os`, and `pytest` passes for me, and it seems like it would be good to run all the tests by default? However, if you usually run the tests in parallel then introducing `FORCE_TEST_REPLACE_BINARY_PATH` may make sense...

python/test/test_polars.py
```diff
@@ -543,21 +544,33 @@ def test_polars_threshold():
...
-        pl.len().dp.noise(scale=1.),
+        expr,
     )
+
     assert str(m_expr((pl.LazyFrame(dict()), pl.all()))) == "len().testing!:noise_plugin()"
 
-    del os.environ["OPENDP_LIB_PATH"]
+    # check that local paths in new expressions get overwritten
+    os.environ["OPENDP_POLARS_LIB_PATH"] = __file__
+    assert str(pl.len().dp.noise(scale=1.)) == f"len().{__file__}:noise([null, dyn float: 1.0])"
```
mccalluc, line 570:
```suggestion
    # check that local paths in new expressions get overwritten
    assert str(pl.len().dp.noise(scale=1.)) == f"len().testing!:noise([null, dyn float: 1.0])"
```

Unless there's really something different about `__file__` here, maybe just leave the envvar as is, to keep it simple?

.github/workflows/smoke-test.yml
```diff
@@ -193,6 +193,11 @@ jobs:
         # (Even in unittests we should use "importskip" rather than "skipif".)
         run: coverage run -m pytest -v ${{ ! contains(matrix.extra-install, 'polars') && '-k "not lazyframe and not polars and not pca"' || ''}}
 
+      - name: Test replace-binary-path
+        if: ${{ contains(matrix.extra-install, 'polars') }}
+        # Messes with envvars, so can't be run in parallel!
+        run: FORCE_TEST_REPLACE_BINARY_PATH=1 pytest -k 'test_replace_binary_path'
```
mccalluc, line 199:
```suggestion
        run: coverage run -m pytest -v ${{ ! contains(matrix.extra-install, 'polars') && '-k "not lazyframe and not polars and not pca"' || 'not test_replace_binary_path'}}

      - name: Test replace-binary-path
        if: ${{ contains(matrix.extra-install, 'polars') }}
        # Messes with envvars, so can't be run in parallel!
        run: pytest -k 'test_replace_binary_path'
```
Maybe just use `-k` instead of introducing a new envvar?

mccalluc (approved):
Unless you're running tests in parallel locally, I'd think it would be simpler to just use `-k` to select tests rather than introducing `FORCE_TEST_REPLACE_BINARY_PATH`... but maybe you do parallelize tests locally?

I'm really fine with it as-is... but if you did want to make a few more tweaks, that would be ok too.

.github/workflows/smoke-test.yml
Shoeboxam: I want to avoid needing "special -k" configuration on the happy path. Without the env var, `pytest -n auto` fails.

PR #1841, opened 2024-07-23
Polars: test pickle bombs

### LINKED ISSUE #1840: Polars: test pickle bombs
We should have CI tests that ensure that manipulation of LazyFrames never triggers pickle deserialization. If this ever changes in Polars, we'll then be alerted by failing CI.

Shoeboxam:
Closes #1840

### DIFF
--- a/python/test/test_polars.py
+++ b/python/test/test_polars.py
@@ -548,3 +548,72 @@ def test_replace_binary_path():
     assert str(m_expr((pl.LazyFrame(dict()), pl.all()))) == "len().testing!:noise_plugin()"
 
     del os.environ["OPENDP_LIB_PATH"]
+
+
+def test_pickle_bomb():
+    pl = pytest.importorskip("polars")
+
+    from polars._utils.parse import parse_into_list_of_expressions  # type: ignore[import-not-found]
+    from polars._utils.wrap import wrap_expr  # type: ignore[import-not-found]
+    from opendp._lib import lib_path
+    import io
+    import re
+    import pickle
+
+    # modified from https://intoli.com/blog/dangerous-pickles/
+    poison_binary = b'c__builtin__\neval\n(V1 / 0\ntR.'
+
+    # poison_binary raises a ZeroDivisionError if it is ever unpickled
+    with pytest.raises(ZeroDivisionError):
+        pickle.loads(poison_binary)
+
+    # craft an expression that contains a poisoned pickle binary
+
+    py_exprs = parse_into_list_of_expressions((pl.len(), pl.lit("Laplace"), pl.lit(0.7)))
+
+    # Replicates parts of register_plugin_function from Polars,
+    # to allow injection of the specially-crafted pickle binary
+    bomb_expr = wrap_expr(
+        pl.polars.register_plugin_function(
+            plugin_path=str(lib_path),
+            function_name="noise",
+            args=py_exprs,
+            kwargs=poison_binary,
+            is_elementwise=True,
+            input_wildcard_expansion=False,
+            returns_scalar=False,
+            cast_to_supertype=False,
+            pass_name_to_apply=False,
+            changes_length=False,
+        )
+    )
+
+    # craft a lazyframe that contains a poisoned pickle binary
+    lf_domain, lf = example_lf()
+    bomb_lf = lf.select(bomb_expr)
+
+    # ensure that ser/de round-trip of expression does not trigger pickle
+    ser_expr = bomb_expr.meta.serialize()
+    pl.Expr.deserialize(io.BytesIO(ser_expr))
+
+    # ensure that ser/de round-trip of lazyframe does not trigger pickle
+    ser_lf = bomb_lf.serialize()
+    pl.LazyFrame.deserialize(io.BytesIO(ser_lf))
+
+    # OpenDP explicitly rejects any pickled data it finds
+    err_msg_re = re.escape("OpenDP does not allow pickled keyword arguments as they may enable remote code execution.")
+    with pytest.raises(dp.OpenDPException, match=err_msg_re):
+        dp.m.make_private_expr(
+            dp.expr_domain(lf_domain, grouping_columns=[]),
+            dp.partition_distance(dp.symmetric_distance()),
+            dp.max_divergence(T=float),
+            bomb_expr,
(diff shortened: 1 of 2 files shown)


Shoeboxam:
* **#1841** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1841?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

python/test/test_polars.py
```diff
@@ -548,3 +548,69 @@ def test_replace_binary_path():
     assert str(m_expr((pl.LazyFrame(dict()), pl.all()))) == "len().testing!:noise_plugin()"
 
     del os.environ["OPENDP_LIB_PATH"]
+
+
+def test_pickle_bomb():
+    pl = pytest.importorskip("polars")
+
+    from polars._utils.parse import parse_into_list_of_expressions  # type: ignore[import-not-found]
```
mccalluc, line 556:
```suggestion
    from polars._utils.parse import parse_into_list_of_expressions  # type: ignore[import-not-found]
    from polars._utils.wrap import wrap_expr
```
Do all the imports at the top?

I'm wary of the imports from `_utils`: Definitely wouldn't rely on this outside of tests, but maybe if it helps us write a clean test, it's worth the risk? Any way we could use a higher-level, public interface?

python/test/test_polars.py
```diff
@@ -548,3 +548,69 @@ def test_replace_binary_path():
...
+
+    poison_binary = b'c__builtin__\neval\n(V1 / 0\ntR.'
+
+    # poison_binary raises a ZeroDivisionError if it is ever unpickled
+    with pytest.raises(ZeroDivisionError):
+        pickle.loads(poison_binary)
+
+    # craft an expression that contains a poisoned pickle binary
+    from polars._utils.wrap import wrap_expr
+
```
mccalluc, line 569:
```suggestion
```

python/test/test_polars.py
```diff
@@ -548,3 +548,69 @@ def test_replace_binary_path():
...
+    pyexprs = parse_into_list_of_expressions((pl.len(), pl.lit("Laplace"), pl.lit(0.7)))
+    bomb_expr = wrap_expr(
+        pl.polars.register_plugin_function(
+            plugin_path=str(lib_path),
+            function_name="noise",
+            args=pyexprs,
```
mccalluc, line 575:
```suggestion
    py_exprs = parse_into_list_of_expressions((pl.len(), pl.lit("Laplace"), pl.lit(0.7)))
    bomb_expr = wrap_expr(
        pl.polars.register_plugin_function(
            plugin_path=str(lib_path),
            function_name="noise",
            args=py_exprs,
```

rust/src/polars/mod.rs
```diff
@@ -57,6 +57,10 @@ where
                 return Ok(None);
             }
 
+            if !kwargs.is_empty() {
+                return fallible!(FailedFunction, "OpenDP does not allow pickled keyword arguments as they may enable remote code execution.");
```
mccalluc, line 61: My understanding is that pickle is specific to python, but polars also supports R: If any eventual R code might also exercise this path, should the error message be more language-agnostic?

python/test/test_polars.py
```diff
@@ -548,3 +548,69 @@ def test_replace_binary_path():
...
+    assert err.value.message == err_msg
+
+    with pytest.raises(dp.OpenDPException) as err:
+        dp.m.make_private_lazyframe(
+            lf_domain,
+            dp.symmetric_distance(),
+            dp.max_divergence(T=float),
+            bomb_lf,
+        )
+    assert err.value.message == err_msg
```
mccalluc, line 616: use the `match` kwarg?
```suggestion
    err_msg_re = re.escape("OpenDP does not allow pickled keyword arguments as they may enable remote code execution.")
    with pytest.raises(dp.OpenDPException, match=err_msg_re) as err:
        dp.m.make_private_expr(
            dp.expr_domain(lf_domain, grouping_columns=[]),
            dp.partition_distance(dp.symmetric_distance()),
            dp.max_divergence(T=float),
            bomb_expr,
        )

    with pytest.raises(dp.OpenDPException, match=err_msg_re) as err:
        dp.m.make_private_lazyframe(
            lf_domain,
            dp.symmetric_distance(),
            dp.max_divergence(T=float),
            bomb_lf,
        )
```

mccalluc (commented):
A couple little tweaks. Only real concern is with using non-public methods, but they might be the best option here?

python/test/test_polars.py
```diff
@@ -548,3 +548,69 @@ def test_replace_binary_path():
...
+
+def test_pickle_bomb():
+    pl = pytest.importorskip("polars")
+
+    from polars._utils.parse import parse_into_list_of_expressions  # type: ignore[import-not-found]
+    from opendp._lib import lib_path
+    import io
+    import pickle
+
+    poison_binary = b'c__builtin__\neval\n(V1 / 0\ntR.'
```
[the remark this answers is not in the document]
Shoeboxam: I'm not sure how I would produce this with pickle.dumps.

python/test/test_polars.py
Shoeboxam: To use a higher-level Polars API, I'd have to further reverse-engineer the pickle protocol. I'd rather not spend more time on that, when this is so simple already.

python/test/test_polars.py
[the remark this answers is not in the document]
Shoeboxam: Left a comment to a blog post!

rust/src/polars/mod.rs
Shoeboxam: Polars doesn't directly support plugins from R at this point, but if you know how things work, you can still write those DSL nodes for plugins into expressions in R. Now that we don't use pickling, R will be able to fill kwargs fields in these new DSL nodes with empty strings.


Ultimately, the slot to fill in these DSL nodes is binary pickle data, so if someone were to craft a DSL node with data in there from R, the error is still appropriate.

mccalluc (approved):
Thanks! Sorry for the quibbles.

damienbfs (approved):
Thanks a lot for adding the test!

PR #1844, opened 2024-07-23
Add `polars` to `extras`

mccalluc:
Builds on:
- #1452

Newly introduced notebooks will need to be updated.

One change is to move `extras.py` to `extras/__init__.py`: This doesn't require any code changes, but it makes the sphinx rendering more consistent.


Shoeboxam (dismissed):
I love the consistency. Approved pending CI.

mccalluc:
One question this raises for me is whether `Margin` really belongs under `polars`. Waiting to see if tests pass...

Shoeboxam (approved):
I think keeping it explicitly a *Polars* margin is ideal, because it should remain consistent with the Polars Margin struct. We'll start to have problems if its fields stray from what with_margin takes.

PR #1858, opened 2024-07-30
Just enable all features at start of test run

### LINKED ISSUE #1852: Consolidate `enable_features` in tests
It looks like python tests are handling this on a piecemeal basis, but if any particular test wanted to make sure a particular feature was _not_ enabled, it couldn't assume a clean slate, and instead would need to do that itself. (FWIW, we have precisely one call to `disable_features` in the python tests.)

So let's just turn on all the features in the test startup code and clean up the 30 references in the tests.

mccalluc:
- Fix #1852


Shoeboxam (approved):
Approved, just need to rebase.

Shoeboxam (approved):
These look good to me!

PR #1861, opened 2024-07-31
Fix dp.np_array2_domain -> dp.numpy.array2_domain

### LINKED ISSUE #1859: CI notebook fails because method name changed
```
def make_theil_sen_percentiles(runs: int=1):
    # Defines type of input. 
    input_domain=dp.np_array2_domain(T=float, num_columns=2),
```
It's now under `extras`. I'll fix.

mccalluc:
- Fix #1859


raprasad (approved):
Approved

PR #1866, opened 2024-07-31
Refactor extras.sklearn for consistency

mccalluc:
Follow-up to:
- #1452

Introduces `sklearn.pca` and puts everything PCA-related under it. The idea is that other sklearn topics might get their own submodules. This structure is really driven by the documentation: I've moved the private methods along with the public, but they could live in a separate file/directory.


python/src/opendp/extras/sklearn/pca/__init__.py
Shoeboxam, line 1: Technically make_private_pca (and all other make_ functions in this module) don't need sklearn. Although I guess the organizational convenience is better than being technically tighter.

Shoeboxam: I think it would be better to leave `make_pca` in its current location, and add a `decomposition` module for the PCA class, to more closely match the SciKit-Learn API.

Shoeboxam (changes requested):

python/src/opendp/extras/sklearn/pca/__init__.py
Shoeboxam: I feel less strongly about the location of `make_` APIs, but we might as well organize SciKit-Learn mimicry in the same way SciKit-Learn does.

Shoeboxam: https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html

mccalluc: > I think it would be better to leave `make_pca` in its current location

Not sure if you mean the state on main, or the state in this branch? PCA should now be in the right place, though.

Shoeboxam (approved):
Thanks for the iterations. Locations of public code seems good to me at this point. 
Conflict should resolve once GitHub rebases this PR in merge.

Issue #1830, opened 2024-07-22
Extend coverage to include raise and except


Shoeboxam:
_Originally posted by @mccalluc in https://github.com/opendp/opendp/pull/1823#pullrequestreview-2191712841_

This PR also highlights that we've configured some exceptions in `.coveragerc`:
```
exclude_also =
  # TODO: Improve coverage and shorten this list!
  ^\s*raise\b
  ^\s*except\b
  os\.environ
  ^\s*\.\.\.\s*$
```

mccalluc:
We'll first remove these from the ignore list, and add line-by-line ignores where needed, and then more slowly clean those up.

PR #1867, opened 2024-07-31
Add checks on "except" lines for test coverage

mccalluc:
- Toward #1830

My sense is that many of these will either be hard to test, or not a big win even when they are tested.
Still, I think it makes sense to improve coverage a little here before trying to handle `raise`: There are 106 lines where coverage is skipped because of `raise`, vs 42 that are touched in this PR for `except`, and where the exception is one our own code has thrown, covering the `except` will necessarily also cover the `raise`.

Not a high priority.


Shoeboxam (approved):
Pragma slightly better than exclude. Not a big deal, but I'm fine with it. Needs rebase.

mccalluc:
> Pragma slightly better than exclude. Not a big deal, but I'm fine with it.

If I'm understanding you correction, you're right, this is a very small change, but it's a prereq to improving coverage line by line.

Shoeboxam (approved):

PR #1875, opened 2024-08-01
remove old dataframe docs

### LINKED ISSUE #1874: Remove the dataframe section from `transformations/index.rst`
We want to encourage people to use polars, although I suppose R users might still use these methods?
A PR will be easy, and we can close if it's not the right direction.

mccalluc:
- Fix #1874

If we should keep them around as a reference for R users, I'd be ok with that: Just close if this isn't the right direction.


Shoeboxam (approved):
I agree it's best we de-emphasize these. We could merge it now, or wait until Polars docs go up. No preference.

Issue #927, opened 2023-08-18
Document new release process


andrewvyrros:
`maintainer-notes.rst` is woefully out of date.

mccalluc:
- #1817 got the release process docs up to date
- #1879 removes the rest of the out of date maintainer-notes

PR #1879, opened 2024-08-02
Clean up old maintainer notes

mccalluc:
Most of the material in maintainer-notes was
- covered by the workflow docs
- or covered by the contribution process
- or out of date. (#927)

Also: de-emphasized the rust and proof init docs, but linked them from the appropriate sections of development-environment.


mccalluc:
> Why de-emphasize these sections of the docs? The proofs intro in particular seems important, and it now has an external link from the invites so I'd be careful that the URL doesn't change.

Let me move the init sections back where they were for URL stability. For someone trying to get started, I worry that this page can be intimidating. I guess I'd like a path where there's enough information about Rust so they can build the software, without getting in mired in other details.

mccalluc:
(The new mini-toctrees I think are an improvement because they offer more information at the point where there might be questions just about Rust or about proof writing.)

mccalluc:
Not urgent, but would like to get this merged before adding a note on the cargo build for R:
- #1918

Shoeboxam (approved):
Changes look good now. I think the READMEs are a better location for the maintainer notes anyways.

PR #1877, opened 2024-08-02
remove references to binder

mccalluc:
- Fix #1873
- also, fix #903, or at least render it moot
- close #955, by dropping binder

If it turns out we _don't_ want to do this, that would also be a good thing to learn.

PR #1880, opened 2024-08-02
naively merge notebook tests

mccalluc:
This is a proposal: It makes python development a little easier by consolidating 2 virtual environments. (You'd still have a separate environment for the docs build.) It does bump up the minimum Python version that is tested, and that might not be acceptable.

- Towards #739
- Assumes #1877 goes forward: `requirments_notebooks.txt` is referenced by `.binder/requirements.txt`


mccalluc:
Recent versions of ipykernel require Python>=3.10.
- Do not want to give up testing against a range of Python versions!
- Try pinning to an older version of ipykernel?
- Give up on having one requirements.txt for all dev environments?
- See if there are any tools in pip-compile to support fallback dependencies?

.github/workflows/smoke-test.yml
```diff
@@ -137,7 +137,7 @@ jobs:
     strategy:
       matrix:
         python-version:
-          - '3.9'
+          - '3.10'
```
Shoeboxam, line 140: I'm reluctant to drop support for 3.9 due to a frozen version in a dev-dependency (ipython) that doesn't span 3.8-3.12. 

3.8 wasn't even EOL yet, but since it hits EOL in October, and it really simplified things, we dropped it.

I don't know what to suggest here, aside from installing from the requirements-dev.in on 3.9 or having separate pip-compile files.

Shoeboxam (changes requested):
Sorry to be a bummer here.

.github/workflows/smoke-test.yml
mccalluc: ipython 8.18 might be the last one that supports older versions?

mccalluc:
I think it would be good to first simplify dependencies with
- #1990

before making them more complicated with this PR.

mccalluc:
> I think it would be good to first simplify dependencies with
> 
> - #1990
> 
> 
> before making them more complicated with this PR.

That's closed now, so this should be simpler.

mccalluc:
I'm going to close this:
- The parent issue is closed.
- Some of this is address by porting notebooks to rst, instead of worrying more about the notebook environment
- The tradeoffs in this PR (in particular, dropping 3.9 from the test matrix) aren't worth any gain in simplicity.

Issue #1885, opened 2024-08-05
Exception in Windows opendp installation


K-Meech:
**Describe the bug**
Opendp throws an error when imported in a fresh Windows installation (see below)

**To Reproduce**
I installed opendp as follows on a Windows 11 computer:
```
conda create -n opendp python=3.10
pip install opendp
```
When I `import opendp` it throws:

```
  File "C:\Users\kimme\anaconda3\envs\opendp\lib\site-packages\opendp\_lib.py", line 31, in _load_library
    raise Exception(f"Expected exactly one binary to be present. Got: {lib_dir_file_names}")
Exception: Expected exactly one binary to be present. Got: []
```

Shoeboxam:
This will be fixed in 0.11, which we are now in the process of releasing. 

Until then, `pip install opendp==0.10.0a20240804001` should work properly.

Thanks for reporting this!

K-Meech:
Thanks @Shoeboxam ! Using version `0.10.0a20240804001` works great.

mccalluc:
👍 closing.

Issue #1758, opened 2024-06-29
PyPi OpenDP 0.10.0 on Windows Install Problem


joshua-oss:
`pip install opendp` into a fresh Conda environment succeeds, but attempting to import the library throws exception: "Expected exactly one binary to be present".

**To Reproduce** (device is Intel x64)
1. conda create env --name windp
2. conda activate windp
3. conda install pip
4. pip install opendp
5. python
6. >> from opendp.transformations import make_sum, make_clamp

**Result**

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\joshuaa\AppData\Local\miniconda3\envs\windp\Lib\site-packages\opendp\__init__.py", line 1, in <module>
    from opendp.mod import Transformation, Measurement, OpenDPException, UnknownTypeException
  File "C:\Users\joshuaa\AppData\Local\miniconda3\envs\windp\Lib\site-packages\opendp\mod.py", line 13, in <module>
    from opendp._lib import AnyMeasurement, AnyTransformation, AnyDomain, AnyMetric, AnyMeasure, AnyFunction
  File "C:\Users\joshuaa\AppData\Local\miniconda3\envs\windp\Lib\site-packages\opendp\_lib.py", line 46, in <module>
    lib, lib_path = _load_library()
                    ^^^^^^^^^^^^^^^
  File "C:\Users\joshuaa\AppData\Local\miniconda3\envs\windp\Lib\site-packages\opendp\_lib.py", line 31, in _load_library
    raise Exception(f"Expected exactly one binary to be present. Got: {lib_dir_file_names}")
Exception: Expected exactly one binary to be present. Got: []

mccalluc:
@joshua-oss, thank you for the bug report. I'm realizing we auto-closed this issue when the fix was merged to main, but we neglected to post instructions here for using the nightly build. We are now preparing the release 0.11 release which will include the fix.

Until that is out, `pip install opendp==0.10.0a20240804001` should work properly.

PR #1886, opened 2024-08-05
report dir name if binary not found

mccalluc:
Proposal. Would help debug cases like 
- #1758 
- #1885
 
Hopefully we don't hit this error again, but if we did, we could ask the user to list what was in that directory. It's also possible that someday there will be a problem in the logic to determine the lib_dir, and this would help us debug that.


Shoeboxam (approved):
I'm convinced!

PR #1894, opened 2024-08-05
ci: fix r-docs build

### LINKED ISSUE #1893: ci: always build R API docs from stable

Shoeboxam:
Closes #1893

Since no ref is stated in the checkout when building R docs, it defaults to main. However, the ideal behavior is for the docs site to show stable. 

Separately, the r-docs build began to fail because it was pointed at main, where we don't check in generated code.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -32,6 +32,8 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
+        with:
+          ref: stable
 
       - name: Set up R
         uses: r-lib/actions/setup-r@v2


Shoeboxam:
* **#1894** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1894?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

mccalluc (approved):
Approve, but does the same change need to also be make for the other checkout in this workflow?
```
      - name: Checkout repository
        uses: actions/checkout@v4
        with:
          # Make sure we get all refs needed to build docs for different versions
          fetch-depth: 0
```

Shoeboxam:
Nope, just the R docs one. The other checkout builds docs for all tagged versions, so it's insensitive to the checkout branch. It pulls down refs for all versions (fetch-depth: 0)

PR #1897, opened 2024-08-06
Update regression notebook

GKD-stack:


mccalluc (approved):
This PR is second pass, responding to comments from the first, so it doesn't need a deep content review. I've made some cosmetic changes, and set it to auto-merge.

PR #1836, opened 2024-07-22
Docs: Polars GroupBy

GKD-stack:
closes #1835

PR #1899, opened 2024-08-06
Edits to merged 1837 keys

GKD-stack:


mccalluc:
Notebook still fails, but farther along:
```
count_age_q = (
    context.query().group_by("AGE").agg(
        pl.len().dp.noise().alias("DP Count")
    ).sort("AGE"))
count_age_dp = count_age_q.release().collect()
```
```
OpenDPException:
  MakeMeasurement("The key-set of {:?} is private and cannot be released without filtering. Please pass a filtering threshold into make_private_lazyframe.")
```
Just filed
- #1916

for the untemplated error message.

mccalluc:
@Shoeboxam took on this notebook, renamed it to `group-by.ipynb`, made some tweaks, and it was merged with #1836.

I'm going to look at the diff here, compare it to the current state of `group-by.html`, and file a follow-up issue if there are any bits that should still make it in.

mccalluc:
Extracted a couple points into
- #1925 
 
Closing!

PR #1902, opened 2024-08-06
Old DF deprecation notes

### LINKED ISSUE #1698: Deprecate dataframe methods?
With the polars API released, we can probably deemphasize the older dataframe methods. I'm not sure how far this should go, but some possibilities include:
- [x] #1874
- [ ] Or just adding a note explains that polars is a better approach?
- [x] #1876
- [ ] Spit out deprecation warnings if these functions are used
- [ ] ...?

### LINKED ISSUE #1876: Adding deprecation notes in the docstrings for dataframe methods
Limit the deprecation notes to python by using the code inclusion machinery?

mccalluc:
- Fix #1698 
- Fix #1876

Assuming that marking these as deprecated is the right track, some questions:
-  I was initially thinking that different methods might have different notes, but right now they are all the same... so maybe it would be better to just make it a boolean on the bootstrap?
- After coding this, I did my homework, and it looks like [PEP 702](https://peps.python.org/pep-0702) will introduce `@deprecated` in 3.13, but we could use the [reference implementation](https://typing-extensions.readthedocs.io/en/latest/#typing_extensions.deprecated) now: Probably preferable to generate that instead, if you're ok with another dependency?


mccalluc:
CI fails because of doctest failures:
```
249     >>> # chain with more transformations from the trans module
250     >>> chained = (
Expected nothing
Got:
    make_split_lines is deprecated: Use the polars plugin instead
```
so it's working as expected, but before investing time in a fix, should see if this is the right direction.

mccalluc:
Not strictly a polars-docs issue, but related to how we steer folks to the new API, so asking for review to help set direction.

mccalluc:
- replaced by #1928

Issue #1684, opened 2024-06-17
`from __future__ import annotations` and then `Union` -> `|`


mccalluc:
Just to be consistent, let's move to the pipe syntax everywhere it doesn't cause problems.

> It is supported if you use `from __future__ import annotations`. However you still have to be careful, as this can cause some weird edge cases with Sphinx if you use it in conjunction with TypeVars.

_Originally posted by @Shoeboxam in https://github.com/opendp/opendp/pull/1604#discussion_r1641861373_

mccalluc:
Not just sphinx: Trying to make this change:
```diff
- Chain = Union[tuple[Domain, Metric], Transformation, Measurement, "PartialChain"]
+ Chain = tuple[Domain, Metric] | Transformation | Measurement | "PartialChain"
```
produces this error:
```
TypeError: unsupported operand type(s) for |: 'types.UnionType' and 'str'
```
Sorting this out isn't feeling like a high priority.

PR #1903, opened 2024-08-07
Change `Union` to `|`

mccalluc:
- Towards #1684

docs build and tests pass locally, though that's no guarantee of how it will work on CI.

I would be inclined to close the issue. The remaining ones might be a little trickier, but I'd guess that the tooling will improve over time, so we shouldn't jump through hoops for every last instance now.

### DIFF
--- a/python/src/opendp/_convert.py
+++ b/python/src/opendp/_convert.py
@@ -1,4 +1,5 @@
-from typing import Sequence, Union, cast
+from __future__ import annotations
+from typing import Sequence, cast
 from inspect import signature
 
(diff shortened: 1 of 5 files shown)


mccalluc:
At this point, my preference would just be to say that we stay with `Union` and `Optional`: If they work in a particular bit of new code, great, but it just doesn't seem like this is worth the hassle. Suggest closing #1684, and don't think about it for a year, at least.

Shoeboxam:
Closing by recommendation of @mccalluc

Issue #1736, opened 2024-06-27
Data loaders for queryables should give human-readable error messages


Shoeboxam:
Shared by @shlomihod 

```python
import opendp.prelude as dp
import polars as pl
dp.enable_features("contrib")


input_domain = dp.lazyframe_domain(
        [dp.series_domain("A", dp.option_domain(dp.atom_domain(T=dp.i32))),
        dp.series_domain("B", dp.atom_domain(T=dp.i32))
        ])
input_metric = dp.symmetric_distance()
output_measure = dp.max_divergence(T=float)
d_in = 1
eps1 = 1.0
eps21 = eps22 = 0.5
eps2 = eps21 + eps22

data = pl.LazyFrame({"A": pl.Series([100, 1, 300], dtype=pl.Int32), "B": pl.Series([4, 4, 6], dtype=pl.Int32)})

overall_pipeline = dp.c.make_sequential_composition(input_domain, input_metric, output_measure, d_in,
                                                    d_mids=[eps1, eps2])

overall_pipline_comp = overall_pipeline(data)


specific_pipeline = dp.c.make_sequential_composition(input_domain, input_metric, output_measure, d_in,
                                                                d_mids = [eps21, eps22])
        
specific_pipeline_comp = specific_pipeline(overall_pipline_comp)
```

PR #1906, opened 2024-08-08
Add test for data loader error

mccalluc:
- Towards #1736

I'd be curious to understand what the expected behavior is.

Right now, the error comes from a `ctypes.cast`, where instead of getting a pointer, it gets a dict. Extracting the pointer from that dict...
```diff
 def _wrap_in_slice(ptr, len_: int) -> FfiSlicePtr:
+    if isinstance(ptr, dict):
+        ptr = ptr['value']
     return FfiSlicePtr(FfiSlice(ctypes.cast(ptr, ctypes.c_void_p), len_))
```
Gives us a different error:
```
E       opendp.mod.OpenDPException:
E         FFI("Error when deserializing LazyFrame. This may be because you're using features from Polars that are not currently supported. Semantic(None, "invalid type: integer `12`, expected enum")")
```
but all the other tests are still passing.

### DIFF
--- a/python/test/test_usability.py
+++ b/python/test/test_usability.py
@@ -115,15 +115,23 @@ def test_string_instead_of_tuple_for_margin_key():
     # counts.release().collect()
 
 
-def test_data_loader_error_is_human_readable():
+@pytest.mark.parametrize(
+    "domain", [dp.lazyframe_domain([]), dp.series_domain("A", dp.atom_domain(T=bool))])
+def test_polars_data_loader_error_is_human_readable(domain):
     pytest.importorskip("polars")
-    for domain in [
-        dp.lazyframe_domain([]),
-        dp.series_domain("A", dp.atom_domain(T=bool)),
-        dp.expr_domain(dp.lazyframe_domain([])),
-    ]:
-        overall_pipeline = dp.c.make_sequential_composition(
-            domain, dp.symmetric_distance(), dp.max_divergence(), d_in=1,
-            d_mids=[1.])
-        with pytest.raises(ValueError, match="expected Polars *"):
-            overall_pipeline("I'm not the right type!")
+    overall_pipeline = dp.c.make_sequential_composition(
+        domain, dp.symmetric_distance(), dp.max_divergence(), d_in=1,
+        d_mids=[1.])
+    with pytest.raises(ValueError, match="expected Polars *"):
+        overall_pipeline("I'm not the right type!")
+
+
+def test_polars_expr_loader_error_is_human_readable():
+    pl = pytest.importorskip("polars")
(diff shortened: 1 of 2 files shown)


mccalluc:
Tests failing for Python 3.9:
```
     def _lazyframe_to_slice(val) -> FfiSlicePtr:
>       state = val.__getstate__()
E       AttributeError: 'Queryable' object has no attribute '__getstate__'
```

[doc say](https://docs.python.org/3/library/pickle.html#object.__getstate__):
> Changed in version 3.11: Added the default implementation of the `__getstate__()` method in the [object](https://docs.python.org/3/library/functions.html#object) class.

At some point here we're trying to pass the wrong kind of object. 3.12 fails later in a way we expect; 3.9 fails earlier because it can not pickle / unpickle generic objects, I think? How we want to handle this case depends on how we want to handle the big picture.

Shoeboxam:
* **#1906** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1906?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Shoeboxam:
@mccalluc I pushed a fix to this, would you mind "reviewing"?

python/test/test_usability.py
```diff
@@ -113,3 +113,17 @@ def test_string_instead_of_tuple_for_margin_key():
...
     # counts.release().collect()
+
+
+def test_data_loader_error_is_human_readable():
+    pytest.importorskip("polars")
+    for domain in [
+        dp.lazyframe_domain([]),
+        dp.series_domain("A", dp.atom_domain(T=bool)),
+        dp.expr_domain(dp.lazyframe_domain([])),
+    ]:
```
mccalluc, line 124: Move this list to `@pytest.mark.parametrize` around the function?

python/test/test_usability.py
```diff
@@ -113,3 +113,17 @@ def test_string_instead_of_tuple_for_margin_key():
     )
     counts.summarize()
     # counts.release().collect()
+
+
+def test_data_loader_error_is_human_readable():
+    pytest.importorskip("polars")
```
mccalluc, line 119: Drop this line? I think you want to check there is nice behavior, whether or not polars is installed.

mccalluc (commented):
Assuming the tests pass with the latest change, a couple thoughts on the test, but looks good overall.

python/test/test_usability.py
Shoeboxam: The type check attempts to import Polars, which fails. I added this to get CI back on track.

mccalluc (commented):
Looks good to me! Since I'm the original author, I can't approve, but you can.

Shoeboxam (approved):

PR #1907, opened 2024-08-08
Remove unused TODO files in docs

mccalluc:
We've changed direction from when these files were created; The current plan is to save the discussion of polars for the tabular data section.

- see #1780


Shoeboxam (approved):
Works for me. I assume you've moved any material worth saving into the new structure.

Issue #1910, opened 2024-08-08
After Sphinx upgrade, `make html` has a html_sidebars error


mccalluc:
Working on this issue:
- #1795

I just upgraded all the docs dependencies, but it looks like there is a change in how it handles `html_sidebars`: My guess is that formerly, if it saw a string, it would change it to a list containing a single string, but now trying to run `make html` produces this error:
```
Configuration error:
Values in 'html_sidebars' must be a list of strings. At least one pattern has a string value: '**'. Change to `html_sidebars = {'**': ['sidebar-nav-bs.html']}`.
make: *** [html] Error 2
```

Easy enough if we are in `conf.py`, but the Makefile is trying to override it like so:
```
... -D 'html_sidebars.**'=sidebar-nav-bs.html ...
```
There are [docs](https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-D) for this parameter, and while it can handle lists _or_ dicts, not clear if it handles nested structures.

I think for now I'll try pinning sphinx to keep the current behavior, but try to get the latest version of nbconvert.

One solution might be to drop the version section:
- #1904 goes part way

... although that PR only drops the releases block, not the branches.

PR #1911, opened 2024-08-08
Resolve pandoc version warning

### LINKED ISSUE #1795: "You are using an unsupported version of pandoc (2.9.2.1)"
Not causing any obvious problems, but it probably is tech debt to address. In the [CI log](https://github.com/opendp/opendp/actions/runs/9944344652/job/27470211534):
```
/home/runner/.local/lib/python3.10/site-packages/nbsphinx/__init__.py:1058: RuntimeWarning: You are using an unsupported version of pandoc (2.9.2.1).
Your version must be at least (2.14.2) but less than (4.0.0).
Refer to https://pandoc.org/installing.html.
```

mccalluc:
- Fix #1795
- Rebuild `requirements.txt` to get the latest version of `nbconvert`, but pin `sphinx` to avoid #1910
- Build docs locally, and they look ok.


Shoeboxam (approved):
It seems the only material change here is pinning sphinx. The follow up issue covers why.

mccalluc:
> It seems the only material change here is pinning sphinx

I'd say the real change is upgrading nbconvert. Pinning sphinx just maintains the status quo there, but we need to be explicit to keep the status quo.

Issue #1908, opened 2024-08-08
Polars privacy map fail but measurement goes through


damienbfs:
**Bug Description**

We construct a polars measurement on a dataset with two columns: "region" and "income". The polars plan computes the average income by region and adds noise with an arbitrary scale of one.

The lazyframe domain we use has a margin for the "region", but only includes `max_partition_length` and omits `max_num_partitions`.

BUG: The measurement can be constructed and executed on data but the privacy map fails with `OpenDPException: 
  FailedFunction("max_num_partitions must be known when the metric is not sensitive to ordering (SymmetricDistance)")`

Since the privacy map fails, we expected the measurement to also fail. Is this a bug or is the current behaviour to be expected?

**To Reproduce**

````python
import opendp.prelude as dp
import polars as pl

dp.enable_features("contrib")

# Minimal dataframe
NB_REGIONS = 3
NB_ROWS = 6

df = pl.DataFrame({"region": [1, 1, 2, 2, 3, 3], "income": [1000.0, 1000.0, 2000.0, 2000.0, 3000.0, 3000.0]}, schema={"region": pl.Int64, "income": pl.Float64})
lf = df.lazy()

# Minimal domain
lf_domain = dp.lazyframe_domain([
    dp.series_domain("region", dp.atom_domain(T=dp.i64)),
    dp.series_domain("income", dp.atom_domain(T=dp.f64))
])
lf_domain = dp.with_margin(lf_domain, by=[], public_info="lengths", max_partition_length=NB_ROWS)

# Group-by query: average income per region, add noïse with scale of 1.0
income_lower_bound, income_upper_bound = 1_000, 100_000
plan = lf.group_by("region").agg([
    pl.col("income").dp.mean(bounds=(income_lower_bound, income_upper_bound), scale=1.0)
]).sort("income")

# Bad domain: add a margin that does not include max_num_partitions
bad_domain = dp.with_margin(lf_domain, by=["region"], public_info="lengths", max_partition_length=NB_ROWS)

# Construct and collect a measurement with bad_domain works
bad_meas = dp.make_private_lazyframe(bad_domain, dp.symmetric_distance(), dp.max_divergence(T=float), plan)
bad_res = bad_meas(lf).collect()

# Privacy map fails
bad_meas.map(1)
```

mccalluc:
Used your sample code to make a test that reproduces the behavior. 
- #1912

Draft PR until we understand the desired behavior.

PR #1912, opened 2024-08-09
reproducer for max_num_partitions

mccalluc:
- towards #1908

Test that reproduces the behavior: This might be of some use as it, but we should also clarify the desired behavior.


Shoeboxam:
Closing in favor of #2059

PR #1848, opened 2024-07-23
docs: Data Preparation

GKD-stack:
closes #1846

PR #1913, opened 2024-08-09
Add docs: data prep notebook

### LINKED ISSUE #1846: Notebook/Page: Data Preparation (use with_columns/select/filter)

mccalluc:
- Fix #1846

Copied notebook from
- #1848, which had gotten untidy

Vikrant wasn't sure about using "limitations" to describe the functionality. Filing as draft and will re-review from the top.


mccalluc (commented):
I think this much will pass CI, but is the content right?
- `with_columns` is used for post-processing here, so I mention that the DP guarantee will still hold under post-processing... but maybe an application of `with_columns` earlier in the process would be interesting?
- I don't think the `filter` chart actually demonstrates the problem of increased noise in small buckets... Maybe make an example with smaller partitions so it does become an issue?

Shoeboxam:
I started refactoring this notebook. The big change here is that we want to be showing how these methods can be applied as pre-processing, for data prep, not as post-processing. Showing how you can do post-processing is not that interesting for us to document, as Polars documentation already covers this. I was hoping to get documentation related to with_columns and filter (and ultimately select as a transformation) before applying DP aggregations. In that setting, OpenDP applies additional constraints that you wouldn't typically have in Polars (like how with_columns/select must preserve row alignment and how filtering discards domain descriptors).

mccalluc:
@Shoeboxam 
> I started refactoring this notebook

Do you have some work in progress that should either be committed here, or used in place of this notebook?

Shoeboxam:
Closed in favor of https://github.com/opendp/opendp/commit/dda74a942570057a9a6c374a8dfa5e36cd01aff4, specifically https://docs.opendp.org/en/stable/getting-started/tabular-data/preparing-microdata.html

PR #1898, opened 2024-08-06
1720 noise comparison w/ edits

GKD-stack:
- Fix #1720 
- Fix #1891

PR #1914, opened 2024-08-10
Update noise comparison notebook

### LINKED ISSUE #1891: Follow-up on "Noise comparison" comments
After merge, comments came in on
- #1735

Review and integrate feedback.

mccalluc:
- Fix #1891

Copy just the notebook from
- #1898

Unwanted merge cruft was introduced in that PR, but the notebook looks good.


mccalluc:
Started to make some copy edits, and then realized that's not where I should invest my time.

Shoeboxam (changes requested):
It took me a few scans to identify the structure in this notebook. I don't think this notebook should be un-orphaned in 0.11. The structure is currently:

* Laplace vs. Gaussian (theory intro)
    * Key differences
    * When Gaussian mechanism is better than Laplace
    * When to use
    * Summary
* Specifying Laplace or Gaussian with the Context API (code demo)
    * Laplace
    * Gaussian
    * Comparison

There are some issues with the current code demonstration:
* differences between the mechanism are not shown
* exact data is used, which we're trying to avoid
* the code could be significantly cut down/simplified
    * de-duplicate code by abstracting query logic
    * remove unnecessary margin descriptors
    * the formatting is very difficult to read
    * visualizing one release does not give you a sense of how the variance changes, which is the big payoff

I think it should be restructured and consolidated. Here's a potential structure:

Rename overall notebook to something like "Comparison of Output Perturbation Mechanisms"
* Distribution
    * theory: Gaussian preserves nice statistical normality property
    * practice: plot each distribution
* Privacy Measure
    * theory: Laplace doesn't need delta but Gaussian composes better
    * practice: set up two contexts, one for each measure. compare accuracy when varying the number of compositions
* Input Metric
    * theory: L2 can be smaller
    * practice: compare accuracy for a count query over a margin with low per-partition contribution

I don't think we should tell people which mechanism to use, we should just clearly explain the differences.

Down the line we'll follow up with a "Comparison of Private Selection Mechanisms" docs page comparing RNM exponential/permute n' flip to RNM Gumbel, which has some similar tradeoffs. Potentially more for online private selection and private candidate selection when we finish hashing out that API structure.

Shoeboxam:
While this is Polars-centric, we now have: https://docs.opendp.org/en/stable/api/user-guide/polars/gaussian-noise.html from #2200

PR #1915, opened 2024-08-12
Change `master` to `main` in two notebooks

mccalluc:
I changed the default branch for consistency while working on https://github.com/opendp/dp-test-datasets/pull/1 ... but I should have kept the old one around and made these fixes first.


Shoeboxam (approved):
Thanks!

PR #1834, opened 2024-07-22
Docs: Computing Fundamental Statistics Notebook

GKD-stack:
closes #1833

PR #1920, opened 2024-08-14
Polars: select margin defaults

### LINKED ISSUE #1919: Polars: select already knows max_num_partitions

Shoeboxam:
Closes #1919

Adds default attributes to margins in select.
This prevents users from needing to specify silly obvious margin attributes when making select queries.

### DIFF
--- a/rust/src/measurements/make_private_lazyframe/select/mod.rs
+++ b/rust/src/measurements/make_private_lazyframe/select/mod.rs
@@ -62,10 +62,6 @@ where
         .margins
         .entry(BTreeSet::new())
         .or_insert_with(Default::default);
-    margin.public_info.get_or_insert(MarginPub::Keys);
-
-    // now that the domain is set up, we can clone it for use in the closure
-    let margin = margin.clone();
 
     if margin.max_partition_contributions.is_some() {
         return fallible!(
@@ -82,6 +78,11 @@ where
             "There is only one partition in select, so both max_influenced_partitions and max_num_partitions must either be unset or one"
         );
     }
+
+    // use the knowledge that we're only working with a single partition
+    margin.public_info.get_or_insert(MarginPub::Keys);
+    margin.max_influenced_partitions = Some(1);
+    margin.max_num_partitions = Some(1);
 
     let t_group_by = Transformation::new(
         expr_domain.clone(),


Shoeboxam:
* **#1920** <a href="https://app.graphite.dev/github/pr/opendp/opendp/1920?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

mccalluc (commented):
As with 
- #1922 
 
I'm guessing the notebook examples are motivating this? And that we'd want to update the examples to use this? I feel like there either needs to be a follow-up issue to use this in the python examples, or the examples need to be updated in this PR. Probably the former?

mccalluc:
> I've got the stack all lined up now. Notebook is motivating this. I don't see any usages in main that would be improved by this, but it simplifies the notebook in the Docs PR at the tip of the stack.

That PR is
- #1834
I think? Notebook CI failed there... but it looks like it was just a [network timeout to download the data](https://github.com/opendp/opendp/actions/runs/10415153163/job/28845436831?pr=1834#step:10:41)? Restarting that job manually...

mccalluc (approved):
Approved to get things rolling, but I'm uncomfortable here. I think right now we want to prioritize work for the meeting, but in the future I'd really like to have tests and documentation updates go along with the code changes. Some motivation:

- Ideally, we can articulate the desired behavior before implementing it. Otherwise I know that I can end up just writing tests that capture the new behavior instead of thinking about what the behavior _should_ be
- For the reviewer, it makes it easier to understand what change the PR is really making, and might make it easier to think about edge cases that were overlooked, for example.
- Similarly, for anyone reviewing the git history, a change in the Rust code will be easier to understand if there is accompanying documentation or test updates in the same commit.
- Process-wise, I also worry about things getting lost. When I'm looking at the notebook at the tip of the stack I know that I'm not going to remember to check if it's exercising this code, if that's even possible to differentiate from the other changes.
- ... Or it's possible that the part of the notebook that exercises this will be edited out by the time we get to review that.
- ... Or that some intervening PR reverts this or somehow interferes with the desired behavior, and we don't notice.
- Thinking about the project goals at a high level: We've made a choice to invest time in proofs, because we want people to be able to trust our code. Tests are much weaker -- they only check one case -- but if we're a kind of project that wants to inspire confidence in out implementations, then I think it makes sense to also invest the time in traditional unit testing of code changes.

Sorry if this is a bit long: For this week
- Tell me on the PR if you want a less picky review!
- I'll generally approve on a re-review
- Another approach might be to merge down from the top of the stack in some cases, to get the code that exercises an update along with the update in one squash commit on main.

Shoeboxam:
I realize that at this point I can still run the notebook without this fix. I feel like I had a code example that required this at some point, but can no longer replicate it. Looking through the codebase holistically, I can't anticipate a scenario where it would be necessary either.

Closing this, but still keeping in mind your comments @mccalluc in mind for future work.

PR #1927, opened 2024-08-15
Add documentation for python vs r cargo build

### LINKED ISSUE #1683: R development process needs docs
My understanding is that developers working in both R and Python are going to need separate cargo builds for the two languages, with different feature sets enabled, and I don't think our documentation currently reflects that.

This is related to but distinct from
- #1675 

ie: I could understand everything, without there being docs that would help anyone else; or there could be great docs that I'm able to follow, but I still might not understand how it works.

### LINKED ISSUE #1918: Update development-environment.rst to explain R vs python builds
We need different builds for Python and R development. 

For Python:
```
cargo build --all-features
```

For R:
```
carbo build --features untrusted,bindings
```

`--all-features` includes Polars, but it causes R to error with: `symbol not found in flat namespace '_PyBaseObject_Type'`

This does mean that you can't be developing in Python and R at the same time: Instead, rebuilds will be needed if you swap contexts.

- #1879 also updates this file: maybe get that merged first to avoid conflicts?

mccalluc:
- Fix #1918
- Fix #1683

### DIFF
--- a/docs/source/contributing/development-environment.rst
+++ b/docs/source/contributing/development-environment.rst
@@ -56,12 +56,11 @@ Now run ``cargo build`` in the ``rust`` subdirectory of the repo:
 This will compile a debug build of the OpenDP shared library, placing it in the directory ``opendp/rust/target/debug``. 
 (The specific name of the library file will vary depending on your platform.)
 
-Substitute ``cargo build`` with ``cargo test`` to test, or ``cargo check`` to regenerate Python and R bindings and check syntax.
+Substitute ``cargo build`` with ``cargo test`` to test, or ``cargo check`` to check syntax.
 
-In the above commands, the features ``untrusted`` and ``bindings`` are enabled.
-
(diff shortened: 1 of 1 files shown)


docs/source/contributing/development-environment.rst
```diff
@@ -200,13 +214,20 @@ The source code and developer documentation is
 R Setup
 -------
 
+First, build a debug binary that works with R. (Note that the resulting binary will not work with Python.)
```
Shoeboxam, line 217: It will mostly work with python, just not Polars. I guess user doesn't really need to know, though.

Shoeboxam (approved):
Long overdue, thanks for clarifying this in the docs.

docs/source/contributing/development-environment.rst
mccalluc: I guess we could say "not suitable for Python development" to be more strictly accurate? We can revisit the wording later. Thanks for the clarification!

PR #1933, opened 2024-08-15
Fix `assert_feature` check

### LINKED ISSUE #1930: Python `assert_feature` ineffective if Python is optimized
`assert_features` uses `assert`:
```
def assert_features(*features: str) -> None:
    for feature in features:
        assert feature in GLOBAL_FEATURES, f"Attempted to use function that requires {feature}, but {feature} is not enabled. See https://github.com/opendp/opendp/discussions/304, then call enable_features(\"{feature}\")"
```

[`assert` is not guaranteed to run](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement):
>  The current code generator emits no code for an assert statement when optimization is requested at compile time

Is someone is running optimized, they can use any method without needing to opt in. Probably not what we want.

mccalluc:
- Fix #1930

Optimized python now fails where it should:
```
$ python -O
Python 3.11.7 (main, Dec  4 2023, 18:10:11) [Clang 15.0.0 (clang-1500.1.0.2.5)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> space = (dp.atom_domain(T=float), dp.absolute_distance(T=float))
>>> laplace_mechanism = space >> dp.m.then_laplace(scale=1.)
...
opendp.mod.OpenDPException:
  Attempted to use function that requires "contrib", but not enabled. See https://github.com/opendp/opendp/discussions/304, then call enable_features("contrib")
```

### DIFF
--- a/python/src/opendp/mod.py
+++ b/python/src/opendp/mod.py
@@ -735,8 +735,10 @@ def disable_features(*features: str) -> None:
 
 
 def assert_features(*features: str) -> None:
-    for feature in features:
-        assert feature in GLOBAL_FEATURES, f"Attempted to use function that requires {feature}, but {feature} is not enabled. See https://github.com/opendp/opendp/discussions/304, then call enable_features(\"{feature}\")"
+    missing_features = [f for f in features if f not in GLOBAL_FEATURES]
+    if missing_features:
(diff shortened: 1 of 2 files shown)


Shoeboxam (approved):
Good catch!

PR #1938, opened 2024-08-15
Add check for absolute links to docs.opendp.org

### LINKED ISSUE #1046: Relative or absolute URLs in notebook markdown?
Right now we have a mixture: In `docs/` there are:
- 34 matches for `](https://docs.opendp.org` 
- 22 matches for `](..`

If we rearrange the docs, we might want to:
- Change everything to relative urls, so broken links can be caught early (once #1043 is in place)
- Then, when we think the structure is stable
  - change everything to absolute urls, so links will still work in binder or colab
  - ... or find a way to set baseurl in those environments, and keep the relative urls in the docs.

mccalluc:
- Fix #1046
- And add a CI check to make sure we don't introduce any new ones.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -229,8 +229,9 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Look for URLs that should be relative
-        # API index page excluded: It points to R docs which are copied into place after build.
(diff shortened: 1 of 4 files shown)


Shoeboxam (approved):
Thanks for the fixes!

PR #1940, opened 2024-08-16
Fix site search to exclude transclusions

### LINKED ISSUE #1934: Included code files show up in site search
- Search for "atom_domain": https://docs.opendp.org/en/stable/search.html?q=atom_domain
- second result is [`<no title>`](https://docs.opendp.org/en/stable/getting-started/code/typical-workflow-framework.html)
- I think we just need a `:noindex:` at the top of the file?

mccalluc:
- Fix #1934

An alternative approach would be to add titles on these pages and fix them so the user could follow the link and get a page that didn't look broken.

Another option might be to name them something other than `rst`, though I think we went through this before, and this works well with our tooling.


mccalluc:
smoke-test is not running: Maybe I've somehow messed up the formatting so it doesn't parse, but github isn't giving good feedback? I'll try starting this over in a new PR.

Shoeboxam (approved):
How would you feel about adding a general exclude for `**/code/` in `exclude_patterns`? I don't see why we'd want these code backing pages to be generated or accessible at all. The exclude might let us drop both :orphan: and :nosearch: tags (and maybe avoid other strange behaviors we don't know about).

This is ok too, but will require ongoing maintenance of tags and will continue resulting in dummy html pages getting generated.

mccalluc:
> How would you feel about adding a general exclude for `**/code/` in exclude_patterns?

Sounds good. I'm going to change this to draft for now: May end up implementing that in a new PR.

mccalluc:
- Replaced by #2018. Closing.

PR #1929, opened 2024-08-15
deprecation machinery

Shoeboxam:
Closes #1928

Might consider renaming docstring.rs to attributes.rs.

I am no longer tracking/maintaining this PR. In Rust parlance, ownership goes to @mccalluc: feel free to mutate or drop/replace the PR, as you see fit.

Issue #859, opened 2023-07-10
Remove current frame constructors


Shoeboxam:
With the advent of Polars, existing dataframe constructors are inferior

mccalluc:
Change of direction: For the near future, we just want to deprecate these:
- #1928

Maybe they'd get removed for a 1.0 release, but at the very least we'd need to implement Polars for R, and that's not on the immediate roadmap. Feel free to reopen if we will get to it in the near future.

PR #1961, opened 2024-08-19
Add deprecation notes on old dataframe functions

### LINKED ISSUE #1928: Deprecation machinery

mccalluc:
- Replaces #1929 (want to make myself the author, so it's clear that review is needed from Mike.)
- Fixes #1928 ... which replaces a number of older issues:
  - #1876
  - #1698
  - #859

### DIFF
--- a/python/setup.cfg
+++ b/python/setup.cfg
@@ -19,6 +19,11 @@ long_description_content_type = text/markdown
 keywords = 
 	differential privacy
 
+[project]
+dependencies = [
+	'deprecated'
+]
+
 [options]
 zip_safe = false
 python_requires = >=3.9
--- a/python/test/__init__.py
+++ b/python/test/__init__.py
@@ -1,6 +1,5 @@
 import opendp.prelude as dp
 
-
 # We don't have tests that require a feature *not* to be enabled,
 # so more stable just to enable everything at the start.
 dp.enable_features('floating-point', 'contrib', 'honest-but-curious')
--- a/python/test/__init__.py
+++ b/python/test/__init__.py
@@ -3,4 +3,4 @@ import opendp.prelude as dp
 
 # We don't have tests that require a feature *not* to be enabled,
 # so more stable just to enable everything at the start.
-dp.enable_features('floating-point', 'contrib', 'honest-but-curious', 'floating-point')
+dp.enable_features('floating-point', 'contrib', 'honest-but-curious')
--- a/rust/opendp_tooling/src/bootstrap/docstring.rs
+++ b/rust/opendp_tooling/src/bootstrap/docstring.rs
@@ -1,6 +1,6 @@
 use std::{collections::HashMap, env, path::PathBuf};
 
-use darling::{Error, Result};
+use darling::{Error, FromMeta, Result};
 use proc_macro2::{Literal, Punct, Spacing, TokenStream, TokenTree};
 use quote::format_ident;
 use syn::{
@@ -18,6 +18,13 @@ pub struct BootstrapDocstring {
     pub arguments: HashMap<String, String>,
     pub generics: HashMap<String, String>,
     pub returns: Option<String>,
+    pub deprecated: Option<String>,
+}
+
+#[derive(Debug, FromMeta, Clone)]
+pub struct DeprecationArguments {
+    pub since: Option<String>,
+    pub note: Option<String>,
 }
 
 impl BootstrapDocstring {
@@ -26,6 +33,23 @@ impl BootstrapDocstring {
         output: &ReturnType,
         path: Option<(&str, &str)>,
     ) -> Result<BootstrapDocstring> {
+        // look for this attr:
+        // #[deprecated(note="please use `new_method` instead")]
+        let deprecated = attrs
+            .iter()
+            .find(|attr| {
+                attr.path.get_ident().map(ToString::to_string).as_deref() == Some("deprecated")
+            })
+            .map(|attr| {
+                // TODO: more nicely format this message
+                Result::Ok(
+                    DeprecationArguments::from_meta(&attr.parse_meta()?)?
+                        .note
+                        .unwrap_or_default(),
+                )
+            })
+            .transpose()?;
+
         let mut doc_sections = parse_docstring_sections(attrs)?;
 
         if let Some(sup_elements) = parse_sig_output(output)? {
@@ -67,6 +91,7 @@ impl BootstrapDocstring {
                 .map(parse_docstring_args)
                 .unwrap_or_else(HashMap::new),
             returns: doc_sections.remove("Returns"),
+            deprecated,
         })
     }
 }
--- a/rust/opendp_tooling/src/codegen/python.rs
+++ b/rust/opendp_tooling/src/codegen/python.rs
@@ -64,7 +64,9 @@ fn generate_module(
     let constructor_mods = ["combinators", "measurements", "transformations"];
 
     let extra_imports = if constructor_mods.contains(&module_name) {
-        r#"from opendp.core import *
+        r#"from deprecated.sphinx import deprecated
+
+from opendp.core import *
 from opendp.domains import *
 from opendp.metrics import *
 from opendp.measures import *"#
@@ -228,10 +230,11 @@ def {then_name}(
         String::new()
     };
 
+    // TODO: If match for "polars" substring, show python specific message.
     let deprecated = func
         .deprecated
         .as_ref()
-        .map(|msg| format!("@deprecated(\"{msg}\")\n"))
+        .map(|_| format!("@deprecated(version='0.11.1', reason='Use :mod:`opendp.extras.polars` instead')\n"))
         .unwrap_or_default();
 
     format!(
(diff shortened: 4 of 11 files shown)


python/setup.cfg
```diff
@@ -19,6 +19,11 @@ long_description_content_type = text/markdown
 keywords = 
 	differential privacy
 
+[project]
+dependencies = [
+	'deprecated'
```
mccalluc, line 24: This seems to be the first runtime python dependency that we've added that isn't an "extra". I think this is fine... but it wouldn't be impossible to factor out just the code we need from the library.

python/test/__init__.py
```diff
@@ -1,6 +1,5 @@
 import opendp.prelude as dp
 
-
 # We don't have tests that require a feature *not* to be enabled,
 # so more stable just to enable everything at the start.
-dp.enable_features('floating-point', 'contrib', 'honest-but-curious', 'floating-point')
\ No newline at end of file
+dp.enable_features('floating-point', 'contrib', 'honest-but-curious')
```
mccalluc, line 5: Unrelated cleanup while in the neighborhood.

mccalluc:
@Shoeboxam , I think this is looking ok. Questions to consider:
- Is the `panic` ok? Or do we want to avoid that?
- Would it be better to pull the deprecation message into a constant? Or maybe more indirection just makes it harder to read?
- Could you remind me what was wrong that `TODO: more nicely format this message` reminds us to fix. Is it just the text of the message?

rust/opendp_tooling/src/codegen/python.rs
```diff
@@ -228,9 +230,23 @@ def {then_name}(
         String::new()
     };
 
+    let deprecated = func
+        .deprecated
+        .as_ref()
+        .map(|msg| {
+            if msg.contains("Polars") {
+                format!(
+                    "@deprecated(version='0.11.1', reason='Use `opendp.extras.polars` instead')\n"
```
Shoeboxam, line 239: You could change the deprecated attribute in Function to:

```rust
struct Deprecated {
	version: String,
	note: String,
}
```

Then here in this code, you'd craft the message accordingly:

```rust
format!("@deprecated(version={}, reason={})\n" deprecated.version, deprecated.note)
```

This way you don't need to update the code here when you add another deprecation.

rust/opendp_tooling/src/lib.rs
```diff
@@ -27,6 +27,8 @@ pub struct Function {
     pub supports_partial: bool,
     // whether to generate FFI
     pub has_ffi: bool,
+    // deprecation warning if applicable
+    pub deprecated: Option<String>,
```
Shoeboxam, line 31: This is where the `Deprecated` struct I suggested above would go.

rust/src/transformations/dataframe/apply/mod.rs
```diff
@@ -59,6 +60,7 @@ where
     )
 }
 
+#[deprecated(note = "Use Polars instead")]
```
Shoeboxam, line 63: Add `version="0.11.1"` here, so that our bootstrap macro can read it later.

python/setup.cfg
```diff
@@ -22,6 +22,8 @@ keywords =
 [options]
 zip_safe = false
 python_requires = >=3.9
+install_requires =
+    deprecated
```
Shoeboxam, line 26: I was hesitant to take on an external dep, but after reading through it, the Sphinx integration is really useful.

rust/opendp_tooling/src/bootstrap/docstring.rs
```diff
@@ -26,6 +33,23 @@ impl BootstrapDocstring {
...
+            .find(|attr| {
+                attr.path.get_ident().map(ToString::to_string).as_deref() == Some("deprecated")
+            })
+            .map(|attr| {
+                // TODO: more nicely format this message
+                Result::Ok(
+                    DeprecationArguments::from_meta(&attr.parse_meta()?)?
+                        .note
+                        .unwrap_or_default(),
+                )
```
Shoeboxam, line 49: If you're building a deprecated struct here, then something like this could work:
```suggestion
                let meta = DeprecationArguments::from_meta(&attr.parse_meta()?)?;
                Result::Ok(Deprecated {
                    version = meta.version.unwrap_or_default(),
                    note = meta.note.unwrap_or_default()
                })
```

Shoeboxam (changes requested):
I left a few suggestions on how the data should flow through.

mccalluc:
Accepted most of the suggestions.
- Ended up calling the field and the struct `deprecation` and `Deprecation` respectively: In context, a noun felt more idiomatic.
- Python and Rust have different names for the parameters: `version` vs `since`, which makes the code a little more confusing.

rust/opendp_tooling/src/lib.rs
```diff
@@ -7,6 +7,12 @@ pub mod proven;
 // logic for generating bindings from the core structures below
 pub mod codegen;
 
+#[derive(Debug)]
+pub struct Deprecation {
+    version: String,
```
Shoeboxam, line 12: Minor nit, this could be "since" as well, to match, but not blocking.

rust/opendp_tooling/src/codegen/python.rs
```diff
@@ -228,9 +230,20 @@ def {then_name}(
         String::new()
     };
 
+    let deprecated_decorator = func
+        .deprecation
+        .as_ref()
+        .map(|deprecation| {
+            format!(
+                "@deprecated(version=\"{}\", reason=\"{}\")\n",
+                deprecation.version, deprecation.note
```
Shoeboxam, line 239: If "since" is not set, then the code will generate `@deprecated(version="", reason="xyz")`. I think it should either error or not include `version=""`. Similarly for reason.

Shoeboxam (changes requested):
Sorry to nit so much! It's looking close though.

rust/opendp_tooling/src/lib.rs
mccalluc: I don't have a preference either way: Will change.

rust/opendp_tooling/src/codegen/python.rs
mccalluc: Good catch!

mccalluc: hmm: I'm slightly confused. I've gone as far as adding a `panic!` above this line, and I don't see the error come through when I `cargo check`, although there are a lot of warning about unused variables.

If this is something we want to validate, maybe we should try to catch it sooner? Like having a `Deprecation::new()` that checks values when the struct is created?

mccalluc:
I also realize that we should add notes in the R code generation as well. Will do that...

rust/opendp_tooling/src/codegen/python.rs
Shoeboxam: It would be better to throw a darling error back in `bootstrap/docstring.rs` when you first parse the deprecated attribute.
https://github.com/opendp/opendp/pull/1961/files#diff-54f929d780d7b93aface00e19d0f0016d7bdc78b0edf0ef3a4c04b2face20ca8R49-R50

Might look like this:
```rust
                    since: meta.since.ok_or_else(|| Error::custom("`since` must be specified").with_span(&attr))?,
                    note: meta.note.ok_or_else(|| Error::custom("`note` must be specified").with_span(&attr))?,
```

Shoeboxam: If it compiles with `with_span`, the error will point at the attribute.

mccalluc: I've made that change, but neither `cargo build --all-features` nor `cargo check --all-features` report an error with `since` deleted in one deprecation. Would I look somewhere else to see this error reported?

mccalluc:
Until I can confirm that missing parameters do produce errors, moving back to draft.

Current unit testing takes the Deprecation struct as input, so it can't check validation that comes earlier.

mccalluc:
> Until I can confirm that missing parameters do produce errors

It works.
```
error: `since` must be specified
  --> src/transformations/dataframe/apply/mod.rs:63:1
   |
63 | #[deprecated(note = "Use Polars instead")]
   | ^

error: could not compile `opendp` (lib) due to 1 previous error
```

My guess is that earlier I was looking at a function that was not bootstrapped, so while it is recognized by Rust as deprecated, our extra check never ran.

rust/src/transformations/dataframe/apply/mod.rs
```diff
@@ -14,6 +14,7 @@ use super::{DataFrame, DataFrameDomain};
 #[cfg(feature = "ffi")]
 mod ffi;
 
+#[deprecated(note = "Use Polars instead", since = "0.11.1")]
```
Shoeboxam, line 17: Can these be changed to since 0.12? I think we shouldn't retrospectively deprecate.

Shoeboxam (changes requested):
Looks good! Minor ask but the core is great. I think we move forward with deprecating R dataframe functionality.

rust/src/transformations/dataframe/apply/mod.rs
mccalluc: done! thanks.

Shoeboxam (approved):
Thanks for pushing this through!

PR #1968, opened 2024-08-21
Add note about required feature on each method doc

### LINKED ISSUE #1662: Flag (and explain?) `honest-but-curious` (and `contrib`?) in API docs

mccalluc:
- fix #1662
- Replaces #1664... Unless having an explanation each place we require `honest-but-curious` would be particularly useful?
- #1987 would let us better test these changes, but doesn't need to block.

I don't want people to just blindly copy and paste code, but if there are limitations on the regime under which the function works, we should mention that.

We also want to nudge ourselves towards getting the proofs filled in so the caveats go away.

These could go at the end, but I think the prominence is not a bad thing. Initially I was thinking about some kind of badge, but that's more of a hassle, and not any more useful.

Would we like to rename this from `flag_check` to `feature_check`? A little more consistent, but that introduces more changes so I didn't want to jump into it.


mccalluc:
realize that inside the function we want to assert features, but outside we want to enable features. Another commit!

mccalluc:
Last commit changed `assert_features` -> `dp.enable_features` in the source code, rather than the docstring. Reverting, and making the change in the right place.

rust/opendp_tooling/src/codegen/r/r.rs
```diff
@@ -229,7 +230,7 @@ fn generate_doc_block(
     };
 
     format!(
-        r#"{description}
+        r#"{flag_doc}{description}
```
Shoeboxam, line 233: I'd rather not make the requirements the first line in the doc. In particular, in R, the first line is the only thing that shows in the reference index: https://docs.opendp.org/en/stable/api/r/reference/index.html (a lot of these could be better anyways, but that's separate).

What if you instead add the set of required features in `BootstrapDocstring::from_attrs`? You might need to pass the list of features into that function. Then you wouldn't need lang-specific codegen and could choose where it gets placed in the resulting docstring.

Shoeboxam (changes requested):
Thanks for doing this, but I have some doubts about the details.

rust/opendp_tooling/src/codegen/r/r.rs
mccalluc: > you wouldn't need lang-specific codegen

Though both the Python and R messages are `enable_features("some_feature")` right now, this could be different between target languages. (Even with Python, initially I had `dp.enable_features("some_feature")`.) But maybe I'm misunderstanding your concern?

Moving this off the first line does make sense: Let me see about the end of the description.

mccalluc:
@Shoeboxam 
- How would you feel about adding https://docs.rs/pretty_assertions as a dev dependency? The changes can be hard to spot, and so I end up copy and pasting and then looking at the git diff, which is not the best order of operations. (The string replacement is in there so you can just copy the output into the source code.)
- I might be misunderstanding your concern, but although the `enable_features` code example is the same for python and R right now, I don't think we should commit ourselves to that. Even for the python, I was wondering if it should have a `dp.` prefix. I think it's fine for now, but I wouldn't want to commit to an approach that would make it harder if they did need to diverge.
- Besides being at the bottom, is there anything else we should have? Maybe it deserves it's own section header, or a link to the feature list.

Shoeboxam:
>     * How would you feel about adding [docs.rs/pretty_assertions](https://docs.rs/pretty_assertions) as a dev dependency? The changes can be hard to spot, and so I end up copy and pasting and then looking at the git diff, which is not the best order of operations. (The string replacement is in there so you can just copy the output into the source code.)

Dev dependency is fine.

>     * I might be misunderstanding your concern, but although the `enable_features` code example is the same for python and R right now, I don't think we should commit ourselves to that. Even for the python, I was wondering if it should have a `dp.` prefix. I think it's fine for now, but I wouldn't want to commit to an approach that would make it harder if they did need to diverge.

It doesn't have to be code, does it? Just a list will do. I'd assume people are familiar with `enable_features`.

>     * Besides being at the bottom, is there anything else we should have? Maybe it deserves it's own section header, or a link to the feature list.

It could be very repetitive. Also, linking to the feature list will make us want to substitute versioned links, and I don't think we want to spend time on that.

mccalluc:
> It doesn't have to be code, does it? Just a list will do. I'd assume people are familiar with `enable_features`.

Got it: Sorry to go down a different path with this. I'll aim for something like:

> Required features: `contrib`

or

> Required features: `contrib`, `honest-but-curious`

mccalluc:
- Replaced by #2014

Issue #1977, opened 2024-08-23
Usability: When grouping by single columns, easy to forget the closing comma in the tuple


mccalluc:
It's easy to type
```
("ILOSTAT"): dp.polars.Margin(public_info="keys", max_partition_length=60_000_000)
```
when you mean
```
("ILOSTAT", ): dp.polars.Margin(public_info="keys", max_partition_length=60_000_000)
```
(note the trailing comma that makes it a tuple.)

Downsteam, it complains that you didn't specify a margin, but it really looks like you did!

Options:
- Silently transform strings to tuples?
- Warn earlier if there is a non-tuple key?

mccalluc:
@saratuftstsp is interested in this. Maybe think about the alternatives and make a suggestion?

PR #1981, opened 2024-08-26
Add an `xfail` test of mistakenly using a string for a margin key.

mccalluc:
- Towards #1977

(For fix, leaning towards failing early if the key is not a tuple.)

### DIFF
--- a/python/test/test_context_usability.py
+++ b/python/test/test_context_usability.py
@@ -85,10 +85,7 @@ def test_query_dir():
     assert 'count' in query_dir
     assert 'laplace' in query_dir
 
-@pytest.mark.xfail(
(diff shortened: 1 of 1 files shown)


Shoeboxam (approved):
Why not also fix the issue? Checking for non-tuple keys seems like a pretty straightforward fix. I'd also be ok with taking a single string and automatically upgrading it. Up to your judgement.

mccalluc:
> Why not also fix the issue? Checking for non-tuple keys seems like a pretty straightforward fix.

Will do: That's the approach I'd prefer. Since there could have been some discussion about the right fix, but there isn't really any question that it's a bug, though it would be good to propose just this much to begin.

Issue #1976, opened 2024-08-23
If CSV has encoding problem, Polars loads a dataset with zero rows


mccalluc:
Hit this during the hackathon, and it was confusing. It's not really our problem... but with DP people aren't supposed to look at the data, so maybe we need to work harder to parse the CSV than polars out of the box, which assumes user interaction.

Shoeboxam:
The behavior I'm seeing with your xfail example is that it parses any rows with unusual characters (like é) as null. Values without those characters parse correctly.

mccalluc:
I thought I saw in the hackathon that no rows would load if there was a misencoding anywhere. I can expand the test.

SalilVadhan:
It happened on the squirrel dataset.


-----
Salil Vadhan (he/they)

mccalluc:
Solution proposed in 
- https://github.com/opendp/dp-creator-ii/pull/29

is to use `encoding="utf8-lossy"`: It's not perfect, but unless two grouping strings are identical apart from characters that are misencoded, it wouldn't harm the statistics. Somehow getting the correct encoding and re-encoding the CSV feels like a big hassle for very little gain.

mccalluc:
Close in favor of
- #2234

PR #1983, opened 2024-08-26
Add`xfail` for mis-encoded CSV loading

mccalluc:
- Towards #1976

If we decide to implement a friendlier CSV loader, we can test it here.


Shoeboxam (approved):
Approved, but not closing the issue yet because docs changes would be good as well.

PR #1987, opened 2024-08-27
Add tests of code generation

### LINKED ISSUE #1969: Unit testing of code generation
In the before-times, the checked-in python and R code were implicit tests of the code generation. If something changed in the wrong way, we'd see it in those files.

Now, for example with
- #1968

it's hard for a code reviewer to see what is actually changing, and unless the changes are breaking, they won't be caught by CI.

We should:
- [ ] test python code generation 
- [ ] test R code generation

Probably just have a big string in the Rust source, and make sure newly generated code matches that.

mccalluc:
- Fix #1969

If something went really wrong in the code generation, we'd be able to tell because Python and R would stop working, but right now that there is no testing that changes smaller than that are working as intended.
- motivating example: #1968
- also: #1961 (Does have a [test of the warning itself](https://github.com/opendp/opendp/pull/1961/files#diff-a2afdeca1db58e51ac0dcbf12d3951a80ae0b7ae38226d2a7e8ed3be09c15a9fR327), but doesn't check the generated code.)

(If tests pass, I believe this is ready to go, but I'd have a preference for merging both of those first, and then rebasing or merging main to this branch and updating the assertions here. Any future codegen work will update these tests in the same PR.)

This isn't trying to give us 100% test coverage of code generation (other inputs would generate other paths, and there's also the code that generates the wrapping module), but it's better than nothing.


rust/opendp_tooling/src/codegen/test.rs
```diff
@@ -0,0 +1,36 @@
...
+use crate::{Argument, Function};
+use std::collections::HashMap;
+
+#[test]
+fn test_python_code_generation() {
+    // TODO: What is the least we can fill in below to get it to run?
+    let argument = Argument {
+        name: None,
+        c_type: None,
+        rust_type: None,
```
mccalluc, line 11: Write a rust function in a string, then use macro quote for a token stream than can be passed to bootstrap. (Usually called by compiler.) ... Or not, because it writes to the file system.

perhaps write a test for `from_ast`?

rust/opendp_tooling/src/codegen/test.rs
```diff
@@ -0,0 +1,36 @@
+use crate::codegen::python::generate_function;
+use crate::{Argument, Function};
+use std::collections::HashMap;
+
+#[test]
+fn test_python_code_generation() {
+    // TODO: What is the least we can fill in below to get it to run?
+    let argument = Argument {
+        name: None,
```
mccalluc, line 9: set this

rust/opendp_tooling/src/codegen/test.rs
```diff
@@ -0,0 +1,36 @@
...
+        c_type: None,
```
mccalluc, line 10: "double"

rust/opendp_tooling/src/codegen/test.rs
mccalluc: "f64"

rust/opendp_tooling/src/codegen/test.rs
```diff
@@ -0,0 +1,36 @@
...
+        rust_type: None,
+        hint: None,
+        description: None,
```
mccalluc, line 13: any string

rust/opendp_tooling/src/codegen/test.rs
```diff
@@ -0,0 +1,36 @@
...
+        default: None,
```
mccalluc, line 14: stronger test if we fill this in.
`Some(Value::Float(23.4))`

rust/opendp_tooling/src/codegen/test.rs
```diff
@@ -0,0 +1,36 @@
...
+        is_type: true,
+        do_not_convert: false,
+        example: None,
+    };
+
+    let function = Function {
+        name: "fake_function".to_string(),
+        description: Some("fake description".to_string()),
+        features: vec!["fake_feature".to_string()],
+        args: vec![],
```
mccalluc, line 25: use the argument defined above

rust/opendp_tooling/src/codegen/test.rs
```diff
@@ -0,0 +1,36 @@
...
+        derived_types: vec![],
+        ret: argument,
```
mccalluc, line 27: try to reuse, if that doesn't work, give it a separate argument.

Shoeboxam (approved):
Thanks for writing this! I'm ok with this so far, but I'm guessing you're planning to update these tests to cover more of the functionality specifically related to things that don't get tested by the existing tests, like adjustments to the docstrings.

PR #2003, opened 2024-08-31
Polars: add cut transformation and categorical dtype

### LINKED ISSUE #894: Polars: add `make_expr_cut`
**Description**
- **Class**: Expr
- **DSL Variant:** Function: Cut 
   - **Purpose**: for multi-quantiles

--- 

Shoeboxam:
Closes #894

### DIFF
--- a/python/test/test_polars.py
+++ b/python/test/test_polars.py
@@ -2,6 +2,7 @@ import pytest
 import opendp.prelude as dp
 import os
 import warnings
+import re
 
 
 def test_polars_version():
@@ -579,7 +580,6 @@ def test_pickle_bomb():
     from polars._utils.wrap import wrap_expr  # type: ignore[import-not-found]
     from opendp._lib import lib_path
     import io
-    import re
     import pickle
 
     # modified from https://intoli.com/blog/dangerous-pickles/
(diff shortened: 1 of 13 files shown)


Shoeboxam:
* **#2004** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2004?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2003** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2003?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

rust/src/domains/polars/series/mod.rs
```diff
@@ -215,12 +215,64 @@ impl<T: CheckAtom + PrimitiveDataType> SeriesAtomDomain for OptionDomain<AtomDom
...
+}
+
+impl CategoricalDomain {
+    /// Only use this constructor if you know both the category set,
+    /// as well as how the categories are encoded.
+    ///
+    /// Typically when Polars constructs categorical data,
+    /// it assigns indices by the order encountered in the data, making the encoding data-dependent.
+    /// An example where this can be called is for categorical data emitted by the Polars cut expression,
+    /// where the categories and encoding are pre-determined by the expression (the bin edges).
```
mccalluc, line 231: This is interesting. I had tried to use category at some point in a docs example, for something like state abbreviations, but it sounds like we would never recommend that. (But this isn't bootstrapped, so it's not showing up in the generated code in any case.)

Maybe the actionable thing here is to note in our Polars table that categories aren't available, and explain why not? File an issue from this comment if that seems like a good route.

mccalluc (dismissed):
Looks good. Think I've reviewed bits of this before, though not in this PR? Let me know if anything deserves a closer look.

rust/src/domains/polars/series/mod.rs
Shoeboxam: I've gone ahead and fleshed this PR out to actually add categoricals. You can use categorical data with unknown encodings, so long as you don't read or change the encoding.

python/test/test_polars.py
```diff
@@ -692,4 +692,52 @@ def test_csv_bad_encoding_loading():
...
+
+    lf_domain = dp.lazyframe_domain([dp.series_domain("A", dp.categorical_domain())])
+    assert str(lf_domain) == "FrameDomain(A: cat; margins=[])"
+
+    lf_domain = dp.lazyframe_domain([dp.series_domain("A", dp.option_domain(dp.categorical_domain()))])
+    assert str(lf_domain) == "FrameDomain(A: cat; margins=[])"
+
+    lf = pl.LazyFrame([pl.Series("A", ["Texas", "New York", None], dtype=pl.Categorical)])
+
+    with pytest.raises(dp.OpenDPException):
```
mccalluc, line 709: Also check the error message? Want to be sure that it makes sense to the user.

(I also see elsewhere in this PR that the categorical is wrapped by optional... but I guess that doesn't mean we can fill it with nulls? -- If that does enable anything, special that too.)

mccalluc (commented):
Thanks for adding Categoricals. I think #2025 can cover any docs.

Where we have a `pytest.raises`, I'd have a preference to also check the text of the error message, to make sure we're hitting the error we think we should hit, and to encourage us to think about the usefulness of the error to end users.

(I wouldn't make this a rule on every project: if the error classes are fine grained, that is enough, but we only have a few, and we know that the embedded Rust errors make it harder to understand, so I think we need an extra checks to make sure it's informative.)

(But if that bar is too high, and it would discourage tests at least this good, please ignore.)

python/test/test_polars.py
Shoeboxam: I explicitly check both now.

mccalluc (approved):
Thanks!

PR #2010, opened 2024-09-03
Add "--package" to "cargo publish"

### LINKED ISSUE #1996: Cargo.toml default-members breaks publish
We'd typically expect `cargo publish rust/Cargo.toml` to publish just the opendp Rust crate. Since we've added `default-members` to the Cargo.toml, this now also tries to run the publish command on the tooling crate, resulting in the following ambiguity:

https://github.com/opendp/opendp/actions/runs/10608370064/job/29404634243

We should either explicitly specify `-p opendp` now when publishing, or drop the default members key and explicitly add a second `cargo test` where appropriate.

mccalluc:
- Fix #1996

(and hopefully fix nightly, and get the warnings on docs for older versions.)


Shoeboxam (approved):
Great! Onwards to nightlies!

PR #2014, opened 2024-09-03
Add required features note

### LINKED ISSUE #1662: Flag (and explain?) `honest-but-curious` (and `contrib`?) in API docs

mccalluc:
- Fix #1662
- Replace #1968
- Introduce pretty diffs in rust tests
- Add a test of docstring construction, but since it is only testing the new code, it's pretty under-powered.

This also removes a bit of special logic for the "Features" section in docstrings that was only used on one function. My sense is that keeping both it and this new feature warning could be confusing. In the long run, it would be good to explain places where `honest-but-curious` is required, but we should make sure that everything is in sync.

- See #1664


Shoeboxam (approved):
The general pattern here has been-- we have structured information in Rust, lets show it in the documentation from other languages.

PR #2026, opened 2024-09-11
pick just one binary

### LINKED ISSUE #2024: Python package fails to load in development environment
Now that `default-members = [".", "opendp_derive", "opendp_tooling"]`, an additional `opendp_derive.dylib` (mac) is emitted in the artifacts folder. This prevents the Python package from loading in dev environments, as the Python package expects exactly one matching binary.

One way to fix this would be to update the Python code to ignore this binary.

mccalluc:
- Fix #2024

### DIFF
--- a/python/src/opendp/_lib.py
+++ b/python/src/opendp/_lib.py
@@ -29,7 +29,7 @@ def _load_library():
         lib_dir_file_names = [
             p for p in lib_dir.iterdir()
             if p.suffix in {".so", ".dylib", ".dll", ".pyd"}
-            and p.stem == 'libopendp']
+            and p.stem.replace("lib", "") == "opendp"]
         if len(lib_dir_file_names) != 1:
             raise Exception(f"Expected exactly one binary to be present in {lib_dir}. Got: {lib_dir_file_names}")
         


python/src/opendp/_lib.py
```diff
@@ -26,7 +26,10 @@ def _load_library():
         lib_dir = Path(__file__).parent / ".." / ".." / ".." / 'rust' / 'target' / build_dir  # pragma: no cover
 
     if lib_dir.exists():
-        lib_dir_file_names = [p for p in lib_dir.iterdir() if p.suffix in {".so", ".dylib", ".dll", ".pyd"}]
+        lib_dir_file_names = [
+            p for p in lib_dir.iterdir()
+            if p.suffix in {".so", ".dylib", ".dll", ".pyd"}
+            and p.stem == 'libopendp']
```
Shoeboxam, line 32: If I recall correctly, on windows the stem doesn't include "lib". I think this will work:
```suggestion
            and p.stem.replace("lib", "") == "opendp"]
```

mccalluc: Thanks!

Shoeboxam (approved):

PR #2040, opened 2024-09-13
Document supported polars methods

### LINKED ISSUE #2025: Polars: documentation page with tables of supported APIs
Move this information, cut down and repurpose:
https://docs.google.com/spreadsheets/d/1w4H-qrSe9pkvNT2NhowfiIHNWXSseOyMX2nu5H6-RkQ/edit?gid=0#gid=0

mccalluc:
- Fix #2025

Draft for now:
- Is this the right place for this information? I think at one time there was an idea of adding a page in the tabular data section, but this seems more like reference material.
- Is a fourth column with doctests useful? Maybe they are useful, but cramming them in a table is a mistake.
- How should we explain where these can be used? They can't be used right after the `dp`, but they will work if `noise` or another method intervenes.
- Does it make sense to do the same thing for `LazyFrame` methods?

### DIFF
--- a/python/src/opendp/extras/polars/__init__.py
+++ b/python/src/opendp/extras/polars/__init__.py
@@ -70,7 +70,7 @@ class DPExpr(object):
 
     A few ``Expr`` aggregation methods are also available:
 
-    .. list-table:: Supported Polars ``Expr`` Methods
+    .. list-table:: Supported Polars ``Expr`` Aggregation Methods
         :header-rows: 1
 
         * - Method
--- a/python/src/opendp/extras/polars/__init__.py
+++ b/python/src/opendp/extras/polars/__init__.py
@@ -45,7 +45,7 @@ class DPExpr(object):
     >>> pl.len().dp
     <opendp.extras.polars.DPExpr object at ...>
 
-    In addition to the DP-specific methods documented below, many Polars ``Expr`` methods are also supported.
+    In addition to the DP-specific methods documented below, some Polars ``Expr`` methods are also supported.
     For these, the best documentation is the `official Polars documentation <https://docs.pola.rs/api/python/stable/reference/expressions/index.html>`_.
 
     .. list-table:: Supported Polars ``Expr`` Methods
@@ -77,13 +77,10 @@ class DPExpr(object):
           - Comments
         * - `len <https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.len.html>`_
           - Number of rows, including nulls
-        * - `count <https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.count.html>`_
+        * - `count <https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.count.html>`_
           - Number of rows, not including nulls
-        * - `n_unique <https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.n_unique.html>`_
+        * - `n_unique <https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.n_unique.html>`_
           - Number of unique rows
-
-
-              
     '''
     def __init__(self, expr):
         """Apply a differentially private plugin to a Polars expression."""
--- a/python/src/opendp/extras/polars/__init__.py
+++ b/python/src/opendp/extras/polars/__init__.py
@@ -54,30 +54,43 @@ class DPExpr(object):
         * - Method
           - Operators
           - Use for
+          - Usage
         * - `alias <https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.alias.html>`_
           -
           - Renaming the expression
+          - 
+            >>> import polars as pl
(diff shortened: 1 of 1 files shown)


python/src/opendp/extras/polars/__init__.py
```diff
@@ -44,6 +44,53 @@ class DPExpr(object):
...
+    In addition to the DP-specific methods documented below, many Polars ``Expr`` methods are also supported.
+    For these, the best documentation is the official Polars documentation.
+
+    .. list-table:: Supported Polars ``Expr`` Methods
+        :header-rows: 1
+
+        * - Method
+          - Operators
+          - Use for
+          - Usage
```
Shoeboxam, line 57: To save space, how would you feel about condensing into two columns:

* method (with link)
* notes

Where for the most part, notes talks about privacy-specific behavior, but if space allows, also some pointers. 

Some suggestions:

# Context Transformations

.
* with_columns
* expressions must not reorder or add/remove rows

.
* filter
* margin descriptors may be lost; consider group_by instead

# Context Aggregations

.
* select
* can specify multiple expressions

.
* group_by/agg
* can specify multiple grouping keys and expressions



# Expression Transformations
.
* eq, ne, lt, le, gt, ge
* Or use operators: ``==`` ``!=`` ``<`` ``<=`` ``>`` ``>=``

.
* clip
* establishes bounds descriptor

.
* fill_null
* if fill value is non-null, establishes non-nullity descriptor

.
* fill_nan
* if fill value is non-null and non-nan, establishes non-nan descriptor

# Expression Aggregators
.
* len
* number of rows, including nulls

.
* count
* number of rows, not including nulls

.
* n_unique
* number of unique rows

I think code examples are already covered by Polars. Please don't hesitate to disagree if you feel differently!

Shoeboxam (dismissed):
Just sharing thoughts, I'm flexible.

mccalluc:
Good call.
- I've broken this into two separate tables, because they work differently, though this doc doesn't really explain the difference.
- I'm not covering the few methods you list above for which we have an actual implementation: Including them in the table seems redundant.
- For me the code example would really have been a test for us. There's not room in a table to give an example that's useful to users.

mccalluc:
Tons of CI failures; Re-running now: Probably a github problem?

mccalluc:
Re-run also failed? re-running again...

python/src/opendp/extras/polars/__init__.py
mccalluc: Thanks. Should be addressed.

python/src/opendp/extras/polars/__init__.py
```diff
@@ -44,6 +44,43 @@ class DPExpr(object):
...
+    .. list-table:: Supported Polars ``Expr`` Methods
+        :header-rows: 1
+
+        * - Method
+          - Comments
+        * - `alias <https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.alias.html>`_
+          - Rename the expression
+        * - `eq, ne, lt, le, gt, ge <https://docs.pola.rs/api/python/stable/reference/expressions/operators.html#comparison>`_
+          - Comparison operators may be more readable: ``==`` ``!=`` ``<`` ``<=`` ``>`` ``>=``
+        * - `and, or, xor <https://docs.pola.rs/api/python/stable/reference/expressions/operators.html#conjunction>`_
```
Shoeboxam, line 60:
```suggestion
        * - `and_, or_, xor <https://docs.pola.rs/api/python/stable/reference/expressions/operators.html#conjunction>`_
```

python/src/opendp/extras/polars/__init__.py
```diff
@@ -44,6 +44,43 @@ class DPExpr(object):
...
+        :header-rows: 1
+
+        * - Method
+          - Comments
+        * - `len <https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.len.html>`_
+          - Number of rows, including nulls
+        * - `count <https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.count.html>`_
+          - Number of rows, not including nulls
+        * - `n_unique <https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.n_unique.html>`_
+          - Number of unique rows
```
Shoeboxam, line 83: I had previously written this to give examples, but we don't have n_unique. Instead, add `sum`?

Shoeboxam (changes requested):
Some small changes. Looking good.

python/src/opendp/extras/polars/__init__.py
Shoeboxam: Bump on this ^^

Shoeboxam (changes requested):
I guess my previous comment didn't go in as a review? My bad.

mccalluc:
Sorry to miss that. Fixing now...

Shoeboxam (approved):
Thanks!

PR #2059, opened 2024-10-02
check max_changed_partitions at construction-time

### LINKED ISSUE #1908: Polars privacy map fail but measurement goes through

Shoeboxam:
Close #1908

### DIFF
--- a/python/test/test_polars.py
+++ b/python/test/test_polars.py
@@ -104,7 +104,7 @@ def test_private_lazyframe_explicit_sum(measure):
     pl_testing = pytest.importorskip("polars.testing")
 
     lf_domain, lf = example_lf(
-        margin=["B"], public_info="keys", max_partition_length=50
+        margin=["B"], public_info="keys", max_partition_length=50, max_num_partitions=10,
     )
 
     expr = pl.col("A").fill_null(0.0).clip(0.0, 1.0).sum().dp.noise(0.0)
@@ -131,7 +131,7 @@ def test_private_lazyframe_sum(measure):
     pl_testing = pytest.importorskip("polars.testing")
 
     lf_domain, lf = example_lf(
-        margin=["B"], public_info="keys", max_partition_length=50
+        margin=["B"], public_info="keys", max_partition_length=50, max_num_partitions=10,
     )
     expr = pl.col("A").fill_null(0.0).dp.sum((1.0, 2.0), scale=0.0)
     plan = seed(lf.collect_schema()).group_by("B").agg(expr).sort("B")
@@ -156,7 +156,7 @@ def test_private_lazyframe_mean(measure):
     pl_testing = pytest.importorskip("polars.testing")
 
     lf_domain, lf = example_lf(
-        margin=["B"], public_info="lengths", max_partition_length=50
+        margin=["B"], public_info="lengths", max_partition_length=50, max_num_partitions=10,
     )
 
     expr = pl.col("A").fill_null(0.0).dp.mean((1.0, 2.0), scale=0.0)
@@ -822,8 +822,6 @@ def test_to_physical_unordered():
 
 
 def test_float_sum_with_unlimited_reorderable_partitions():
-    pl = pytest.importorskip('polars')
-    
     lf_domain = dp.lazyframe_domain([
         dp.series_domain("region", dp.atom_domain(T=dp.i64)),
         dp.series_domain("income", dp.atom_domain(T=dp.f64))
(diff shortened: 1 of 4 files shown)


Shoeboxam:
* **#2059** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2059?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

python/test/test_polars.py
```diff
@@ -821,3 +821,23 @@ def test_to_physical_unordered():
...
+        dp.series_domain("region", dp.atom_domain(T=dp.i64)),
+        dp.series_domain("income", dp.atom_domain(T=dp.f64))
+    ])
+    lf_domain = dp.with_margin(lf_domain, by=["region"], public_info="lengths", max_partition_length=6)
+
+    from opendp.domains import _lazyframe_from_domain
+    lf = _lazyframe_from_domain(lf_domain)
+
+    # sum of income per region, add noise with scale of 1.0
+    plan = lf.group_by("region").agg([
```
mccalluc, line 837: Slight preference for keeping the `importorskip` as late as possible, so the first part of the test runs whether or not polars is installed. If there's an indirect dependency on polars, that could be handled with `optional_dependency`.

```suggestion
    lf_domain = dp.lazyframe_domain([
        dp.series_domain("region", dp.atom_domain(T=dp.i64)),
        dp.series_domain("income", dp.atom_domain(T=dp.f64))
    ])
    lf_domain = dp.with_margin(lf_domain, by=["region"], public_info="lengths", max_partition_length=6)

    from opendp.domains import _lazyframe_from_domain
    lf = _lazyframe_from_domain(lf_domain)

    # sum of income per region, add noise with scale of 1.0
    pl = pytest.importorskip('polars')
    plan = lf.group_by("region").agg([
```

python/test/test_polars.py
```diff
@@ -821,3 +821,23 @@ def test_to_physical_unordered():
...
+        pl.col("income").dp.sum(bounds=(1_000, 100_000), scale=1.0)
+    ])
+
+    # since there are an unknown number of partitions, and each partition has non-zero sensitivity, sensitivity is undefined
+    with pytest.raises(dp.OpenDPException, match='.*max_num_partitions must be known when the metric is not sensitive to ordering.*'):
```
mccalluc, line 842:
```suggestion
    with pytest.raises(dp.OpenDPException, match='max_num_partitions must be known when the metric is not sensitive to ordering'):
```
regex can match substring.

rust/src/transformations/make_stable_expr/expr_sum/mod.rs
```diff
@@ -125,6 +125,11 @@ where
         _ => TI::Sum::inf_cast(*d_in)?.inf_mul(&l.alerting_abs()?.total_max(u)?),
     };
 
+    // 'mnp_check: this invariant is used later
```
mccalluc, line 128: I'm expecting to see `'mnp_check` in the code here, but it's just in an error message. Is that right?

mccalluc (commented):
Some little comments, but the biggest headache seems to be hashbrown: Whatever the fix for that is, it might belong in its own PR that others can be based on.

rust/src/transformations/make_stable_expr/expr_sum/mod.rs
Shoeboxam: yep, the error message should never be seen by the user, and it references another location by a comment.

mccalluc (approved):
Thanks!

PR #2084, opened 2024-10-10
simplify sum traits

### LINKED ISSUE #2083: `make_sum` trait simplification

Shoeboxam:
Close #2083

### DIFF
--- a/rust/src/transformations/sum/int/monotonic/mod.rs
+++ b/rust/src/transformations/sum/int/monotonic/mod.rs
@@ -5,10 +5,8 @@ use crate::{
     domains::{AtomDomain, VectorDomain},
     error::Fallible,
     metrics::{AbsoluteDistance, IntDistance, SymmetricDistance},
-    traits::Number,
+    traits::{Integer, Number},
 };
-
-use super::AddIsExact;
 
 #[cfg(feature = "ffi")]
 mod ffi;
@@ -26,7 +24,7 @@ mod ffi;
 ///
 /// # Generics
 /// * `T` - Atomic Input Type and Output Type
-pub fn make_bounded_int_monotonic_sum<T>(
+pub fn make_bounded_int_monotonic_sum<T: Integer>(
     bounds: (T, T),
 ) -> Fallible<
     Transformation<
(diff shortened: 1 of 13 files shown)


Shoeboxam:
* **#2084** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2084?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

rust/src/transformations/sum/int/monotonic/mod.rs
```diff
@@ -110,29 +102,13 @@ where
...
-        fn is_monotonic(_: (Self, Self)) -> bool {
-            true
-        }
-    })+)
+/// # Proof Definition
+/// Returns true if and only if the signs of `a` and `b` agree,
+/// assuming that values of zero always agree.
+pub(crate) fn signs_agree<T: Number>((a, b): (T, T)) -> bool {
+    let _0 = T::zero();
+    a == _0 || b == _0 || (a > _0) == (b > _0)
```
mccalluc, line 110:
```suggestion
    (a == _0 && b == _0) || (a > _0) == (b > _0)
```
or
```suggestion
    (a == b) || (a > _0) == (b > _0)
```
Maybe I'm missing something, but it looks like this could return true if `a == 0`, and not even look at `b`, which doesn't seem right.

mccalluc (commented):
I might be confused, but there could be a small bug. Consider adding a test that will fail before fixing.

rust/src/transformations/sum/int/monotonic/mod.rs
Shoeboxam: This is the correct behavior for this function! If either value is zero, then the sign will always agree. That is, -4 and 0 are considered to have agreement in sign, as 0 agrees to anything.

mccalluc: ah, got it:
> assuming that values of zero always agree

mccalluc (approved):
Looking at this again, I think I interpreted it as
> assuming that values of zero always agree _with each other_

which makes less sense than 
> assuming that values of zero always agree _with anything_

Sorry!

PR #2086, opened 2024-10-10
disable ffi and partial features on crates.io

### LINKED ISSUE #2085: Consider disabling partial and ffi features in docs.rs documentation build
For both partial and ffi functions, the documentation is repetitive to the original functions, which makes the documentation more confusing to navigate.

Shoeboxam:
Closes #2085

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -131,7 +131,8 @@ partials = ["derive", "opendp_derive/partials"]
 crate-type = ["rlib", "cdylib", "staticlib"]
 
 [package.metadata.docs.rs]
-features = ["derive", "untrusted", "ffi", "polars"]
+no-default-features = true
+features = ["use-openssl", "derive", "untrusted", "polars"]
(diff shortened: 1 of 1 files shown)


Shoeboxam:
* **#2086** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2086?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

mccalluc (approved):
Maybe add a comment on `use-openssl`? It feels like the odd one out, but I'm not super familiar with this code.

I also don't have a clear sense right now of what the default features actually get us. Would there be any point in changing the default features, so we could use the default more often?

Approving, but feel free to add a comment, if you think it would be useful -- It might not!

Shoeboxam:
I anticipate people always using the default features unless there's a really good reason not to. The default/non-default configuration is also pretty conservative in terms of privacy (flags are opt-in, openssl is more secure than not).

PR #2087, opened 2024-10-11
Polars: proof definitions

Shoeboxam:
Towards #2080

### DIFF
--- a/rust/src/domains/polars/expr/mod.rs
+++ b/rust/src/domains/polars/expr/mod.rs
@@ -117,7 +117,7 @@ pub struct ExprDomain {
 
 impl ExprDomain {
     /// # Proof Definition
-    /// Returns an expr domain with members matching those of `frame_domain`.
+    /// Returns an [`ExprDomain`] with members matching those of `frame_domain`.
     pub fn new<F: Frame>(frame_domain: FrameDomain<F>, context: ExprContext) -> ExprDomain {
         Self {
             frame_domain: frame_domain.cast_carrier(),
--- a/rust/src/domains/polars/expr/mod.rs
+++ b/rust/src/domains/polars/expr/mod.rs
@@ -5,14 +5,14 @@ use std::fmt::{Debug, Formatter};
 
 use crate::core::{Metric, MetricSpace};
 use crate::metrics::{
-    AbsoluteDistance, ChangeOneDistance, HammingDistance, InsertDeleteDistance, LInfDistance,
-    LpDistance, Parallel, PartitionDistance, SymmetricDistance,
+    ChangeOneDistance, HammingDistance, InsertDeleteDistance, LInfDistance, LpDistance, Parallel,
+    PartitionDistance, SymmetricDistance,
 };
 use crate::traits::ProductOrd;
 use crate::transformations::DatasetMetric;
 use crate::{core::Domain, error::Fallible};
 
-use super::{DslPlanDomain, Frame, FrameDomain, Margin, NumericDataType, SeriesDomain};
+use super::{DslPlanDomain, Frame, FrameDomain, Margin, SeriesDomain};
 
 #[cfg(feature = "ffi")]
 mod ffi;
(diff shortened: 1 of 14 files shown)


Shoeboxam:
* **#2087** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2087?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

rust/src/domains/polars/expr/mod.rs
```diff
@@ -41,11 +41,13 @@ pub enum ExprContext {
 }
 
 impl ExprContext {
-    fn get_plan(&self, val: &(DslPlan, Expr)) -> DslPlan {
-        let (lp, expr) = val.clone();
-        let frame = LazyFrame::from(lp);
+    /// # Proof Definition
+    /// Given a context (`self`), logical plan (`lp`) and expression,
```
mccalluc, line 45: Not particular to this PR, but would `plan` be more readable than `lp`, at the cost of two more characters?

rust/src/domains/polars/expr/mod.rs
```diff
@@ -54,11 +56,14 @@ impl ExprContext {
                         .map(col)
                         .collect::<Vec<_>>(),
                 )
-                .agg([expr]),
+                .agg([expr.clone()]),
```
mccalluc, line 59: I don't think `expr` is particularly heavyweight, but is there an argument for doing it once up top? Probably depends on context? Mostly for my own learning.

rust/src/domains/polars/expr/mod.rs
```diff
@@ -104,23 +116,36 @@ pub struct ExprDomain {
 }
 
 impl ExprDomain {
+    /// # Proof Definition
+    /// Returns an expr domain with members matching those of `frame_domain`.
```
mccalluc, line 120:
```suggestion
    /// Returns an [`ExprDomain`] with members matching those of `frame_domain`.
```

rust/src/domains/polars/expr/mod.rs
```diff
@@ -69,6 +74,13 @@ impl ExprContext {
...
     }
 
+    /// # Proof Definition
+    /// Only return without an error if the expression context `self`
+    /// allows for row ordering or the number of rows to be changed.
+    ///
+    /// This is used to disallow operations like sorting, shuffling or filtering
+    /// when row alignment must not change,
+    /// like when augmenting a new column via `with_column`.
     pub fn break_alignment(&self) -> Fallible<()> {
```
mccalluc, line 84: Not the subject of this PR, but would this be more clear with a `check_` name?

rust/src/domains/polars/expr/mod.rs
```diff
@@ -178,8 +205,17 @@ impl Debug for ExprDomain {
     }
 }
 
+/// OuterMetric encodes the relationship between
```
mccalluc, line 208: not a proof definition?

mccalluc (approved):
All looks good. Had a couple questions while in the neighborhood, but I think this would be fine to merge as-is.

rust/src/domains/polars/expr/mod.rs
Shoeboxam: The overhead is ultimately the same, as only one branch of the match will ever be taken at a time.

rust/src/domains/polars/expr/mod.rs
Shoeboxam: Hmm, it could be something like, `check_alignment_can_be_broken`.

rust/src/domains/polars/expr/mod.rs
Shoeboxam: Yeah, this one isn't worded as a proof definition. There are some cases where a proof definition on the trait itself could make sense (for marker traits), but I think most of the time it's the methods, consts and associated types that need definitions.

PR #2091, opened 2024-10-11
Polars: reuse margin descriptors

### LINKED ISSUE #1707: Polars: reuse margin descriptors when possible
Imagine you specify a margin descriptor like so:

```python
    margins={(): dp.Margin(max_partition_len=10_000)}
```

Then the max partition length should be known in the following query:

```python
(
    context.query()
    .group_by("A")
    .agg(pl.col("B").fill_null(0).dp.sum((0, 10)))
    .release().collect()
)
```

Shoeboxam:
Close #1707


Shoeboxam:
* **#2091** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2091?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

docs/source/getting-started/assessing-utility/noise.ipynb
```diff
@@ -184,7 +184,7 @@
     "    privacy_loss=dp.loss_of(epsilon=1.0),\n",
     "    split_evenly_over=10,\n",
     "    margins={\n",
-    "        (\"ILOSTAT\", ): dp.polars.Margin(public_info=\"keys\", max_partition_length=60_000_000),\n",
+    "        \"ILOSTAT\": dp.polars.Margin(public_info=\"keys\"),\n",
```
mccalluc, line 187: My preference would be to keep tuple keys as the style we present in the documentation, but support strings because it's easy to forget that comma. I think in the docs having tuples in some cases and strings in others may actually be a little confusing to the user.

(Using list keys probably isn't a good option, because they're mutable, right?)

rust/src/domains/polars/frame/mod.rs
```diff
@@ -389,3 +461,47 @@ impl Margin {
             .min(l_1)
     }
 }
+
+/// # Proof Definition
+/// Return a subset of `sets` whose intersection is a superset of `elements`.
```
mccalluc, line 466: Expecting a statement here in terms of the parameters `sets` and `not_covered`, rather than `elements`.

rust/src/domains/polars/frame/test.rs
```diff
@@ -39,3 +39,136 @@ fn test_margin() -> Fallible<()> {
...
+        lf_domain
+            .get_margin(BTreeSet::from(["A".to_string()]))
+            .public_info,
+        Some(MarginPub::Keys)
+    );
+    Ok(())
+}
+
+#[test]
+fn test_option_or() {
```
mccalluc, line 169:
```suggestion
fn test_option_min() {
```

rust/src/transformations/make_stable_expr/expr_len/mod.rs
```diff
@@ -45,10 +45,7 @@ where
     );
 
     // we only care about the margin that matches the current grouping columns
```
mccalluc, line 47:
```suggestion
```
Less explanation needed now?

mccalluc (commented):
Might have typos on one proof definition and a test name, but biggest worry is with inconsistent of margin keys in the docs. If we do want to move towards string keys, then maybe having comma-delimited strings would be good if there is more than one key? But my slight preference is to stick with tuples as keys.

Not something I would block on, but I think consistency in presentation is good, whatever we decide on, though we can be a little more flexible in what we accept to make life easier for the user (ie, transform strings to tuples.)

docs/source/getting-started/assessing-utility/noise.ipynb
Shoeboxam: Changed all to tuples. Once passed into the lib, they can never change, and I remember seeing pandas representing grouping keys as tuples.

rust/src/domains/polars/frame/test.rs
```diff
@@ -39,3 +39,136 @@ fn test_margin() -> Fallible<()> {
...
+        &["A"],
+        Margin::default()
+            .with_max_num_partitions(10)
+            .with_max_influenced_partitions(3),
+    )?
+    .with_margin(
+        &["B"],
+        Margin::default()
+            .with_max_num_partitions(10)
+            .with_max_influenced_partitions(3),
```
mccalluc, line 102: Slightly stronger test if these are set to different values than the first margin. Not that it's likely, but if the code just used the first margin twice, these tests wouldn't catch the problem.

rust/src/domains/polars/frame/test.rs
```diff
@@ -39,3 +39,136 @@ fn test_margin() -> Fallible<()> {
...
+            .max_num_partitions,
+        Some(100)
+    );
+
+    assert_eq!(
+        lf_domain
+            .get_margin(BTreeSet::from(["B".to_string()]))
+            .max_influenced_partitions,
+        Some(3)
+    );
```
mccalluc, line 117: Just to check the extreme cases, would it be good to check what the behavior is on an empty set? or if a member of the set is not in the margins?

rust/src/transformations/make_stable_expr/expr_len/mod.rs
```diff
@@ -44,11 +44,7 @@ where
         input_domain.context.clone(),
     );
 
-    // we only care about the margin that matches the current grouping columns
-    let margin_id = input_domain.context.grouping_columns()?;
-    let public_info = (input_domain.frame_domain.margins.get(&margin_id))
-        .map(|m| m.public_info.clone())
-        .unwrap_or_default();
+    let margin = input_domain.active_margin()?;
```
mccalluc, line 47: Move down so the definition is right before the usage?

mccalluc (commented):
Maybe some more testing of edge cases? Since the tests have a parallel structure, could imagine factoring out a helper function, but leave that to your judgement. Kind of skimming the details of the implementation: Let me know if you'd like a closer look at anything in particular.

mccalluc (dismissed):
Thanks for the tweaks!

PR #2092, opened 2024-10-12
Polars: remove sort postprocessor

### LINKED ISSUE #1951: Polars: remove sort postprocessor
Since we shuffle dataframe collection to conceal the natural ordering of data, sort as a postprocessor becomes wasted compute. It's also misleading that this doesn't error.

It would be possible to skip the shuffle if the sort covers all grouping keys, but we should not prioritize building this functionality without a good use case.

Shoeboxam:
Close #1951


Shoeboxam:
* **#2092** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2092?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

mccalluc (commented):
From the issue description, I was thinking of this as an internal change, but it's also a change in how we suggest people use the library. Would it be good to add a test of the behavior if `sort()` is still included, and assert that the right error message comes up? Assuming it would be an error now: I'm not sure.

mccalluc (approved):
Thanks! Looks good.

PR #2093, opened 2024-10-12
Polars: update honest-but-curious note on OnceFrame.lazy

### LINKED ISSUE #1982: Polars: Adjust conditions for "honest-but-curious" on `OnceFrame.lazy`
OnceFrames ensure that the computation is only executed once, and that the computation is order-insensitive. 

The `lazy` method of a OnceFrame allows you to retrieve the underlying LazyFrame, provided that you opt-in to honest-but-curious. The conditions for this flag are stated essentially as, only execute once, and shuffle the output. This doesn't state that extending the compute plan with order-sensitive computations before shuffling is not allowed, but it should.

Shoeboxam:
Close #1982

### DIFF
--- a/python/src/opendp/extras/polars/__init__.py
+++ b/python/src/opendp/extras/polars/__init__.py
@@ -324,7 +324,7 @@ class OnceFrame(object):
         
         To ensure that row ordering is not observed:
         1. Do not extend the compute plan with order-sensitive computations.
-        2. Shuffle the output once collected.
+        2. Shuffle the output once collected `(in Polars sample all, with shuffle enabled) <https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.sample.html>`_.
         """
         from opendp._data import onceframe_lazy
 
--- a/python/src/opendp/extras/polars/__init__.py
+++ b/python/src/opendp/extras/polars/__init__.py
@@ -22,7 +22,6 @@ from opendp.mod import (
     Domain,
     Measurement,
     OpenDPException,
-    assert_features,
     binary_search,
     binary_search_chain,
 )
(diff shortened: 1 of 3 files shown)


Shoeboxam:
* **#2093** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2093?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

python/src/opendp/extras/polars/__init__.py
```diff
@@ -321,11 +320,14 @@ def lazy(self):
 
         Requires "honest-but-curious" because the privacy guarantees only apply if:
         1. The LazyFrame (compute plan) is only ever executed once.
-        2. The analyst does not observe ordering of rows in the output. To ensure this, shuffle the output.
+        2. The analyst does not observe ordering of rows in the output. 
+        
+        To ensure that row ordering is not observed:
+        1. Do not extend the compute plan with order-sensitive computations.
+        2. Shuffle the output once collected.
```
mccalluc, line 327:
```suggestion
        2. `Shuffle <https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.shuffle.html>`_ the output once collected.
```

python/src/opendp/extras/polars/__init__.py
```diff
@@ -321,11 +320,14 @@ def lazy(self):
...
         """
         from opendp._data import onceframe_lazy
 
-        assert_features("honest-but-curious")
```
mccalluc, line 328: Shouldn't this still be required? Or am I missing something?

mccalluc (changes requested):
Maybe link to polars docs, and keep the `assert_features`?

python/src/opendp/extras/polars/__init__.py
Shoeboxam: It's already asserted on onceframe_lazy

mccalluc: I'd have a slight preference to keep it. The cost is negligible, and if some moderately paranoid person is reviewing the code in the future, it will save them them trouble of confirming that the wrapped function has the assert.

mccalluc (approved):
I'd prefer a little redundancy in this case, but not a blocker.

PR #2089, opened 2024-10-11
Polars: error if string given as margin identifier

Shoeboxam:
Close #1977

PR #2096, opened 2024-10-13
Polars: add strptime

### LINKED ISSUE #1755: Polars: year, month, timezone, offsets, etc
Polars: year, month, timezone, offsets, etc
**Description**
- **Class**: Expr
- **DSL Variant:** Function: TemporalExpr
--- 

Shoeboxam:
Towards #1755

### DIFF
--- a/rust/src/domains/polars/datetime/mod.rs
+++ b/rust/src/domains/polars/datetime/mod.rs
@@ -1,0 +1,23 @@
+use polars::prelude::TimeUnit;
+
+use crate::{core::Domain, error::Fallible};
+
+/// A domain that represents a datetime.
+/// Number of milli/nano/micro seconds since Unix epoch.
+///
+/// Consider the set of all possible time points distinct for each setting of time_zone.
+/// Calculations that convert between time zones map between these disjoint sets of time points.
+#[derive(Debug, Clone, PartialEq)]
+pub struct DatetimeDomain {
+    pub time_unit: TimeUnit,
+    /// See https://docs.pola.rs/user-guide/transformations/time-series/timezones/
+    pub time_zone: Option<String>,
+}
+
+impl Domain for DatetimeDomain {
+    type Carrier = u64;
+
+    fn member(&self, _val: &Self::Carrier) -> Fallible<bool> {
+        Ok(true)
+    }
+}
--- a/rust/src/transformations/make_stable_expr/namespace_str/expr_strptime/mod.rs
+++ b/rust/src/transformations/make_stable_expr/namespace_str/expr_strptime/mod.rs
@@ -1,0 +1,115 @@
+use polars::prelude::*;
+use polars_plan::dsl::Expr;
+
+use crate::core::{Function, MetricSpace, StabilityMap, Transformation};
+use crate::domains::{ExprDomain, OuterMetric, SeriesDomain, WildExprDomain};
+use crate::error::*;
+use crate::polars::{literal_value_of, ExprFunction};
+use crate::transformations::DatasetMetric;
+
+use super::StableExpr;
+
+#[cfg(test)]
+mod test;
+
+/// Make a Transformation that returns a `strptime` expression.
+///
+/// # Arguments
+/// * `input_domain` - Expr domain
+/// * `input_metric` - The metric under which neighboring LazyFrames are compared
+/// * `expr` - The str.strptime expression
+pub fn make_expr_strptime<M: OuterMetric>(
+    input_domain: WildExprDomain,
+    input_metric: M,
+    expr: Expr,
+) -> Fallible<Transformation<WildExprDomain, ExprDomain, M, M>>
+where
+    M::InnerMetric: DatasetMetric,
+    M::Distance: Clone,
+    (WildExprDomain, M): MetricSpace,
+    (ExprDomain, M): MetricSpace,
+    Expr: StableExpr<M, M>,
+{
+    let Expr::Function {
+        input: inputs,
+        function: FunctionExpr::StringExpr(StringFunction::Strptime(to_type, mut strptime_options)),
+        ..
+    } = expr
+    else {
+        return fallible!(MakeTransformation, "expected str.strptime expression");
+    };
+
+    let Ok([input, ambiguous]) = <&[_; 2]>::try_from(inputs.as_slice()) else {
+        return fallible!(
+            MakeTransformation,
+            "str.strptime must have two arguments, found {}",
+            inputs.len()
+        );
+    };
+
+    let t_prior = input.clone().make_stable(input_domain, input_metric)?;
+    let (middle_domain, middle_metric) = t_prior.output_space();
+
+    if strptime_options.format.is_none() {
+        return fallible!(MakeTransformation, "format must be specified; otherwise Polars will attempt to infer the format from the data, resulting in an unstable transformation");
+    }
+
+    if matches!(to_type, DataType::Time) && !strptime_options.exact {
+        return fallible!(
+            MakeTransformation,
+            "non-exact not implemented for Time data type"
+        );
+    }
+
+    // Strict casting makes the transformation unstable: errors tell you things about private data.
+    // Could throw an error if strict, but it is the default, so for ease-of-use it is forced to be non-strict.
+    // It is also ok for overflow to wraparound (but may be undesirable for users).
+    strptime_options.strict = false;
+
+    // never raise on error
+    let ambiguous = lit(match literal_value_of::<String>(ambiguous) {
+        Ok(Some(a)) if ["earliest", "latest"].contains(&a.as_str()) => a,
+        _ => "null".to_string(),
+    });
+
+    let mut output_domain = middle_domain.clone();
+    let series_domain = &mut output_domain.column;
+    let name = series_domain.field.name.as_ref();
+
+    // check input and output types
+    if series_domain.field.dtype != DataType::String {
+        return fallible!(
+            MakeTransformation,
+            "str.strptime input dtype must be String, found {}",
+            series_domain.field.dtype
+        );
+    }
+
+    if !matches!(
+        to_type,
+        DataType::Time
+            | DataType::Datetime(TimeUnit::Microseconds | TimeUnit::Milliseconds, _)
+            | DataType::Date
+    ) {
+        return fallible!(
+            MakeTransformation,
+            "str.strptime output dtype must be Time, micro- or milli-second Datetime or Date, found {}",
+            to_type
+        );
+    }
+
+    *series_domain = SeriesDomain::new_from_field(Field::new(name, to_type.clone()))?;
+
+    t_prior
+        >> Transformation::new(
+            middle_domain.clone(),
+            output_domain,
+            Function::then_expr(move |expr| {
+                expr.str()
+                    .strptime(to_type.clone(), strptime_options.clone(), ambiguous.clone())
+            }),
+            middle_metric.clone(),
+            middle_metric,
+            StabilityMap::new(Clone::clone),
+        )?
+}
--- a/rust/src/transformations/make_stable_expr/namespace_str/expr_strptime/mod.rs
+++ b/rust/src/transformations/make_stable_expr/namespace_str/expr_strptime/mod.rs
@@ -85,6 +85,12 @@ where
         );
     }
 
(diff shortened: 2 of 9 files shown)


Shoeboxam:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/opendp/opendp/2096?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2096** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2096?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#2095** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2095?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2091** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2091?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

rust/src/domains/polars/datetime/mod.rs
```diff
@@ -0,0 +1,18 @@
+use polars::prelude::TimeUnit;
+
+use crate::{core::Domain, error::Fallible};
+
+/// A domain that represents a datetime.
+#[derive(Debug, Clone, PartialEq)]
+pub struct DatetimeDomain {
+    pub time_unit: TimeUnit,
+    pub time_zone: Option<String>,
```
mccalluc, line 9: This seems odd, but I'm not sure I understand the goal. Expected a composition of Date and Time. Are we expecting to get anything out of the timezone string? Should there be hours ahead or behind UTC?

rust/src/transformations/make_stable_expr/expr_str/expr_strptime/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+    ) {
+        return fallible!(
+            MakeTransformation,
+            "str.strptime dtype must be Time, Datetime or Date, found {}",
+            to_type
+        );
+    }
+    // Strict casting makes the transformation unstable: errors tell you things about private data.
+    // Could throw an error if strict, but it is the default, so for ease-of-use it is forced to be non-strict.
+    // It is also ok for overflow to wraparound.
```
mccalluc, line 61: What is the range of dates which we can represent? If dates earlier than 1970 wrap around, that is going to be undesirable in many cases.

rust/src/transformations/make_stable_expr/expr_str/expr_strptime/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+            "non-exact not implemented for Time data type"
+        );
+    }
+
+    // never raise on error
+    let ambiguous_str = literal_value_of(ambiguous)
+        .ok()
+        .flatten()
+        .unwrap_or("raise".to_string());
+    let ambiguous = if ["earliest".to_string(), "latest".to_string()].contains(&ambiguous_str) {
```
mccalluc, line 80: "earliest" and "latest" don't occur elsewhere in this PR. Are they coming from a library?

rust/src/transformations/make_stable_expr/expr_str/expr_strptime/test.rs
```diff
@@ -0,0 +1,87 @@
...
+                DataType::Date,
+                StrptimeOptions {
+                    format: Some("%Y-%m-%d".to_string()),
+                    strict: false,
+                    exact: false,
+                    cache: true,
+                },
+                lit("null"),
+            )
+            .alias("date"),
```
mccalluc, line 69: Do you need the alias if the name doesn't change?

rust/src/transformations/make_stable_expr/expr_str/expr_strptime/test.rs
```diff
@@ -0,0 +1,87 @@
...
+┌──────────┬─────────────────────┬────────────┐
+│ time     ┆ datetime            ┆ date       │
+│ ---      ┆ ---                 ┆ ---        │
+│ time     ┆ datetime[ns]        ┆ date       │
+╞══════════╪═════════════════════╪════════════╡
+│ 14:00:00 ┆ 2023-08-12 14:00:00 ┆ 2023-08-12 │
+│ 15:00:00 ┆ 2024-09-13 15:00:00 ┆ 2024-09-13 │
+│ 16:00:00 ┆ 2025-10-14 16:00:00 ┆ 2025-10-14 │
+│ 16:00:00 ┆ null                ┆ null       │
+└──────────┴─────────────────────┴────────────┘"#;
```
mccalluc, line 84: This kind of test is going to be fragile if the data changes... but it might be easier than constructing a dataframe to check against.

rust/src/transformations/make_stable_expr/expr_str/expr_strptime/test.rs
```diff
@@ -0,0 +1,87 @@
...
+            "2023-08-12 14:00:00".to_string(),
+            "2024-09-13 15:00:00".to_string(),
+            "2025-10-14 16:00:00".to_string(),
+            "2025-13-14 16:00:00".to_string()
+        ],
+        "date" => &[
+            "2023-08-12".to_string(),
+            "2024-09-13".to_string(),
+            "2025-10-14".to_string(),
+            "2025-13-14".to_string()
```
mccalluc, line 31: Should you test behavior with bad data? Ill-formated, day-of-month out of range, overflow, underflow.

mccalluc (commented):
Main concern is how ill-formed / out-of-range dates get handled.

rust/src/transformations/make_stable_expr/expr_str/expr_strptime/mod.rs
mccalluc: Never mind: found https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.str.strptime.html

rust/src/domains/polars/datetime/mod.rs
Shoeboxam: * Time (i64): number of nanoseconds since midnight
* Datetime (i64): number of milli/nano/micro seconds since Unix epoch. Time zone and time unit are encoded as part of the type/shared amongst all rows in a column
* Date (i32): number of days since Unix epoch.

Time and Date don't carry any extra type information/domain descriptors, so they can be represented via AtomDomain.
Datetime is separate as it isn't a composition of date and time, it is its own data format. It is also different in that it has type information (zone and unit) that needs to be known to interpret the meaning of the data. For the purposes of OpenDP, I consider the set of time points spanned by the domain to be distinct for each time zone. Updated the documentation.

rust/src/transformations/make_stable_expr/expr_str/expr_strptime/mod.rs
Shoeboxam: Dates and datetimes are signed offsets from the epoch, so dates earlier than 1970 don't wrap until much earlier time points.

Shoeboxam:
I got rid of the fixtures, rebased on top of the refactored ExprDomain, and updated the tests to edge cases per your suggestions @mccalluc. Thanks for the thoughtful review!

rust/src/transformations/make_stable_expr/namespace_str/expr_strptime/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    if !matches!(
+        to_type,
+        DataType::Time
+            | DataType::Datetime(TimeUnit::Microseconds | TimeUnit::Milliseconds, _)
+            | DataType::Date
+    ) {
+        return fallible!(
+            MakeTransformation,
+            "str.strptime output dtype must be Time, micro- or milli-second Datetime or Date, found {}",
```
mccalluc, line 96: The comment above also mentions nanoseconds. Add to this message?

rust/src/transformations/make_stable_expr/namespace_str/expr_strptime/test.rs
```diff
@@ -0,0 +1,162 @@
...
+        StrptimeOptions {
+            format: Some("%Y-%m-%dT%H:%M:%SZ".to_string()),
+            strict: false,
+            exact: true,
+            cache: true,
+        },
+        lit("null"),
+    ));
+
+    println!("{:?}", lf.collect());
```
mccalluc, line 94: Stricter test here?

rust/src/transformations/make_stable_expr/namespace_str/expr_strptime/test.rs
```diff
@@ -0,0 +1,162 @@
...
+    assert_eq!(format!("{:?}", observed), expected);
+    Ok(())
+}
+
+#[test]
+fn test_strptime_datetime() -> Fallible<()> {
+    let lf = df!("datetime" => &[
+        "+262142-01-01T00:00:00Z",
+        "+262143-01-01T00:00:00Z",
+        "+262144-01-01T00:00:00Z",
```
mccalluc, line 80: These are unusual years: Any reason to pick these particular three? I feel like it's trying to exercise something particular, but I'm not sure what.

rust/src/transformations/make_stable_expr/namespace_str/expr_strptime/mod.rs
mccalluc: Oh, the absence of nano is probably related to https://github.com/pola-rs/polars/issues/19928. Maybe file an issue in this repo to track that?

rust/src/transformations/make_stable_expr/namespace_str/expr_strptime/test.rs
```diff
@@ -0,0 +1,162 @@
...
+                format: Some("%Y-%m-%dT%H:%M:%S.fZ".to_string()),
+                strict: false,
+                exact: true,
+                cache: true,
+            },
+            lit("null"),
+        ));
+
+    let expected = DataFrame::new(vec![Series::new("t", vec![None::<NaiveTime>])])?;
+    assert_eq!(lf.collect()?, expected);
```
mccalluc, line 115: If I'm reading this correctly, there's no data in the expected frame. Is this because there's a problem with the decimal seconds? This behavior is really different from the integer seconds that follow. Consider splitting this into two tests, and either in the test name or a comment, explain why this one is different.

rust/src/transformations/make_stable_expr/namespace_str/mod.rs
```diff
@@ -0,0 +1,52 @@
...
+    match str_function {
+        #[cfg(feature = "contrib")]
+        StringFunction::Strptime(_, _) => expr_strptime::make_expr_strptime(input_domain, input_metric, expr),
+
+        expr => fallible!(
+            MakeTransformation,
+            "Expr is not recognized at this time: {:?}. {}If you would like to see this supported, please file an issue.",
+            expr,
+            get_disabled_features_message()
+        )
```
mccalluc, line 50: This error seems to be conflating two separate things. Maybe they just don't have `contrib` enabled? Or maybe even with `contrib`, it doesn't match?

Maybe have two separate `fallible!`s, one guarded with `#[cfg(feature = "contrib")]`, and one with the negation?

Or if it's better to keep it simple, one error message that makes it clear that there are two different possible root causes?

mccalluc (commented):
Thanks for the explanations. Reviewed from the top, and there are still some things that seem odd to me, but I might be confused.

Thinking about scope: I'm not sure if this would be exposed on the python side with this PR, or if there is follow-up work that will do that. If there isn't a follow-up, then it would be nice to have some tests on the Python side, even if we're holding off on docs until the feature set for this release is stabilized.

rust/src/transformations/make_stable_expr/namespace_str/expr_strptime/mod.rs
Shoeboxam: I left an error condition in the code.

rust/src/transformations/make_stable_expr/namespace_str/mod.rs
Shoeboxam: This is what the `get_disabled_features_message()` call does. If there are disabled features it interpolates that explanation into the error message.

mccalluc (approved):
Just looking at the latest changes, and the explanations are very helpful!

PR #2097, opened 2024-10-14
Polars: add date and time components

### LINKED ISSUE #1755: Polars: year, month, timezone, offsets, etc

Shoeboxam:
Close #1755

### DIFF
--- a/python/src/opendp/extras/polars/__init__.py
+++ b/python/src/opendp/extras/polars/__init__.py
@@ -73,6 +73,12 @@ class DPExpr(object):
           - Fill missing values with provided value
         * - `lit <https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.lit.html>`_
           - Return an expression representing a literal value
+        * - `str.strptime <https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.str.strptime.html>`_
+          - Parse a string into a date, time or datetime
+        * - `date components <https://docs.pola.rs/api/python/stable/reference/expressions/temporal.html>`_
+          - ``year``, ``iso_year``, ``quarter``, ``month``, ``week``, ``weekday``, ``day``, ``ordinal_day``
+        * - `time components <https://docs.pola.rs/api/python/stable/reference/expressions/temporal.html>`_
+          - ``hour``, ``minute``, ``second``, ``millisecond``, ``microsecond``, ``nanosecond``
 
     A few ``Expr`` aggregation methods are also available:
 
--- a/rust/src/transformations/make_stable_expr/namespace_dt/expr_datetime_component/mod.rs
+++ b/rust/src/transformations/make_stable_expr/namespace_dt/expr_datetime_component/mod.rs
@@ -1,0 +1,123 @@
+use polars::prelude::*;
+use polars_plan::dsl::Expr;
+use polars_plan::prelude::{ApplyOptions, FunctionOptions};
+
+use crate::core::{Function, MetricSpace, StabilityMap, Transformation};
+use crate::domains::{ExprDomain, OuterMetric, WildExprDomain};
+use crate::error::*;
+use crate::polars::ExprFunction;
+use crate::transformations::DatasetMetric;
+
+use super::StableExpr;
+
+#[cfg(test)]
+mod test;
+
+/// Make a Transformation that returns an expression that extracts a datetime component.
+///
+/// # Arguments
+/// * `input_domain` - Expr domain
+/// * `input_metric` - The metric under which neighboring LazyFrames are compared
+/// * `expr` - The datetime component expression
+pub fn make_expr_datetime_component<M: OuterMetric>(
+    input_domain: WildExprDomain,
+    input_metric: M,
+    expr: Expr,
+) -> Fallible<Transformation<WildExprDomain, ExprDomain, M, M>>
+where
+    M::InnerMetric: DatasetMetric,
+    M::Distance: Clone,
+    (WildExprDomain, M): MetricSpace,
+    (ExprDomain, M): MetricSpace,
+    Expr: StableExpr<M, M>,
+{
+    let Expr::Function {
+        input: inputs,
+        function: FunctionExpr::TemporalExpr(temporal_function),
+        ..
+    } = expr
+    else {
+        return fallible!(MakeTransformation, "expected datetime component expression");
+    };
+
+    let Some((to_dtype, _)) = match_datetime_component(&temporal_function) else {
+        return fallible!(
+            MakeTransformation,
+            "expected datetime component, found {:?}",
+            temporal_function
+        );
+    };
+
+    let Ok([input]) = <&[_; 1]>::try_from(inputs.as_slice()) else {
+        return fallible!(
+            MakeTransformation,
+            "{} must have one arguments, found {}",
+            temporal_function,
+            inputs.len()
+        );
+    };
+
+    let t_prior = input.clone().make_stable(input_domain, input_metric)?;
+    let (middle_domain, middle_metric) = t_prior.output_space();
+
+    let in_dtype = middle_domain.column.dtype();
+    if !matches!(
+        in_dtype,
+        DataType::Time | DataType::Datetime(_, _) | DataType::Date
+    ) {
+        return fallible!(
+            MakeTransformation,
+            "expected a temporal input type, got {}",
+            in_dtype
+        );
+    }
+
+    let mut output_domain = middle_domain.clone();
+    output_domain.column.set_dtype(to_dtype)?;
+
+    t_prior
+        >> Transformation::new(
+            middle_domain.clone(),
+            output_domain,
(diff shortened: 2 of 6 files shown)


python/src/opendp/extras/polars/__init__.py
```diff
@@ -67,6 +67,12 @@ class DPExpr(object):
           - Fill missing values with provided value
         * - `lit <https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.lit.html>`_
           - Return an expression representing a literal value 
+        * - `str.strptime <https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.str.strptime.html>`_
+          - Parse a string into a date, time or datetime
+        * - `date components <https://docs.pola.rs/api/python/stable/reference/expressions/temporal.html>`_
+          - year, iso_year, quarter, month, week, weekday, day, ordinal_day
+        * - `time components <https://docs.pola.rs/api/python/stable/reference/expressions/temporal.html>`_
+          - hour, minute, second, millisecond, microsecond, nanosecond
```
mccalluc, line 75: Double backticks around these names?

rust/src/transformations/make_stable_expr/namespace_dt/expr_datetime_component/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+
+#[cfg(test)]
+mod test;
+
+/// Make a Transformation that returns an expression that extracts a datetime component.
+///
+/// # Arguments
+/// * `input_domain` - Expr domain
+/// * `input_metric` - The metric under which neighboring LazyFrames are compared
+/// * `expr` - The dt expression
```
mccalluc, line 23:
```suggestion
/// * `expr` - The datetime expression
```

rust/src/transformations/make_stable_expr/namespace_dt/expr_datetime_component/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+    (ExprDomain, M): MetricSpace,
+    Expr: StableExpr<M, M>,
+{
+    let Expr::Function {
+        input: inputs,
+        function: FunctionExpr::TemporalExpr(temporal_function),
+        ..
+    } = expr
+    else {
+        return fallible!(MakeTransformation, "expected dt expression");
```
mccalluc, line 41:
```suggestion
        return fallible!(MakeTransformation, "expected datetime expression");
```

rust/src/transformations/make_stable_expr/namespace_dt/expr_datetime_component/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+        Quarter => (DataType::Int8, Some(4)),
+        Month => (DataType::Int8, Some(12)),
+        Week => (DataType::Int8, Some(53)),
+        WeekDay => (DataType::Int8, Some(7)),
+        Day => (DataType::Int8, Some(31)),
+        OrdinalDay => (DataType::Int16, Some(366)),
+        Hour => (DataType::Int8, Some(24)),
+        Minute => (DataType::Int8, Some(60)),
+        Second => (DataType::Int8, Some(60)),
+        Millisecond => (DataType::Int32, None),
```
mccalluc, line 127:
```suggestion
        Millisecond => (DataType::Int32, Some(1000)),
```
just continuing the pattern?

rust/src/transformations/make_stable_expr/namespace_dt/expr_datetime_component/test.rs
```diff
@@ -0,0 +1,33 @@
...
+
+use super::*;
+
+#[test]
+fn test_make_expr_components() -> Fallible<()> {
+    let (lf_domain, lf) = example_dt_lf()?;
+    let lf_plan = parse_example_dt_lf(lf.clone()).with_columns([
+        col("time").dt().second(),
+        col("datetime").dt().ordinal_day(),
+        col("date").dt().quarter(),
```
mccalluc, line 15:
```suggestion
        col("time").dt().second().alias("second"),
        col("datetime").dt().ordinal_day().alias("day"),
        col("date").dt().quarter().alias("quarter"),
```
so the headers on the table make sense?

rust/src/transformations/make_stable_expr/namespace_dt/expr_datetime_component/test.rs
```diff
@@ -0,0 +1,33 @@
...
+use crate::transformations::make_stable_expr::namespace_dt::test::{
+    example_dt_lf, parse_example_dt_lf,
+};
+use crate::transformations::make_stable_lazyframe;
+
+use super::*;
+
+#[test]
+fn test_make_expr_components() -> Fallible<()> {
+    let (lf_domain, lf) = example_dt_lf()?;
```
mccalluc, line 11: Slight preference to keep the example data in the same file, even at the cost of some redundancy. I want to be able to look at the test and see that it's doing the right thing, rather than making opaque assertions.

rust/src/transformations/make_stable_expr/namespace_dt/test.rs
```diff
@@ -0,0 +1,78 @@
...
+            "2024-09-13 15:00:45".to_string(),
+            "2025-10-14 16:00:56".to_string(),
+            "2025-13-14 16:00:39".to_string()
+        ],
+        "date" => &[
+            "2023-08-12".to_string(),
+            "2024-09-13".to_string(),
+            "2025-10-14".to_string(),
+            "2025-13-14".to_string()
+        ]
```
mccalluc, line 35: I don't think these four rows are making a test that's any stronger than just one line? I'm sort of confused about the times that are almost but not quite the same: Is that to exercise anything particular?

To avoid distracting the user, I'd suggest one really boring date: `2000-01-01 00:00:00`, and any others should try to hit particular edge cases or errors.

mccalluc (commented):
Main comment is just to simplify the test fixture (unless there's something special about this set of times and dates that I'm not seeing.)

Shoeboxam:
Rebased on top of the ExprDomain and SeriesDomain refactors, and comments addressed in the rebase, so new changes aren't in a separate commit.

rust/src/transformations/make_stable_expr/namespace_dt/expr_datetime_component/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+        IsoYear => (DataType::Int32, None),
+        Quarter => (DataType::Int8, Some(4)),
+        Month => (DataType::Int8, Some(12)),
+        Week => (DataType::Int8, Some(53)),
+        WeekDay => (DataType::Int8, Some(7)),
+        Day => (DataType::Int8, Some(31)),
+        OrdinalDay => (DataType::Int16, Some(366)),
+        Hour => (DataType::Int8, Some(24)),
+        Minute => (DataType::Int8, Some(60)),
+        Second => (DataType::Int8, Some(60)),
```
mccalluc, line 117: Looking at the next file, I think quarter is 1-indexed, while hour is 0-indexed. Should the values here for time units be one less?

rust/src/transformations/make_stable_expr/namespace_dt/expr_datetime_component/test.rs
```diff
@@ -0,0 +1,138 @@
...
+│ ---      ┆ ---                 ┆ ---        ┆ ---       ┆ ---          ┆ ---          │
+│ time     ┆ datetime[ms]        ┆ date       ┆ i8        ┆ i16          ┆ i8           │
+╞══════════╪═════════════════════╪════════════╪═══════════╪══════════════╪══════════════╡
+│ 00:01:02 ┆ 2000-01-02 03:04:05 ┆ 2000-01-02 ┆ 0         ┆ 2            ┆ 1            │
+└──────────┴─────────────────────┴────────────┴───────────┴──────────────┴──────────────┘"#;
+    assert_eq!(format!("{:?}", actual), expected);
+    Ok(())
+}
+
+fn test_temporal_op_schema<const L: usize>(df: DataFrame, ops: [TemporalFunction; L]) {
```
mccalluc, line 53: Slight preference to reserve `test_` for tests, and a helper like this might be 
```suggestion
fn assert_temporal_op_schema<const L: usize>(df: DataFrame, ops: [TemporalFunction; L]) {
```

Imagining something looking at the code for a little bit and at first wondering where the `#[test]` is.

mccalluc (approved):
There might be an innocuous off-by-one problem, and slight preference for a different name for the helper function, but nothing that needs to block if I'm confused.

rust/src/transformations/make_stable_expr/namespace_dt/expr_datetime_component/mod.rs
Shoeboxam: No, this tracks the number of unique values, not the max value.

PR #2098, opened 2024-10-16
feat: Polars: default to unknown dtype

Shoeboxam:
Close #2094

PR #2104, opened 2024-10-20
Polars: private grouping with explicit key sets

### LINKED ISSUE #1964: Polars: support grouping by explicit key sets
We support thresholding keys, and releasing key-sets in the clear. It would also be useful to support grouping by explicit key sets. 

This might be done by:
1. preparing a dummy dataset with the cartesian product of grouping keys in grouping columns, as well as noised default values
2. left join on the dummy dataset with the DP release
3. shadow noisy samples

The following does the same but as an honest-but-curious combinator:
```python
def make_protect_keys(meas, encodings):
    """makes the resulting key-set include the cartesian product of encodings over the given grouping keys"""

    # to satisfy honest but curious, input meas must compute noisy histogram with Laplace noise in polars with no postprocessing
    dp.assert_features("honest-but-curious")

    if meas.output_measure != dp.max_divergence(T=float):
        raise ValueError("key protection is only implemented for Laplace")

    # add the same amount of noise as the meas does
    m_lap = dp.m.make_laplace(
        dp.vector_domain(dp.atom_domain(T=int)),
        dp.l1_distance(T=int),
        scale=dp.summarize_polars_measurement(meas)["scale"][0],
    )

    zeros = [0] * np.prod([len(vals) for vals in encodings.values()])
    df_zeros = pl.DataFrame(
        dict(zip(encodings, tuple(combo))) for combo in product(*encodings.values())
    )

    def function(lf):
        df_left = df_zeros.with_columns(pl.Series("len", m_lap(zeros)))
        df_noisy = meas(lf).collect()
        return df_left.join(df_noisy, on=list(encodings.keys()), how="left").select(
            *encodings.keys(), pl.col.len_right.fill_null(pl.col.len).alias("len")
        )

    return dp.m.make_user_measurement(
        meas.input_domain, meas.input_metric, meas.output_measure, function, meas.map
    )
```

Shoeboxam:
Close #1964


Shoeboxam:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/opendp/opendp/2104?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2104** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2104?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#2098** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2098?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2097** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2097?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2096** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2096?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2095** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2095?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2091** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2091?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

docs/source/getting-started/tabular-data/grouping.ipynb
```diff
@@ -161,7 +161,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Protected Group Keys\n",
+    "## Stable Group Keys\n",
```
mccalluc, line 164: The paragraph which follows doesn't mention anything about stability, so I'm not really sure why that's the term we're using.

docs/source/getting-started/tabular-data/grouping.ipynb
```diff
@@ -9,9 +9,9 @@
     "The OpenDP Library allows you to compute statistics over grouped data.\n",
     "We'll examine three settings with progressively more public information:\n",
     "\n",
-    "- Protected group keys: `public_info=None`\n",
-    "- Public group keys: `public_info=\"keys\"`\n",
-    "- Public group lengths: `public_info=\"lengths\"`\n",
+    "- Stable group keys\n",
+    "- Explicit group keys\n",
+    "- Invariant group keys\n",
```
mccalluc, line 14: Is the `public_info` part no longer accurate or relevant? Or maybe it just doesn't belong in these titles? If someone is reading code, and trying to understand the semantics, I think it could be useful to highlight this part, but your call.

docs/source/getting-started/tabular-data/grouping.ipynb
```diff
@@ -387,7 +387,93 @@
...
+    "\n",
+    "query_age_ilostat.summarize()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This query works by conducting a left join where the left dataset is your explicit list of grouping keys and the right dataset contains the results of the grouped aggregation.\n",
+    "Grouping keys that don't exist in the original data are imputed as if the differentially private statistic were computed on an empty data partition."
```
mccalluc, line 469: Would we ever use the polars syntax for joins? Guessing that's not in the picture right now. It's not immediately obvious how the code above is a join. Consider pulling out an expression and assigning it to `left` or `right`, or adding comments in the code.

docs/source/getting-started/tabular-data/grouping.ipynb
```diff
@@ -387,7 +387,93 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Invariant Group Keys\n",
```
mccalluc, line 476: Paragraph below doesn't use the term "invariant", so it feels like there's a disconnect.

docs/source/getting-started/tabular-data/grouping.ipynb
```diff
@@ -614,9 +700,9 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Public Group Sizes\n",
+    "## Invariant Group Sizes\n",
     "\n",
-    "It is also possible to declare partition sizes as public information.\n",
+    "It is also possible to declare partition sizes as data invariants.\n",
```
mccalluc, line 705: FWIW, unless it just wasn't accurate, I think the former description was more clear.

python/src/opendp/extras/polars/__init__.py
```diff
@@ -864,6 +866,45 @@ def with_columns_seq(  # type: ignore[empty-body]
             """
             ...
 
+        def join(  # type: ignore[empty-body]
```
mccalluc, line 869: This seems like a big addition, but it isn't mentioned in the docs in this PR. Maybe in a follow-up?

python/src/opendp/extras/polars/__init__.py
```diff
@@ -864,6 +866,45 @@ def with_columns_seq(  # type: ignore[empty-body]
...
+            ...
+
+        def with_keys(
+            self,
+            keys: _LazyFrame,
+            on: list[str] | None = None,
+        ) -> LazyFrameQuery:
+            """
+            Shorthand to join with an explicit key-set.
+            """
```
mccalluc, line 896: `with_keys` is in the notebook, once. Consider using the full version with `join` first, and then demonstrating the shortcut?

mccalluc: ... but unless you think this is really a big improvement, I wonder if it would be better just not to add shortcuts that diverge farther from the out-of-the-box polars syntax? Constraining users to a polars-like syntax may not be a bad thing.

python/test/test_polars.py
```diff
@@ -1,8 +1,8 @@
 import pytest
 import opendp.prelude as dp
-import os
 import warnings
 import re
+import os
```
mccalluc, line 5: It hasn't felt like a priority for me, but if we wanted cleaner diffs, we could use isort, perhaps as a precommit hook. I won't file an issue right now, but feel free.

python/test/test_polars.py
```diff
@@ -681,16 +680,16 @@ def test_pickle_bomb():
     pl.LazyFrame.deserialize(io.BytesIO(ser_lf))
 
     # OpenDP explicitly rejects any pickled data it finds
-    err_msg_re = re.escape("OpenDP does not allow pickled keyword arguments as they may enable remote code execution.")
-    with pytest.raises(dp.OpenDPException, match=err_msg_re):
+    err_msg = "OpenDP does not allow pickled keyword arguments as they may enable remote code execution."
+    with pytest.raises(dp.OpenDPException, match=err_msg):
```
mccalluc, line 684: I would kind of like to keep `re` in the name. I've forgotten that it's not a string match and been a little confused debugging a test.

python/test/test_tests.py
```diff
@@ -18,7 +18,6 @@ def test_TODOs():
     assert str(dp.smoothed_max_divergence()) == 'SmoothedMaxDivergence'
 
     # TODO: Add meaningful tests of partial constructors:
-    dp.m.then_private_expr(None, None) # type: ignore[arg-type]
```
mccalluc, line 21: Maybe there's a change to the generated code and this method is no longer available? Otherwise, this test seems like it could stay around.

rust/src/accuracy/polars/mod.rs
```diff
@@ -97,65 +97,87 @@ pub fn summarize_lazyframe(lazyframe: &LazyFrame, alpha: Option<f64>) -> Fallibl
...
+    if let Some(MatchGroupBy {
+        aggs: exprs,
+        key_sanitizer,
+        ..
+    }) = match_group_by(logical_plan.clone())?
+    {
+        let threshold = if let Some(KeySanitizer::Filter(predicate)) = key_sanitizer {
+            Some(
+                is_threshold_predicate(predicate.clone())
+                    .ok_or_else(|| err!(FailedFunction, "predicate is not a valid filter"))?,
```
mccalluc, line 110: Include the predicate in the error message?

mccalluc (commented):
It feels like more changes to the docs might be good to keep up with the library, but this is subjective. I also just skimmed the Rust code and should give it a closer look tomorrow. I'm going to mark this as a comment, but will request re-review so that I look at it again soon, even if doc changes end up in a separate PR.

docs/source/getting-started/tabular-data/grouping.ipynb
Shoeboxam: It is still accurate and relevant, but I wanted to change how this was communicated for two reasons. 

1. I also feel that I should more strongly caution against using "public_info" and more clearly associate it with what is generally considered to be a more negative word "invariant".
2. Now that we have more options for privately releasing group keys (stable under zCDP and explicit), public_info is only specified when working with the third category (invariants).

docs/source/getting-started/tabular-data/grouping.ipynb
Shoeboxam: Added an example using a right join to more thoroughly illustrate the concept.

python/src/opendp/extras/polars/__init__.py
Shoeboxam: It's a join specifically for privatization, not for microdata. Although I did add an example in the grouping notebook.

python/src/opendp/extras/polars/__init__.py
Shoeboxam: I was curious how you'd feel about this! Some considerations:
1. Writing a left join is more difficult in the context API: see the complexity of this implementation, where you have to go under the hood. This gives an easier shorthand to write a left join.
2. Left joins are more likely to be supported by database backends.
3. I wanted this to be easy to use... with the Polars API the key set needs to be lazy, the user has to specify they want a right join and they need to specify join keys. 

Added an example with explicit join to the notebook.

mccalluc: Sounds good. Maybe record these motivations in a comment?

python/test/test_polars.py
```diff
@@ -190,14 +190,21 @@ def test_stable_expr():
...
+        dp.len(scale=1.0)
+    )
+
+    e_plan = m_len(pl.LazyFrame(dict()))
+
+    pl_testing.assert_frame_equal(e_plan.plan, pl.LazyFrame(dict()))
+
+    print(e_plan.fill)
+    assert bool(re.match("len().*:noise_plugin()", str(e_plan.expr)))
+    assert bool(re.match("0.*:noise_plugin()", str(e_plan.fill)))
```
mccalluc, line 207: Pretty sure the `bool` isn't needed: `assert` is happy with truth-y.

mccalluc (commented):
A couple comments, but not done. Will put it back on my plate for a full review.

python/src/opendp/extras/polars/__init__.py
Shoeboxam: Added

mccalluc (dismissed):
Latest changes look good. Thanks!

Issue #2069, opened 2024-10-04
Stabilize `sample_bernoulli_exp`


Shoeboxam:
Remove \contrib in `sample_bernoulli_exp.tex`

PR #2115, opened 2024-10-22
review: sample_discrete_gaussian

Shoeboxam:
Towards #2069


Shoeboxam:
* **#2115** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2115?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

ChristianLebeda (changes requested):
I found two issues in the proof. Both are easy to fix. 

In Lemma 2.3, $\sigma$ is sometimes used in place of $t$. This should be updated since $\tau \neq 1/\sigma$ in [CKS20]. I marked all incorrect occurrences of $\sigma$ in red in the attached PDF.

There are some formatting issues in Theorem 2.4. Brackets are missing around the exponents in the $\LaTeX$ source code. Specifically, `e^-\frac{[val1]}{[val2]}` should be replaced by `e^{-\frac{[val1]}{[val2]}}`. As an alternative you could use `\exp\left(-\frac{[val1]}{[val2]}\right)` here.

[lebeda_review_sample_discrete_gaussian.pdf](https://github.com/user-attachments/files/17936887/lebeda_review_sample_discrete_gaussian.pdf)

Shoeboxam:
Thanks @ChristianLebeda! Good eye on the exponent as well.

annabelle400:
The mathematical properties are established in the proof. They are a little tricky to connect to the actual code. Could these be inserted with comments? 

Do we also need to know the preconditions to sample_discrete-Laplace and sample_bernouilli_exp  function calls? Are these explained in Lem2.2-2.3?

Shoeboxam:
@annabelle400 I made a number of revisions to clean up the formatting of the proof and some minor changes to try to make the code easier to map to the proof.

There are links in the proof document that you can click through to get the proof definition (precondition/postcondition) for these functions. I'm trying to lean on the linking instead of duplicating the definition again in each proof that uses it, to ease maintenance. As of yesterday's 0.12 library release, the links are not working, I have a fix in #2212 for 0.12.1.

annabelle400:
Hi Michael,

Thanks for this! I added review comments in a GitHub comment. Maybe I put it in the wrong place? Let me know and l’ll try to submit in the right place!

All the best,
Annabelle

On 4 Mar 2025, at 3:15 pm, Michael Shoemate ***@***.***> wrote:

﻿

@Shoeboxam<https://github.com/Shoeboxam> requested your review on: #2115<https://github.com/opendp/opendp/pull/2115> review: sample_discrete_gaussian.

Shoeboxam:
@annabelle400 When you click on files changed, do you see a green "Finish your review" button on the right-hand side? You may still need to submit your review.

annabelle400:
Hmm, I can’t really see where you mean. But maybe the request is to check recent changes? 

Annabelle

Shoeboxam:
@annabelle400: When you write comments to a pull request like this one, they won't actually be visible to anyone else until you submit a review. Until then, those comments remain "pending". To submit a review, click on "Files Changed", and then on the green button with wording like "Submit review".

annabelle400 (approved):
I think the proof is fine now.

PR #2119, opened 2024-10-22
review: sample_uniform_ubig_below

### LINKED ISSUE #2070: Stabilize `sample_uniform_ubig_below`
Remove \contrib in `sample_uniform_ubig_below.tex`

Shoeboxam:
Closes #2070

### DIFF
--- a/rust/src/traits/samplers/uniform/mod.rs
+++ b/rust/src/traits/samplers/uniform/mod.rs
@@ -63,7 +63,7 @@ pub fn sample_uniform_uint_below<T: Integer + Unsigned + FromBytes<N>, const N: 
 /// Sample an integer uniformly from `[0, upper)`
 ///
 /// # Proof Definition
-/// For any non-negative setting of `upper`,
+/// For any non-zero `upper`,
 /// return either `Err(e)` if there is insufficient system entropy,
 /// or `Some(sample)`, where `sample` is uniformly distributed over `[0, upper)`.
 pub fn sample_uniform_ubig_below(upper: UBig) -> Fallible<UBig> {
--- a/rust/src/traits/samplers/uniform/mod.rs
+++ b/rust/src/traits/samplers/uniform/mod.rs
@@ -63,7 +63,7 @@ pub fn sample_uniform_uint_below<T: Integer + Unsigned + FromBytes<N>, const N: 
 /// Sample an integer uniformly from `[0, upper)`
 ///
 /// # Proof Definition
-/// For any non-zero `upper`,
+/// For any positive setting of `upper`,
 /// return either `Err(e)` if there is insufficient system entropy,
 /// or `Some(sample)`, where `sample` is uniformly distributed over `[0, upper)`.
 pub fn sample_uniform_ubig_below(upper: UBig) -> Fallible<UBig> {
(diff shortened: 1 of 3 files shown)


Shoeboxam:
* **#2119** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2119?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

rust/src/traits/samplers/uniform/sample_uniform_ubig_below.tex
```diff
@@ -1,72 +1,60 @@
...
+For any setting of the input parameter \texttt{upper}, 
+\texttt{sample\_uniform\_ubig\_below} either 
+\begin{itemize} 
+    \item raises an exception if there is a lack of system entropy, 
+    \item returns \texttt{out} where \texttt{out} is uniformly distributed between $[0, upper)$. 
+\end{itemize} 
+ 
+\section{Proof} 
+\begin{proof} 
+\texttt{byte\_len} is the fewest number of bytes necessary to represent \texttt{upper},
```
tudorcebere, line 38: Hi @Shoeboxam, this looks great. A few comments on the clarity.

`bit_len` is the fewest bytes necessary to represent `upper`. Ideally, we would want to uniformly sample bit_len bits and do rejection sampling over that by comparing with `upper`. However, we don't have a method to draw random bits, and we can only do multiples of 8 bits. What the algorithm does is to pull the smallest multiple of 8 bigger than `bit_len`, fill the bits with random numbers and essentially throw away the padding pits added to satisfy the multiple of 8. Then we do rejection sampling as intended.

A nice note is that it does not waste random bits when `upper` is a power of 2. 

My main confusion was why we need all of this and not sample directly `bit_len`. It was that we have efficient sampling at a byte level and need to throw the difference away.

Overall it looks good, besides this minor confusion, it looks great!

Shoeboxam: Looking back at this, I responded in the diffs but not the comments. I had updated the writeup to make the byte-alignment point more explicit: we only have efficient byte-level sampling, so we sample the smallest byte-aligned superset of the intended bit range and then do rejection sampling against upper.

Shoeboxam:
I adjusted this implementation to stay closer in structure to the analogous cleanup in [#2120](https://github.com/opendp/opendp/pull/2120). The main substantive change is that the rejection threshold is now expressed in terms of the full sampled range, which avoids unnecessary rejection and wasted entropy.

rust/src/traits/samplers/uniform/mod.rs
```diff
@@ -63,17 +63,18 @@ pub fn sample_uniform_uint_below<T: Integer + Unsigned + FromBytes<N>, const N:
 /// Sample an integer uniformly from `[0, upper)`
 ///
 /// # Proof Definition
-/// For any non-negative setting of `upper`,
+/// For any positive setting of `upper`,
 /// return either `Err(e)` if there is insufficient system entropy,
 /// or `Some(sample)`, where `sample` is uniformly distributed over `[0, upper)`.
 pub fn sample_uniform_ubig_below(upper: UBig) -> Fallible<UBig> {
     // ceil(ceil(log_2(upper)) / 8)
     let byte_len = upper.bit_len().div_ceil(8);
```
zachratliff22: Looking at the [Dashu's`bit_len` implementation](https://docs.rs/dashu-base/0.4.3/src/dashu_base/bit.rs.html#32-46) the bit length of a positive number is described as “_the floored base-2 logarithm of the number plus one._” This differs from `ceil(ceil(log_2(upper)) / 8)` stated in the proof. In particular, at byte boundaries such as `upper = 256`, the implementation uses one more byte than the proof states. Can we correct the proof to match the implementation?

zachratliff, line 71: @Shoeboxam the proof looks great! The main (minor) change is that the implementation seems to differ slightly from what is described in the PDF. Looking at the [Dashu's`bit_len` implementation](https://docs.rs/dashu-base/0.4.3/src/dashu_base/bit.rs.html#32-46) the bit length of a positive number is described as “_the floored base-2 logarithm of the number plus one._” i.e. `floor(log_2(upper)) + 1`. This differs from `ceil(ceil(log_2(upper)) / 8)` stated in the proof (and in the code comment). In particular, at byte boundaries such as `upper = 256`, the implementation uses one more byte than the proof states.

zachratliff (changes requested):

tholop:
Thank you for the review, @zachratliff! @Shoeboxam, it sounds like there is a minor fix to do before we can merge.

PR #2120, opened 2024-10-22
review: sample_uniform_uint_below

### LINKED ISSUE #2071: Stabilize `sample_uniform_uint_below`
Remove \contrib from `sample_uniform_uint_below.tex`

Shoeboxam:
Closes #2071

### DIFF
--- a/rust/src/traits/samplers/uniform/sample_uniform_uint_below.tex
+++ b/rust/src/traits/samplers/uniform/sample_uniform_uint_below.tex
@@ -1,70 +1,64 @@
-\documentclass{article}
-\input{../../../lib.sty}
-
-\title{\texttt{fn sample\_uniform\_uint\_below}}
-\author{\MichaelShoemateAuthor}
-
-\begin{document}
-\maketitle
-
-\contrib
-
-This document proves that the implementation of \rustdoc{traits/samplers/uniform/fn}{sample\_uniform\_uint\_below} in \asOfCommit{mod.rs}{f5bb719} 
-satisfies its definition.
-
-\texttt{sample\_uniform\_uint\_below} uses rejection sampling. 
-In each round all bits of the integer are filled randomly, drawing an unsigned integer uniformly at random.
-The algorithm returns the sample, modulo the upper bound, so long as the sample is not one of the final "div" largest integers.
-
-\subsection*{PR History}
-\begin{itemize}
-    \item \vettingPR{473}
-\end{itemize}
-
-\subsection{Hoare Triple}
-\subsection*{Precondition}
-\subsubsection*{Compiler-verified}
-\begin{itemize}
-    \item \texttt{upper} must be of type \texttt{T}.
-    \item \texttt{T} has traits \texttt{Integer + Unsigned + FromBytes<N>}, which narrows valid types to u16, u32, u64, u128, usize.
-    \item By the definition of \texttt{FromBytes}, \texttt{N} is the number of bytes in \texttt{T}.
-\end{itemize}
-
-\subsubsection*{Caller-verified}
-None
-
-\subsection*{Pseudocode}
-
-\lstinputlisting[language=Python,firstline=2]{./pseudocode/sample_uniform_uint_below.py}
-
-\subsection*{Postcondition}
-\begin{theorem}
-For any setting of the input parameter \texttt{upper},
-\texttt{sample\_uniform\_uint\_below} either
-\begin{itemize}
-    \item raises an exception if there is a lack of system entropy,
-    \item returns \texttt{out} where \texttt{out} is uniformly distributed between $[0, upper)$.
-\end{itemize}
-\end{theorem}
-\begin{proof} 
-\label{unsigned-integer-proof}
-By the postcondition of \rustdoc{traits/samplers/uniform/fn}{sample\_from\_uniform\_bytes},
-then \texttt{sample} is a sample between zero and \texttt{T.MAX} inclusive, the greatest representable number of type \texttt{T}.
-
-You could sample one of \texttt{upper} values uniformly at random by rejecting \texttt{sample} if it is larger than \texttt{upper}.
-That is, only return \texttt{sample} if \texttt{sample} is less than \texttt{upper}.
-
-It is equivalent to extend the acceptance region, 
-by returning \texttt{sample \% 2} if \texttt{sample} is less than \texttt{sample * 2}, 
-so long as \texttt{sample * 2 <= T.MAX}.
-This reduces the rejection rate, which increases algorithm performance.
-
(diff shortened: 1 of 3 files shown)


Shoeboxam:
* **#2120** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2120?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

rust/src/traits/samplers/uniform/sample_uniform_uint_below.tex
```diff
@@ -1,73 +1,64 @@
...
+\label{unsigned-integer-proof} 
+By the postcondition of \rustdoc{traits/samplers/uniform/fn}{sample\_from\_uniform\_bytes}, 
+then \texttt{sample} is a sample between zero and \texttt{T.MAX} inclusive, the greatest representable number of type \texttt{T}. 
+ 
+You could sample one of \texttt{upper} values uniformly at random by rejecting \texttt{sample} if it is larger than \texttt{upper}. 
+That is, only return \texttt{sample} if \texttt{sample} is less than \texttt{upper}. 
+ 
+It is equivalent to extend the acceptance region,  
+by returning \texttt{sample \% 2} if \texttt{sample} is less than \texttt{sample * 2},  
+so long as \texttt{sample * 2 <= T.MAX}.
```
ChristianLebeda, line 53: There are some typos in this statement. I think the intended meaning was: 
by returning \texttt{sample \% upper} if \texttt{sample} is less than \texttt{upper * 2},  
so long as \texttt{upper * 2 <= T.MAX}.

rust/src/traits/samplers/uniform/sample_uniform_uint_below.tex
```diff
@@ -1,73 +1,64 @@
...
+    \item returns \texttt{out} where \texttt{out} is uniformly distributed between $[0, upper)$. 
+\end{itemize} 
+ 
+\subsection{Proof} 
+\begin{proof}  
+\label{unsigned-integer-proof} 
+By the postcondition of \rustdoc{traits/samplers/uniform/fn}{sample\_from\_uniform\_bytes}, 
+then \texttt{sample} is a sample between zero and \texttt{T.MAX} inclusive, the greatest representable number of type \texttt{T}. 
+ 
+You could sample one of \texttt{upper} values uniformly at random by rejecting \texttt{sample} if it is larger than \texttt{upper}.
```
ChristianLebeda, line 48: I suggest updating this to:
"if it is larger than or equal to \texttt{upper}."

ChristianLebeda (changes requested):
I spotted a few issues. All are easy to fix.

I highlighted two sentences that can be rephrased. 

As a minor note, the style of the PDF differs from other proofs because the Hoare Triple and Proof are subsections in the latex source instead of sections. This prints labels 0.1 and 0.2 instead of 1 and 2.

I believe there is a bug in the implementation. The check on this line should be strictly less than rather than "<=" - https://github.com/opendp/opendp/blob/f5bb719bc24aee4f9abad41b43a6005de1e9690e/rust/src/traits/samplers/uniform/mod.rs#L146
This bug gives a slightly higher probability to sampling 0.

Shoeboxam:
@ChristianLebeda Thanks for the review!

The sampler was not biased before this change. In the old version, the accepted samples still formed a set whose size was a multiple of upper, so reducing modulo upper gave each output in [0, upper) the same number of accepted preimages. The issue was efficiency, not correctness: the threshold was based on MAX instead of the full draw space size MAX + 1, so it rejected some values unnecessarily and wasted a small amount of entropy. For wide integer types like u64 this is negligible, but it is more noticeable for small types like u8.

I originally wrote it this way because it was simple to reason about, and directly working with the full draw space MAX + 1 can overflow for fixed-width integer types. I’ve now adjusted the sampler to equivalently use front-rejection, which avoids overflow and removes the unnecessary rejection.

I also added the explicit precondition upper > 0, since zero was never a valid input and previously could panic.

For the other two comments, I updated “larger than upper” to “larger than or equal to upper”, and I fixed the proof typo from sample % 2 / sample * 2 to sample % upper / upper * 2.

ChristianLebeda (dismissed):
All concerns resolved.

ChristianLebeda:
Thank you for the explanation and for updating the proof and code @Shoeboxam 

The proof looks good to me now so I changed my review to approve.

I like the trick with front-rejection, it's very clean.

zachratliff (approved):
The proof looks great to me! I think there's just one minor typo that appears throughout the proof/code that caused me some confusion. It seems that we you want $N$ to be the number of bytes in $\texttt{T}$ but then you repeatedly use $2^N$ as the number of representable values in $\texttt{T}$.

rust/src/traits/samplers/uniform/sample_uniform_uint_below.tex
```diff
@@ -1,70 +1,57 @@
...
-\subsection*{Postcondition}
-\begin{theorem}
-For any setting of the input parameter \texttt{upper},
-\texttt{sample\_uniform\_uint\_below} either
-\begin{itemize}
-    \item raises an exception if there is a lack of system entropy,
-    \item returns \texttt{out} where \texttt{out} is uniformly distributed between $[0, upper)$.
-\end{itemize}
-\end{theorem}
+In each round all bits of the integer are filled randomly, drawing an unsigned integer uniformly at random from a space of size $2^N$.
```
zachratliff, line 14: $2^{8N}$? $N$ is defined as the number of bytes in $\texttt{T}$

rust/src/traits/samplers/uniform/sample_uniform_uint_below.tex
```diff
@@ -1,70 +1,57 @@
...
-Therefore conditioning \texttt{sample} on being less than \texttt{T.MAX - T.MAX \% upper} 
-results in \texttt{sample \% upper} being an unbiased, uniformly distributed sample.
-
-\noindent Therefore, for any value of \texttt{upper}, the function satisfies the postcondition.
-\end{proof}
-
-\end{document}
+\label{unsigned-integer-proof} 
+By the postcondition of \rustdoc{traits/samplers/uniform/fn}{sample\_from\_uniform\_bytes}, 
+\texttt{sample} is drawn uniformly from the $2^N$ representable values of type \texttt{T}, namely $\{0, \ldots, \texttt{T.MAX}\}$.
```
zachratliff, line 45: Since $N$ is defined as the number of _bytes_ in the generic type $\texttt{T}$, shouldn't all instances of $2^N$ be replaced with $2^{8N}$?

tholop:
Great, thanks @zachratliff. @Shoeboxam there is also a small tweak to do before we can merge.

PR #2124, opened 2024-10-22
review: conservative_discrete_laplacian_tail_to_alpha

### LINKED ISSUE #2061: Stabilize `conservative_discrete_laplacian_tail_to_alpha`
Remove \contrib in `conservative_discrete_laplacian_tail_to_alpha.tex`

Shoeboxam:
Close #2061

### DIFF
--- a/rust/src/accuracy/tail_bounds/conservative_discrete_laplacian_tail_to_alpha.tex
+++ b/rust/src/accuracy/tail_bounds/conservative_discrete_laplacian_tail_to_alpha.tex
@@ -1,41 +1,58 @@
-\documentclass{article}
-\input{../../lib.sty}
-
-\title{\texttt{fn conservative\_discrete\_laplacian\_tail\_to\_alpha}}
-\author{\MichaelShoemateAuthor}
-
-\begin{document}
-\maketitle
-
-\contrib
-
-Proof for \rustdoc{accuracy/tail\_bounds/fn}{conservative\_discrete\_laplacian\_tail\_to\_alpha}.
-
-\begin{definition}
-    Define $Y \sim \mathcal{L}_\mathbb{Z}(0, s)$, a random variable following the discrete laplace distribution:
-    \begin{equation}
-        \forall x \in \mathbb{Z} \qquad \Pr[X = x] = \frac{e^{1/s} - 1}{e^{1/s} + 1} e^{-|x|/s}
-    \end{equation}
-\end{definition}
-
-\begin{theorem}
-    Assume $X \sim \mathcal{L}_\mathbb{Z}(0, s)$, and $t > 0$.
-    \begin{equation}
-        \alpha = P[X > t] = \frac{e^{-t/s}}{e^{-1/s} + 1}
-    \end{equation}
-\end{theorem}
-
-\begin{proof}
-    \begin{align*}
-        \alpha &= P[X > t] \\
-        &= \sum_{x=t + 1}^{\infty} \frac{e^{1/s} - 1}{e^{1/s} + 1} e^{-|x|/s} \\
-        &= \frac{e^{1/s} - 1}{e^{1/s} + 1} \sum_{x=t + 1}^{\infty} e^{-x/s} 
-            && \text{since } t > 0 \\
-        &= \frac{e^{1/s} - 1}{e^{1/s} + 1} \frac{e^{(1-(t + 1))/s}}{e^{1/s} - 1} 
-            && \text{since } \sum_{x=t}^\infty p^x = \frac{p^t}{1 - p} \text{ if } |p| < 1 \text{, let } p = e^{1/s}\\
-        &= \frac{e^{-t/s}}{e^{1/s} + 1}
-    \end{align*}
-
-\end{proof}
-
+\documentclass{article} 
+\input{../../lib.sty} 
+ 
+\title{\texttt{fn conservative\_discrete\_laplacian\_tail\_to\_alpha}} 
+\author{\MichaelShoemateAuthor} 
+ 
+\begin{document} 
+\maketitle 
+ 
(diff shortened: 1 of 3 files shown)


Shoeboxam:
* **#2124** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2124?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

annabelle400 (commented):
In the proof of the theorem, isn't p = 1/e^{1/s} ?

The statement of the theorem doesn't seem to match the last line of the proof?

I think with the above value for p, the proof does yield the last line.

Can you say what scale and tail are in the theorem? I assume s, t ?

I am not sure now whether the code matches the last line of the proof, or the statement of the theorem on which it depends. Please check and clarify.

rust/src/accuracy/tail_bounds/conservative_discrete_laplacian_tail_to_alpha.tex
```diff
@@ -1,41 +1,60 @@
...
-\maketitle
-
-\contrib
-
-Proof for \rustdoc{accuracy/tail\_bounds/fn}{conservative\_discrete\_laplacian\_tail\_to\_alpha}.
-
-\begin{definition}
-    Define $Y \sim \mathcal{L}_\mathbb{Z}(0, s)$, a random variable following the discrete laplace distribution:
-    \begin{equation}
-        \forall x \in \mathbb{Z} \qquad \Pr[X = x] = \frac{e^{1/s} - 1}{e^{1/s} + 1} e^{-|x|/s}
```
tudorcebere, line 17: nit: In the Definition block, the random variable is defined as Y, but the pmf directly below it uses X (Pr[X=x]=…). These should be made consistent.

rust/src/accuracy/tail_bounds/conservative_discrete_laplacian_tail_to_alpha.tex
```diff
@@ -1,41 +1,60 @@
...
-    \end{equation}
-\end{definition}
-
-\begin{theorem}
-    Assume $X \sim \mathcal{L}_\mathbb{Z}(0, s)$, and $t > 0$.
-    \begin{equation}
-        \alpha = P[X > t] = \frac{e^{-t/s}}{e^{-1/s} + 1}
```
tudorcebere, line 24: The exponent in the denominator contains a wrong negative sign, should be exp(1/s) + 1.

rust/src/accuracy/tail_bounds/conservative_discrete_laplacian_tail_to_alpha.tex
```diff
@@ -1,41 +1,60 @@
...
-\end{theorem}
-
-\begin{proof}
-    \begin{align*}
-        \alpha &= P[X > t] \\
-        &= \sum_{x=t + 1}^{\infty} \frac{e^{1/s} - 1}{e^{1/s} + 1} e^{-|x|/s} \\
-        &= \frac{e^{1/s} - 1}{e^{1/s} + 1} \sum_{x=t + 1}^{\infty} e^{-x/s} 
-            && \text{since } t > 0 \\
-        &= \frac{e^{1/s} - 1}{e^{1/s} + 1} \frac{e^{(1-(t + 1))/s}}{e^{1/s} - 1} 
-            && \text{since } \sum_{x=t}^\infty p^x = \frac{p^t}{1 - p} \text{ if } |p| < 1 \text{, let } p = e^{1/s}\\
```
tudorcebere, line 35: Here we have a bug (not problematic): you state that |p| < 1and you let p = exp(1/s), with s>0. I think you want to make it p = exp(-1/s), no?

rust/src/accuracy/tail_bounds/pseudocode/conservative_discrete_laplacian_tail_to_alpha.py
```diff
@@ -0,0 +1,6 @@
+# type: ignore
+def conservative_discrete_laplacian_tail_to_alpha(scale: f64, tail: u32) -> f64:
+    t = f64.neg_inf_cast(tail)
+    numer = t.neg_inf_div(-scale).inf_exp()
```
tudorcebere: Is this actually correct?

So neg_inf_div rounds towards -inf if I understand correctly, giving us a lower bounds on -t/s. When we exponentiate this, aren't we underestimating exp(-t/s)?

tudorcebere (changes requested):
The proof looks great! I would be tempted to make it more similar to the Gaussian one, where there are some explanations provided and a bit of context. I am happy to give it a try to add those. I also left a few comments (the floating points one might be wrong)

Shoeboxam:
Thanks for all the comments. I fixed the proof and normalized it to the newer template style from #2699. 

* corrected the pmf normalization sign
* corrected the theorem statement for P[X > t]
* clarified that s = scale and t = tail
* fixed the geometric-series substitution to use p = e^{-1/s}

On the rounding question: Tudor’s concern was valid for the pseudocode, but not for the Rust implementation itself. The implementation was already using the right conservative directions; the pseudocode was too indirect and made the numerator look downward-rounded. I rewrote it to directly mirror the implementation, as it should be.

I think this was likely a partial-update from an earlier iteration: parts of the proof seem to match an older version of the function. This would explain why the pseudocode no longer mirrored how the implementation actually rounded.

annabelle400:
Dear Michael,

Thanks for these! I am travelling until early May. Would it be ok to look at these when I get back?

Thanks!
Annabelle

On 22 Apr 2026, at 4:58 am, Michael Shoemate ***@***.***> wrote:

﻿

@Shoeboxam<https://github.com/Shoeboxam> requested your review on: opendp/opendp#2124<https://github.com/opendp/opendp/pull/2124> review: conservative_discrete_laplacian_tail_to_alpha.

Shoeboxam:
@annabelle400 Of course. Safe travels!

annabelle400 (approved):
Looks ok now.

PR #2125, opened 2024-10-22
review: conservative_continuous_gaussian_tail_to_alpha

### LINKED ISSUE #2060: Stabilize `conservative_continuous_gaussian_tail_to_alpha`
Remove `\contrib` in `conservative_continuous_gaussian_tail_to_alpha.tex`.

Shoeboxam:
Closes #2060

### DIFF
--- a/rust/src/accuracy/tail_bounds/conservative_continuous_gaussian_tail_to_alpha.tex
+++ b/rust/src/accuracy/tail_bounds/conservative_continuous_gaussian_tail_to_alpha.tex
@@ -1,72 +1,97 @@
-\documentclass{article}
-\input{../../lib.sty}
-
-\title{\texttt{fn conservative\_continuous\_gaussian\_tail\_to\_alpha}}
-\author{\MichaelShoemateAuthor}
-
-\begin{document}
-\maketitle
-
-\contrib
-
-Proof for \rustdoc{accuracy/tail\_bounds/fn}{conservative\_continuous\_gaussian\_tail\_to\_alpha}.
-
-\begin{definition}
-    \label{gaussian}
-    Define $X \sim \mathcal{N}(0, s)$, a random variable following the continuous gaussian distribution:
-    \begin{equation}
-        f(x) = \frac{1}{\sigma \sqrt{2 \pi}} e^{-\frac{1}{2}\left( \frac{x}{\sigma}\right)^2}
-    \end{equation}
-\end{definition}
-
-\begin{definition}
-    \label{erf}
-    The error function is defined as:
-    \begin{equation}
-        \mathrm{erf}(z) = \frac{2}{\sqrt{\pi}} \int_{0}^{z} e^{-t^2} dt
-    \end{equation}
-\end{definition}
-
-\begin{definition}
-    \label{erfc}
-    The complementary error function is defined as:
-    \begin{equation}
-        \mathrm{erfc}(z) = 1 - \mathrm{erf}(z)
-    \end{equation}
-\end{definition}
-
-\begin{lemma}
-    \label{erfc-err}
-    The implementation of \texttt{erfc} differs from a conservatively rounded implementation by no greater than one 32-bit float ulp.
-\end{lemma}
-
-\begin{proof}
-    The following code conducts an exhaustive search.
-    % the adjacent erfc_err_analysis.py file is checked-in, and runnable
-    \label{sec:erfc-err-check}
-    \lstinputlisting[language=Python,firstline=2]{./erfc_err_analysis.py}
-
-    Upon completion, the greatest discovered error is at most 1 ulp.
-\end{proof}
-
+\documentclass{article} 
+\input{../../lib.sty} 
+ 
+\title{\texttt{fn conservative\_continuous\_gaussian\_tail\_to\_alpha}} 
+\author{\MichaelShoemateAuthor} 
+ 
+\begin{document} 
+\maketitle 
+ 
+Proof for \rustdoc{accuracy/tail\_bounds/fn}{conservative\_continuous\_gaussian\_tail\_to\_alpha}. 
+ 
+ 
+\section{Hoare Triple} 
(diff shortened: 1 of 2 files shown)


Shoeboxam:
* **#2125** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2125?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

annabelle400 (commented):
The implementation for this relies on the implementation of the  erf function, together with rounding.

The proof consists of definitions and a lemma statement that says this.
There is included code that checks these assumptions which looks good!

Other than that, it would be good to match up the statement of the postcondition:

Should it read...
Pr[x>= t] ?

but perhaps because of the definition of t with rounding, it should be 
a strict inequality. Please could you confirm.

Also "at least" is easier to read than "no smaller than".

rust/src/accuracy/tail_bounds/conservative_continuous_gaussian_tail_to_alpha.tex
```diff
@@ -1,72 +1,98 @@
...
+    \begin{equation} 
+        \alpha = P[X \ge t] = \frac{1}{2} \mathrm{erfc}\left(\frac{t}{\sigma \sqrt{2}}\right) 
+    \end{equation} 
+\end{theorem} 
+ 
+\begin{proof} 
+    \begin{align*} 
+        \alpha &= P[X \ge t] \\ 
+        &= \frac{1}{\sigma \sqrt{2 \pi}} \int_{t}^{\infty} e^{-\frac{1}{2}\left( \frac{t}{\sigma}\right)^2} dt && \text{by } \ref{gaussian} \\ 
+        &= \frac{1}{2} \left(1 + \mathrm{erf}\frac{t}{\sigma \sqrt{2}}\right) && \text{by } \ref{erf} \\
```
tudorcebere: This should be \frac{1}{2} \left(1 - \mathrm{erf}\frac{t}{\sigma \sqrt{2}}\right), no?

rust/src/accuracy/tail_bounds/conservative_continuous_gaussian_tail_to_alpha.tex
```diff
@@ -1,72 +1,98 @@
...
+    Assume $X \sim \mathcal{N}(0, s)$, and $t > 0$. 
+    \begin{equation} 
+        \alpha = P[X \ge t] = \frac{1}{2} \mathrm{erfc}\left(\frac{t}{\sigma \sqrt{2}}\right) 
+    \end{equation} 
+\end{theorem} 
+ 
+\begin{proof} 
+    \begin{align*} 
+        \alpha &= P[X \ge t] \\ 
+        &= \frac{1}{\sigma \sqrt{2 \pi}} \int_{t}^{\infty} e^{-\frac{1}{2}\left( \frac{t}{\sigma}\right)^2} dt && \text{by } \ref{gaussian} \\
```
tudorcebere: nit: don't use t for both the integral bound and the integration variable.

rust/src/accuracy/tail_bounds/conservative_continuous_gaussian_tail_to_alpha.tex
```diff
@@ -1,72 +1,98 @@
...
+    The following code conducts an exhaustive search. 
+    % the adjacent erfc_err_analysis.py file is checked-in, and runnable 
+    \label{sec:erfc-err-check} 
+    \lstinputlisting[language=Python,firstline=2]{./erfc_err_analysis.py} 
+ 
+    Upon completion, the greatest discovered error is at most 1 ulp. 
+\end{proof} 
+ 
+\begin{theorem} 
+    Assume $X \sim \mathcal{N}(0, s)$, and $t > 0$.
```
tudorcebere: You are using \sigma later on.

rust/src/accuracy/tail_bounds/pseudocode/conservative_continuous_gaussian_tail_to_alpha.py
```diff
@@ -0,0 +1,12 @@
...
+    # the SQRT_2 constant is already rounded down
+    SQRT_2_CEIL = SQRT_2.next_up_()
+
+    t = tail.neg_inf_div(scale).neg_inf_div(SQRT_2_CEIL)
+    # round down to nearest smaller f32
+    t = f64(f32.neg_inf_cast(t))
+    # erfc error is at most 1 f32 ulp (see erfc_err_analysis.py)
+    t = f32.inf_cast(erfc(t)).next_up_()
+
+    f64(t).inf_div(2.0)
```
tudorcebere: Missing a return here

tudorcebere (commented):
The proof looks good! Left some small comments, but the corectness shouldn't change.

Shoeboxam:
Thanks for the reviews! Changes are in the latest commit (there's two commits now).

@annabelle400 
The function’s contract is intentionally kept as a bound on Pr[X > tail] for the ideal continuous Gaussian reference distribution. The thresholded discrete gaussian noise privacy map that calls it needs $\Pr[Z_j > T - Δ_∞]$. I use the tail bound for the continuous distribution as a (very slightly) conservative overestimation for the discrete gaussian tail bound when scale is large because the discrete tail bound isn't tractably computable in that regime. A guarantee on $\Pr[X \ge tail]$ would actually be different, even though we're talking about a continuous distribution, due to the float discretization.

In the proof I use $\Pr[X > tail] = \Pr[X >= tail]$ only as an internal step because that reference distribution is continuous; I’m not making that claim about the float implementation or any discretized/postprocessed sampler.

I added a comment in the proof to this effect, it's confusingly specific otherwise and I had forgotten about this rationale myself!

@tudorcebere 
I updated the proof to restate everything in terms of scale/tail, fixed the sign in the erf step, avoided reusing the integration variable, and added the missing return in the pseudocode.

PR #2129, opened 2024-10-24
non-honest-but-curious plugin APIs

### LINKED ISSUE #2128: non-"honest-but-curious" plugin functions
There are legitimate library plugins that don't require honest-but-curious, but all currently require honest-but-curious.

Shoeboxam:
Closes #2128


rust/opendp_tooling/src/codegen/r/mod.rs
```diff
@@ -41,7 +41,12 @@ const BLACKLIST: &'static [&'static str] = &[
     "new_queryable",
     "new_privacy_profile",
     "user_domain",
-    "_user_domain_descriptor",
+    "_make_transformation",
+    "_make_measurement",
+    "_extrinsic_domain",
+    "_extrinsic_domain_descriptor",
+    "_extrinsic_distance",
+    "_extrinsic_divergence",
```
mccalluc, line 49: Not sure how to be confident that this is up to date. If there is something that's already checking this, add a note at the top?

rust/src/domains/ffi.rs
```diff
@@ -297,8 +297,8 @@ pub extern "C" fn opendp_domains__vector_domain(
     match atom_domain.type_.contents {
         TypeContents::GENERIC { name: "AtomDomain", .. } => 
             dispatch!(monomorphize_all, [(atom_domain.carrier_type, @primitives)], (atom_domain, size)),
-        TypeContents::PLAIN("UserDomain") => monomorphize_user_domain(atom_domain, size),
-        _ => fallible!(FFI, "VectorDomain constructor only supports AtomDomain or UserDomain inner domains")
+        TypeContents::PLAIN("ExtrinsicDomain") => monomorphize_user_domain(atom_domain, size),
+        _ => fallible!(FFI, "Inner domain of VectorDomain must be AtomDomain or ExtrinsicDomain (created via user_domain)")
```
mccalluc, line 301: Should the `user_name` function be renamed to `extrinsic_domain` to match? Things were in sync, but at least the names seem a little more confusing now.

rust/src/domains/ffi.rs
```diff
@@ -358,9 +358,9 @@ pub extern "C" fn opendp_domains__map_domain(
     match (&key_domain.type_.contents, &value_domain.type_.contents) {
         (TypeContents::GENERIC { name: "AtomDomain", .. }, TypeContents::GENERIC { name: "AtomDomain", .. }) => 
             dispatch!(monomorphize, [(key_domain.carrier_type, @hashable), (value_domain.carrier_type, @primitives)], (key_domain, value_domain)),
-        (TypeContents::GENERIC { name: "AtomDomain", .. }, TypeContents::PLAIN("UserDomain")) => 
+        (TypeContents::GENERIC { name: "AtomDomain", .. }, TypeContents::PLAIN("ExtrinsicDomain")) => 
             dispatch!(monomorphize_extrinsic, [(key_domain.carrier_type, @hashable)], (key_domain, value_domain)),
-        _ => fallible!(FFI, "MapDomain constructor only supports AtomDomain or UserDomain inner domains")
+        _ => fallible!(FFI, "Value domain of MapDomain must be AtomDomain or ExtrinsicDomain (created via user_domain)"),
```
mccalluc, line 363: I think `user_domain` is a python function, and isn't part of the rust code, so should it be in a rust error message? More often we worry about rust details bleeding through into python, but it happens the other way, too. This probably isn't something to get hung up about, but with possibly more language bindings we'd want to avoid error messages that hard code one particular naming convention. (In C# the function would probably be `UserDomain`.)

rust/src/domains/ffi.rs
```diff
@@ -481,21 +481,67 @@ pub extern "C" fn opendp_domains__user_domain(
...
+pub extern "C" fn opendp_domains___extrinsic_domain_descriptor(
     domain: *mut AnyDomain,
 ) -> FfiResult<*mut ExtrinsicObject> {
-    let domain = try_!(try_as_ref!(domain).downcast_ref::<UserDomain>()).clone();
+    let domain = try_!(try_as_ref!(domain).downcast_ref::<ExtrinsicDomain>()).clone();
     FfiResult::Ok(util::into_raw(domain.element.value.clone()))
 }
+
+#[bootstrap(
+    name = "_extrinsic_domain",
```
mccalluc, line 504: In the last PR I mentioned being concerned about this as a possible backdoor. Scenario: A is using opendp+pysyft to access a dataset controlled by B. As B is reviewing the code from A, `contrib` or `honest-but-curious` will set off an alarm. A just wants to get the work done, and isn't even malicious, and but uses `_extrinsic_something`. Underscores are private-ish, but wouldn't necessarily set off any alarms for B. Neither would `extrinsic`.

I don't have any great ideas. Two so-so ideas:
- These methods go in a private module and we make it harder to access them from `prelude`.
- Instead of being globals, the features are scoped, and they can be disabled in an appropriate `with` block.

rust/src/measures/ffi.rs
```diff
@@ -314,7 +314,42 @@ pub extern "C" fn opendp_measures__user_divergence(
...
+    name = "_extrinsic_divergence",
+    arguments(descriptor(rust_type = "String"))
+)]
+/// Construct a new ExtrinsicDivergence, a privacy measure defined from a bindings language.
+/// This is meant for internal use, as it does not require "honest-but-curious",
+/// unlike `user_divergence`.
+///
+/// Any two instances of ExtrinsicDivergence are equal if their string descriptors are equal.
+///
+/// The essential requirement of a privacy measure is that it is closed under postprocessing.
```
mccalluc, line 330: Here and elsewhere would it be good to just reference the corresponding docs for the `user` function to avoid the copy and paste?

mccalluc (commented):
some concerns, kind of overlapping with the other PR. If it needs to be done, and this is the best way of doing it, I shouldn't block, but it feels a little odd to have this architecture to control access, and then make a path that goes around it.

rust/opendp_tooling/src/codegen/r/mod.rs
Shoeboxam: If these are missing, the R package will generate code that attempts to bind to these functions, and that generated code will raise errors/cause failures.

rust/src/domains/ffi.rs
Shoeboxam: Moved into separate module.

rust/opendp_tooling/src/codegen/python.rs
```diff
@@ -131,6 +131,7 @@ We suggest importing under the conventional name ``dp``:
             "{}{}",
             "The ``transformations`` module provides functions that deterministicly transform datasets.",
             special_boilerplate("transformations".to_string())),
+        "internal" => "The ``internal`` module provides functions that can be used to construct library primitives without the use of the \"honest-but-curious\" flag.".to_string(),
```
mccalluc, line 134: Should it be `_internal`? Or is it even rendered by Sphinx? Whatever works.

mccalluc (approved):
My hope is that Sphinx doesn't even generate docs for `_internal`, but if it does, the docstring you've provided seems good. Approved, assuming that is working as intended.

rust/opendp_tooling/src/codegen/python.rs
Shoeboxam: This is right! The replacement only happens on file write, I think.

PR #2137, opened 2024-11-04
docs: utf8-lossy in notebooks

mccalluc:
- Towards #1976

Unless you think this is something that can be applied automatically?


Shoeboxam (changes requested):
We still need ignore_errors for when numeric values fail to parse. `utf8-lossy` is take-it-or-leave it- it doesn't give us the needed stability guarantee but doesn't hurt it either.

I don't have vision into what happens to your data before it is passed into the library, and bringing the csv loader into the library is a bit of a wolf in sheep's clothing.

mccalluc:
Close in favor of
- #2234

PR #2139, opened 2024-11-04
feat: add `__version__`

### LINKED ISSUE #2103: Add `__version__`
Seems like a standard way for libraries to provide their version info to users at runtime.

Motivation: Just starting to try to get the logger working locally, and running into a little speed bump. `deserialize.py` has 
```
import pkg_resources
```
But when I try following the steps in the github workflow locally:
```
$ cd test && pytest -v
...
E   ModuleNotFoundError: No module named 'pkg_resources'
```
It looks like this is [no longer the preferred way of doing things](https://stackoverflow.com/questions/7446187/no-module-named-pkg-resources). Explicitly downgrading is an option... but do we even need this? What are we using it for?
```python
try:
    OPENDP_VERSION = pkg_resources.get_distribution("opendp").version
except pkg_resources.DistributionNotFound:
    OPENDP_VERSION = "development"
...
def make_load_ast(obj, force=False):
    if obj["version"] != OPENDP_VERSION and not force:
        raise ValueError(
            f"OpenDP version in parsed object ({obj['version']}) does not match the current installation ({OPENDP_VERSION}). Set `force=True` to try to load anyways."
        )
```
Feels like a simpler approach would be:
```diff
def make_load_ast(obj, force=False):
-    if obj["version"] != OPENDP_VERSION and not force:
+    if obj["version"] != opendp.__version__ and not force: 
        raise ValueError(
-            f"OpenDP version in parsed object ({obj['version']}) does not match the current installation ({OPENDP_VERSION}). Set `force=True` to try to load anyways."
+            f"OpenDP version in parsed object ({obj['version']}) does not match the current installation ({opendp.__version__}). Set `force=True` to try to load anyways."
        )
```

mccalluc:
- Maybe fix #2103?
- Is this the right place for it? Maybe it belongs in `opendp`, not `opendp.prelude`?
- Right now I can't say that I understand the conditions when `VERSION` is not the version number.
- Do we need to clean up other places where `VERSION` is being read directly?


Shoeboxam (commented):
It feels strange to me to have __VERSION__ in the prelude, but I'm also unsure where it should go, as we have multiple packages sharing the OpenDP namespace.

python/src/opendp/prelude.py
```diff
@@ -34,6 +34,10 @@
     >>> type(dp.c.make_basic_composition)
     <class 'function'>
 '''
+from opendp._lib import get_opendp_version as __get_opendp_version
+
+
+__version__ = __get_opendp_version()
```
mccalluc, line 40: Define in mod instead: That's where top-level public stuff goes.

mccalluc:
`__version__` isn't picked up by `import *` because it starts with an underscore, so I added an `__all__`. Looking around noticed that there was only one other file where we import everything that didn't already have `__all__`, so fixed that too. This did surface a couple places where we were getting names (`np` and `re`) by accident, so tidy that up.

One question: Are `ELEMENTARY_TYPES`, `INTEGER_TYPES`, `NUMERIC_TYPES`, `HASHABLE_TYPES`, `PRIMITIVE_TYPES` intended to be part of the public interface, or not? I would guess not: Happy to clean that up in this PR, or can file a follow-up issue.

python/src/opendp/typing.py
```diff
@@ -24,6 +24,80 @@
 from opendp._lib import ATOM_EQUIVALENCE_CLASSES, import_optional_dependency
 
 
+__all__ = [
+    # TODO: Are these intended to be part of the public interface?
```
Shoeboxam, line 28: They don't need to be, so IMO we don't make them part of the public interface.

Shoeboxam (changes requested):
Thanks for finding these. Feel free to fix the re-exports and drop visibility on those values here.

python/src/opendp/typing.py
```diff
@@ -24,6 +24,80 @@
 from opendp._lib import ATOM_EQUIVALENCE_CLASSES, import_optional_dependency
 
 
+__all__ = [
```
Shoeboxam, line 27: I think we don't have to have version in prelude; could we avoid having `__all__` instead as it duplicates names? Or will that always cause imports to be re-exported?

Shoeboxam (changes requested):

python/src/opendp/typing.py
mccalluc: Sorry, I'm not quite understanding this.

> I think we don't have to have version in prelude

so `__version__` would be in `mod.py`, but unlike other names there, it would need to be explicitly imported? 
```
from opendp.mod import __version__
# or...
from opendp.mod
print(mod.__version__)
```
Apart from this, end users never need to use `mod` explicitly, so it feels weird to require it here.

mccalluc: I'll take a stab at this, but wouldn't hurt to talk things through.

python/src/opendp/mod.py
```diff
@@ -10,13 +10,37 @@
...
+    'GLOBAL_FEATURES',
+    'enable_features',
+    'disable_features',
+    'assert_features',
+    'binary_search_chain',
+    'binary_search_param',
+    'binary_search',
+    'exponential_bounds_search',
+    '__version__',
+]
```
Shoeboxam, line 42: Do we still need this? It would be nice to not have to maintain it.

mccalluc: It's not absolutely necessary, but I think we do need to maintain a list of exports somewhere, for a couple reasons:
- Without `__all__`, `from opendp.mod import *` brings in all of the names, including this file's own imports, which clearly don't belong in this namespace.
- `import *` only brings in names without underscores, and doesn't include `__version__`.

An alternative would be to list everything to import instead of using `*`, but splitting it between two files seems like it would be more annoying. Other possibilities:
- Don't touch `mod.py` in this PR, and instead define `__version__` in `prelude.py`.
- Don't add `__all__`, and instead add an explicit import: `from opendp.mod import __version__`.

Shoeboxam: Great point!

Shoeboxam (dismissed):
Thanks for your patience with my reviews!

mccalluc:
To make sure it's doing the job, I've merged this branch into
- #2228

and it can be re-reviewed as part of that.

PR #2142, opened 2024-11-12
add modular perturbative noise mechanisms

### LINKED ISSUE #458: Support modular metric spaces

Shoeboxam:
Close #458


fleetingfound:
I have made a pull request (#2369) which introduces a `radius` parameter for noise mechanisms. The sampling algorithms and privacy maps have been updated accordingly. Please see the pull request for details.

## To do

- Tests
- FFI
- Documentation
- Proofs for:
  - `NoisePrivacyMap` for `ZExpFamily<2>` with `Approximate<ZeroConcentratedDivergence>`
  - `NoisePrivacyMap` for `ZExpFamily<1>` with `Approximate<MaxDivergence>`
  - `conservative_discrete_gaussian_tail_to_alpha_lower`
  - `SampleDiscreteNoise` for `ZExpFamily<1>`
  - `SampleDiscreteNoise` for `ZExpFamily<2>`

I plan to work on the proofs.

docs/source/api/user-guide/measurements/additive-noise-mechanisms.ipynb
```diff
@@ -452,6 +453,44 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Arithmetic: Standard vs. Modular\n",
+    "\n",
+    "For integer-valued queries on cyclic domains, you may want additive noise to wrap around the native integer representation instead of living on the integers.\n",
+    "Set `modular=True` on `absolute_distance`, `l1_distance`, or `l2_distance` to enable wrapped Laplace or Gaussian noise on integer domains.\n",
+    "\n",
+    "For example, `dp.absolute_distance(T=int, modular=True)` uses modular distance on the native `i32` representation in Python, so the mechanism operates modulo `2^32`.\n",
+    "This is useful when the released statistic already lives in a ring, such as a fixed-width counter."
```
mccalluc, line 467: I had the sense earlier that you might use modular even on a non-cyclic domain: Clipping introduces one kind of bias, and wrapping around introduces another... but if the data actually falls inside the specified range, there were advantages to modular noise?

If I've got this wrong, or if that is a thing, but it's a different thing than what you're describing here, great.

docs/source/api/user-guide/measurements/additive-noise-mechanisms.ipynb
```diff
@@ -452,6 +453,44 @@
...
+   "source": [
+    "# wrapped scalar Laplace on the native i32 ring\n",
+    "input_space = dp.atom_domain(T=int), dp.absolute_distance(T=int, modular=True)\n",
+    "modular_lap = dp.m.make_laplace(*input_space, scale=1.)\n",
+    "print(modular_lap)\n",
+    "\n",
+    "# wrapped vector Gaussian on the same ring\n",
+    "input_space = dp.vector_domain(dp.atom_domain(T=int)), dp.l2_distance(T=float, modular=True)\n",
+    "modular_gauss = dp.m.make_gaussian(*input_space, scale=2.)\n",
+    "print(\"rho:\", modular_gauss.map(d_in=1.))\n"
```
mccalluc, line 496: Just for parallelism with the example above, also print the original object? Might not make sense: your call.
```suggestion
    "print(modular_gauss)\n",
    "print(\"rho:\", modular_gauss.map(d_in=1.))\n"
```

python/test/test_metrics.py
```diff
@@ -96,3 +96,32 @@ def test_bound():
...
+
+
+def test_modular_metric_constructors():
+    m_abs = dp.absolute_distance(T=int, modular=True)
+    m_l1 = dp.l1_distance(T=int, modular=True)
+    m_l2 = dp.l2_distance(T=float, modular=True)
+
+    assert "modulo" in str(m_abs)
+    assert "modular" in str(m_l1)
+    assert "modular" in str(m_l2)
```
mccalluc, line 108: Slight preference to check the whole string, unless you think that's too fragile. Depends on whether it's more like to catch real unintended changes... or just add spurious failures.

python/test/test_metrics.py
```diff
@@ -96,3 +96,32 @@ def test_bound():
...
+def test_modular_noise_metrics():
+    meas_abs = dp.m.make_laplace(
+        dp.atom_domain(T=int),
+        dp.absolute_distance(T=int, modular=True),
+        scale=1.0,
+    )
+    assert isinstance(meas_abs(0), int)
+    assert meas_abs.map(1) == 1.0
+
+    meas_vec = dp.m.make_gaussian(
```
mccalluc, line 120: Slight preference to break up tests where the pieces are independent, so if there is a failure you have as much information as possible.
```suggestion
def test_laplace_modular_noise():
    meas_abs = dp.m.make_laplace(
        dp.atom_domain(T=int),
        dp.absolute_distance(T=int, modular=True),
        scale=1.0,
    )
    assert isinstance(meas_abs(0), int)
    assert meas_abs.map(1) == 1.0


def test_gaussian_modular_noise():
    meas_vec = dp.m.make_gaussian(
```

R/opendp/tests/testthat/test-mod.R
```diff
@@ -5,3 +5,35 @@ test_that("binary search", {
...
+
+  meas_abs <- make_laplace(
+    atom_domain(.T = "i32"),
+    absolute_distance(.T = "i32", modular = TRUE),
+    1.
+  )
+  expect_type(meas_abs(arg = 0L), "integer")
+  expect_equal(meas_abs(d_in = 1L), 1.0)
+
+  meas_vec <- make_gaussian(
```
mccalluc, line 32: Ditto: consider breaking up tests into smaller parts that are independent.

rust/src/measurements/noise/distribution/gaussian/mod.rs
```diff
@@ -79,10 +83,18 @@ where
...
-        _input_metric: &L2Distance<RBig>,
+        input_metric: &L2Distance<RBig>,
         _outut_measure: &ZeroConcentratedDivergence,
     ) -> Fallible<PrivacyMap<L2Distance<RBig>, ZeroConcentratedDivergence>> {
-        let ZExpFamily { scale } = self.clone();
+        if self.divisor.is_some() != input_metric.modular() {
+            return fallible!(
+                MakeMeasurement,
+                "divisor ({}) must be set if and only if the input metric is modular ({})",
+                self.divisor.is_some(),
```
mccalluc, line 93: Would it make sense to show the actual divisor in the error message? Just to give the user more information for debugging.
```suggestion
                "divisor ({}) must be set if and only if the input metric is modular ({})",
                self.divisor,
```

rust/src/measurements/noise/distribution/gaussian/NoisePrivacyMap_for_ZExpFamily2.tex
```diff
@@ -57,21 +57,63 @@ \subsection*{Postcondition}
...
+    By the definition of modular $L_2$ distance, there exists a lift
+    $t \in \mathbb{Z}^d$ such that $\pi(t) = r$ and
+    \[
+        \sum_{i=1}^d \frac{t_i^2}{\sigma^2} \le \left(\frac{\texttt{d\_in}}{\sigma}\right)^2.
+    \]
+
+    Choose any lift $a \in \mathbb{Z}^d$ of $x$.
+    Then $a + t$ is a lift of $x'$.
+    Let $P$ be the law of $a + Y$ on $\mathbb{Z}^d$, and $Q$ the law of $a + t + Y$.
+    Their pushforwards under $\pi$ are exactly the laws of $M_m(x)$ and $M_m(x')$.
```
mccalluc, line 92: Looking across other proofs, "pushforward", "law", and "lift" are all new terms. Do they need definitions, or can this be explained in the language we use across other proofs?

rust/src/measurements/noise/distribution/laplace/mod.rs
```diff
@@ -87,10 +90,18 @@ where
...
-        _input_metric: &L1Distance<RBig>,
+        input_metric: &L1Distance<RBig>,
         _output_measure: &MaxDivergence,
     ) -> Fallible<PrivacyMap<L1Distance<RBig>, MaxDivergence>> {
-        let ZExpFamily { scale } = self.clone();
+        if self.divisor.is_some() != input_metric.modular() {
+            return fallible!(
+                MakeMeasurement,
+                "divisor ({}) must be set if and only if the input metric is modular ({})",
+                self.divisor.is_some(),
```
mccalluc, line 100: ditto on divisor in error message?

rust/src/measurements/noise/nature/integer/test.rs
```diff
@@ -22,6 +29,26 @@ fn test_make_int_to_bigint() -> Fallible<()> {
...
+
+    let t_cast = make_int_to_bigint::<i8, 2, f64>(space)?;
+    assert_eq!(
+        t_cast.invoke(&vec![i8::MIN, -1, 0, i8::MAX])?,
+        vec![
+            IBig::from(0),
+            IBig::from(127),
+            IBig::from(128),
+            IBig::from(255)
+        ]
```
mccalluc, line 47: Naively, I'd expect make_int_to_bigint on 0 to return (bigint) 0. If this is functioning as intended, maybe think about the name?

rust/src/measurements/noise/nature/integer/test.rs
```diff
@@ -58,5 +91,16 @@ fn test_then_saturating_cast() -> Fallible<()> {
...
+
+    let f_i8_mod = then_saturating_cast::<i8>(true);
+    assert_eq!(
+        f_i8_mod.eval(&vec![
+            IBig::from(0),
+            IBig::from(127),
+            IBig::from(128),
+            IBig::from(255)
+        ])?,
+        vec![i8::MIN, -1, 0, i8::MAX]
```
mccalluc, line 103: ditto.  Not the behavior I'd expect from a cast.

mccalluc (commented):
Lots of little comments: Nothing that is absolutely essential, and if nothing here resonates with you, happy to approve on re-review. Highlights:
### Language in proofs
You're using new terms here. Even if they are appropriate for this mathematics, would it be worth defining them to make sure the reader has the right understanding?
### Test style
- Where possible, partition big tests into smaller independent tests: Test failures will be more informative.
- We have a pattern, particularly in rust, of providing an input vector and comparing the output to an expected output vector. Here, the length is just 4, but there are longer examples. For readability, consider transposing, so you have input-output pairs.
### Casting semantics
- I would expect small integers to be the same, but just big, after cast. Maybe "cast" isn't the right word? or maybe I'm just confused.

PR #2150, opened 2024-11-14
fix coverage check in CI, RDP function FFI

### LINKED ISSUE #2149: Incomplete coverage doesn't cause CI failure

Shoeboxam:
Closes #2149

The following changes were made in the process of fixing coverage:
Drive-by: FFI for creating RDP privacy curves from Python.
Drive-by: Allow partial chain to be the first element of the query.

Two commits, the first shows that CI fails if coverage is not complete, and the second expands coverage.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -199,7 +199,9 @@ jobs:
         # We only expect test coverage with all extra dependencies.
         # "polars" implies "scikit-learn": see setup.cfg.
         if: ${{ contains(matrix.extra-install, 'polars') && contains(matrix.python-version, '3.12') }}
-        run: coverage report | tee /tmp/coverage.txt
+        run: |
+          set -o pipefail
+          coverage report | tee /tmp/coverage.txt
 
       - name: Upload coverage
         if: ${{ contains(matrix.extra-install, 'polars') && contains(matrix.python-version, '3.12') }}
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -200,8 +200,8 @@ jobs:
         # "polars" implies "scikit-learn": see setup.cfg.
         if: ${{ contains(matrix.extra-install, 'polars') && contains(matrix.python-version, '3.12') }}
         run: |
-          set -o pipefail
-          coverage report | tee /tmp/coverage.txt
+          coverage report
+          touch /tmp/coverage.txt
 
       - name: Upload coverage
         if: ${{ contains(matrix.extra-install, 'polars') && contains(matrix.python-version, '3.12') }}
--- a/python/src/opendp/context.py
+++ b/python/src/opendp/context.py
@@ -716,7 +716,7 @@ class PartialChain(object):
                 meas = _cast_measure(self(param), output_measure, d_out)
                 return meas.check(d_in, d_out)
         
-        param = binary_search(predicate, T=T)
+        param = binary_search(predicate)
         chain = self.partial(param)
         chain.param = param
         return chain
@@ -726,7 +726,6 @@ class PartialChain(object):
         if isinstance(other, (Transformation, Measurement, PartialConstructor)):
             return PartialChain(lambda x: self(x) >> other)
 
-        print(self, other)
         raise ValueError("At most one parameter may be missing at a time")
     
     def __rrshift__(self, other: Union[tuple[Domain, Metric], Transformation, Measurement]):
--- a/python/src/opendp/context.py
+++ b/python/src/opendp/context.py
@@ -44,6 +44,7 @@ from opendp.mod import (
     Domain,
     Measurement,
     Metric,
+    PartialConstructor,
     Queryable,
     Transformation,
     Measure,
@@ -697,7 +698,7 @@ class PartialChain(object):
 
     def __call__(self, v):
         """Returns the transformation or measurement with the given parameter."""
-        return self.partial(v) # pragma: no cover
+        return self.partial(v)
 
     def fix(self, d_in: float, d_out: float, output_measure: Optional[Measure] = None, T=None):
         """Returns the closest transformation or measurement that satisfies the given stability or privacy constraint.
@@ -720,11 +721,12 @@ class PartialChain(object):
         chain.param = param
         return chain
 
-    def __rshift__(self, other: Union[Transformation, Measurement]):
+    def __rshift__(self, other: Union[Transformation, Measurement, PartialConstructor]):
         # partials may be chained with other transformations or measurements to form a new partial
-        if isinstance(other, (Transformation, Measurement)): # pragma: no cover
+        if isinstance(other, (Transformation, Measurement, PartialConstructor)):
             return PartialChain(lambda x: self(x) >> other)
 
+        print(self, other)
         raise ValueError("At most one parameter may be missing at a time")
     
     def __rrshift__(self, other: Union[tuple[Domain, Metric], Transformation, Measurement]):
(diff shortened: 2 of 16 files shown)


Shoeboxam:
* **#2150** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2150?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

.github/workflows/smoke-test.yml
```diff
@@ -199,7 +199,9 @@ jobs:
         # We only expect test coverage with all extra dependencies.
         # "polars" implies "scikit-learn": see setup.cfg.
         if: ${{ contains(matrix.extra-install, 'polars') && contains(matrix.python-version, '3.12') }}
-        run: coverage report | tee /tmp/coverage.txt
+        run: |
+          set -o pipefail
+          coverage report | tee /tmp/coverage.txt
```
mccalluc, line 204: I remember adding the `tee`: Sorry that it broke things!

I had considered just redirecting the output to a file, but then you don't see the output if it fails, so this seems like the best solution.

python/src/opendp/context.py
```diff
@@ -697,7 +697,6 @@ def __init__(self, f, *args, **kwargs):
 
     def __call__(self, v):
         """Returns the transformation or measurement with the given parameter."""
-        # TODO: Can we exercise this?
         return self.partial(v) # pragma: no cover
```
mccalluc, line 700:
```suggestion
        return self.partial(v)
```
Do you mean to include this as tested code now? The TODO isn't very useful, so no complaints about removing it: we can always find the `pragma`.

python/src/opendp/context.py
```diff
@@ -723,11 +722,16 @@ def predicate(param):
 
     def __rshift__(self, other: Union[Transformation, Measurement]):
         # partials may be chained with other transformations or measurements to form a new partial
-        # TODO: Can we exercise this?
         if isinstance(other, (Transformation, Measurement)): # pragma: no cover
```
mccalluc, line 725:
```suggestion
        if isinstance(other, (Transformation, Measurement)):
```
ditto: Maybe covered now?

python/test/context/test_context.py
```diff
@@ -265,6 +265,18 @@ def test_approx_to_approx_zCDP():
     assert release == [1, 2, 3]
 
 
+def test_agg_input():
+    dp.enable_features("contrib")
```
mccalluc, line 269:
```suggestion
```
Should be covered in `__init__.py`:
```
# We don't have tests that require a feature *not* to be enabled,
# so more stable just to enable everything at the start.
dp.enable_features('floating-point', 'contrib', 'honest-but-curious')
```

python/test/test_comb.py
```diff
@@ -110,7 +120,10 @@ def test_make_approximate():
         dp.c.make_fix_delta(dp.c.make_zCDP_to_approxDP(dp.m.make_gaussian(*input_space, 10.)), delta=1e-6)
     ])
 
-    print(meas.map(1.))
+    assert meas.map(1.) == (0.5299414688369495, 1e-06)
```
mccalluc, line 123: Not an absolute blocker, but prefer an expression that explains where this number comes from. I imagine a future maintainer who is hitting a test failure here: They shouldn't "fix" it by just changing the assertion to whatever number they have, and we should set a good example.

python/test/test_comb.py
```diff
@@ -120,7 +133,7 @@ def test_make_pureDP_to_zCDP():
         dp.m.make_gaussian(*input_space, 10.)
     ])
 
-    print(meas.map(1.))
+    assert meas.map(1.) == 0.010000000000000002
```
mccalluc, line 136: Maybe round to make this less fragile? Unless it really should be exactly this?

python/test/test_core.py
```diff
@@ -328,4 +328,8 @@ def test_pointer_classes_dont_iter():
     # We override __iter__ so as to make this infinite loop/lock impossible to accidentally trigger
     with pytest.raises(ValueError):
         [*dp.atom_domain(T=bool)]
-    
\ No newline at end of file
+
+
+def test_erfc():
+    from opendp._data import erfc
+    assert erfc(0.5) == 0.4795001222363462
```
mccalluc, line 335: This might be the exception: The math is in `erfc`, and you assert what the result should be.

mccalluc (dismissed):
Wonder if a couple of the `pragma`s can be cleared, and a preference for explicit expressions, but nothing that needs to block. Sorry for introducing the bug in the first place!

python/test/test_comb.py
Shoeboxam: It works out to exactly this every time, due to the float representation of .01 being greater than .01.

python/src/opendp/context.py
```diff
@@ -709,24 +709,30 @@ def fix(self, d_in: float, d_out: float, output_measure: Optional[Measure] = Non
...
-    def __rshift__(self, other: Union[Transformation, Measurement]):
+    def __rshift__(self, other: Union[Transformation, Measurement, PartialConstructor]):
         # partials may be chained with other transformations or measurements to form a new partial
-        # TODO: Can we exercise this?
-        if isinstance(other, (Transformation, Measurement)): # pragma: no cover
-            return PartialChain(lambda x: self.partial(x) >> other)
+        if isinstance(other, (Transformation, Measurement, PartialConstructor)):
+            return PartialChain(lambda x: self(x) >> other)
 
+        print(self, other)
```
mccalluc, line 729:
```
Name                    Stmts   Miss  Cover   Missing
-----------------------------------------------------
src/opendp/context.py     336      1    99%   729
```
```suggestion
```
... but still need a test that reaches the following `ValueError`, or a pragma

mccalluc (commented):
maybe the print was added for debugging?

.github/workflows/smoke-test.yml
mccalluc, line 204: Realized this can be much simpler!
```suggestion
          coverage report
          touch /tmp/coverage.txt
```
We don't actually need the report: We just use the artifact to confirm that it ran somewhere in the test matrix.

python/src/opendp/context.py
mccalluc, line 729: Correction: all `raise` lines are ignored during the coverage check.
```
exclude_also =
  # TODO: Improve coverage and shorten this list!
  ^\s*raise\b
  os\.environ
  ^\s*\.\.\.\s*$
```
so, no `pragma` is needed here... although it obviously would be good to expand coverage and make sure we reach this when appropriate.

mccalluc (commented):
Earlier review was misguided. Sorry!

mccalluc (approved):
Thanks. Looks good. Raises a few higher level questions for me:
- Bare `print` in tests: Is it useful for me to ask for more explanation on these during review? 
- Making sure typing will work for library users. Should we have a goal not to use any `pragma: ignore` in our tests where we exercise the API?
- testing exceptions: Should we stop excluding these from code coverage? Probably add a lot of ignores now, and incrementally add tests, particularly on new code.

Maybe talk about these at the next stand up?

Shoeboxam:
* In this case the same output was tested elsewhere, I was considering the test to be whether or not the query would run at all under the conditions presented to it. Mentioning the print is good though and its frequent inclusion in these diffs is my oversight.
* I don't know why that ignore was there-- it passes without that ignore. Removed it.
* My attention is drawn to other things atm. If you want to work that change, then you're welcome to.

mccalluc:
Thanks for the response. I'll give this a quick look and hopefully approve.

- We have #1830 already

and in 2025 I could image switching gears for a bit and tackling that and some other tidying.

mccalluc (approved):
Looks good. Thanks!

PR #2154, opened 2024-11-20
docs: Feedback links

### LINKED ISSUE #2127: Review 3rd party feedback services for integration with docs
We can't offer real-time support, and an LLM is not going to work with the complexity of the API, but could a third party widget on the docs site help? We do have email and slack for support, but those are relatively high friction. Imagining something like:

> Was this helpful? `Yes` `No`
> Anything else we should know? `______________`
> Include an email if you'd like a response: `______________`

And it gets sent to our slack or email, and there's no software for us to maintain.

mccalluc:
- Fix #2127. Surveyed other options, but this seems good enough. 

(If it's not good enough, then that's also good to know! I don't have any hard requirements right now to constrain the search.)

If you're working on this branch note that the iframe is a 403 if you're just loading it from `file:///`: It needs to be served by http.

Looks like this:

Screenshot 2024-11-20 at 9 08 46 AM
Screenshot 2024-11-20 at 9 09 08 AM

I've changed the font and text color to match our site, but it's still obviously a google form.

Questions to consider:
- Privacy wise, are we ok with a google form? One alternative would be to write out the html for the iframe only after the user clicks... but I think that would be a bit for work for not much gain.
- Are these questions the questions we want to ask?
- Whose responsibility is it to look at the responses?

### DIFF
--- a/docs/source/_templates/questions-feedback.html
+++ b/docs/source/_templates/questions-feedback.html
@@ -4,6 +4,12 @@
     -->
     <details>
         <summary>Questions or feedback?</summary>
-        iframe goes here
+        <iframe
+            src="https://docs.google.com/forms/d/e/1FAIpQLSe65Pu58WubBORSEpzlXS2ZlXGr88AjXAwd3MNRAb2PMew2_w/viewform?embedded=true"
+            width="640" height="570"
+            frameborder="0"
+            marginheight="0"
+            marginwidth="0"
+        >Loading…</iframe>
     </details>
 </div>
--- a/docs/source/_templates/questions-feedback.html
+++ b/docs/source/_templates/questions-feedback.html
@@ -1,0 +1,9 @@
+<div class="admonition warning" style="position: relative;">
+    <!--
+     Without the style, the region for the article title overlaps, and the link is unclickable.
+    -->
+    <details>
+        <summary>Questions or feedback?</summary>
+        iframe goes here
+    </details>
+</div>
--- a/docs/source/_templates/questions-feedback.html
+++ b/docs/source/_templates/questions-feedback.html
@@ -4,12 +4,14 @@
     -->
     <details>
         <summary>Questions or feedback?</summary>
-        <iframe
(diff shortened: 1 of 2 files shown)


mccalluc:
From Sharon on slack:
> Hmm, if that was the intent, a form doesn't seem like the right mechanism IMO.  Maybe something else can be explored

mccalluc:
Talked with Annie:
- These questions are good; could be simplified to just a thumbs up / down.
- Would like to know what page users are on.

Next steps:
- Google forms can be prepopulated, but I don't know that the prepopulation is transparent in the URL. If it is, great!
- If not, look at qualtrics: We've used it before, and billing is already set up.

Moving back to draft.

mccalluc:
Yes: Prefilled fields encoding isn't too tricky:

`space then/dash` becomes:

`https://docs.google.com/forms/d/e/1FAIpQLSeL-KZ_K1Yz5SOjGo0wYOLUEVbtOwBU1Ty6WeoRJbFWeEmD5w/viewform?entry.1628573312=space+then/dash`

Shoeboxam:
From running this myself, the 403 goes away if you update the settings of the form to allow folks from outside of Harvard to see it. However, it comes with a lot of invasive tracking that I can't disable through any cookies or javascript manipulation. 

Regardless, we should just provide invite links to the OpenDP Slack, email and github discussions so that we 
- don't dilute engagement 
- don't add more things to keep track of

Shoeboxam (changes requested):
From running this myself, the 403 goes away if you update the settings of the form to allow folks from outside of Harvard to see it. However, it comes with a lot of invasive tracking that I can't disable through any cookies or javascript manipulation. 

Regardless, we should just provide invite links to the OpenDP Slack, email and github discussions so that we 
- don't dilute engagement 
- don't add more things to keep track of

mccalluc:
Updated the content, copying over the section from the README. I'd suggest some changes both here and there, though I don't want this to get bogged down.
- Drop twitter and github discussions
- Add security@opendp.org

Happy to make additional changes here, or in a follow-up PR, or just leave it as it: your call.

Shoeboxam (approved):
This looks great!

PR #2156, opened 2024-11-21
feat: prove make_expr_strptime

Shoeboxam:
Closes #1755

PR #2155, opened 2024-11-21
prove counting_query_stability_map and make_expr_count

Shoeboxam:
Closes #1740

PR #2153, opened 2024-11-20
feat: prove `get_margin` and `find_min_covering`

Shoeboxam:
Closes #1707

PR #2157, opened 2024-11-28
Polars: simplify ExprDomain

Shoeboxam:
When writing proofs for Polars functionality, I found that there were too many descriptors carried around in ExprDomain leading to more complexity in proofs. Before this PR ExprDomain held a frame domain, with information for all columns and margins, as well as information about the context.

Since expressions always produce a single column, the refactored ExprDomain only holds a single series domain instead of all series domains.

When tracking margins in a row-by-row context, maintaining metadata about margins that is consistent with all other margins is tricky, and the descriptors quickly become extremely pessimistic to the point where they are not useful. Due to this, I've completely removed access to margin metadata when in the row-by-row context.

When tracking margins in a grouping context, only the margin about the current grouping needs to be known. Therefore in the grouping context, ExprDomains may hold one margin. This is useful for calibrating sensitivity in aggregators.

By this logic, proof for output domain correctness need only consider one series domain and how the margin might change in the grouping context (which for most transformations will not change).

This PR also adds a `WildExprDomain` used to start expression transformations/measurements when a column has not yet been selected.`WildExprDomain` carries descriptors for all the select-able series domains, and its carrier type is just a plan, which simplifies how expression transformations/measurements are called (no more tuple of a plan and `all()`)

### DIFF
--- a/python/test/test_polars.py
+++ b/python/test/test_polars.py
@@ -51,8 +51,11 @@ def example_lf(margin=None, **kwargs):
 
 
 def test_expr_domain():
-    lf_domain, _ = example_lf()
-    dp.expr_domain(lf_domain)
+    series_domains, _ = example_series()
+    dp.wild_expr_domain(series_domains)
+    dp.wild_expr_domain(series_domains, by=["A", "B"])
+    dp.wild_expr_domain(series_domains, by=["A", "B"], max_num_partitions=10)
+    dp.wild_expr_domain(series_domains, by=["A", "B"], public_info="keys")
 
 
 def test_domains():
@@ -85,15 +88,6 @@ def test_lazyframe_ffi():
     t_ident = (lf_domain, dp.symmetric_distance()) >> dp.t.then_identity()
 
     pl_testing.assert_frame_equal(t_ident(lf).collect(), lf.collect())
-
-
-def test_expr_ffi():
-    """ensure that expr domain's carrier type can be passed to/from Rust"""
-    pl = pytest.importorskip("polars")
-    lf_domain, lf = example_lf()
-    expr_domain = dp.expr_domain(lf_domain, grouping_columns=[])
-    t_ident = (expr_domain, dp.symmetric_distance()) >> dp.t.then_identity()
-    assert str(t_ident((lf, pl.col("A")))[1]) == str(pl.col("A"))
 
 
 @pytest.mark.parametrize(
@@ -188,14 +182,14 @@ def test_cast():
 
 def test_stable_expr():
     pl = pytest.importorskip("polars")
-    domain = dp.expr_domain(example_lf()[0])
+    domain = dp.wild_expr_domain(example_series()[0])
     with pytest.raises(dp.OpenDPException):
         dp.t.make_stable_expr(domain, dp.symmetric_distance(), pl.col("A").cast(int))
 
 
 def test_private_expr():
     pl = pytest.importorskip("polars")
-    domain = dp.expr_domain(example_lf(margin=[])[0], grouping_columns=[])
+    domain = dp.wild_expr_domain(example_series()[0], by=[])
     with pytest.raises(dp.OpenDPException):
         dp.m.make_private_expr(
             domain,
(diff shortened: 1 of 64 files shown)


Shoeboxam:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/opendp/opendp/2157?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2096** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2096?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#2097](https://github.com/opendp/opendp/pull/2097) <a href="https://app.graphite.dev/github/pr/opendp/opendp/2097?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2156](https://github.com/opendp/opendp/pull/2156) <a href="https://app.graphite.dev/github/pr/opendp/opendp/2156?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#2157** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2157?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/opendp/opendp/2157?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2095** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2095?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#2155](https://github.com/opendp/opendp/pull/2155) <a href="https://app.graphite.dev/github/pr/opendp/opendp/2155?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#2091** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2091?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#2153](https://github.com/opendp/opendp/pull/2153) <a href="https://app.graphite.dev/github/pr/opendp/opendp/2153?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`

<h2></h2>

python/test/test_polars.py
```diff
@@ -129,6 +123,7 @@ def test_private_lazyframe_explicit_sum(measure):
 def test_private_lazyframe_sum(measure):
     pl = pytest.importorskip("polars")
     pl_testing = pytest.importorskip("polars.testing")
+    dp.enable_features("rust-stack-trace")
```
mccalluc: If it's useful, I'd prefer to just add this to `test/__init__.py` since it will end up being applied globally.

python/test/test_polars.py
```diff
@@ -188,14 +183,14 @@ def test_cast():
 
 def test_stable_expr():
     pl = pytest.importorskip("polars")
-    domain = dp.expr_domain(example_lf()[0])
+    domain = dp.wild_expr_domain(example_series()[0])
```
mccalluc, line 185: It looks like we never referenced `expr_domain` in the docs, except for the generated API docs. If users would never need to use `expr_domain` or `wild_expr_domain` directly, then maybe they should be private? (Or else docs that explain when they would use it.)

(Started wondering about whether this is the best name in a public interface, but if it's something that only gets used internally, it's a fine name.)

python/test/test_polars.py
```diff
@@ -610,23 +605,23 @@ def test_polars_threshold_rho():
 def test_replace_binary_path():
     import os
     pl = pytest.importorskip("polars")
-    expr = pl.len().dp.noise(scale=1.)
+    expr = dp.len(scale=1.)
```
mccalluc, line 607: `pl.len().dp.noise()` comes up several times in the docs: Should those occurrences be cleaned up too?

mccalluc (commented):
I see there are already changes on this branch, so I'll submit just this much, and look at the Rust code on a second pass. On the python side I'm wondering if some of this is implementation details that don't need to be part of the public API (`wild_expr`), and at the same time, whether some of the changes that look like simplifications to the API should be applied more widely in the public docs (`pl.len().dp.noise` -> `dp.len`).

python/test/test_polars.py
Shoeboxam: I will add a docs page for it

mccalluc (approved):
Last commit looks good. Let me know if I should be looking at anything else? Approved.

PR #2158, opened 2024-11-28
Polars: simplify SeriesDomain

Shoeboxam:
SeriesDomain had a drawback where you could update the element domain and forget to update the dtype, or vice-versa.

The refactored code retrieves the dtype from the element domain, and eliminates the dtype descriptor. This also eliminates the use of the Polars Field struct, shortening attribute access. I also added some useful helpers for setting the dtype or element domain that better reuses existing code.

The code also fixes a bug in the implementation of PartialEq for SeriesDomain-- two series domains that differ in their nullity bit would still be considered equal. This doesn't result in any potential exploits though, since our idiom is to set the input space to the previous domain's output space.

### DIFF
--- a/python/src/opendp/extras/polars/__init__.py
+++ b/python/src/opendp/extras/polars/__init__.py
@@ -420,7 +420,7 @@ class DPExpr(object):
         │ ---    │
         │ f64    │
         ╞════════╡
-        │ ...    │
+        │ ...... │
         └────────┘
 
         Privately estimates the numerator and denominator separately, and then returns their ratio.
--- a/rust/src/domains/polars/series/mod.rs
+++ b/rust/src/domains/polars/series/mod.rs
@@ -33,24 +33,26 @@ mod test;
 /// ```
 #[derive(Clone)]
 pub struct SeriesDomain {
-    /// The name of the series and type of underlying data.
-    pub field: Field,
+    /// The name of series in the domain.
+    pub name: SmartString,
     /// Domain of each element in the series.
     pub element_domain: Arc<dyn DynSeriesElementDomain>,
-    /// Indicates if data can contain null values.
+    /// Indicates if elements can be null.
     pub nullable: bool,
 }
 
-impl core::cmp::PartialEq for SeriesDomain {
+impl PartialEq for SeriesDomain {
     fn eq(&self, other: &Self) -> bool {
-        self.field.eq(&other.field) && self.element_domain.eq(&self.element_domain)
+        self.name == other.name
+            && self.element_domain.eq(&other.element_domain)
+            && self.nullable == other.nullable
     }
 }
 
 impl Domain for SeriesDomain {
     type Carrier = Series;
     fn member(&self, value: &Self::Carrier) -> Fallible<bool> {
-        if &self.field != &*value.field() {
+        if &self.name != &value.name() {
             return Ok(false);
         }
 
@@ -74,7 +76,7 @@ impl Domain for SeriesDomain {
             }};
         }
 
-        match self.field.dtype {
+        match self.dtype() {
             DataType::UInt8 => atom_member!(u8, UInt8Type),
             DataType::UInt16 => atom_member!(u16, UInt16Type),
             DataType::UInt32 => atom_member!(u32, UInt32Type),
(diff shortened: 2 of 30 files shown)


Shoeboxam:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/opendp/opendp/2158?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2158** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2158?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/opendp/opendp/2158?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2096** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2096?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2 other dependent PRs ([#2097](https://github.com/opendp/opendp/pull/2097) <a href="https://app.graphite.dev/github/pr/opendp/opendp/2097?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2156](https://github.com/opendp/opendp/pull/2156) <a href="https://app.graphite.dev/github/pr/opendp/opendp/2156?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#2157** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2157?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2095** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2095?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#2155](https://github.com/opendp/opendp/pull/2155) <a href="https://app.graphite.dev/github/pr/opendp/opendp/2155?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#2091** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2091?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#2153](https://github.com/opendp/opendp/pull/2153) <a href="https://app.graphite.dev/github/pr/opendp/opendp/2153?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`

<h2></h2>

python/src/opendp/extras/polars/__init__.py
```diff
@@ -421,7 +421,7 @@ def mean(
         │ ---    │
         │ f64    │
         ╞════════╡
-        │ ...    │
+        │ ...... │
```
mccalluc, line 424: doesn't hurt anything, but `...` is sufficient. Doubled, this is essentially a regex like `.*.*`.

rust/src/domains/polars/series/mod.rs
```diff
@@ -189,19 +203,22 @@ impl SeriesDomain {
     /// If the domain of elements is of type `AtomDomain<T>`, then returns the domain as that type,
     /// otherwise returns an error.
     pub fn atom_domain<T: 'static + CheckAtom>(&self) -> Fallible<&AtomDomain<T>> {
+        self.element_domain::<AtomDomain<T>>()
+    }
+
+    /// # Proof Definition
+    /// If the domain of elements is of type `D`, then returns the domain as that type,
+    /// otherwise returns an error.
+    pub fn element_domain<D: 'static>(&self) -> Fallible<&D> {
```
mccalluc, line 212: Does this need to be public? If possible, it's nice to keep the API small, so the user has an easier time finding the method they actually need.

mccalluc (approved):
Looks good to me.

You mention
> The code also fixes a bug in the implementation of PartialEq for SeriesDomain-- two series domains that differ in their nullity bit would still be considered equal.

Even if the fix is easy, wouldn't hurt to add a test for this: It's always possible that a future refactor could restore the earlier bug, and we'd like to catch it if it came back.

python/src/opendp/extras/polars/__init__.py
Shoeboxam: I set float-precision to zero above to prevent random dataframe sizes. With this configuration, it can return `-inf`, which is four characters, leaving three spaces. But the doctest requires four spaces. So I doubled the `...` to keep things loosen the doctest requirement on spacing without breaking the table

rust/src/domains/polars/series/mod.rs
Shoeboxam: It can be public-- I was going to expose this same method in FFI.

PR #2172, opened 2024-12-06
dev: Warn about bare `print`

### LINKED ISSUE #2169: subprocess test to check for stray `print` in tests
If we have a `print` in a test, it should also include an explanation to a future dev about what to expect to see. Seems like a subprocess test would be easy:
```
! egrep 'print\([^"f]' -r python/test --exclude '*.pyc'
```

(`f` for f-strings: I'm sure this could be done better.)

(There are actually a lot of matches right now: Don't know whether it makes sense to remove them all, or add some placeholder explanation.)

(The same thing might be done outside the tests: There, my sense is that `print` is never good, and occurrences should be replaced with logger calls.)

mccalluc:
- Fix #2169

The main idea is to catch these before code review, but obviously not a high priority.

Where there are bare `print` statements in the tests currently, providing a placeholder for now seems like a conservative approach.

`_test.py` seems like an old demo. If it's important, could make it into a real test.


Shoeboxam (commented):
The new test looks good to me, but do you think you could handle some of the TODOs?

mccalluc:
- About half seem stable, so I replaced the `print` with an `assert`
- A few cases were just redundant, ie there was already an `assert`, so I removed those.
- For most of the remainder, I just used the expression as the label, which seems good enough.

This does add several tests where we assert equality to a particular float with a lot of decimal points. In similar situations I've asked if we could have a closed form expression instead, but these are stronger tests than they were before, and maybe that's enough for now.

python/test/test_subprocesses.py
```diff
@@ -7,6 +7,10 @@
     'flake8 linting': 'cd .. && flake8 . --count --show-source --statistics',
     # Had non-reproducible errors between local runs, so disable caching.
     'mypy type checking': 'mypy . --cache-dir=/dev/null',
+    # Looking for cases where first character is not `'`, `"`, or `f`
+    # (for f-string). Negative lookahead would be more precise.
+    '"print" in test/ should be explained': "! egrep 'print\\(([^\"'\"'\"'f]' -r test",
+    '"print" in src/ should be log instead': "! egrep '^\\s+print\\(' -r src",
```
mccalluc, line 13: These could be done in python, like in the [example data PR](https://github.com/opendp/opendp/pull/2229/files#diff-a2d90aa4999938f1e470e0ed44ad7e58f4ac72e154368d73576ff56aa2fe7e6eR14-R18), but I don't have a strong preference.

Shoeboxam (approved):
Looks good! Thanks for strengthening the tests.

Issue #2164, opened 2024-12-03
Use more ABCs for types


mccalluc:
Prompted by
- #2089

```python
Optional[dict[tuple[str, ...], Margin]]
```
Users might want to provide a [mapping](https://docs.python.org/3/glossary.html#term-mapping) other than a `dict`, or a [sequence](https://docs.python.org/3/glossary.html#term-sequence) other than a `tuple`. Using a base class for the annotation wouldn't change runtime behavior, but it would save users who are doing something a little unusual from getting type checking errors on their own code that uses OpenDP.

- See also: #2151

Shoeboxam:
ABCs are the norm now. Thanks for bringing this up!

PR #2177, opened 2024-12-09
list -> Sequence

mccalluc:
This isn't a high priority, and this isn't a global replacement, but it does tweak the signatures on many public methods. The advantage is that users could supply values that aren't exactly `list`s but are good enough, and with this change they should get fewer spurious typing errors.

- Towards #2164


Shoeboxam (approved):
The preference for slicing over mutation with pop is interesting. All looks fine to me, though.

PR #2181, opened 2024-12-10
docs: Check for docstrings that are out of sync

### LINKED ISSUE #1395: Check that docstrings agree with signatures

mccalluc:
- Fix #1395

After looking for a library that would handle this, and coming back empty-handed, decided just to wrap my own, and it didn't turn out to be too complicated.

Can imagine expanding this to check that we do have the parameters in the docstring, and that we have some description for each parameter, and maybe check for return and typing info.

Going to branch from here and strengthen the tests, and start filling in some gaps in the docs.

### DIFF
--- a/python/test/test_ast.py
+++ b/python/test/test_ast.py
@@ -13,7 +13,7 @@ class Function(NamedTuple):
 
 def get_params_from_node_docstring(node):
     docstring = ast.get_docstring(node)
-    return set(re.findall(r':param (\w+):', docstring))
+    return set(re.findall(r':param (\w+):', docstring))  # type: ignore[arg-type]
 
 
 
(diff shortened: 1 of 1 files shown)


Shoeboxam (approved):
Short and sweet, looks good to me. Thanks!

PR #2182, opened 2024-12-10
docs: Checking docstrings, continued

mccalluc:
- Builds on #2181: Use that machinery to find things that can be improved, and make the fixes.
- If there is a docstring, param tags must match signature.
- Public functions must have docstrings. (From just the AST, it is hard to identify functions that are purely internal, so make many of them explicitly private to exclude them from the check.)
- Check for unknown Sphinx directives.
- Check for consistency between type annotations and docs
- Check that directives are in usual order

Several blocks of code are commented out, with the number of warns that arise if they are uncommented: I didn't want to introduce even more changes that would make this hard to review and merge, but there is a lot of follow-up work that could be done.


Shoeboxam (changes requested):
As in the case of testing there are no prints, this seems like a nice test/nice to have, but could you work on filling the missing documentation in?

mccalluc:
Fill in TODOs, though not with anything particularly useful in most cases. In the future, I'd hope that checks like this would nudge us to provide something that does more than repeat the method name on new code, but for many of these, maybe that's enough.

Also [removed](https://github.com/opendp/opendp/pull/2182/files#diff-d8b5490a98400c6e71ac63fd5226bf1165657e2c8f969f6a4487cb6b08a5a921L440) one `eval` method that seems to be unused?

python/src/opendp/typing.py
```diff
@@ -368,6 +368,11 @@ def parse_or_infer(
         raise UnknownTypeException("either type_name or public_example must be passed")  # pragma: no cover
 
     def substitute(self: Union["RuntimeType", str], **kwargs):
+        '''
+        TODO
```
mccalluc, line 372: @Shoeboxam , could you suggest a docstring here?

mccalluc:
rtype checking may already be working on this branch, but if other checks are strengthened, we might get more errors on it.

mccalluc:
Going forward: don't be strict about :rtype: and :type: annotations -- @mccalluc should remove any TODOs for those in this PR.

python/src/opendp/typing.py
Shoeboxam: Substitutes any generic type parameters according to the passed keyword arguments.

python/src/opendp/extras/polars/__init__.py
```diff
@@ -895,6 +895,9 @@ def with_keys(
         ) -> LazyFrameQuery:
             """
             Shorthand to join with an explicit key-set.
+
+            :param keys:
+            :param on:
```
Shoeboxam, line 859:
```suggestion
            :param keys: lazyframe containing a key-set whose columns correspond to the grouping keys
            :param on: optional, the names of columns to join on. Useful if the key dataframe contains extra columns
```

python/src/opendp/extras/numpy/__init__.py
```diff
@@ -179,6 +179,7 @@ def _sscp_domain(
     :param p: designates L`p` norm
     :param size: number of rows in data
     :param num_features: number of rows/columns in the matrix
+    :param T:
```
Shoeboxam, line 182:
```suggestion
    :param T: the type of data elements in the array
```

python/src/opendp/context.py
```diff
@@ -569,7 +574,11 @@ def make(*args, **kwargs) -> "Query":
         return make
 
     def new_with(self, *, chain: Chain, wrap_release=None) -> "Query":
-        """Convenience constructor that creates a new query with a different chain."""
+        """Convenience constructor that creates a new query with a different chain.
+        
+        :param chain:
+        :param wrap_release:
```
Shoeboxam, line 580:
```suggestion
        :param chain: the prior query. Either a metric space or transformation
        :param wrap_release: a function to apply to apply to releases
```

python/src/opendp/context.py
```diff
@@ -541,6 +544,8 @@ def make(*args, **kwargs) -> "Query":
             """Wraps the ``make_{name}`` constructor to allow one optional parameter and chains it to the current query.
 
             This function will be called when the user calls ``query.{name}(...)``.
+
+            :param kwargs:
```
Shoeboxam, line 548:
```suggestion
            :param args: arguments for the constructor function
            :param kwargs: keyword arguments for the constructor function
```

python/src/opendp/extras/polars/__init__.py
```diff
@@ -1041,6 +1044,8 @@ def agg(
         ) -> LazyFrameQuery:
             """
             Compute aggregations for each group of a group by operation.
+
+            :param named_aggs:
```
Shoeboxam, line 1007:
```suggestion
            :param aggs: expressions to apply in the aggregation context
            :param named_aggs: named/aliased expressions to apply in the aggregation context
```

python/src/opendp/extras/polars/__init__.py
```diff
@@ -1065,7 +1070,10 @@ def release(self) -> OnceFrame:
             raise ImportError(ERR_MSG)
 
         def summarize(self, alpha: float | None = None):
-            """Summarize the statistics released by this query."""
+            """Summarize the statistics released by this query.
+            
+            :param alpha:
```
Shoeboxam, line 1034:
```suggestion
            :param alpha: optional. A value in [0, 1] denoting the statistical significance.
```

python/src/opendp/mod.py
```diff
@@ -275,11 +293,14 @@ def __call__(self, arg):
         return transformation_invoke(self, arg)
 
     def map(self, d_in):
-        """Map an input distance `d_in` to an output distance."""
+        """Map an input distance `d_in` to an output distance.
+        
+        :param d_in: Input distance
```
Shoeboxam, line 298:
```suggestion
        :param d_in: Input distance. An upper bound on how far apart neighboring datasets can be with respect to the input metric
```

python/src/opendp/mod.py
```diff
@@ -487,23 +525,37 @@ class Domain(ctypes.POINTER(AnyDomain)): # type: ignore[misc]
     _type_ = AnyDomain
 
     def member(self, val):
+        '''
+        Check if ``val`` is a member of the domain.
+        
+        :param val:
```
Shoeboxam, line 531:
```suggestion
        :param val: a value to be checked for membership in `self`
```

python/src/opendp/mod.py
```diff
@@ -487,23 +525,37 @@ class Domain(ctypes.POINTER(AnyDomain)): # type: ignore[misc]
...
+        Carrier type of domain
+        '''
         from opendp.domains import domain_carrier_type
         from opendp.typing import RuntimeType
         return RuntimeType.parse(domain_carrier_type(self))
     
     @property
     def descriptor(self) -> Any:
+        '''
+        Descriptor of domain
```
Shoeboxam, line 557:
```suggestion
        Descriptor of domain. Used to retrieve the descriptor associated with domains defined in Python 
```

python/src/opendp/mod.py
```diff
@@ -636,10 +700,20 @@ def __init__(self, curve):
         self.curve = curve
 
     def delta(self, epsilon):
+        '''
+        Returns the delta that corresponds to this epsilon.
+        
+        :param epsilon:
```
Shoeboxam, line 706:
```suggestion
        :param epsilon: Allowance for a multiplicative difference, or max divergence, in the distributions of releases on adjacent datasets
```

python/src/opendp/mod.py
```diff
@@ -636,10 +700,20 @@ def __init__(self, curve):
...
+        '''
         from opendp._data import privacy_profile_delta
         return privacy_profile_delta(self.curve, epsilon)
 
     def epsilon(self, delta):
+        '''
+        Returns the epsilon that corresponds to this delta.
+        
+        :param delta:
```
Shoeboxam, line 715:
```suggestion
        :param epsilon: Allowance for an additive difference between the distributions of releases on adjacent datasets
```

Shoeboxam (changes requested):
Lots of small cleanups in here and some future-proofing.

python/src/opendp/extras/polars/__init__.py
mccalluc: Since `1 - alpha` comes up frequently, I'll add that here as well.

python/src/opendp/typing.py
mccalluc, line 372:
```suggestion
        Substitutes any generic type parameters according to the passed keyword arguments
```

Shoeboxam (approved):

python/src/opendp/mod.py
```diff
@@ -1053,17 +1127,21 @@ def check_type(v):
...
             raise TypeError("unable to infer type `T`; pass the type `T` or bounds")  # pragma: no cover
 
     # core search functionality
     def signed_band_search(center, at_center, sign):
-        """identify which band (of eight) the decision boundary lies in, 
-        starting from `center` in the direction indicated by `sign`"""
+        """Identify which band (of eight) the decision boundary lies in.
+
+        :param center: Start here
+        :param at_center:
```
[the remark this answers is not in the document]
mccalluc: Thanks! I'll apply this suggestion on #2242 so the change doesn't clear your approval.

PR #2187, opened 2024-12-12
Reposition version warning after page load

### LINKED ISSUE #2152: Sticky CSS for "This documentation is for an old version of OpenDP" warning
If you open an anchor link you won't see it. Can we specify the `position` so it will always be at the top of the window?

(I think the nav bar at the top has changed, and for some older versions was much longer, so should probably not try to inject anything there.)

mccalluc:
- Fix #2152

(I spent a little time trying to get a pure-css solution, but I had some trouble because of wrapping elements around this div.)


Shoeboxam (approved):
Looks good!

PR #2189, opened 2024-12-12
feat: if randomgen is missing, prompt for scikit-learn

### LINKED ISSUE #1975: Missing randomgen error message should ask for scikit-learn

mccalluc:
- Fix #1975, I think?


Shoeboxam (changes requested):
I'm not sure why the issue asks for scikit-learn. I think it should ask for numpy, as numpy is the minimal feature that adds randomgen.

Shoeboxam (approved):
Thanks!

PR #2192, opened 2024-12-13
feat: Integrate logger

### LINKED ISSUE #2180: Integrate `opendp-logger` into this repo
- Follow-up for https://github.com/opendp/opendp-logger/issues/10

It was initially separate, I think, because it added some complexity, but we've seen cases where it would be useful, and having it integrated would nudge us to better document it, and to better support it in the main codebase. One possible path would be:
- Copy and paste the code and tests into this repo
- Add docs
- Update the opendp-logger README with a deprecation notice.
- Push last version to pypi with notice
- Archive opendp-logger.

but:
- We aren't going to support cross-version compatibility on these serializations, so #2103 should be a prereq, so we can warn if we try to deserialize a serialization coming from a different version.
- Right now, it's enabled with `enable_logging()`... but could we make this `enable_features("logging")` instead, just to keep the interface simple?
- ... unless we just want to have it on by default? Instead of decorating and redefining all these functions at runtime, what if we just made this part of the generated code? 
  - Or go in a different direction, and move the logic into a [custom JSONEncoder](https://docs.python.org/3/library/json.html#json.JSONEncoder).
- Is the interface what we want going forward? 
  - Was supporting pickle considered?
  - Or is part of the story behind this feature the idea that users can use it to construct transformations as data structures which we can then deserialize, and a certain amount of human-writability, if not -readability, is desired?
  - In which case, would it make sense to have the interface return or read a dict, and leave serialization to JSON to the user?

mccalluc:
- Fix #2180

Draft
- Would like to figure out the right way to add `__version__`: #2139
- Needs docs
- Needs to be added to prelude
- Maybe add this to `enable_feature`?
- Test error behavior: what does it do with Context or Polars?

Before moving this out of draft, I'd like to try other approaches: Pickle or JSONEncoder. Before committing this to the library I'd like to have more confidence that this API won't cramp us.


mccalluc:
The one example In the old repo used the deprecated dataframe API. Trying to make a new example I'm hitting errors. I'm not sure if it's a bug, or if what I'm trying to do is just out of scope. My guess is that the logic that that identifies the functions to wrap may be out of date, but I'm not sure, and I could use some help understanding how this is intended to work, in any case.

There are also lines of code that weren't exercised in the original test, and it would be good to extend test coverage over them as well.

mccalluc:
The example should work.

The wrapping is probably happening, but the prelude has the unwrapped one. (On a second import it would probably work?)

For now, change the import order in the example and see if it can work and give us a green checkmark.

Then, think about how this should actually work?
- Have `enable_logging` overwrite the prelude when logging turned on?
- Updating the codegen seems less bad than it did earlier. Perhaps less complex, by some metric.

mccalluc:
Close in favor of 
- #2228

Adding it to the code generation will be easier to understand than redefinition.

PR #2197, opened 2024-12-13
docs: update venv instructions with more descriptive names

### LINKED ISSUE #2195: `.venv-docs` and `.venv-src`
Right now, the instructions have you create `.venv` and `venv`, and I know which is which, but it would be better just to have explicit names for these to avoid confusion.

mccalluc:
- Fix #2195

Low priority!

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -4,10 +4,13 @@
 
 [flake8]
 exclude =
-    # suggested virtualenv for python/
-    .venv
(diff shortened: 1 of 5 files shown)


Shoeboxam (approved):
Nice to have a name that isn't just .venv, it clears up a potential confusion between the command and name.

PR #2199, opened 2024-12-17
Polars 0.12 polish

### LINKED ISSUE #2198: Polars: 0.12 polish

Shoeboxam:
Close #2198

* Adds some missing descriptor FFI and domains for the new categorical and temporal dtypes.
* Adds ability for Polars Context to recognize new temporal dtypes from schema
* Enforce that grouping by a categorical dtype requires known encoding, so that row ordering is not leaked via data-dependent encoding
* `Type` registry now has separate cfg list, instead of mocking types to be added to the `Type` registry
* Dis-allow Categorical from fill_null
* Drop all data descriptors when applying `not_`
* Important: check that input data to sum is not null and not nan
* Fix expected output dtype of converting Date to physical.


Shoeboxam:
* **#2200** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2200?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2199** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2199?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/opendp/opendp/2199?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rust/src/domains/polars/datetime/ffi.rs
```diff
@@ -0,0 +1,53 @@
...
+    arguments(time_zone(default = b"null")),
+    generics(DI(suppress)),
+    returns(c_type = "FfiResult<AnyDomain *>")
+)]
+/// Construct an instance of `DatetimeDomain`.
+///
+/// Documentation on valid time zones can be found [in the Polars documentation](https://docs.pola.rs/user-guide/transformations/time-series/timezones/).
+///
+/// # Arguments
+/// * `time_unit` - One of `ns`, `us` or `ms`, corresponding to nano-, micro-, and milliseconds
```
mccalluc, line 24: If it's possible with the macro, would we want to provide a default? (I would lean towards milliseconds.)

rust/src/measurements/make_private_lazyframe/group_by/mod.rs
```diff
@@ -70,6 +70,18 @@ where
...
 
     let by = match_grouping_columns(keys.clone())?;
+
+    by.iter().try_for_each(|name| {
+        let series_domain = middle_domain.series_domain(name.clone())?;
+        let Ok(domain) = series_domain.element_domain::<CategoricalDomain>() else {
+            return Ok(())
+        };
+        if domain.encoding().is_none() {
+            return fallible!(MakeMeasurement, "Categorical encoding is data-dependent, which may reveal sensitive record ordering. Convert {} to string before grouping.", name);
```
mccalluc, line 80:
```suggestion
            return fallible!(MakeMeasurement, "Categorical encoding is data-dependent, which may reveal sensitive record ordering. Cast {} to string before grouping.", name);
```

Just to be consistent with the method they will need to use.

rust/src/measurements/make_private_lazyframe/group_by/pseudocode/make_private_group_by.py
```diff
@@ -12,18 +12,18 @@ def make_private_group_by(
     t_prior = input_expr.make_stable(input_domain, input_metric)  # |\label{line:tprior}|
     middle_domain, middle_metric = t_prior.output_space()
 
-    grouping_columns = match_grouping_columns(keys)  # |\label{line:groupcols}|
+    by = match_grouping_columns(keys)  # |\label{line:groupcols}|
```
mccalluc, line 15: Very slight preference to keep the longer name for clarity.

rust/src/transformations/make_stable_expr/expr_binary/mod.rs
```diff
@@ -52,13 +52,19 @@ where
...
+    let right_series = &t_right.output_domain.column;
 
-    let left_nullable = t_left.output_domain.column.nullable;
-    let right_nullable = t_right.output_domain.column.nullable;
+    if matches!(left_series.dtype(), DataType::Categorical(_, _)) {
+        return fallible!(MakeTransformation, "{} cannot be applied to categorical data, because it may trigger a data-dependent CategoricalRemappingWarning in Polars", op);
+    }
+    if matches!(right_series.dtype(), DataType::Categorical(_, _)) {
+        return fallible!(MakeTransformation, "{} cannot be applied to categorical data, because it may trigger a data-dependent CategoricalRemappingWarning in Polars", op);
+    }
```
mccalluc, line 63: Would it make sense to combine these two with an `||` so the return doesn't need to be repeated?

Hoping that a python user wouldn't be able to hit these errors?

rust/src/transformations/make_stable_expr/expr_boolean_function/mod.rs
```diff
@@ -81,13 +81,12 @@ where
         data_column.nullable = false;
     }
 
-    if matches!(bool_function, Not) && data_column.dtype() != DataType::Boolean {
-        // under these conditions, the expression performs a bitwise negation
-        data_column.drop_bounds()?;
+    data_column.set_dtype(if matches!(bool_function, Not) {
+        // under these conditions, the expression performs a bitwise negation and all descriptors are dropped
+        data_column.dtype()
```
mccalluc, line 86: So just by itself this statement:
```
data_column.set_dtype(data_column.dtype())
```
would drop descriptors? I'm a little surprised that it changes anything. Might be working correctly, but the explicit `drop_bounds` in the previous state seems more clear.

rust/src/transformations/make_stable_expr/expr_sum/mod.rs
```diff
@@ -49,9 +49,30 @@ where
 
     let (by, input_margin) = middle_domain.context.grouping("sum")?;
 
+    if middle_domain.column.nullable {
+        return fallible!(
+            MakeTransformation,
+            "input data might contain nulls. Preprocess your data with `.fill_null`."
```
mccalluc, line 55: Should the column name be included in the message? It might not be obvious downstream.

rust/src/transformations/make_stable_expr/expr_sum/mod.rs
```diff
@@ -49,9 +49,30 @@ where
...
+    let nan = match dtype {
+        Float32 => middle_domain.column.atom_domain::<f32>()?.nullable(),
+        Float64 => middle_domain.column.atom_domain::<f64>()?.nullable(),
+        _ => false,
+    };
+
+    if nan {
+        return fallible!(
+            MakeTransformation,
+            "input data might contain nans. Preprocess your data with `.fill_nan`."
```
mccalluc, line 72: Include column name in message?

mccalluc (commented):
Lots of little comments. 

Biggest concern is about casting: My understanding is that the API is going to be a little more restrictive now with the column casting required. We should make sure that the error message you'd get with old code is very easy to understand, and that the fix is clear. If there are multiple columns that need casting, would it be possible to collect them all, instead giving the user just one column error at a time?

docs/source/getting-started/tabular-data/grouping.ipynb
```diff
@@ -858,6 +858,8 @@
     "\n",
     "query_work_hours = (\n",
     "    context.query()\n",
+    "    .with_columns(HWUSUAL=pl.col.HWUSUAL.cast(int))\n",
+    "    .filter(pl.col(\"HWUSUAL\") != 99)\n",
```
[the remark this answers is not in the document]
Shoeboxam: I backed out this change, I shouldn't have made it because it breaks the invariant above, and is already covered in the data preprocessing.

python/test/test_polars.py
```diff
@@ -779,6 +780,27 @@ def test_categorical_domain():
             )
 
 
+def test_categorical_domain_with_encoding():
+    pl = pytest.importorskip("polars")
+
+    lf = pl.LazyFrame([pl.Series("A", ["Texas", "New York", None], dtype=pl.Categorical)])
+
+    encoding = ["Texas", "New York"]
```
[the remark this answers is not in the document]
Shoeboxam: changed to categories

python/test/test_polars.py
```diff
@@ -790,17 +812,15 @@ def test_categorical_context():
     context = dp.Context.compositor(
         data=lf,
         privacy_unit=dp.unit_of(contributions=1),
-        privacy_loss=dp.loss_of(epsilon=1.0),
+        privacy_loss=dp.loss_of(epsilon=1.0, delta=1e-6),
```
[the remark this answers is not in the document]
Shoeboxam: I added a check that it fails without casting, which corresponds to the previous code, and then tested a fix that involves casting.

rust/src/measurements/make_private_lazyframe/group_by/pseudocode/make_private_group_by.py
Shoeboxam: I'm going to hit this again before too long with expression grouping keys and expressions in margins.

docs/source/getting-started/tabular-data/grouping.ipynb
```diff
@@ -858,10 +858,8 @@
     "\n",
     "query_work_hours = (\n",
     "    context.query()\n",
-    "    .with_columns(HWUSUAL=pl.col.HWUSUAL.cast(int))\n",
-    "    .filter(pl.col(\"HWUSUAL\") != 99)\n",
```
mccalluc: Isn't it important to filter out the 99s? Or is that done elsewhere? We don't want them in the mean.

docs/source/getting-started/tabular-data/grouping.ipynb
```diff
@@ -875,7 +873,7 @@
     "The mean is computed by dividing separate sum and length estimates.\n",
     "Since the partition length is marked as public information, the length is released in the clear, without any noise.\n",
     "\n",
-    "Reading the table, the smallest suitable noise scale parameter for the sum is 5762. \n",
+    "Reading the table, the smallest suitable noise scale parameter for the sum is 7056. \n",
```
mccalluc, line 876: Preference to keep exact numbers out of the human-written text: It's really easy to re-evaluate the notebook and forget to make the update here.

python/test/test_polars.py
```diff
@@ -816,6 +817,17 @@ def test_categorical_context():
...
+    try:
+        (
+            context.query()
+            .group_by("B")
+            .agg(dp.len())
+            .release()
+        )
+        assert False, "unreachable"
+    except dp.OpenDPException as err:
+        assert "Categories are data-dependent" in str(err)
```
mccalluc, line 829: more idiomatic:
```
with pytest.raises(dp.OpenDPException, match=r'Categories are data-dependent'):
    context.query()
        .group_by("B")
        .agg(dp.len())
        .release()
```

mccalluc (commented):
Maybe a couple more tweaks. Thanks!

docs/source/getting-started/tabular-data/grouping.ipynb
mccalluc: (But this would be a slightly bigger re-write, so it doesn't need to be in this PR.)

docs/source/getting-started/tabular-data/grouping.ipynb
Shoeboxam: It's filtered above, before building the context. It's done that way to preserve the margin metadata

mccalluc (approved):
Thanks for the last tweak. Approved.

PR #2207, opened 2024-12-19
feat: Polars select transformation

Shoeboxam:
Close #1709

PR #2209, opened 2024-12-20
feat: Polars expressions in group_by context

### LINKED ISSUE #2208: Group by expressions

Shoeboxam:
Closes #2208


Shoeboxam:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/opendp/opendp/2209?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2209** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2209?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/opendp/opendp/2209?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

python/src/opendp/context.py
```diff
@@ -383,7 +383,7 @@ def compositor(
         split_evenly_over: Optional[int] = None,
         split_by_weights: Optional[list[float]] = None,
         domain: Optional[Domain] = None,
-        margins: Optional[MutableMapping[tuple[str, ...], Margin]] = None,
+        margins: Optional[Union[MutableMapping[tuple[str, ...], Margin], Sequence[tuple[tuple[Any, ...], Margin]]]] = None,
```
mccalluc, line 386: I worry about introducing unions, but if this is the right direction, the `:param:` doc below needs to be updated to explain the new option.

docs/source/getting-started/tabular-data/preparing-microdata.ipynb
```diff
@@ -69,7 +69,7 @@
        "  white-space: pre-wrap;\n",
        "}\n",
        "</style>\n",
-       "<small>shape: (4, 2)</small><table border=\"1\" class=\"dataframe\"><thead><tr><th>HWUSUAL</th><th>len</th></tr><tr><td>cat</td><td>u32</td></tr></thead><tbody><tr><td>&quot;(0, 20]&quot;</td><td>6407</td></tr><tr><td>&quot;(20, 40]&quot;</td><td>54209</td></tr><tr><td>&quot;(40, 60]&quot;</td><td>15472</td></tr><tr><td>&quot;(98, inf]&quot;</td><td>119814</td></tr></tbody></table></div>"
```
mccalluc, line 72: Preference to take some time early in this quarter to do some cleanup, including converting notesbooks to RST: Will make reviews a bit easier.

python/test/test_polars.py
```diff
@@ -48,16 +48,20 @@ def example_lf(margin=None, **kwargs):
...
         lf_domain = dp.with_margin(lf_domain, by=margin, **kwargs)
     return lf_domain, lf
 
 
 def test_expr_domain():
     series_domains, _ = example_series()
     dp.wild_expr_domain(series_domains)
-    dp.wild_expr_domain(series_domains, by=["A", "B"])
-    dp.wild_expr_domain(series_domains, by=["A", "B"], max_num_partitions=10)
-    dp.wild_expr_domain(series_domains, by=["A", "B"], public_info="keys")
```
mccalluc, line 60: Would it be good to retain the old test to make sure it continues to work as well?

mccalluc (commented):
I've only skimmed the Python code, and should come back to this later... maybe in January.
- I worry a little about unions in type signatures: It puts more of a burden on the code downstream to handle every possibility.
- I worry a little about changes in the tests: Unless the existing behavior is changing, could you add new assertions, but leave the old, to make sure the old functionality is still there?

python/src/opendp/context.py
Shoeboxam: Refactored to take a list of margins, and to include by in the margin.

python/src/opendp/context.py
```diff
@@ -395,26 +394,32 @@ def compositor(
...
+            # allows dictionaries of {[by]: [margin]}
+            if isinstance(margins, MutableMapping):
+                margins = [replace(margin, by=by) for by, margin in margins.items()]
             
             pl = import_optional_dependency("polars")
-            for by, margin in iter_margins:
-                if not isinstance(by, tuple):
-                    msg = by if isinstance(by, str) else "your-column"
+            for margin in margins:
+                if not isinstance(margin.by, Sequence) or isinstance(margin.by, str):
```
mccalluc, line 417: I wonder if we want to be stricter here: I know that it can be confusing when a string is used when a list is expected, and you get a set of single characters.

Maybe this should be a `set` instead? I think the semantics would be closer to what we want (order doesn't matter, columns can be given twice, no closing comma that is easy to forget.)

mccalluc (commented):
May `by` should be a `set`?

python/src/opendp/context.py
Shoeboxam: Polars expressions are not hashable, unfortunately.

mccalluc (approved):
Looking just at the latest changes, seems good to me: Thanks for the change. "Grouping" is a slightly shorter / simpler word than "aggregation", but I don't have a strong preference.

Not really about this PR, but is it meaningful or useful to specify a margin on the same set of columns more than once, but perhaps with different public info on each instance? If it's not already, should it be an error?

Shoeboxam:
I switched to "aggregation" because grouping might imply expressions in group_by. But expressions in group_by don't belong in grouping, they belong in row by row.

Shoeboxam:
You also can't add the same margin twice:
https://github.com/opendp/opendp/blob/b6249cb3233e02eaf4709c02e50458573a34079d/rust/src/domains/polars/frame/mod.rs#L222

Shoeboxam:
There are settings where margin descriptors can be "at odds", though. I always just take the tightest inferrable descriptor.

For instance, public length when grouping by A, B, but only public key when grouping by A. OpenDP will infer grouping by A also has public lengths.

PR #2210, opened 2024-12-20
feat: Polars select/with_columns postprocessors

### LINKED ISSUE #2183: Polars: select postprocessor
SQL queries parsed by Polars always have a `.select` postprocessor that selects the columns of interest.

Shoeboxam:
Close #2183


Shoeboxam:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/opendp/opendp/2210?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2210** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2210?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/opendp/opendp/2210?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2209** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2209?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

rust/src/measurements/make_private_lazyframe/postprocess/test.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    assert!(make_private_lazyframe(
+        lf_domain,
+        SymmetricDistance,
+        MaxDivergence::default(),
+        query,
+        None,
+        None,
+    )
+    .is_err());
```
mccalluc, line 73: In the python code, I'd expect a stricter assertion in a test like this,
- to make sure we're getting the error we expect,
- and to make sure the error message is readable.

I'm not certain if that needs to apply here, though.

mccalluc (commented):
Would you have any interest in implementing a test coverage check in Rust in January? There are a few different execution paths here, and nothing looks suspicious, but it would be something that could give us more confidence.

rust/src/measurements/make_private_lazyframe/postprocess/test.rs
```diff
@@ -0,0 +1,81 @@
...
+        HashSet::from([col("B")]),
+        Margin::default()
+            .with_public_lengths()
+            .with_max_partition_length(5),
+    )?;
+    let lf = df!("A" => [0, 1, 1], "B" => [true, false, true])?.lazy();
+
+    let query = lf
+        .group_by([col("B")])
+        .agg([col("A").dp().sum((0, 1), Some(1.))])
```
mccalluc, line 32: Since this much is the same between the two tests, consider pulling it out, to emphasize what is different?

rust/src/measurements/make_private_lazyframe/postprocess/test.rs
```diff
@@ -0,0 +1,81 @@
...
+        .select([col("A").head(None)]);
```
mccalluc, line 64: Is this a pattern that end users might use? On the [python side](https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.head.html), it doesn't look like the argument is optional, so it might cause problems, though maybe just with type checking.

mccalluc (approved):
Perhaps a tweak to the test, but seems ok? (Just looking at the whole PR; diff since last review isn't available.)

rust/src/measurements/make_private_lazyframe/postprocess/test.rs
Shoeboxam: Well, taking the first five groups breaks the privacy guarantee, so it isn't something we can allow. But users can do `.release().collect().select([pl.col("A").head()])` instead. 

The argument is optional- there is a default of 5? Also those are docs for the DataFrame, not for Expr, but it doesn't matter!

mccalluc (approved):
Thanks. Looks good.

PR #2218, opened 2025-01-03
docs: fix PCA API docs

### LINKED ISSUE #2217: `opendp.extras.sklearn.decomposition.PCA` not in API docs
`opendp.extras.sklearn.decomposition` only covers `PCAEpsilons` and `make_private_pca`. My guess is that docs for `PCA` aren't generated because it's a subclass of a class that is only determined at runtime. Possible fixes:
- Add sklearn to the docs venv, so the superclass exists at doc generation time?
- Instead of a dynamic base class, try changing the base class dynamically? 
- Or redefining the class if extras are installed?
- Rewrite source code at doc generation time to hide the dynamic superclass? (This is probably a bad idea!)

Similar issue was closed when we made the module public: That was one step in the right direction.
- #1798:

mccalluc:
- Fix #2217

Instead of making the base class dynamic (which causes trouble for Sphinx), make it just a regular class, but update the class hierarchy if the optional install is installed.

The core is getting rid of 
```python
_decomposition = import_optional_dependency('sklearn.decomposition', False)
if _decomposition is not None:
    class _SKLPCA(_decomposition.PCA): # type: ignore[name-defined]
        ...
else: # pragma: no cover
    class _SKLPCA(object): # type: ignore[no-redef]
        ...
```
and replacing it with
```
_decomposition = import_optional_dependency('sklearn.decomposition', False)
if _decomposition is not None:
    PCA = type('PCA', (_decomposition.PCA,), dict(PCA.__dict__))  # type: ignore[assignment,misc]
```

Since it's now documented by Sphinx, fill in docstring a little: could still use more documentation on parameters, but this is a start.

### DIFF
--- a/python/src/opendp/extras/sklearn/decomposition/__init__.py
+++ b/python/src/opendp/extras/sklearn/decomposition/__init__.py
@@ -165,11 +165,13 @@ class PCA():
         n_changes: int = 1,
         whiten: bool = False,
     ) -> None:
-        if self.__class__.__bases__ == (object,):
-            raise ImportError(
-                "The optional install scikit-learn is required for this functionality"
-            )
-        super().__init__(
+        # Error if constructor called without depency installed:
+        import_optional_dependency('sklearn.decomposition')
+        
+        # The zero-argument form of super() does not work,
+        # (I believe) because the type argument is determined lexically,
+        # and it doesn't see our redefinition.
+        super(PCA, self).__init__(
             n_components or n_features,
             whiten=whiten,
         )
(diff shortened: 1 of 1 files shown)


Shoeboxam (approved):
I was not aware of the three-parameter functionality of `type`. Of separate interest, would this be a better way to define the LazyFrameQuery class?

mccalluc:
> I was not aware of the three-parameter functionality of `type`. Very slick. Of separate interest, would this be a better way to define the LazyFrameQuery class?

Don't know, but I just filed #2231. Thanks!

PR #2223, opened 2025-01-07
feat: expression filtering

### LINKED ISSUE #899: Add `make_filter`
- also related to timing attacks

Shoeboxam:
Closes #899

### DIFF
--- a/docs/source/api/user-guide/polars/expressions/manipulation.ipynb
+++ b/docs/source/api/user-guide/polars/expressions/manipulation.ipynb
@@ -376,6 +376,38 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "## Drop NaNs\n",
+    "`.drop_nans` removes NaN float values from the data. Not to be confused with `.drop_nulls`. \n",
+    "While this proves that the data does not have NaNs, it also drops data invariants on dataset size.\n",
+    "This expression is not row-by-row, so may only be used in an aggregation context.\n",
+    "\n",
+    "See `.drop_nulls` for a usage example.\n",
+    "\n",
+    "## Drop Nulls\n",
+    "`.drop_nulls` removes null values from the data. Not to be confused with `.drop_nans` for float data. \n",
+    "While this proves that the data does not have nulls, it also drops data invariants on dataset size.\n",
+    "This expression is not row-by-row, so may only be used in an aggregation context.\n",
+    "\n",
+    "In the following example, all nans and nulls are dropped from the `HWUSUAL` column."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "(\n",
+    "    context.query()\n",
+    "    .select(pl.col.HWUSUAL.drop_nans().drop_nulls().dp.sum((0, 100)))\n",
+    "    .summarize()\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
     "\n",
     "## Fill NaN\n",
     "`.fill_nan` replaces NaN float values. Not to be confused with `.fill_null`.\n",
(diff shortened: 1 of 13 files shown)


rust/src/transformations/make_stable_expr/expr_drop_nan_or_null/test.rs
```diff
@@ -0,0 +1,42 @@
...
+    let series_domain = SeriesDomain::new("", AtomDomain::<f32>::default());
+    let lf_domain = LazyFrameDomain::new(vec![series_domain])?.with_margin(
+        HashSet::new(),
+        Margin::default().with_max_partition_length(5),
+    )?;
+    let lf = df!("" => &[1.0, f32::NAN])?.lazy();
+
+    let lf_filter =
+        lf.clone()
+            .select([col("").drop_nans().drop_nulls().dp().sum((0.0, 1.0), None)]);
```
mccalluc, line 26: I see the NaN, but does the `drop_nulls` here actually do anything? Would it make sense to test `drop_nans` and `drop_nulls` separately?

rust/src/transformations/make_stable_expr/expr_fill_nan/mod.rs
```diff
@@ -33,12 +33,19 @@ where
         return fallible!(MakeTransformation, "expected fill_nan expression");
     };
 
+    // only enforce row-by-row context if the fill expression is not broadcastable
+    let expr_domain = if fill.clone().meta().root_names().len() > 0 {
+        input_domain.as_row_by_row()
+    } else {
+        input_domain.clone()
+    };
```
mccalluc, line 41: Only because there isn't a test that has been updated right beside this file, I'm wondering if both cases are tested?

docs/source/api/user-guide/polars/expressions/manipulation.ipynb
```diff
@@ -372,6 +372,38 @@
     "The output type is categorical, but with a data-independent encoding, meaning OpenDP allows grouping by these keys."
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Drop NaNs\n",
+    "`.drop_nans` removes NaN float values from the data. Not to be confused with `.drop_nulls`. \n",
+    "While this proves that the data does not have NaNs, it also drops data invariants on dataset size.\n",
```
mccalluc, line 381: I think I know what you mean, but "proves" feels odd. Could you say something like "This will guarantee that the downstream data will not have NaNs"?

docs/source/api/user-guide/polars/expressions/manipulation.ipynb
```diff
@@ -435,6 +467,10 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "If the fill expression is not broadcast-able (references other columns), \n",
```
mccalluc, line 470: If we don't rely on the notion of "broadcasting" down the road, could we just say "If the fill expression does not reference other columns ..."?  It feels like we're introducing jargon that isn't necessary.

mccalluc (commented):
As usual, some comments on docs and tests, but I didn't really get into the rust code and understand it. Let me know if there's anything particular I should focus on in a second pass.

rust/src/transformations/make_stable_expr/expr_drop_nan_or_null/test.rs
Shoeboxam: OpenDP assumes data could be nan or null. Dropping nulls leaves nans, and vice versa. Good point about splitting the tests. I added more tests.

rust/src/transformations/make_stable_expr/expr_fill_nan/mod.rs
Shoeboxam: Added.

rust/src/transformations/make_stable_expr/expr_fill_nan/mod.rs
```diff
@@ -36,8 +36,16 @@ where
     // only enforce row-by-row context if the fill expression is not broadcastable
     let expr_domain = if fill.clone().meta().root_names().len() > 0 {
         input_domain.as_row_by_row()
-    } else {
+    } else if let Expr::Literal(value) = fill.clone() {
+        if !value.is_scalar() {
+            return fallible!(MakeTransformation, "fill expression must be broadcastable");
```
mccalluc, line 41: You rephrased this in the docs, but "broadcastable" in an internal error message is ok with me, if you want to keep it.

mccalluc (approved):
Doing a quick skim this time, but the changes look good. Thanks!

PR #2224, opened 2025-01-07
feat: polars enum dtype

Shoeboxam:
Closes #1721

PR #2222, opened 2025-01-07
feat: Polars replace

Shoeboxam:
Closes #1752

PR #2225, opened 2025-01-07
feat: polars array dtype

### LINKED ISSUE #2221: Polars: support array dtypes

Shoeboxam:
Closes #2221

### DIFF
--- a/python/src/opendp/extras/polars/__init__.py
+++ b/python/src/opendp/extras/polars/__init__.py
@@ -34,6 +34,7 @@ from opendp.domains import (
     categorical_domain,
     datetime_domain,
     enum_domain,
+    array_domain,
 )
 from opendp.measurements import make_private_lazyframe
 
@@ -609,16 +610,22 @@ def _lazyframe_domain_from_schema(schema) -> Domain:
 
 def _series_domain_from_field(field) -> Domain:
     """Builds the broadest possible SeriesDomain that matches a given field."""
+    name, dtype = field
+    return series_domain(name, option_domain(_domain_from_dtype(dtype)))
+
+
+def _domain_from_dtype(dtype) -> Domain:
+    """Builds the broadest possible Domain that matches a given dtype."""
     import polars as pl
 
-    name, dtype = field
     if dtype == pl.Categorical:
-        return series_domain(name, option_domain(categorical_domain()))
+        return categorical_domain()
     if dtype == pl.Enum:
-        return series_domain(name, option_domain(enum_domain(dtype.categories)))
+        return enum_domain(dtype.categories)
     if dtype == pl.Datetime:
-        dt_domain = datetime_domain(dtype.time_unit, dtype.time_zone)
-        return series_domain(name, option_domain(dt_domain))
+        return datetime_domain(dtype.time_unit, dtype.time_zone)
+    if dtype == pl.Array:
+        return array_domain(_domain_from_dtype(dtype.inner), dtype.size)
 
     T = {
         pl.UInt32: "u32",
@@ -638,9 +645,7 @@ def _series_domain_from_field(field) -> Domain:
     if T is None:
         raise ValueError(f"unrecognized dtype: {dtype}")  # pragma: no cover
 
-    element_domain = option_domain(atom_domain(T=T, nullable=T in {"f32", "f64"}))
-    return series_domain(name, element_domain)
-
+    return atom_domain(T=T, nullable=T in {"f32", "f64"})
 
 _LAZY_EXECUTION_METHODS = {
     "collect",
--- a/python/test/test_polars.py
+++ b/python/test/test_polars.py
@@ -1146,5 +1146,17 @@ def test_array_domain():
     # check that domain is as expected
     assert observed == expected
 
+@pytest.mark.xfail(reason="broken until https://github.com/pola-rs/polars/issues/20162 is fixed")
+def test_array_domain_query():
+    pl = pytest.importorskip("polars")
+
+    # this triggers construction of a lazyframe domain from the schema
+    context = dp.Context.compositor(
+        data=pl.LazyFrame(pl.Series("alpha", [["A", "B", "C"]] * 100, dtype=pl.Array(pl.String, 3))),
+        privacy_unit=dp.unit_of(contributions=1),
+        privacy_loss=dp.loss_of(epsilon=1.0, delta=1e-7),
+        split_evenly_over=1,
+    )
+    
     # this is broken until https://github.com/pola-rs/polars/issues/20162 is fixed
-    # context.query().with_columns(pl.col.alpha.explode()).select(dp.len()).release().collect()
+    context.query().with_columns(pl.col.alpha.explode()).select(dp.len()).release().collect()
--- a/python/test/test_polars.py
+++ b/python/test/test_polars.py
@@ -1145,3 +1145,6 @@ def test_array_domain():
 
     # check that domain is as expected
     assert observed == expected
+
+    # this is broken until https://github.com/pola-rs/polars/issues/20162 is fixed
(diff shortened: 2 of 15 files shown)


Shoeboxam:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/opendp/opendp/2225?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2225** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2225?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/opendp/opendp/2225?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2224** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2224?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2223** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2223?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2222** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2222?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2210** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2210?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2209** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2209?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

python/test/test_polars.py
```diff
@@ -1127,3 +1127,22 @@ def test_enum_domain():
...
+    # this triggers construction of a lazyframe domain from the schema
+    context = dp.Context.compositor(
+        data=pl.LazyFrame(pl.Series("alpha", [["A", "B", "C"]] * 100, dtype=pl.Array(pl.String, 3))),
+        privacy_unit=dp.unit_of(contributions=1),
+        privacy_loss=dp.loss_of(epsilon=1.0, delta=1e-7),
+        split_evenly_over=1,
+    )
+    observed = context.accountant.input_domain
+    expected = dp.lazyframe_domain([
+        dp.series_domain("alpha", dp.option_domain(dp.array_domain(dp.atom_domain(T=str), 3))),
```
mccalluc, line 1144: To start the review, just copied this test out and ran it without any other changes, to understand what this PR gives us.

Not obvious why it should be wrapped in an option domain, but probably fine.

python/src/opendp/extras/polars/__init__.py
```diff
@@ -611,16 +612,22 @@ def _lazyframe_domain_from_schema(schema) -> Domain:
 
 def _series_domain_from_field(field) -> Domain:
     """Builds the broadest possible SeriesDomain that matches a given field."""
+    name, dtype = field
+    return series_domain(name, option_domain(_domain_from_dtype(dtype)))
```
mccalluc, line 616: Follow-up to the comment on the test: This is where the `option_domain` comes from, but I don't really understand why it's needed.

rust/src/domains/polars/array/mod.rs
```diff
@@ -0,0 +1,53 @@
...
+
+/// A domain that represents enum data.
+///
+/// Enum data is ostensibly a string,
+/// however the data is stored as a vector of indices into an encoding.
+/// This gives memory speedups when the number of unique values is small.
+///
+/// Differs from the ArrayDomain in that the categories are fixed and known.
+#[derive(Clone, Debug)]
+pub struct ArrayDomain {
```
mccalluc, line 20: Did you mean to change the name of the struct to something like `EnumDomain`? If this is the right name, maybe more explanation about how this `ArrayDomain` differs from some other `ArrayDomain`?

rust/src/measurements/make_private_expr/expr_report_noisy_max/mod.rs
```diff
@@ -101,6 +101,26 @@ where
         );
     }
 
+    use DataType::*;
+    if let DataType::Array(dtype, _) = middle_domain.column.dtype() {
```
mccalluc, line 105:
```suggestion
    if let DataType::Array(dtype, _) != middle_domain.column.dtype() {
```
Slight preference to flatten the nested ifs, for readability: First one check, and fail if needed, and then a second check, and another fail.

rust/src/transformations/make_stable_expr/namespace_arr/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+        function: FunctionExpr::ArrayExpr(array_function),
+        ..
+    } = &expr
+    else {
+        return fallible!(MakeTransformation, "expected array expression");
+    };
+
+    fallible!(
+        MakeTransformation,
+        "Expr is not recognized at this time: {:?}. Waiting for: https://github.com/pola-rs/polars/pull/20421",
```
mccalluc, line 39: Will it be a problem if that doesn't get addressed? If the maintainers aren't interested, would you take a different approach?

Would it be worth adding an expected-failure test that exercises this branch? I'm not sure exactly what it is that is blocked.

mccalluc (commented):
Only thing that seems like it might be a mistake is the naming around 
> Differs from the ArrayDomain in that the categories are fixed and known.

but I might just be confused.

python/test/test_polars.py
Shoeboxam: This is for a dataframe column where each row is an array-- like the score vector emitted by the discrete quantile scoring transformation. Rows can be null instead of being an array. OpenDP makes as few assumptions about the data as possible, therefore it assumes that rows can be null instead of being an array.

python/src/opendp/extras/polars/__init__.py
Shoeboxam: Yep, input data could have null rows (or at least, a potential adjacent dataset could have a new record with null in the array column).

rust/src/domains/polars/array/mod.rs
Shoeboxam: Updated docs. These docs should talk about ArrayDomain.

rust/src/measurements/make_private_expr/expr_report_noisy_max/mod.rs
Shoeboxam: Switched to a let-else binding.

rust/src/transformations/make_stable_expr/namespace_arr/mod.rs
Shoeboxam: If it doesn't get addressed then I'll have to pick apart the debug stringification. :/

python/test/test_polars.py
```diff
@@ -1145,3 +1145,6 @@ def test_array_domain():
 
     # check that domain is as expected
     assert observed == expected
+
+    # this is broken until https://github.com/pola-rs/polars/issues/20162 is fixed
+    # context.query().with_columns(pl.col.alpha.explode()).select(dp.len()).release().collect()
```
mccalluc, line 1150: Would prefer an xfail in a case like this, so if we upgrade, and it starts working, we'll be prompted to update the test to reflect the new, correct behavior.

mccalluc (approved):
Just looking at the last commit, seems good. I like `xfail`, but not a requirement.

mccalluc (commented):
Thanks for `xfail`. I'll rerun the tests, in case it is some kind of flakiness:

https://github.com/opendp/opendp/actions/runs/13668830906/job/38215436358#step:5:143
```
error: linking with `cc` failed: exit status: 1
```

restarted...

PR #2228, opened 2025-01-08
feat: Opendp object serialization

### LINKED ISSUE #2103: Add `__version__`

### LINKED ISSUE #2180: Integrate `opendp-logger` into this repo

mccalluc:
- Replace #2192 
- Fix #2180
- Fix #2103 
- Follow-up issue: #2246

This takes a different approach to the serialization of OpenDP objects, which at least at first seems like it might be simpler.

- Tweak the python.rs code gen to add a dictionary on objects.
  - The original code had lists of the classes and modules to wrap. At least for now I seem to be getting by without that.
  - Calling the dictionary `__opendp_dict__` instead of `log`, just to make it clear that this is special, and internal.
- Build on stdlib json serialization/deserialization, with the details in private functions
- list vs. tuple: seems simpler to let lists flow straight through and treat tuples as the special case.


mccalluc:
Talked with Mike, and this seems like a good direction, though it still needs a lot of work. Removing Mike as reviewer for now.

mccalluc:
A lot more coverage.
- What is still missing?
- With `__opendp_dict__` being more widely available, maybe `__eq__` methods should use it instead?
- Polars serialization: Does it make sense to automatically replace the location of the binary just before deserialization? Or maybe it should be stripped before serialization?

python/src/opendp/context.py
```diff
@@ -367,6 +367,21 @@ def __init__(
         self.d_mids = d_mids
         self.d_out = d_out
         self.space_override = space_override
+        self.__opendp_dict__ = {
+            'func': 'Context',
+            'module': 'context',
+            'kwargs': {
+                'accountant': accountant,
+                'queryable': queryable,
```
mccalluc, line 375: This has internal state that can't be serialized.

same for accountant.

python/test/test_serialization.py
```diff
@@ -0,0 +1,76 @@
...
+                privacy_loss=dp.loss_of(epsilon=1.0),
+                split_evenly_over=1,
+            )
+        ]
+    ],
+)
+def test_serializable(_readable_name, dp_obj):
+    serialized = dp.serialize(dp_obj)
+    deserialized = dp.deserialize(serialized)
+    assert dp_obj == deserialized
```
mccalluc, line 48: test equality of serializations, rather than equality of objects.

python/src/opendp/mod.py
```diff
@@ -432,6 +436,17 @@ class Queryable(object):
...
         self.value = value
         self.query_type = query_type
+        self.__opendp_dict__ = {
+            'func': 'Queryable',
+            'module': 'mod',
+            'kwargs': {
+                'value': value,
+                'query_type': query_type,
+            },
+        }
```
mccalluc, line 446: In the future, some queryables may be serializable, but it would be universal. Worth adding a special case error message.

python/src/opendp/mod.py
```diff
@@ -651,6 +666,7 @@ def _depends_on(self, *args):
 class PartialConstructor(object):
     def __init__(self, constructor):
         self.constructor = constructor
+        self.__opendp_dict__ = {}
```
mccalluc, line 669: Might work if constructor was in the dict, but constructor not serializable.

python/src/opendp/mod.py
```diff
@@ -1122,4 +1141,49 @@ def exception_predicate(v):
...
+        module = importlib.import_module(f"opendp.{dp_dict['module']}")
+        func = getattr(module, dp_dict["func"])
+        return func(**dp_dict.get("kwargs", {}))
+    if _TUPLE_FLAG in dp_dict:
+        return tuple(dp_dict[_TUPLE_FLAG])
+    pl = import_optional_dependency('polars', raise_error=False)
+    if pl is not None and _POLARS_FLAG in dp_dict:
+        from io import StringIO
+        # TODO: update the references to the binary to point
+        # to the correct location in this environment.
```
mccalluc, line 1178: No need to change, but should comment in the docs about the envvar that controls this.

python/src/opendp/mod.py
```diff
@@ -1122,4 +1141,49 @@ def exception_predicate(v):
...
+        return pl.Expr.deserialize(StringIO(dp_dict[_POLARS_FLAG]), format="json")
+    return dp_dict
+
+def serialize(dp_obj):
+    return json.dumps(dp_obj, cls=_Encoder)
+
+def deserialize(dp_json):
+    return json.loads(dp_json, object_hook=_deserialization_hook)
```
mccalluc, line 1186: Try wrapping an opendp serialization in a dict, and make sure the whole datastructure can make the round trip.

rust/opendp_tooling/src/codegen/python.rs
```diff
@@ -200,8 +200,16 @@ def {then_name}(
 
 {doc_params}{example}
     """
-    return PartialConstructor(lambda {dom_met}: {name}(
+    output = PartialConstructor(lambda {dom_met}: {name}(
 {args}))
+    output.__opendp_dict__ = {{
+            'func': '{then_name}',
+            'module': 'measurements',
```
Shoeboxam, line 207: This is hardcoded when it should be based on the module, `module_name` is passed in above.

python/test/test_serialization.py
```diff
@@ -0,0 +1,76 @@
...
+            dp.m.approximate(dp.m.max_divergence()),
+            dp.m.user_divergence("user_divergence"),
+            # Measurements:
+            dp.m.make_gaussian(atom, dp.absolute_distance(int), 1),
+            dp.m.then_gaussian(1),
+            # Compositions:
+            (dp.vector_domain(dp.atom_domain(T=int)), dp.symmetric_distance())
+            >> dp.t.then_clamp((0, 10))
+            >> dp.t.then_sum()
+            >> dp.m.then_laplace(scale=5.0),
```
Shoeboxam, line 35: Should also test that serde works for amplification combinator. Given that this works for the chainer combinator, I am optimistic that this is already working as intended.

Shoeboxam (dismissed):
Great progress!

rust/opendp_tooling/src/codegen/python.rs
mccalluc:
```suggestion
            'module': '{module_name}',
```

python/src/opendp/mod.py
mccalluc: done! (or a tuple, just to kill two birds with one stone.)

python/src/opendp/mod.py
```diff
@@ -1122,4 +1141,49 @@ def exception_predicate(v):
...
+            return obj
+        if isinstance(obj, list):
+            return [self.default(value) for value in obj]
+        pl = import_optional_dependency('polars', raise_error=False)
+        if pl is not None and isinstance(obj, pl.Expr):
+            return {_POLARS_FLAG: obj.meta.serialize(format="json")}
+        raise Exception(f'OpenDP JSON Encoder does not handle {obj}')  # pragma: no cover
+    
+def _deserialization_hook(dp_dict):
+    if 'func' in dp_dict:
```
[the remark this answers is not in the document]
mccalluc: done.

python/src/opendp/mod.py
mccalluc: Done.

python/test/test_serialization.py
mccalluc: done.

python/src/opendp/context.py
mccalluc: removed this block.

python/src/opendp/mod.py
mccalluc: special message added, though not sure it shows in every desired case.

python/test/test_serialization.py
```diff
@@ -0,0 +1,76 @@
...
+            dp.Context.compositor(
+                data=[5.0] * 100,
+                privacy_unit=dp.unit_of(contributions=1),
+                privacy_loss=dp.loss_of(epsilon=1.0),
+                split_evenly_over=1,
+            )
```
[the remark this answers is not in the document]
mccalluc: Removed the code that tried to make context serializable, and added test that we get the expected error message. Need to add tests of more classses, including Queryable

python/test/test_serialization.py
mccalluc: added a test: worked out of the box.

python/src/opendp/mod.py
```diff
@@ -1122,4 +1141,49 @@ def exception_predicate(v):
...
+            }
+        if isinstance(obj, tuple):
+            return {_TUPLE_FLAG: obj}
+        if isinstance(obj, (str, int, float, bool, type(None))):
+            return obj
+        if isinstance(obj, list):
+            return [self.default(value) for value in obj]
+        pl = import_optional_dependency('polars', raise_error=False)
+        if pl is not None and isinstance(obj, pl.Expr):
+            return {_POLARS_FLAG: obj.meta.serialize(format="json")}
```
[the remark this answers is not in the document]
mccalluc: lazyframes covered, and switched from json to base64 encoded bytes.

docs/source/api/user-guide/utilities/serialization.rst
```diff
@@ -0,0 +1,74 @@
...
+        >>> import polars as pl
+        >>> import io
+
+        >>> expr = dp.len(scale=1.0)
+        >>> bytes = expr.meta.serialize()
+        >>> new_expr = pl.Expr.deserialize(bytes)
+        >>> type(expr)
+        <class 'polars.expr.expr.Expr'>
+        >>> type(new_expr)
+        <class 'polars.expr.expr.Expr'>
```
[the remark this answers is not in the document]
Shoeboxam: Query's _chain or _lf_plan is serializable, but the query itself is not/doesn't need to be. Using `dp.serialize` seems fine to me, that just checks isinstance Query and serializes the `_chain` or `_lf_plan`.

mccalluc:
Doesn't handle `LazyFrameQuery` yet, but I wonder if that needs to be a blocker: it is a more complicated case.

It feels weird saying that we're serializing the query if we're not capturing all of the state, but conversely if the field that we are _actually_ serializing isn't public, we can't really point the user to that. Some possibilities:
- Make these internals public if we are encouraging users to set them, or read from them?
- Special logic for serializing this object: actually capture the whole state, and not just the constructor arguments?
- Call it something other than "serialization"?

Other notes:

- Filed #2244, in part because I found I was doing a lot more imports inside blocks than usual, just because the top of the file was so far away.
- If tweaks might be made to `LazyFrameQuery`, would that be another prompt for #2236, or some other simplification.

python/src/opendp/mod.py
Shoeboxam: Does this need to exist at all?

python/test/test_serialization.py
```diff
@@ -0,0 +1,165 @@
...
+import pytest
+
+import opendp.prelude as dp
+from opendp._lib import import_optional_dependency
+
+atom = dp.atom_domain(bounds=(0., 10.))
+input_space = dp.vector_domain(atom, size=10), dp.symmetric_distance()
+chained = input_space >> dp.t.then_mean() >> dp.m.then_laplace(scale=0.5)
+
+# TODO: Make a fake serialization, and check version mismatch.
```
Shoeboxam, line 13: Could be upgraded to an issue.

Shoeboxam (approved):
Approving, but left a couple comments.

mccalluc:
> Does this need to exist at all?

Not at run time, but it prevents a lot of error from mypy: `"PartialConstructor" has no attribute "__opendp_dict__"`

> Could be upgraded to an issue.

Sorry: I should have cleared the TODO. This is covered by `test_version_mismatch_warning`

Let me fix both of these...

mccalluc:
@Shoeboxam , related to this, did you have any thoughts about pickling UDFs? There might be other reasons not to do that, but it seems like it could work.

Shoeboxam (approved):
I think it would be great to add a pickling feature. Be sure to put it behind honest-but-curious. Thanks for the updates!

PR #2229, opened 2025-01-09
feat: Example data in API

### LINKED ISSUE #2216: Move sample data inside API?
Would it make sense to add a method in extras that provides the sample data?

- Having sample data as part of the API is not unusual for a data-centric library.
- The data itself wouldn't need to be moved into this codebase: behind the scenes it would still do an HTTP GET.
- Having to do the shell-based `wget` to get the example steers us to write notebooks even when there aren't other compelling reasons (ie, visualizations)
- Since it would be in the API, we could use normal docs to describe the data, and it would be findable and consistent.

mccalluc:
- Fix #2216
- No shell commands in notebooks! More readable.
- Descriptions of examples can be found in our API docs
- More feasible to translate notebooks to RST.

I started with the helpers being StringIO, but I actually think now it makes more sense to cache the actual files instead of relying on state in the module. This way, each run should only require one download of each CSV, and the user doesn't need to see the details.


mccalluc:
Hit an error with a rust dependency:
```
Downloaded bytemuck_derive v1.9.0
error: failed to parse manifest at `/home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/bytemuck_derive-1.9.0/Cargo.toml`

Caused by:
  feature `edition2024` is required

  The package requires the Cargo feature called `edition2024`, but that feature is not stabilized in this version of Cargo (1.83.0 (5ffbef321 2024-10-29)).
  Consider trying a newer version of Cargo (this may require the nightly release).
  See https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#edition-2024 for more information about the status of this feature.
```
Maybe something is unpinned? Or maybe this will be hard to reproduce?

python/src/opendp/extras/examples/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+    This only does what we need. Does not handle:
+    - plain http
+    - redirects
+    - url queries
+    - explicit port numbers
+    '''
+    from urllib.parse import urlparse
+    import http.client
+    parsed = urlparse(url)
+    assert parsed.hostname is not None
```
Shoeboxam, line 28: Raise instead of assert is better?

Shoeboxam (approved):
One other benefit of this is the docstring associated with the function.

python/src/opendp/extras/examples/__init__.py
mccalluc: In general yes, but in this case:
- It's shorter
- I only added it to make mypy happy: It complains about the possibility of `HTTPSConnection(None)`.
- The input is something we control.
- It can make sense to use `assert` when it really is an assertion that just should be true, and not something that depends on outside input.

(But if you'd like to say that we just never use `assert`s because they can be confusing, I'd be on board with that.)

PR #2236, opened 2025-01-22
feat: Simplify Polars OO code

### LINKED ISSUE #2231: Refactor `LazyFrameQuery` with three-parameter `type`?
>  I was not aware of the three-parameter functionality of `type`. Very slick. Of separate interest, would this be a better way to define the LazyFrameQuery class?

_Originally posted by @Shoeboxam in https://github.com/opendp/opendp/pull/2218#pullrequestreview-2551476114_

I don't know, but worth investigating. FWIW: This is the only other place where we have alternate definitions for a class, and in that way it is similar to #2218:

```
$ egrep '^\s*class' -r src | perl -pne 's/\S+:\s*//; s/\(.*//' | sort | uniq --repeated
class LazyFrameQuery
```

mccalluc:
- Inspired by #2218
- Fix #2231

... but in this case it looks like we don't actually need anything from the superclass, so we could just drop it, and we don't need to update the type hierarchy or anything like this.

We do lose type annotations, because the types may not actually be available, but those weren't showing up in the docs in any case, so I'm not sure it's a big loss.


Shoeboxam (dismissed):
The gains we were getting from having separate definitions and subclassing Polars types were largely eliminated by Sphinx not being aware of the subclassing and spotty IDE support.

Thanks for simplifying this!

mccalluc:
Thanks! Resolving the merge conflicts is tricky: What may be easiest is not to worry about the details here, but take the current state in main, and then redo this. Working on it...

mccalluc:
- Replaced by #2251

PR #2238, opened 2025-01-22
docs: Convert notebooks to RST

### LINKED ISSUE #2205: More notebooks to RST
Most of the new Polars docs would benefit from being translated to RST.
- When we run the notebooks, we don't check that the output matches, except to make sure there are no errors.
- When we do want to demonstrate an error, it is much tidier in a doctest.
- The notebook run in CI is a separate action from smoke-test, while doctests in RST are run along with the rest of the tests.
- If we use Sphinx references instead of markdown links, we can catch bad links during the docs build.
- Smaller, easier to read diffs.

mccalluc:
- Fix #2205
- Simpler source code
- Doctests check that the output is what we expect
- Down the road, possible to tab between examples for different languages

The downside is that there are a lot of `...` where we used to have numbers, and perhaps there still should be more: It's nice to give some indication of what the number could be, but since it really could be anything, that makes the test kind of brittle.

Where we do just want to wildcard it, I'm printing something on the line before it. This doesn't work:
```
>>> random()
...
```
because pytest interprets the `...` as a continuation, and then nothing matches the actual output. So instead there are lines like:
```
>>> print('random:', random())
random: ...

### CODE REVIEW

**Review by Shoeboxam — CHANGES_REQUESTED** [review]
While the doctests are nice to have, IMO we care less about their testing functionality than readability. For this reason, it feels more desirable to me to skip the doctest output check in cases where the output is non-deterministic, than to strip out the nondeterministic outputs with ellipses for a slightly stronger test. This is essentially how things are done in the notebooks. 

While sphinx doesn't have a flag built-in for this, sphinx just calls out to doctest, which can be extended: https://stackoverflow.com/a/69780437/10221612

From this, the following snip inserted into `conf.py` should register the flag that can be used to disable the output checker:
```python
import doctest

IGNORE_RESULT = doctest.register_optionflag('IGNORE_RESULT')

OutputChecker = doctest.OutputChecker
class CustomOutputChecker(OutputChecker):
    def check_output(self, want, got, optionflags):
        if IGNORE_RESULT & optionflags:
            return True
        return OutputChecker.check_output(self, want, got, optionflags)

doctest.OutputChecker = CustomOutputChecker
```

I think this results in the ideal behavior where code gets executed, the doctest fails if an exception is raised, and the output check is skipped. This should allow us to use `# doctest: +IGNORE_RESULT` in cases where nondeterministic output is shown for documentation purposes.

mccalluc:
Sounds good. I think I'll see if it's possible to do something a little stricter, ie: wildcard just the numbers in the response, but still check that the rest of the structure is accurate... but if that gets tricky, I'll go with this. Thanks!

mccalluc:
I'm going to 
- close this,
- wait for #2245
- and then translate one notebook at a time, to avoid merge conflicts.

PR #2241, opened 2025-01-22
Fix reported broken links

mccalluc:
https://github.com/opendp/opendp/actions/runs/12848639062/job/35826297642

Two are actually fixed, but just removed the Spectus link, since there's no longer anything about this library on that site.


Shoeboxam (approved):
OpenDP is not mentioned as part of a data release, just "advertised"/linked from what is now the Cuebiq website.

PR #2242, opened 2025-01-22
docs: Check that every parameter has a brief description

mccalluc:
- Builds on #2182
- In some places, rename the parameter for clarity
- A lot of copy and pasting, but I haven't tried to make older docs consistent, so there is some variety that we may not want.


mccalluc:
```suggestion
        :param at_center: how the predicate evaluates at `center`. Search terminates when predicate changes
```

_Originally posted by @Shoeboxam in https://github.com/opendp/opendp/pull/2182#discussion_r1927649861_

Shoeboxam (approved):
Thanks for the docs and checks!

PR #2249, opened 2025-01-29
testing: Fuzzy doctests

### LINKED ISSUE #2245: Doctest flag: be less fussy
While the doctests are nice to have, IMO we care less about their testing functionality than readability. For this reason, it feels more desirable to me to skip the doctest output check in cases where the output is non-deterministic, than to strip out the nondeterministic outputs with ellipses for a slightly stronger test. This is essentially how things are done in the notebooks. 

While sphinx doesn't have a flag built-in for this, sphinx just calls out to doctest, which can be extended: https://stackoverflow.com/a/69780437/10221612

From this, the following snip inserted into `conf.py` should register the flag that can be used to disable the output checker:
```python
import doctest

IGNORE_RESULT = doctest.register_optionflag('IGNORE_RESULT')

OutputChecker = doctest.OutputChecker
class CustomOutputChecker(OutputChecker):
    def check_output(self, want, got, optionflags):
        if IGNORE_RESULT & optionflags:
            return True
        return OutputChecker.check_output(self, want, got, optionflags)

doctest.OutputChecker = CustomOutputChecker
```

I think this results in the ideal behavior where code gets executed, the doctest fails if an exception is raised, and the output check is skipped. This should allow us to use `# doctest: +IGNORE_RESULT` in cases where nondeterministic output is shown for documentation purposes.

_Originally posted by @Shoeboxam in https://github.com/opendp/opendp/pull/2238#pullrequestreview-2574633293_

mccalluc:
- Fix #2245
- Prereq to #2205

This might be needlessly complicated, but the idea is that instead of simply ignoring the result altogether, we can loosen it, so that if the RST has a number preceded by tilde, any number will work, but the rest of the output still needs to match.

FYI: Pytest has a related doctest extension: [NUMBER](https://docs.pytest.org/en/7.1.x/how-to/doctest.html?highlight=doctest#using-doctest-options).

> when enabled, floating-point numbers only need to match as far as the precision you have written in the expected doctest output

The [source code](https://github.com/pytest-dev/pytest/blob/2f369848ea0ead42182efc77000a5ecef629a0ba/src/_pytest/doctest.py#L654-L676) is interesting.

### DIFF
--- a/python/conftest.py
+++ b/python/conftest.py
@@ -11,16 +11,40 @@ def add_dp(doctest_namespace):
 
 
 FUZZY = doctest.register_optionflag('FUZZY')
+FUZZY_DF = doctest.register_optionflag('FUZZY_DF')
+IGNORE = doctest.register_optionflag('IGNORE')
+
+# Related doctest flags:
+# - NUMBER (https://docs.pytest.org/en/7.1.x/how-to/doctest.html?highlight=NUMBER#using-doctest-options)
+#   Floating-point numbers only need to match as far as the precision you have written in the expected doctest output
+# - SKIP (https://docs.python.org/3/library/doctest.html#doctest.SKIP)
+#   Do not run the example at all
+
+def _norm_df(df):
+    df = re.sub(r'╞.*', '', df, flags=re.DOTALL)
+    df = re.sub(r'\s+', ' ', df)
+    df = re.sub(r'shape: \(\d+,', 'shape: (#', df)
+    return df
 
 class CustomOutputChecker(doctest.OutputChecker):
     def check_output(self, want, got, optionflags):
         if FUZZY & optionflags:
+            if optionflags - FUZZY:
+                raise Exception('FUZZY can not be used with other flags')
             # A replacement string has special behavior with backslashes,
             # but the value returned by a lambda does not.
             number_re = lambda _: r'-?\d+(\.\d*)?'
             # Replace each tilde-number with a regex representing any possible number.
             want_re = re.sub(r'\\~(\\-)?\d+(\\\.\d*)?', number_re, re.escape(want.strip()))
             return bool(re.search(want_re, got))
+        if FUZZY_DF & optionflags:
+            if optionflags - FUZZY_DF:
+                raise Exception('FUZZY_DF can not be used with other flags')
+            return _norm_df(want) == _norm_df(got)
+        if IGNORE & optionflags:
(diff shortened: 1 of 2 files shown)


Shoeboxam (changes requested):
I would also add the IGNORE_OUTPUT flag, because there are situations where the number of rows in a table is random.

mccalluc:
I would worry a bit if we totally ignored the output, but here's the option, and another one that is just for dataframes.

python/conftest.py
```diff
@@ -1,6 +1,51 @@
...
+IGNORE = doctest.register_optionflag('IGNORE')
+
+# Related doctest flags:
+# - NUMBER (https://docs.pytest.org/en/7.1.x/how-to/doctest.html?highlight=NUMBER#using-doctest-options)
+#   Floating-point numbers only need to match as far as the precision you have written in the expected doctest output
+# - SKIP (https://docs.python.org/3/library/doctest.html#doctest.SKIP)
+#   Do not run the example at all
+
+def _norm_df(df):
+    df = re.sub(r'╞.*', '', df, flags=re.DOTALL)
```
Shoeboxam, line 24: Ignores all text after header.

Shoeboxam (changes requested):
I'm skeptical how much FUZZY strengthens doctests over IGNORE in the common cases we use doctests, and it adds a significant complexity. For now, don't include FUZZY but do leave a comment mentioning its implementation in this PR.

mccalluc:
For posterity: If a fuzzy numeric doctest ever becomes useful, try this:
```
        if FUZZY & optionflags:
            if optionflags - FUZZY:
                raise Exception('FUZZY can not be used with other flags')
            # A replacement string has special behavior with backslashes,
            # but the value returned by a lambda does not.
            number_re = lambda _: r'-?\d+(\.\d*)?'
            # Replace each tilde-number with a regex representing any possible number.
            want_re = re.sub(r'\\~(\\-)?\d+(\\\.\d*)?', number_re, re.escape(want.strip()))
            return bool(re.search(want_re, got))
```

We're thinking that the cases where we output a single DP mean, for example, aren't that much more effective with this than they are with the new `IGNORE`.

mccalluc:
segmentation fault from python and r linting. Waiting on fixes to be merged to main:
- #2280
- #2278

Shoeboxam (approved):
Thanks for the updates.

PR #2252, opened 2025-01-29
docs: better 404 page + better docs links

### LINKED ISSUE #2215: `api/user-guide/measurements/index.html`: "Refer to the Measurement for an explanation"
Guessing a different link was replaced with a sphinx ref, and result is a little odd.

Maybe the page title should be expanded, so it can stand on its own?

### LINKED ISSUE #2237: Add links to new resources in docs
Not sure exactly where they should go. I also wonder if some of the older links should be trimmed, or maybe just made less prominent.

- [ ] https://registry.oblivious.com/
- [ ] https://pypi.org/project/dp_wizard/

### LINKED ISSUE #2247: Fix redirect for https://docs.opendp.org/en/stable/user/combinators.html
Just goes to https://docs.opendp.org/en/stable/api/user-guide/combinators.html, which is still 404.

### LINKED ISSUE #2248: Friendlier message if we have a JS redirect
Annie writes:

> 404 is scary - that's all I wanted to say. :laughing:

Where we have a JS redirect, leave the scary **404** out of the title on that page. Just say that the site structure has changed and we'll be redirecting.

mccalluc:
These are all small, so thought it'd be ok to lump them together.

- Fix #2248 
- Fix #2247 
- Fix #2237 
- Fix #2215

### DIFF
--- a/docs/404.html
+++ b/docs/404.html
@@ -5,7 +5,7 @@
         Easiest just to write raw HTML and skip Sphinx altogether.
     -->
     <head>
-        <title>404 &mdash; OpenDP</title>
+        <title>Page not found &mdash; OpenDP</title>
         <style>
             * {
                 font-family: -apple-system, BlinkMacSystemFont, Segoe UI, "Helvetica Neue", Arial, sans-serif;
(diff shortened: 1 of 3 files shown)


Shoeboxam (changes requested):

Shoeboxam (approved):
Great! Thanks.

PR #2253, opened 2025-01-29
feat: Serialize query

mccalluc:
I think this demonstrates what we want to do, but I'm not sure what the motivation is. It feels like we're encouraging the user to make a whole big `Context` object and then not using anything but the `Expr`. 
- Is there more that is being captured in the serialization than is shown in this example? 
- Or is it that we want the user to try out their work locally before passing it to a server, so the `Context` is the more natural level to be working with?

If this is essentially on the right path (though I'm not sure that it is!) is the job mostly about syntactic sugar? Maybe a friendlier way deserialize, in particular?


docs/source/api/user-guide/utilities/serialization.rst
```diff
@@ -38,6 +38,50 @@ If this is something you need, please reach out so that we can understand your u
...
+        >>> import polars as pl
+
+        >>> context = dp.Context.compositor(
+        ...     data=pl.LazyFrame({}),
+        ...     privacy_unit=dp.unit_of(contributions=1),
+        ...     privacy_loss=dp.loss_of(epsilon=1.0),
+        ...     split_evenly_over=1,
+        ... ) # TODO: What's the benefit of creating this whole context, if the only thing that is serialized is the Expr?
+        >>> query = context.query().select(dp.len())
+        >>> plan = query._lf_plan  # TODO: Rename to ".opendp_plan"?
```
mccalluc, line 60: better name. `opendp_plan`?

docs/source/api/user-guide/utilities/serialization.rst
```diff
@@ -37,6 +37,51 @@ If this is something you need, please reach out so that we can understand your u
         >>> type(new_obj)
         <class 'opendp.mod.Measurement'>
 
+.. _lazyframe-serialization:
```
Shoeboxam, line 40: Can we put Polars Context first? Polars with context API is where I'd expect people would land first.
Then normal Context API? 

```python
query = context.query().a().b().c()
serialized_meas = dp.serialize(query.resolve())

# something like this:
query_2 = context.query()
query_2.chain = dp.deserialize(serialized_meas)
```

Then framework API?

python/src/opendp/context.py
```diff
@@ -378,6 +378,24 @@ def __repr__(self) -> str:
...
+        for context and full example.
+
+        :param serialized_plan: A plan like that returned by ``query.polars_plan.serialize()``
+        '''
+        import io
+        pl = import_optional_dependency('polars')
+        
+        new_plan = pl.LazyFrame.deserialize(io.BytesIO(serialized_plan))
+        new_query = self.query()
+        assert isinstance(new_query, LazyFrameQuery) # Lets us tighten the output type.
```
Shoeboxam, line 394: Irregardless of typing, I think it would be better to give an error that says the data must be a lazyframe to use this method.

```suggestion
	if not isinstance(new_query, LazyFrameQuery):
		raise ValueError("deserialize_polars_plan may only be used when data is a LazyFrame")
```

Shoeboxam: For the case where a user creates a normal Context, but calls this method and would get an assertion error without explanation.

Shoeboxam (changes requested):
Good direction, left some comments. Thanks.

docs/source/api/user-guide/utilities/serialization.rst
mccalluc: Done, I think.

docs/source/api/user-guide/utilities/serialization.rst
```diff
@@ -37,6 +37,51 @@ If this is something you need, please reach out so that we can understand your u
...
+        ...     split_evenly_over=1,
+        ... )
+        >>> query = context.query().select(dp.len())
+
+        >>> dp.serialize(query) # Directly serializing the query is not supported:
+        Traceback (most recent call last):
+        ...
+        Exception: OpenDP JSON Encoder ...
+
+        >>> serialized_plan = query.polars_plan.serialize() # But this will work!
```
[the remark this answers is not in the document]
mccalluc: Think this is addressed, but will be good to check the phrasing.

Shoeboxam (approved):
Thanks for iterating on this, docs look great. May make changes after trying to make APIs for Query and LazyFrameQuery more similar at some point.

Shoeboxam (approved):
RE: The PR description, I think we already resolved this while chatting and the contents of the PR now justify it.

PR #2256, opened 2025-01-29
feat: Pickle UDFs for serialization

### LINKED ISSUE #2254: Pickle UDFs for serialization
>  I think it would be great to add a pickling feature. Be sure to put it behind honest-but-curious. Thanks for the updates!

_Originally posted by @Shoeboxam in https://github.com/opendp/opendp/pull/2228#pullrequestreview-2582099855_

mccalluc:
- Fix #2254

@Shoeboxam, besides the plugins, I wonder if it would be good to check that other deserialized objects actually behave as expected? Could add more tests here, or separately, or maybe that is overly scrupulous. Your call.

### DIFF
--- a/docs/source/api/user-guide/utilities/serialization.rst
+++ b/docs/source/api/user-guide/utilities/serialization.rst
@@ -52,6 +52,12 @@ are not currently serializable:
         ...
         Exception: OpenDP JSON Encoder currently does not handle instances of <class 'opendp.mod.Queryable'>: It may have state which is not set by the constructor. Error on: Queryable(Q=query_type)
 
+Plugins with user defined functions are serializable,
+but ``honest-but-curious`` must be enabled,
+and due to limitations of python's pickle, ``lambda`` will not work.
+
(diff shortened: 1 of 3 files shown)


mccalluc:
Never mind: Not going to work. Pickled functions aren't wrappers around the bytecode or anything like that: they just contain a reference to the name of the function.
```
>>> def f():
...   return 'hello world'
...
>>> from pickle import dumps
>>> dumps(f)
b'\x80\x04\x95\x12\x00\x00\x00\x00\x00\x00\x00\x8c\x08__main__\x94\x8c\x01f\x94\x93\x94.'
```
new session:
```
>>> from pickle import loads
>>> f = loads(b'\x80\x04\x95\x12\x00\x00\x00\x00\x00\x00\x00\x8c\x08__main__\x94\x8c\x01f\x94\x93\x94.')
  File "<stdin>", line 1, in <module>
AttributeError: Can't get attribute 'f' on <module '__main__' (<class '_frozen_importlib.BuiltinImporter'>)>
```

PR #2257, opened 2025-01-30
testing: Clean up AST tests

mccalluc:
For the reviewer:
- Confirmation that none of the `TODO`s need attention right now?
- ... but any one of them would be a small improvement, and something we might point a new dev at for a warm up?


Shoeboxam (approved):
👍

PR #2258, opened 2025-01-30
feat: when-then-else

mccalluc:


rust/src/transformations/make_stable_expr/expr_ternary/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+            let falsy = t_falsy.invoke(arg)?;
+
+            Ok(ExprPlan {
+                plan: arg.clone(),
+                expr: Expr::Ternary {
+                    predicate: Arc::new(predicate.expr),
+                    truthy: Arc::new(truthy.expr),
+                    falsy: Arc::new(falsy.expr),
+                },
+                fill: None, // TODO
```
mccalluc, line 81: This is actually right: this is run before aggregation, so there's no empty group that needs a default.

python/test/test_polars_ternary.py
```diff
@@ -0,0 +1,79 @@
...
+
+def test_when_then_otherwise_incomplete():
+    pl = pytest.importorskip("polars")
+    lf_domain, lf = example_lf()
+    with pytest.raises(Exception, match=r'unsupported literal value: null'):
+        dp.t.make_stable_lazyframe(
+            lf_domain,
+            dp.symmetric_distance(),
+            lf.select(
+                pl.when(pl.col("A") == 1).then(1).alias('fifty'),
```
Shoeboxam, line 76: Here you can allow the computation to run, but the output domain may now contain null.

rust/src/transformations/make_stable_expr/expr_ternary/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+        .as_ref()
+        .clone()
+        .make_stable(input_domain.as_row_by_row(), input_metric.clone())?;
+    let t_falsy = falsy
+        .as_ref()
+        .clone()
+        .make_stable(input_domain.as_row_by_row(), input_metric.clone())?;
+
+    let (truthy_domain, _truthy_metric) = t_truthy.output_space();
+    let (falsy_domain, _falsy_metric) = t_falsy.output_space();
```
Shoeboxam, line 55: Would be good to check that the metrics match too!

rust/src/transformations/make_stable_expr/expr_ternary/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    if truthy_domain != falsy_domain {
```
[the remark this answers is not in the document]
Shoeboxam: This is just to prevent polars fallible casting!

rust/src/transformations/make_stable_expr/expr_ternary/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+        return fallible!(
+            MakeTransformation,
+            "output domains in ternary must match, instead found {:?} and {:?}",
+            truthy_domain,
+            falsy_domain
+        );
+    }
+
+    if matches!(truthy_domain.column.dtype(), DataType::Categorical(_, _)) {
```
Shoeboxam, line 66: Don't need to check both due to earlier check.

rust/src/transformations/make_stable_expr/expr_ternary/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+        // Since literal categorical values aren't possible,
+        // not clear if this is actually reachable.
+        return fallible!(MakeTransformation, "ternary cannot be applied to categorical data, because it may trigger a data-dependent CategoricalRemappingWarning in Polars");
+    }
+
+    let mut output_domain = truthy_domain.clone();
```
Shoeboxam, line 72: Does the output always take on the name of the truthy value? Might be good to test.

rust/src/transformations/make_stable_expr/expr_ternary/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+    output_domain.column.drop_bounds().ok();
+    output_domain.column.nullable = false;
```
Shoeboxam, line 74:
```suggestion
    output_domain.column.nullable |= falsey_domain.column.nullable;
```

rust/src/transformations/make_stable_expr/expr_ternary/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    if matches!(truthy_domain.column.dtype(), DataType::Categorical(_, _)) {
+        // Since literal categorical values aren't possible,
+        // not clear if this is actually reachable.
+        return fallible!(MakeTransformation, "ternary cannot be applied to categorical data, because it may trigger a data-dependent CategoricalRemappingWarning in Polars");
+    }
+
+    let mut output_domain = truthy_domain.clone();
+    output_domain.column.drop_bounds().ok();
```
Shoeboxam, line 73: In this case, if the truthy domain does not have nans and falsey domain has nans, then the resulting output domain would have made the claim that the data does not have nans. Instead, lets clear those descriptors:

This is a shorthand to completely replace the element domain with the loosest descriptors.
```suggestion
    output_domain.column.set_dtype(output_domain.column.dtype())?;
```

rust/src/transformations/make_stable_expr/expr_ternary/test.rs
```diff
@@ -0,0 +1 @@
+
```
Shoeboxam, line 1: Some tests would be good to add!

Shoeboxam (changes requested):
Looking good, some comments! Thanks.

rust/src/transformations/make_stable_expr/expr_ternary/mod.rs
mccalluc: Added a note to explain.

rust/src/transformations/make_stable_expr/expr_ternary/mod.rs
mccalluc: Added a test that we take the column name from the truthy, just like vanilla polars.

PR #2259, opened 2025-01-31
docs: Add documentation on Queryable

### LINKED ISSUE #1660: Improve documentation for queryables
They are first-class members of the framework... but mostly a developer feature? But the concept of queryable is showing up in user-facing docs, too.

So we should document this better, but not sure right now what that looks like.

mccalluc:
- Fix #1660

### DIFF
--- a/docs/source/api/user-guide/combinators/compositors.rst
+++ b/docs/source/api/user-guide/combinators/compositors.rst
@@ -105,7 +105,7 @@ Sequential composition relaxes the basic compositor, allowing for
 queries to be submitted interactively. That is, you can make submit a
 query, view the output, and then submit another query that uses the
 information gained from the prior release. However, this API still
-requires ``sequentiality``, which we’ll discuss in more detail later.
+requires `sequentiality`, which we’ll discuss in more detail below.
 
 The API for interactive compositors is more verbose than in the
 non-interactive case because you must explicitly pass the input domain,
--- a/docs/source/api/user-guide/combinators/compositors.rst
+++ b/docs/source/api/user-guide/combinators/compositors.rst
@@ -98,6 +98,8 @@ The privacy map simply sums the constituent output distances.
             >>> mean_fraction_meas.map(1)
             3.0
 
+.. _sequential-composition:
+
 Sequential Composition
 ----------------------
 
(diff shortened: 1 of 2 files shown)


docs/source/api/user-guide/combinators/compositors.rst
```diff
@@ -190,7 +192,11 @@ when the analyst may still interact with the first queryable. This is
 the subject of a further line of research on concurrent compositors,
 which we hope to make available in the next library release.
 
-An example of this constraint is demonstrated in the “Nesting” section
```
Shoeboxam, line 193: This documentation is outdated, it's still a limitation, but only for some measures, in particular approximate zero concentrated or approximate renyi DP.

Shoeboxam, line 193: The compositor is *sequential* in the sense that multiple queries are sequentially released on the same dataset.

..note:
	The sequential compositor API makes another distinction 
        between adaptive composition and concurrent composition,
	which varies based on the choice of privacy measure.

	Adaptive composition is subject to the limitation that only one queryable is active at any point in time.
	To satisfy adaptive composition, the compositor locks, or freezes, any queryable it has
	previously spawned when a new query arrives.
	This is because the postprocessing argument doesn't necessarily 
        hold when the analyst may still interact with earlier queryables.

	Concurrent composition lifts this limitation for measures of privacy 
        where we have been able to prove that postprocessing still holds.
	In OpenDP, all privacy measures support concurrent composition,
	except for approximate zCDP and approximate Renyi-DP.

python/src/opendp/mod.py
```diff
@@ -498,6 +498,11 @@ def __iter__(self):
 Transformation = cast(Type[Transformation], Transformation) # type: ignore[misc]
 
 class Queryable(object):
+    '''
+    Queryables are used for :ref:`sequential composition <sequential-composition>`.
```
Shoeboxam, line 502:
```suggestion
    Queryables are used for interactive mechanisms like :ref:`sequential composition <sequential-composition>`.
```

Shoeboxam (changes requested):
Some suggestions mainly on text I left outdated. Thanks.

docs/source/api/user-guide/combinators/compositors.rst
mccalluc: tweaked sphinx syntax and pasted in.

python/src/opendp/mod.py
mccalluc: done.

docs/source/api/user-guide/combinators/compositors.rst
mccalluc: covered by the longer provided text.

Shoeboxam (approved):
Thanks!

PR #2260, opened 2025-01-31
docs: Check for doctrings on classes

mccalluc:
- Prompted by #2259

Make sure we document all of our public classes. The changes here do that a few different ways.
- Fill in new documentation
- Move docstrings up from `__init__` that weren't being rendered. (This is the default, but like everything else [it could be configured](https://stackoverflow.com/questions/5599254/how-to-use-sphinxs-autodoc-to-document-a-classs-init-self-method).)
- Make classes that don't seem intended for end user use private. (Maybe more of this?)

The `__init__` docs weren't being checked, but I could imagine checking that the new class docstrings are consistent with constructor parameters.

### DIFF
--- a/python/test/test_ast.py
+++ b/python/test/test_ast.py
@@ -68,6 +68,9 @@ class Checker():
         if len(self.all_ast_args) and self.all_ast_args[0].arg in ['self', 'cls']:
             # TODO: Confirm that this is a method in a class.
             self.all_ast_args.pop(0)
+        
+        if self.name == '__init__' and self.docstring:
+            self.errors.append('Move docstring up to class')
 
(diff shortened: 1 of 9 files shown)


python/test/test_ast.py
```diff
@@ -196,13 +199,14 @@ def get_errors(self):
 PUBLIC = 'public'
 PRIVATE = 'private'
 
-class Function(NamedTuple):
+class CodeObj(NamedTuple):
```
Shoeboxam, line 202: This is nice to avoid confusion with Function.

Shoeboxam (approved):
Looks good!

PR #2265, opened 2025-02-05
feat: `member` is true or false, with a warning; Expand serialization tests

### LINKED ISSUE #1007: `member()` should return `True` or error, never `False`
https://docs.opendp.org/en/stable/user/programming-framework/supporting-elements.html#domain

What output can we expect from `member()`? Going beyond the examples on the page:
```
f64_atom_domain.member('test')
TypeError: inferred type is String, expected f64. See[ https://github.com/opendp/opendp/discussions/298](https://github.com/opendp/opendp/discussions/298)
```

So we see three possible outcomes: `True`, `False`, or error. @Shoeboxam indicates that the desired behavior from the Rust code is `True` or error, never `False`.

### LINKED ISSUE #2262: Stronger tests on deserialization
Deserialization probably works, but I haven't actually exercised it over much of the codebase. There shouldn't be any surprises here, but better safe than sorry.

mccalluc:
- fix #1007
- Fix #2262

Python is less rigorous than Rust, and I think most users are just going to expect `member` to return a bool. But we do have a side channel to pass more information, if the user wants more information: warnings.

(This is motivated by wanting to write stronger tests of serialization, but its not strictly necessary for that: I could just have `with pytest.warns()` blocks.)

### DIFF
--- a/python/src/opendp/mod.py
+++ b/python/src/opendp/mod.py
@@ -559,6 +559,7 @@ class Domain(ctypes.POINTER(AnyDomain)): # type: ignore[misc]
         '''
         try:
             # TODO: Should we rename the import to "_member"?
+            # https://github.com/opendp/opendp/issues/2268
             from opendp.domains import member
             return member(self, val)
         except Exception as e:
--- a/python/src/opendp/mod.py
+++ b/python/src/opendp/mod.py
@@ -557,8 +557,15 @@ class Domain(ctypes.POINTER(AnyDomain)): # type: ignore[misc]
         
         :param val: a value to be checked for membership in `self`
         '''
-        from opendp.domains import member
-        return member(self, val)
+        try:
+            # TODO: Should we rename the import to "_member"?
+            from opendp.domains import member
+            return member(self, val)
+        except Exception as e:
+            from warnings import warn
+            warn(str(e))
+            return False
+
 
     @property
     def type(self) -> Union["RuntimeType", str]:
--- a/python/src/opendp/mod.py
+++ b/python/src/opendp/mod.py
@@ -551,7 +551,7 @@ class Domain(ctypes.POINTER(AnyDomain)): # type: ignore[misc]
     '''
     _type_ = AnyDomain
 
-    def member(self, val):
+    def member(self, val) -> bool:
         '''
         Check if ``val`` is a member of the domain.
         
@@ -563,7 +563,7 @@ class Domain(ctypes.POINTER(AnyDomain)): # type: ignore[misc]
             return member(self, val)
         except Exception as e:
             from warnings import warn
-            warn(str(e))
+            warn(f'Value ({val}) does not belong to carrier type of {type(self)}. Details: {e}')
             return False
 
 
--- a/python/src/opendp/mod.py
+++ b/python/src/opendp/mod.py
@@ -563,7 +563,7 @@ class Domain(ctypes.POINTER(AnyDomain)): # type: ignore[misc]
             return member(self, val)
         except Exception as e:
             from warnings import warn
-            warn(f'Value ({val}) does not belong to carrier type of {type(self)}. Details: {e}')
+            warn(f'Value ({val}) does not belong to carrier type of {self}. Details: {e}')
             return False
 
 
--- a/python/test/extras/test_ext_domains.py
+++ b/python/test/extras/test_ext_domains.py
@@ -50,7 +50,7 @@ def test_sscp_domain():
     domain = _sscp_domain(num_features=4, T=dp.f32)
     domain.member(np.random.normal(size=(4, 4)).astype(np.float32))
 
-    with pytest.raises(dp.OpenDPException):
+    with pytest.warns(UserWarning, match=r'does not belong to carrier type'):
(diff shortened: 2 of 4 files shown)


python/src/opendp/mod.py
```diff
@@ -551,14 +551,21 @@ class Domain(ctypes.POINTER(AnyDomain)): # type: ignore[misc]
...
         '''
         Check if ``val`` is a member of the domain.
         
         :param val: a value to be checked for membership in `self`
         '''
-        from opendp.domains import member
-        return member(self, val)
+        try:
+            # TODO: Should we rename the import to "_member"?
+            from opendp.domains import member
```
mccalluc, line 562: File follow up issue.

python/src/opendp/mod.py
```diff
@@ -551,14 +551,21 @@ class Domain(ctypes.POINTER(AnyDomain)): # type: ignore[misc]
...
+    def member(self, val) -> bool:
         '''
         Check if ``val`` is a member of the domain.
         
         :param val: a value to be checked for membership in `self`
         '''
-        from opendp.domains import member
-        return member(self, val)
+        try:
+            # TODO: Should we rename the import to "_member"?
```
mccalluc, line 561:
```suggestion
            # TODO: Should we rename the import to "_member"?
            # https://github.com/opendp/opendp/issues/2268
```

python/test/test_serialization.py
```diff
@@ -104,6 +104,76 @@ def test_version_mismatch_warning():
...
+    #     ],
+    # )
+    # def test_serializable_metric(_readable_name, dp_metric, a, b, dist):
+    #     assert dp_metric(a, b) == dist
+
+    #     serialized = dp.serialize(dp_metric)
+    #     deserialized = dp.deserialize(serialized)
+
+    #     assert deserialized(a, b) == dist
+
```
mccalluc, line 162:
```suggestion
```

mccalluc (commented):
a couple follow-ups for me.

python/test/test_serialization.py
```diff
@@ -104,6 +104,59 @@ def test_version_mismatch_warning():
 # in this case it needs to wrap the tests.
 pl = import_optional_dependency('polars', raise_error=False)
 if pl is not None:
+    # member() will warn if instance is not even of the carrier type,
+    # but that behavior is tested elsewhere, and can be ignored here.
+    @pytest.mark.filterwarnings("ignore::UserWarning")
+    @pytest.mark.parametrize(
+        "_readable_name,dp_domain,in_value,out_value",
+        [
+            (str(dp_domain), dp_domain, in_value, out_value)
```
Shoeboxam, line 113: Instead of the comprehension, would it be better to change the repr?

Going by Python conventions, I thought it was the case that repr was more gritty and str gave a pretty summary. Are we following python conventions?

Shoeboxam: Pytest is somehow stringifying the input in a way we don't expect, because we have `__repr__` implemented which should handle repr and str pretty stringification.

Shoeboxam (changes requested):
Consistency is key!

python/test/test_serialization.py
mccalluc: Sorry: I misdiagnosed this. I'll file an issue just to check on the __repr__ (since it seems like something was out of sync when we looked at the code), but it has nothing to do with this.

Pytest by default will only render primitives like strings and numbers. For any other object, it will just sequentially number it to give the test a name... but we can override that with `ids`, which I've just done for all of these tests.

python/test/extras/test_ext_domains.py
```diff
@@ -50,7 +50,7 @@ def test_sscp_domain():
     domain = _sscp_domain(num_features=4, T=dp.f32)
     domain.member(np.random.normal(size=(4, 4)).astype(np.float32))
 
-    with pytest.raises(dp.OpenDPException):
+    with pytest.warns(UserWarning, match=r'does not belong to carrier type'):
         domain.member(np.random.normal(size=(4, 4)))
```
mccalluc, line 54: Make this stronger: check return value.

mccalluc:
@Shoeboxam : Thanks for the nudge to clean up the comprehensions. `ids` is much cleaner, though it means that some of the tests have very long generated names. I don't think the long names get in the way, and it's not worth making the test code more complicated to get simpler strings. 

(Because of all the indentation changes, ignore whitespace in the diff: https://github.com/opendp/opendp/pull/2265/files?diff=unified&w=1)

python/test/test_serialization.py
```diff
@@ -12,39 +12,37 @@
...
+        dp.m.approximate(dp.m.max_divergence()),
+        dp.m.user_divergence("user_divergence"),
+        # Measurements:
+        dp.m.make_gaussian(atom, dp.absolute_distance(float), 1),
+        dp.m.then_gaussian(1),
+        # Compositions:
+        chained,
+        dp.c.make_population_amplification(chained, population_size=100),
     ],
+    ids=lambda arg: str(arg)
```
Shoeboxam, line 43: Slick

Shoeboxam (approved):
Thanks for iterating!

PR #2267, opened 2025-02-06
docs: Add AST checks on the formatting of the docstring

### LINKED ISSUE #1396: `:example:` tags conflict with return type annotations
This documentation:
```
def space_of(T, M=None, infer=False) -> Tuple[Domain, Metric]:
    """A shorthand for building a metric space.

    A metric space consists of a domain and a metric.

    :example:

    >>> from typing import List # in Python 3.9, can just write list[int] below
    ...
    >>> dp.space_of(List[int])
    (VectorDomain(AtomDomain(T=i32)), SymmetricDistance())
    >>> # the verbose form allows greater control:
    >>> (dp.vector_domain(dp.atom_domain(T=dp.i32)), dp.symmetric_distance())
    (VectorDomain(AtomDomain(T=i32)), SymmetricDistance())

    :param T: carrier type (the type of members in the domain)
    :param M: metric type
    :param infer: if True, ``T`` is an example of the sensitive dataset. Passing sensitive data may result in a privacy violation.
    """
```
is rendered as:
Screenshot 2024-03-15 at 4 42 19 PM

For some reason, the return type is inserted below the "Example" header?

Remove the `:example:` tag and we get:
Screenshot 2024-03-15 at 4 43 37 PM

Perhaps a sphinx bug? But I really don't know. I also have not been able to find the docs for `:example:`, ironically. I'd suggest just removing it.

---

Update:
It does work if `:example:` is the last entry. Maybe what we really want is to make sure that this is always the case.

---

Update:
_Now_ I think I understand the conditions for this bug: It shows up when there is a return type annotation (rather than an explicit `:return:` in the docstring); the documentation that is generated from this is inserted between the example header and the example content. I've spent a little time searching the Sphinx issues, and haven't found exactly this issue described. Maybe an interaction with a plugin?

### LINKED ISSUE #1967: Formatting of doc for `debias_randomized_response_bitvec`
The docstring doesn't look right:
- too many backtics
- dollar-signs coming through verbatim
- RST section headers in line

This function is also a little odd, not starting with `make_` or `then_`: is it in the right place? Is that related to the rendering problem?
Screenshot 2024-08-21 at 1 42 52 PM

### LINKED ISSUE #2188: problems with markdown -> rst -> html translation
Rust docs are Markdown, Python/Sphinx is RST (though [there may be other options (SO)](https://stackoverflow.com/questions/2471804/using-sphinx-with-markdown-instead-of-rest)).

My understanding is that the python codegen produces a mix of MD and RST, and then the `docstring` function in `conf.py` attempts to translate that to correct RST, but it can get confused. To a user, at first these look like simple typos in the docs, but it's really a bug in our translation code.

## Examples:

- [x] #1967

https://docs.opendp.org/en/v0.11.1/api/python/opendp.measurements.html#opendp.measurements.debias_randomized_response_bitvec

```
Computes the sum of the answers into a $k$-length vector $Y$
and returns ````math Y\frac{Y-\frac{f}{2}}{1-f} ```` :type f: float 
:raises TypeError: if an argument’s type differs from the expected 
type :raises UnknownTypeException: if a type argument fails to 
parse :raises OpenDPException: packaged error from the core 
OpenDP library
```

- [ ] Latex `$` comes through in param string

https://docs.opendp.org/en/v0.11.1/api/python/opendp.measurements.html#opendp.measurements.make_randomized_response_bitvec

```
f (float) – Per-bit flipping probability. Must be in $(0, 1]$.
```

- [ ] Doctest inlined in example

https://docs.opendp.org/en/v0.11.1/api/python/opendp.accuracy.html#opendp.accuracy.summarize_polars_measurement

```
First, create a measurement with the Polars API: >>> import 
opendp.prelude as dp >>> import polars as pl >>> 
dp.enable_features(“contrib”) … >>> lf =
```

- [x] #1396

Return conflicts with `:example:` -- Probably related.

## Strategy

There are several steps in the translation:
- Writing up Rust docs in a constrained MD syntax that will translate nicely
- Codegen in `python.rs`
- Checking doctring in `test_ast.py`: #2182
- `docstring` function in `conf.py`
- RST to HTML in Sphinx

I wouldn't want to add any complexity to the codegen, and we have limited control with any tool we'd add to Sphinx, so that makes me think the focus should be on the `docstring` function if any work is done here, with a small set of input/output fixtures, so we can be confident that fixing one problem isn't breaking something else.

When we have the docstring checks in place, syntax checks could go there, and we can also just try to fix the source in the Rust docstrings, though I think the fix may not be possible there.

mccalluc:
Left is local docs build; Right is nightly
Screenshot 2025-02-06 at 6 32 11 PM

- If we require all the directives to be in a contiguous block, `Example` comes in the right place.
- Also require doctests to be contiguous, so they don't get split into little boxes.
- Fix #1396

Screenshot 2025-02-06 at 6 36 05 PM

- The math looks better, and the `Raises` is formatted correctly.
- The formatting of the rust link is also fixed with a later commit, after this screenshot was taken.
- Fix #1967

Screenshot 2025-02-06 at 8 59 31 PM

- Not filed separately, but this is last bullet point, so fix #2188 

Also:
- No issue filed, but fix the rendering of numbered lists and bullet lists.
- Fix formatting of links to rust docs...
  - RST does not support nested inline markup
  - There is a [package](https://silverrainz.me/blog/sphinxnotes-comboroles.html) that may add support for this, but I don't think it's worth the complexity.
  - But these links are often (usually?) not working. #1025. Maybe we just drop these links?

### DIFF
--- a/python/src/opendp/context.py
+++ b/python/src/opendp/context.py
@@ -152,7 +152,6 @@ def domain_of(T, infer: bool = False) -> Domain:
 
     >>> dp.domain_of('Option<int>')  # Python's `Optional` is not supported.
     OptionDomain(AtomDomain(T=i32))
-    
     >>> dp.domain_of(dp.i32)
     AtomDomain(T=i32)
 
--- a/python/src/opendp/extras/polars/__init__.py
+++ b/python/src/opendp/extras/polars/__init__.py
@@ -907,7 +907,7 @@ class LazyFrameQuery():
 
         >>> import polars as pl
         >>> data = pl.LazyFrame([pl.Series("convicted", [0, 1, 1, 0, 1] * 50, dtype=pl.Int32)])
-
+        >>>
         >>> context = dp.Context.compositor(
         ...     data=data,
         ...     privacy_unit=dp.unit_of(contributions=1),
@@ -915,12 +915,12 @@ class LazyFrameQuery():
         ...     split_evenly_over=1,
         ...     margins={(): dp.polars.Margin(max_partition_length=1000)},
         ... )
-
+        >>>
         >>> query = context.query().select(
         ...     dp.len(),
         ...     pl.col("convicted").fill_null(0).dp.sum((0, 1))
         ... )
-
+        >>>
         >>> query.summarize(alpha=.05)  # type: ignore[union-attr]
         shape: (2, 5)
         ┌───────────┬──────────────┬─────────────────┬───────┬──────────┐
--- a/python/src/opendp/extras/polars/__init__.py
+++ b/python/src/opendp/extras/polars/__init__.py
@@ -895,13 +895,13 @@ class LazyFrameQuery():
     def summarize(self, alpha: float | None = None):
         """Summarize the statistics released by this query.
 
-        :param alpha: optional. A value in [0, 1] denoting the statistical significance. For the corresponding confidence level, subtract from from 1: for 95% confidence, use 0.05 for alpha.
-
         If ``alpha`` is passed, the resulting data frame includes an ``accuracy`` column.
 
         If a threshold is configured for censoring small/sensitive partitions,
         a threshold column will be included,
         containing the cutoff for the respective count query being thresholded.
+
+        :param alpha: optional. A value in [0, 1] denoting the statistical significance. For the corresponding confidence level, subtract from from 1: for 95% confidence, use 0.05 for alpha.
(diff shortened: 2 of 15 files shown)


Shoeboxam (approved):
Thanks! Looks good, and fixes future problems.

PR #2269, opened 2025-02-07
Remove code that generates links to rust docs

### LINKED ISSUE #1025: Python API Measures: Links from Python to Rust are broken

mccalluc:
- Fix #1025

A proposal. A separate PR (#2267) addressing the formatting, but how useful would these links be, even if the kinks were ironed out? What's the story where a Python user would want to consult the Rust docs?

### DIFF
--- a/rust/build/derive.rs
+++ b/rust/build/derive.rs
@@ -123,7 +123,7 @@ fn parse_file_tree(
 
                 let mut contents = String::new();
                 File::open(&path)?.read_to_string(&mut contents)?;
-                if let Some(funcs) = parse_file(contents, &proof_paths, module_name) {
+                if let Some(funcs) = parse_file(contents, &proof_paths) {
                     matches.extend(funcs);
                 } else {
                     return Ok(None);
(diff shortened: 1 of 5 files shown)


Shoeboxam:
I like the link because it lets you see the meaningful implementation/source code of the function. The links work as expected, it's just that the rust docs build failed.

mccalluc:
ok!

PR #2273, opened 2025-02-10
Add pre-commit hooks, with `rust fmt`

### LINKED ISSUE #2271: Add pre-commit hooks to normal dev environment?
It is one more step at setup, but it would probably save us time overall. Out of the box it would give us some white-space cleaning, and 
- [ ] `cargo fmt` could be added easily.
- [ ] for python, `black` would be analogous, but we'll need to clean up the existing codebase.

mccalluc:
- Fix #2271

Not critical, but I've tripped over this often enough to make it worthwhile. Will address python formatting separately.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,9 +7,9 @@ repos:
     - id: trailing-whitespace
   - repo: local
     hooks:
-      - id: rustfmt
-        name: rustfmt
-        description: Check if all files follow the rustfmt style
-        entry: cargo fmt --all -- --check --color always
+      - id: cargo-fmt
+        name: Format Rust
+        description: Reformat Rust code as needed
+        entry: bash -c 'cd rust && cargo fmt'
         language: system
         pass_filenames: false
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,15 @@
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v2.3.0
+    hooks:
+    - id: check-yaml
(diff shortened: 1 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,15 @@
...
+    hooks:
+    - id: check-yaml
+    - id: end-of-file-fixer
+    - id: trailing-whitespace
+  - repo: local
+    hooks:
+      - id: cargo-fmt
+        name: Format Rust
+        description: Reformat Rust code as needed
+        entry: bash -c 'cd rust && cargo fmt'
```
Shoeboxam, line 13: Check that cargo fmt returns nonzero if changes are made.
- If it returns zero (success) even when changes are made, is there an argument to make it fail?

Otherwise, I've been using this command in my pre-commit, that just checks the format is right:

```shell
cargo fmt --manifest-path=rust/Cargo.toml --all --check
```

Shoeboxam: Also consider adding `--all` so that the check runs for all crates in the workspace.

Shoeboxam (changes requested):

mccalluc:
Maybe something like 
```
    hooks:
      - id: cargo-fmt
        name: Check Rust format
        entry: "bash -c 'pwd && cd rust && cargo fmt --all --check || echo \"run: cargo fmt --all\" && false'"
        language: system
        pass_filenames: false
```
but still needs work. Feel like I'm missing something obvious.

mccalluc:
our dev instructions are already heavy. Don't want to add to the burden.

PR #2275, opened 2025-02-10
Make sure default `cargo test` also passes

### LINKED ISSUE #2272: plain `cargo test` (without `--all-features`) errors
```
$ cargo test
   Compiling ...
error[E0432]: unresolved import `crate::measurements::make_laplace`
 --> src/accuracy/tail_bounds/test.rs:7:5
  |
7 |     measurements::make_laplace,
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^ no `make_laplace` in `measurements`
```

Haven't looked into it, but my guess is that the code that defines `make_laplace` is guarded, but the test code does not have a corresponding guard. Might be considered a special case of
- #1343

mccalluc:
If this isn't something we would expect to pass, it would be good for me to understand that, too.

- Fix #2272

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -122,7 +122,10 @@ jobs:
           key: ${{ runner.os }}-cargo-${{ hashFiles('rust/Cargo.toml') }}
           restore-keys: ${{ runner.os }}-cargo-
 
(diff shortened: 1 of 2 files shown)


Shoeboxam (approved):
👍

PR #2168, opened 2024-12-04
fix CallbackFn segfault

Shoeboxam:
Closes #2167

PR #2278, opened 2025-02-14
fix: replace _dependencies with lifelines

### LINKED ISSUE #2235: core dump from test/test_comb.py::test_renyidp
Hit a CI error after merging to main. Was not able to reproduce with a re-run. Sometimes errors are sporadic, but if we keep a record, over time a pattern may emerge.

https://github.com/opendp/opendp/actions/runs/12814449031/job/35731336189#step:8:375
```
Fatal Python error: Aborted

Current thread 0x00007f2ebb484b80 (most recent call first):
  File "/home/runner/work/opendp/opendp/python/src/opendp/core.py", line 182 in function_eval
  File "/home/runner/work/opendp/opendp/python/src/opendp/mod.py", line 461 in __call__
  File "/home/runner/work/opendp/opendp/python/test/test_comb.py", line 112 in test_renyidp
...
  File "/opt/hostedtoolcache/Python/3.9.21/x64/bin/coverage", line 8 in <module>
/home/runner/work/_temp/d4b1501f-f733-458e-be52-c7428eb711eb.sh: line 1:  1983 Aborted                 (core dumped) coverage run -m pytest -v
test/test_comb.py::test_renyidp 
```

Shoeboxam:
Fix #2235

The plugin APIs embed Python objects inside opaque Rust structs (transformation, measurement, queryable, function, domain, privacy profile, etc).

Imagine a user defines a custom function and uses the plugin API to embed it into a Rust struct.
When execution continues, the Rust struct holds a pointer to memory allocated in Python. 
Unfortunately, Python doesn't know the memory is still being referenced by Rust, 
and the Python refcount to that object naturally drops to zero.
The Python interpreter then frees the object upon the next garbage collection, leading to a use-after-free memory corruption when the Rust struct is used.

Previously, the generated code would add a reference to any Python objects embedded in the Rust struct, to the corresponding opaque struct on the Python side under the `_dependencies` attribute.
That is, the Rust object "depends on" the Python objects in its `_dependencies`,
extending the lifetime of any embedded Python objects to the lifetime of the Rust struct.
Two downsides to this approach:
- tracking the dependencies requires Rust structs built by combinators to inherit dependencies of its arguments, leading to complexity
- there is no natural place to attach dependencies for privacy maps that return user-defined Renyi-DP tradeoff functions

PR #2168 introduced a new approach to maintain the refcount, by adding a "lifeline" to callback functions.
The lifeline is a struct that increments the Python refcount when cloned, and decrements when dropped.
The intention of the PR was to capture a clone of the lifeline in the Rust closure,
so that the Python refcount for the object would remain non-zero until all Rust closures referencing the Python object/holding the lifeline are dropped.
Unfortunately, the lifeline wasn't being captured, leaving the refcount unprotected.

This PR ensures that the lifeline is captured, and adjusts transition functions to use the same mechanism. 
Since all Python objects are now protected via lifelines, this PR also removes the dependencies system.

Disables R linting due to #2279

### DIFF
--- a/rust/src/domains/ffi.rs
+++ b/rust/src/domains/ffi.rs
@@ -500,12 +500,17 @@ pub extern "C" fn opendp_domains__user_domain(
     let identifier = try_!(to_str(identifier)).to_string();
     let descriptor = try_as_ref!(descriptor).clone();
     let element = ExtrinsicElement::new(identifier, descriptor);
-    let member = wrap_func(try_as_ref!(member).clone());
-    let member = Function::new_fallible(move |arg: &ExtrinsicObject| -> Fallible<bool> {
-        member(&AnyObject::new(arg.clone()))?.downcast::<bool>()
+
+    let member_closure = wrap_func(try_as_ref!(member).clone());
+    let member_function = Function::new_fallible(move |arg: &ExtrinsicObject| -> Fallible<bool> {
+        member_closure(&AnyObject::new(arg.clone()))?.downcast::<bool>()
     });
 
-    Ok(AnyDomain::new(ExtrinsicDomain { element, member })).into()
+    Ok(AnyDomain::new(ExtrinsicDomain {
+        element,
+        member: member_function,
+    }))
+    .into()
 }
 
 #[bootstrap(
--- a/rust/src/domains/ffi.rs
+++ b/rust/src/domains/ffi.rs
@@ -7,7 +7,7 @@ use crate::{
     domains::{type_name, AtomDomain, MapDomain, VectorDomain},
     error::Fallible,
     ffi::{
-        any::{AnyDomain, AnyObject, CallbackFn, Downcast},
+        any::{wrap_func, AnyDomain, AnyObject, CallbackFn, Downcast},
         util::{self, c_bool, into_c_char_p, to_str, ExtrinsicObject, Type, TypeContents},
     },
     traits::{CheckAtom, Float, Hashable, Integer, Primitive},
@@ -470,8 +470,7 @@ impl Domain for ExtrinsicDomain {
         identifier(c_type = "char *", rust_type = b"null"),
         member(rust_type = "bool"),
         descriptor(default = b"null", rust_type = "ExtrinsicObject")
-    ),
-    dependencies("c_member")
+    )
 )]
 /// Construct a new UserDomain.
 /// Any two instances of an UserDomain are equal if their string descriptors are equal.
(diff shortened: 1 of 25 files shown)


Shoeboxam:
* **#2278** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2278?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/opendp/opendp/2278?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rust/src/domains/ffi.rs
```diff
@@ -500,10 +499,9 @@ pub extern "C" fn opendp_domains__user_domain(
     let identifier = try_!(to_str(identifier)).to_string();
     let descriptor = try_as_ref!(descriptor).clone();
     let element = ExtrinsicElement::new(identifier, descriptor);
-    let member = try_as_ref!(member).clone();
+    let member = wrap_func(try_as_ref!(member).clone());
     let member = Function::new_fallible(move |arg: &ExtrinsicObject| -> Fallible<bool> {
-        let c_res = (member.callback)(AnyObject::new_raw(arg.clone()));
-        Fallible::from(util::into_owned(c_res)?)?.downcast::<bool>()
+        member(&AnyObject::new(arg.clone()))?.downcast::<bool>()
```
mccalluc, line 504: This is superficial, but with the redefinition of `member` it's trickier to understand the type. I would have trouble explaining whether `callback` is needed or not.

rust/src/ffi/any.rs
```diff
@@ -354,6 +354,9 @@ pub struct CallbackFn {
 // wrap a CallbackFn in a closure, so that it can be used in transformations and measurements
 pub fn wrap_func(func: CallbackFn) -> impl Fn(&AnyObject) -> Fallible<AnyObject> {
     move |arg: &AnyObject| -> Fallible<AnyObject> {
+        // extends the lifetime of func.callback to the lifetime of this closure
+        let _ = &func.lifeline;
```
mccalluc, line 358: Is there any particular relationship between this and the similar line in `ffi.rs`? Like, would it serve any purpose to reference the other case in both locations? Perhaps not- If someone needs to touch this again, the connections might be clear to them.

mccalluc (commented):
Looks good overall. If you have a fix for the R, maybe get that in first, and then pick up that fix in this PR. Or all in one big PR? Whatever makes sense.

> I was able to deterministically reproduce the error by mixing collects in at a certain spot, I'll add it back in

Great. What I meant by marking it was `xfail` was just kind of a test-first approach:
- Make a test that demonstrates the bug behavior, either in a separate PR, or in an earlier commit in the same PR.
- And then fix it: If the xfail test was already in main, the reviewer can see in the diff that the xfail goes away, or if it's in the same PR, a test passes that failed on an earlier commit.

The code at the end is the same, but seeing the change in behavior can help the reviewer be a little more confident.

rust/src/ffi/any.rs
Shoeboxam: They are unrelated, but they do use the same trick!

mccalluc (dismissed):
Thanks!

PR #2280, opened 2025-02-14
fix: compatibility with R lintr 3.2

### LINKED ISSUE #2279: R malformed .lintr file
**Describe the bug**
The R linter has started to fail in CI without any changes and with an unclear error message.

**To Reproduce**
https://github.com/opendp/opendp/actions/runs/13323279467/job/37211670001?pr=2278
https://github.com/opendp/opendp/actions/runs/13310323010/job/37171581468?pr=2214

Loading the dcf file locally produces no error:
```R
read.dcf(".lintr")
```

Running the linter locally produces no error:
```shell
Rscript -e 'lintr::lint_dir("docs/source")'
```

Using R 4.4.2, which matches the version on GitHub Actions.

No clear culprit in lintr changelog:
https://lintr.r-lib.org/news/#lintr-312

Shoeboxam:
Close #2279

### DIFF
--- a/docs/source/.lintr
+++ b/docs/source/.lintr
@@ -5,6 +5,6 @@ linters: all_linters(
     # With "all_linters", these problems are also reported:
     todo_comment_linter = NULL, # TODO comments should be removed.
     undesirable_function_linter = NULL, # Function "library" is undesirable. As an alternative, use roxygen2's @importFrom statement in packages and `::` in scripts, instead of modifying the global search path.
-    one_call_pipe_linter = NULL
+    one_call_pipe_linter = NULL # Avoid pipe |> for expressions with only a single call.
     ) # Closing paren must be indented, and no closing comma above!
 encoding: "UTF-8"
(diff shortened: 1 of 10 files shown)


docs/source/.lintr
```diff
@@ -5,5 +5,6 @@ linters: all_linters(
     # With "all_linters", these problems are also reported:
     todo_comment_linter = NULL, # TODO comments should be removed.
     undesirable_function_linter = NULL, # Function "library" is undesirable. As an alternative, use roxygen2's @importFrom statement in packages and `::` in scripts, instead of modifying the global search path.
+    one_call_pipe_linter = NULL
```
mccalluc, line 8:
```suggestion
    one_call_pipe_linter = NULL # Avoid pipe |> for expressions with only a single call.
```
Unless it would be less trouble to direct users to the other config for explanations on all of these.

mccalluc (commented):
All the changes seem reasonable, but do you understand why this started breaking? Is there something we can pin? Or is pinning isn't the best approach, an additional error message on failure so the next time we won't be surprised, and know the steps to fixing it?

mccalluc (commented):
Changes look good, but now there's a CI problem with python that would need to be fixed before merging to main, so I'll hold off on approval.

Shoeboxam:
@mccalluc The failure was the segfault to be fixed by the child PR.

mccalluc (approved):
Thanks for re-running. Approved!

PR #2288, opened 2025-02-19
fix: always default to include NaN in domains

### LINKED ISSUE #1673: Rename AtomDomain's null terminology
Mike says: nullable came along before polars... so this is right, but maybe the name should change. File issue?

_Originally posted by @mccalluc in https://github.com/opendp/opendp/pull/1590#discussion_r1595755501_

The risk we run is that renaming to NaN is too specific. It's really supposed to encompass any missing value. Perhaps we should rename along the lines of `missingness`.

Shoeboxam:
Fix #1673


Shoeboxam:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/opendp/opendp/2288?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2288** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2288?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/opendp/opendp/2288?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2240** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2240?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

python/src/opendp/extras/numpy/_make_np_count/__init__.py
```diff
@@ -13,6 +13,10 @@ def make_np_count(input_domain: Domain, input_metric: Metric) -> Transformation:
     :param input_metric: instance of `symmetric_distance()`
     """
     dp.assert_features("contrib")
+
+    if not str(input_domain).startswith("NPArray2Domain"):
+        raise ValueError("input_domain must be NPArray2Domain")  # pragma: no cover
```
mccalluc, line 18: With domain objects having more accessible structure now, can we get away from looking for substrings?

python/src/opendp/extras/numpy/_make_np_sscp/__init__.py
```diff
@@ -1,52 +1,38 @@
...
     dp.assert_features("contrib", "floating-point")
 
     if not str(input_domain).startswith("NPArray2Domain"):
-        raise ValueError("input_domain must be a 2d-numpy array domain")  # pragma: no cover
+        raise ValueError(
+            "input_domain must be a 2d-numpy array domain"
+        )  # pragma: no cover
 
+    if input_metric != dp.symmetric_distance():
+        raise ValueError("input metric must be symmetric distance")  # pragma: no cover
```
mccalluc, line 27: For here and below: Include the `input_domain` or `input_metric` in the error message?

Test coverage?

python/src/opendp/extras/numpy/_make_np_sum/__init__.py
```diff
@@ -20,6 +20,12 @@ def make_np_sum(input_domain: Domain, input_metric: Metric) -> Transformation:
 
     dp.assert_features("contrib", "floating-point")
 
+    if not str(input_domain).startswith("NPArray2Domain"):
+        raise ValueError("input_domain must be NPArray2Domain")  # pragma: no cover
+    
+    if input_domain.descriptor.nan:
+        raise ValueError("input_domain must not permit NaN elements")  # pragma: no cover
```
mccalluc, line 27: Ditto: Prefer to always include what we got instead, and to try to have test coverage, unless this is a branch we shouldn't go down unless there's a bug.

python/src/opendp/mod.py
```diff
@@ -633,10 +633,13 @@ def bounds(self) -> tuple[float, float] | None:
...
     
     @property
-    def nullable(self) -> bool:
-        '''Whether the domain includes null values'''
-        from opendp.domains import _atom_domain_get_nullable
-        return _atom_domain_get_nullable(self)
+    def nan(self) -> bool:
+        """Whether the domain includes NaN values
+        
+        Only relevant when the carrier type is a floating point type.
```
mccalluc, line 639:
```suggestion
        Only relevant when the carrier type is a floating point type.
        All other types will always return ``False``.
```

mccalluc (commented):
It's a big change! It's not unreasonable, but it is asking the user for another kwarg in many places, but that might be right call for consistency.

Apart from the code, I'd suggest just taking a day or two away from this, and confirming when you look at it with fresh eyes that it still makes sense.

Other thoughts:
- I think this needs be explained somewhere early in the getting-started docs, along with impute, which is being used more frequently. 
- show the expected error message in a doctest at the same point: Not sure how much instruction should be baked in to the error message, but if someone does get the error message, it'd be nice if they can search the docs and get a full explanation.
- Style: When `nan` stands on its own, should we prefer `NaN`? (Search and replace for `\bnan\b` should be easy.)
- `nullable` had the advantage of looking like an adjective. Would something like `nan_allowed` be ok? though that is getting a little long.
- The logic about `None` seemed weird to me, but I could be thinking about it wrong.

R/opendp/tests/testthat/test-domains.R
```diff
@@ -26,5 +26,5 @@ test_that("map_domain", {
   domain <- map_domain(atom_domain(.T = "String"), atom_domain(.T = "f64"))
   expect_true(member(domain, new_hashtab(c("A", "B"), c(3, 4))))
   expect_equal(domain_carrier_type(domain), "HashMap<String, f64>")
-  expect_equal(domain_debug(domain), "MapDomain { key_domain: AtomDomain(T=String), value_domain: AtomDomain(T=f64) }")
+  expect_equal(domain_debug(domain), "MapDomain { key_domain: AtomDomain(T=String), value_domain: AtomDomain(nan=true, T=f64) }")
```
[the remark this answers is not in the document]
Shoeboxam: It wasn't true by default before for float T.

python/src/opendp/extras/numpy/__init__.py
```diff
@@ -123,6 +125,15 @@ def array2_domain(
     T = dp.RuntimeType.parse(T)
     if T not in ATOM_MAP:
         raise ValueError("T must be an elementary type")  # pragma: no cover
+    
+    if T in {"f32", "f64"}:
+        # if nan is None, then assume NaN values are present
+        if nan is None:
+            nan = True
```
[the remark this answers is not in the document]
Shoeboxam: `nan=True` is only valid for float T, though? Are you saying we should let `nan=True` for non-float T?

python/src/opendp/extras/numpy/_make_np_count/__init__.py
Shoeboxam: Resolved in follow-up PR

python/src/opendp/extras/numpy/__init__.py
```diff
@@ -117,7 +117,9 @@ def array2_domain(
     _check_nonnegative_int(size, "size")
     _check_nonnegative_int(num_columns, "num_columns")
 
-
+    if nan is None:
+        nan = T in {"f32", "f64"}
+    
     T = T or _ELEMENTARY_TYPES.get(origin.dtype.type)
```
mccalluc, line 123: If the order of these lines were switched, the behavior would change:
```
T = T or _ELEMENTARY_TYPES.get(origin.dtype.type)

if nan is None:
    nan = T in {"f32", "f64"}
```
In this order, if T started as `None`, but `origin.dtype.type` was `float`, `nan` will end up `True`.

If it's correct just the way it is, maybe add a comment to explain?

mccalluc (commented):
Could you double check a couple lines? Might be fine as is.

python/src/opendp/extras/numpy/__init__.py
Shoeboxam: Thanks good catch.

mccalluc (dismissed):
Thanks!

PR #2291, opened 2025-02-20
dev: precommit hooks, without `rust fmt`

mccalluc:
- Towards #2271
- Adding Rust formatting will be a follow up: #2273

Realized that out of the box, precommit default would make a lot of whitespace changes, and these will generate lots of merge conflicts eventually.

For the reviewer:
- Is this actually the direction we want to go? If this whitespace formatting is too fussy, maybe we disable these and only use it for code formatting.
- Is this the right time for it, or should we get outstanding changes in place before integrating this?


mccalluc:
our dev instructions are already heavy. Don't want to add to the burden.

PR #2299, opened 2025-02-28
Deprecate margins dict

### LINKED ISSUE #2297: Warn if dict is used for margins
I could be confused (I've been confused before!), but we should check what happens if the old Margin syntax is used. Working on DP Wizard right now, and I suspect old code will still run, but the dict keys are ignored, and you don't get the right results.

- Write a test to simulate this.
- If it runs without error, add a check that the margins are not a dict (or map?), with a message for the user.
- and the test should now fail: change it to `raises`

mccalluc:
- Fix #2297

I'd have a slight preference that there be one right way to use the library, just so the users aren't left wondering if the dict and list somehow work differently. Feel free to close if this seems too fussy.

### DIFF
--- a/python/src/opendp/context.py
+++ b/python/src/opendp/context.py
@@ -407,8 +407,11 @@ class Context(object):
             domain = domain_of(data, infer=True)
 
         if margins:
+
             # allows dictionaries of {[by]: [margin]}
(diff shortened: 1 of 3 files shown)


Shoeboxam (approved):
Thanks!

PR #2303, opened 2025-03-03
Docs: Add a note about tools to consider for param consistency checks

### LINKED ISSUE #2296: AST checks: Reminder of why we're doing this ourselves -- and maybe make it available to others?
Before launching into the AST tests (these check that our docstrings are actually consistent with the code) I did spend some time looking for an off-the-shelf solution. My memory is that while there were some packages that would check other documentation styles, I just couldn't find anything for RST.

- Double check this assumption.
- If it's correct, document the research, explaining why the other options don't quite work.
- Perhaps factor this out into a separate package, so other projects can benefit?

It's not a high priority, but pulling out stable code that isn't part of the core logic for OpenDP would be a small simplification.

mccalluc:
- Fix #2296

I think this has nudged us in a good direction, but in the long run I don't think we need to maintain this tooling. Either docsig or pydoclint might be a good replacement.

### DIFF
--- a/python/test/test_ast.py
+++ b/python/test/test_ast.py
@@ -1,3 +1,50 @@
+'''
+These checks have grown over time, and off-the-shelf tools have been improving at the same time:
+Now that our docs are cleaner, and the tools are also better, it might make sense to drop test_ast.py.
+Sphinx/RST isn't supported by all tools, but we still have a couple good options.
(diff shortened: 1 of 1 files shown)


Shoeboxam (approved):
Thanks for the utility analysis on these. Both of the top options look good/preferable to the others. Would also be good to unload some of the non-library-specific code.

PR #2307, opened 2025-03-04
feat: binary function expressions

### LINKED ISSUE #2196: Add weighted means
Useful for a use-case!

Shoeboxam:
Fix #2196
Makes weighted queries possible, for use-case.

### DIFF
--- a/rust/src/transformations/make_stable_expr/expr_binary/mod.rs
+++ b/rust/src/transformations/make_stable_expr/expr_binary/mod.rs
@@ -1,12 +1,9 @@
 use polars::prelude::*;
 use polars_plan::dsl::Expr;
 use polars_plan::utils::expr_output_name;
-// This code is not particular to Python, so we shouldn't need the higher-level library here.
-// https://github.com/opendp/opendp/issues/2309
-use pyo3_polars::export::polars_core::utils::try_get_supertype;
 
 use crate::core::{Function, MetricSpace, StabilityMap, Transformation};
-use crate::domains::{AtomDomain, ExprDomain, ExprPlan, OuterMetric, SeriesDomain, WildExprDomain};
+use crate::domains::{ExprDomain, ExprPlan, OuterMetric, SeriesDomain, WildExprDomain};
 use crate::error::*;
 use crate::transformations::DatasetMetric;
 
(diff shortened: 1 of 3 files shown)


Shoeboxam:
* **#2307** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2307?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/opendp/opendp/2307?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rust/src/transformations/make_stable_expr/expr_binary/test.rs
```diff
@@ -285,3 +320,84 @@ fn test_xor() -> Fallible<()> {
...
+            Some(f64::NAN),
+            Some(f64::NAN),
+            None,
+            None,
+            Some(f64::NAN),
+            Some(0.)
+        ]
+    )?;
+    test_binary!(get_i32_data, div, [None, Some(1), None])?;
+    Ok(())
```
mccalluc, line 402: If someone did need to debug this, it would be hard to cross reference things. Some ideas for the future:
- limit the use of fixtures to examples where you expect the same behavior for all members
- or, depending on the number of members, the content might be represented in the name: `nan_null_inf_zero`
- or, make the expected outputs part of the fixture, and different tests will pull out different columns
  - since this will be more complex data structure, consider other representations: Delimited strings, or even outside files that are read by the test.

The general idea is to enable someone reading the test in isolation to confirm that the behavior is correct.

I don't know that this needs to block this PR, but it's also the case that I'm not actually closely reviewing the assertions of these tests because of the readability.

rust/src/transformations/make_stable_expr/expr_binary/test.rs
```diff
@@ -285,3 +320,84 @@ fn test_xor() -> Fallible<()> {
...
+            Some(f64::INFINITY),
+            Some(f64::NAN)
+        ]
+    )?;
+    test_binary!(get_i32_data, mul, [Some(0), Some(1), None])?;
+    Ok(())
+}
+
+#[test]
+fn test_div() -> Fallible<()> {
```
mccalluc, line 386: I think we have three different kinds of division now? Test them all?

mccalluc (commented):
How will users know about this? Consider adding docs in this PR or filing a follow-up issue. (There might be a larger feature set you're working on, and it may make sense to wait until it's all in place, as long as we have a reminder to fill the gap.)

rust/src/transformations/make_stable_expr/expr_binary/test.rs
```diff
@@ -77,7 +112,7 @@ macro_rules! test_binary {
             .unwrap();
 
         assert!(output_series.nullable == (out > 0));
-        Ok(())
+        Fallible::Ok(())
```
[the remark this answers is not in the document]
Shoeboxam: Compiler is, for some reason, unable to infer the error type, so this disambiguates it.

mccalluc (approved):
I think the git history is flattened, so I'm not sure exactly what are the latest changes. Approving, but I think comments explaining the places where we're just trying to duplicate the behavior of Polars would be useful, but doesn't need to be this PR.

rust/src/transformations/make_stable_expr/expr_binary/mod.rs
```diff
@@ -69,9 +60,49 @@ where
...
+                }
+                _ => unreachable!("due to above match arm"),
+            };
+
+            let mut series_domain = SeriesDomain::new_from_field(Field::new(
+                left_series.name.clone(),
+                out_dtype.clone(),
+            ))?;
+
+            // output is nullable when op is FloorDivide since a // 0 is null for any a
```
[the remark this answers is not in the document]
Shoeboxam: I just loosened it to be safe.

rust/src/transformations/make_stable_expr/expr_binary/mod.rs
```diff
@@ -69,9 +60,49 @@ where
...
+            let common_dtype = try_get_supertype(&left_series.dtype(), &right_series.dtype())?;
+
+            let out_dtype = match op {
+                Plus | Minus | Multiply | FloorDivide => common_dtype,
+                TrueDivide | Divide => {
+                    if common_dtype.is_float() {
+                        common_dtype
+                    } else {
+                        DataType::Float64
+                    }
```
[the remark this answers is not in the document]
Shoeboxam: After doing some testing, the relationship was more complicated.

Shoeboxam:
@mccalluc For Monday, I replaced the branching logic with a mock execution and flattened everything out/loosened the descriptors/added the remaining unsupported binary ops.

rust/src/transformations/make_stable_expr/expr_binary/mod.rs
```diff
@@ -62,50 +59,61 @@ where
...
+    let mut series_domain = SeriesDomain::new_from_field(field)?;
+
+    // division by zero may introduce null values
+    series_domain.nullable = left_series.nullable
+        || right_series.nullable
+        || matches!(op, FloorDivide | TrueDivide | Divide);
+
+    if matches!(op, EqValidity | NotEqValidity) {
+        series_domain.nullable = false;
+    }
```
mccalluc, line 113: Combining this into one big expression might be a little safer over time. Something like:
```
    series_domain.nullable = ! matches!(op, EqValidity | NotEqValidity) && (
        left_series.nullable
        || right_series.nullable
        || matches!(op, FloorDivide | TrueDivide | Divide));

File: rust/src/transformations/make_stable_expr/expr_binary/test.rs [hunk 7]
```diff
@@ -135,6 +135,18 @@ fn test_eq() -> Fallible<()> {
     )
 }
 
+#[test]
+fn test_eq_missing() -> Fallible<()> {
+    test_binary!(
+        get_f64_data,
+        eq_missing,
+        [
+            true, false, // NaN is equal to NaN?
```
mccalluc, line 144: This is surprising, and doesn't seem to be the case in other implementations. Python for example:
```
>>> float('nan') == float('nan')
False
```
I think this is also the IEEE standard.

I'd suggest filing an issue and linking to it here (and in the test above), and then understanding where this behavior comes from. Even if it's not a mistake, there should still be some explanation of this in the polars docs?

rust/src/transformations/make_stable_expr/expr_binary/test.rs
```diff
@@ -421,3 +445,23 @@ fn test_floor_div() -> Fallible<()> {
...
+            Some(f64::NAN),
+            None,
+            None,
+            Some(f64::NAN),
+            Some(f64::NAN)
+        ]
+    )?;
+    test_binary!(get_i32_data, rem, [None, Some(0), None])?;
+    Ok(())
+}
```
mccalluc, line 467: Not something to block this PR, and there's a longer note in an earlier, but I'm not comfortable with this style of testing.

Since the inputs are separated from the asserted outputs, it's easy to make this just an assertion of the current behavior, without thinking about what that behavior should be. I believe you're being careful, but I'm not really checking that each assertion makes sense, and it may also be hard for a future maintainer.

mccalluc (commented):
Just looking at the most recent changes. The behavior for NaNs is surprising: A comment linking to an issue to get to the root of it would be good.

Comment on boolean expression you can take or leave.

Test style is probably something to talk about: It worries me, but I might be thinking about it wrong.

rust/src/transformations/make_stable_expr/expr_binary/test.rs
Shoeboxam: It's a documented departure from IEEE754. I think they'd rather run a (fast) bitwise comparison.

rust/src/transformations/make_stable_expr/expr_binary/mod.rs
Shoeboxam: This is more difficult to read, IMO. 

Currently, there's some cases that could introduce nulls. And then there's some ops that eliminate all nulls.

Shoeboxam:
Re-requested review, added another test to be safe, to track behavior of overflow. [If this behavior changes](https://github.com/pola-rs/polars/issues/22216#issuecomment-2794881592) then we'll know and can remediate.

mccalluc (approved):
Looks good. Thank you for the explanation on rust behavior.

PR #2317, opened 2025-03-13
feat: Better error message if bad margin

mccalluc:
Filing as draft. I've added a test for a related check that didn't have a python test, but I'm not clear on how we would exercise this new check.


Shoeboxam (approved):
Thanks for doing this!

PR #2319, opened 2025-03-15
fix: openssl compilation in smoke-test

### LINKED ISSUE #2305: rust-test doesn't work with ubunu-latest (24.02)
Currently, `ubuntu-latest` is 24.02: https://github.com/actions/runner-images

For some reason, with this version, the [rust-tests fail](https://github.com/opendp/opendp/actions/runs/13641166881/job/38174429598), because individual tests do not complete.
```
test src/domains/mod.rs - domains::VectorDomain (line 407) has been running for over 60 seconds
test src/domains/polars/frame/mod.rs - domains::polars::frame::FrameDomain (line 68) has been running for over 60 seconds
test src/domains/polars/series/mod.rs - domains::polars::series::SeriesDomain (line 27) has been running for over 60 seconds
```

For now, I've pinned just that action to a lower version.

- follow up to #2304 
  - which fixes #2301

Shoeboxam:
Related to #2305

Ubuntu 24.04 updated glibc beyond 0.39, where glibc changed header functions to add `cstd23` aliases for various functions. Our GitHub actions speeds up the build by reusing/caching rlib files from parent actions, which use Ubuntu 24.04. Therefore the openssl-sys rlib from the cache contains references to glibc `cstd23` functions that are only valid on Ubuntu 24.04, which don't yet exist in the glibc of Ubuntu 22.xx.

This PR standardizes CI to ubuntu-22.04, and updates pip to fix a packaging.licenses docs CI failure.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -284,7 +284,7 @@ jobs:
 
   python-sdist:
     needs: bindings
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
     - uses: actions/checkout@v4
       with:
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -13,7 +13,7 @@ on:
 
 jobs:
   rust-build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     defaults:
       run:
         working-directory: rust
@@ -90,8 +90,6 @@ jobs:
 
   rust-test:
     needs: rust-build
-    # TODO: Upgrade to ubuntu-latest.
-    # https://github.com/opendp/opendp/issues/2305
     runs-on: ubuntu-22.04
     defaults:
       run:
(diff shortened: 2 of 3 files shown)


Shoeboxam:
* **#2319** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2319?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/opendp/opendp/2319?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mccalluc (approved):
This PR looks good, but I think we want to leave this issue open?:
- #2305

rust-test still doesn't work with ubunu-latest, and at some point we'll need to fix this, if not for Ubuntu 24, then Ubuntu 26. It's a little tech debt, not a high priority, but we should keep an issue around. (Removing the TODO is fine.)

If we come back in a few months, maybe the fall, someone else may have hit it by then.

Issue #2323, opened 2025-03-18
More descriptive error messages, with links to docs


mccalluc:
In addition to simply making the error messages more descriptive, and possibly failing earlier (#2315):
- Error messages can contain links to a docs page with more information
- The URL will include the current version number (see `make_prook_link`)
- The page might be at the end of the API user guide.
- We should have a check that any URL anchor in an error corresponds to an anchor on the page.
- One section on the page per error
  - Minimal reproducer python doctest
  - Explanation
  - Fix
- With doctests in place, we can remove some of the usability unit tests.

PR #2324, opened 2025-03-19
docs: Start error reference in User Guide

mccalluc:
- Towards #2323

👉 Big question is whether a reference page for errors is good way to improve usability, compared with all the other things we could spend time on.

I think that adding an example that works will be an important complement to each failing doc test: Coming up with a passing example might take a little time in some cases.

The messages themselves might be enriched: in this case I might have added the bad margin to the error message. If we're visiting these points in the code, maybe we also want to make that update?

It's up to the user to make sure the title of their section actually matches the string used for the URL. While it would be possible for `get_error_url` to read `errors.rst` and confirm that the string matches, that feels a little weird.

### DIFF
--- a/docs/source/api/user-guide/errors.rst
+++ b/docs/source/api/user-guide/errors.rst
@@ -1,0 +1,47 @@
+Errors
+======
+
+OpenDP is designed to prevent operations which would violate the guarantees of differential privacy.
+This means there are a lot of potential errors, some more clear than others.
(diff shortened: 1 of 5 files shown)


mccalluc:
`cargo fmt` error. Should fix in main, and then bring that in here.

mccalluc:
The error that's in this PR right now doesn't really need more explanation... but others might: So don't merge this now, but see if there are places where it would be useful.

Issue #2274, opened 2025-02-10
Add `black` to precommit hooks


mccalluc:
- follow-up to #2273

Will need to reformat existing codebase, so wait until there are few python PRs outstanding.

PR #2327, opened 2025-03-20
Apply black to __init__.py and _convert.py

mccalluc:
- Towards #2274

Working on DP Wizard, I've gotten to like black formatting. Also one Rust whitespace fix.

Plus:
- Consistent python style across OpenDP projects
- Analogous to the `cargo fmt` we're already using on the Rust side
- Getting the format clean is either a pre-req or primary motivation for adding a pre-commit hook 
- If this looks good, plan to make more small PRs to avoid merge conflicts

Minus:
- Benefits are subjective
- Makes `git blame` harder to use
- Might not apply to generated code in the end, and the inconsistency might be a nuissance

### DIFF
--- a/python/src/opendp/__init__.py
+++ b/python/src/opendp/__init__.py
@@ -1,1 +1,6 @@
-from opendp.mod import Transformation, Measurement, OpenDPException, UnknownTypeException
+from opendp.mod import (
+    Transformation,
+    Measurement,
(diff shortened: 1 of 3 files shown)


mccalluc:
Talking with Mike: Our dev instructions are already heavy, don't want to add burden.

PR #2335, opened 2025-03-28
Update use case list

### LINKED ISSUE #2334: Update use case list in docs
> Shall we add the UNHCR use case here? https://docs.opendp.org/en/stable/index.html
> 
> Also, we should soften what we say about Swiss FSO.  I don't think they actually released income statistics with DP; rather they prototyped tools to do this and are exploring whether to bring them into production.
> 
> Let's also separate the New Digital Divide into a bullet of its own and say a bit more about it.  That was not just researchers at MS, but at MS and HBS.
> 
> Here's the link to the UNHCR blog for reference: https://opendp.org/blog/unhcr-uses-differential-privacy-their-microdata-library

mccalluc:
- Fix #2334

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -41,10 +41,12 @@ Who is using OpenDP?
 
 Some of the applications of OpenDP in healthcare, government, and tech include:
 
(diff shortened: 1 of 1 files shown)


Shoeboxam (approved):
Thanks!

Issue #1382, opened 2024-03-14
Context API: register user-defined constructors


Shoeboxam:
We have the following functions in the extrinsics module (that are not public, but could be made public):

* register_transformation
* register_measurement

When someone registers a constructor, there should be a way to add it to the constructors dict in `context.py`.

Shoeboxam:
Closed in #1452

PR #2343, opened 2025-04-10
feat: constructor registration in Context API

### LINKED ISSUE #2338: Context API constructor registration
I thought we already had this!

```python
# example constructor
def make_duplicate(input_domain, input_metric, k):
    assert input_metric == dp.symmetric_distance()
        input_domain, input_metric,
        input_domain, input_metric,
        lambda x: x * k,
        lambda d_in: d_in * k
    )

# proposed new functionality
Context.register(make_duplicate)

# now you can call it in your chain
query = context.query().duplicate(4).count().laplace()
query.release()
```

Shoeboxam:
#1382 was not resolved
closes #2338 

This PR also stops registering constructors in the generated modules, opting instead to register them in the context API.

### DIFF
--- a/docs/source/api/user-guide/plugins/context-api-plugins.rst
+++ b/docs/source/api/user-guide/plugins/context-api-plugins.rst
@@ -1,0 +1,66 @@
+Context API Plugins
+===================
+
+Constructor functions built from plugins can be registered in the Context API.
+
+We'll start by briefly copying over the constructor from the :ref:`measurement-plugin` and registering it.
+The constructor function may now be called from the Context API.
+
+.. tab-set::
+
+  .. tab-item:: Python
+
+    .. code:: python
+
+        >>> def make_int_constant(constant):
+        ...     return dp.m.make_user_measurement(
+        ...         input_domain=dp.atom_domain(T=int),
+        ...         input_metric=dp.absolute_distance(T=int),
+        ...         output_measure=dp.max_divergence(),
+        ...         function=lambda _: constant,
+        ...         privacy_map=lambda _: 0.0,
+        ...     )
+        ...
+        >>> dp.register(make_int_constant)
+        ...
+        >>> context = dp.Context.compositor(
+        ...     data=[1, 2, 3],
+        ...     privacy_unit=dp.unit_of(contributions=36),
+        ...     privacy_loss=dp.loss_of(epsilon=1.0),
+        ...     split_evenly_over=2,
+        ... )
+        >>> context.query().clamp((0, 5)).sum().int_constant("denied").release()
+        'denied'
+
+A drawback of this approach is that the constructor function is not very flexible.
+The input domain and metric are hard-coded, only accepting integers, 
+and can't take into account the output domain and output metric of the previous transformation.
+
+If the first two arguments to the constructor function are the ``input_domain`` and ``input_metric``,
+then they can be omitted when you call the function from the context API, and will be passed up from the previous transformation.
+
+.. tab-set::
+
+  .. tab-item:: Python
+
+    .. code:: python
+
+        >>> def make_anything_constant(input_domain, input_metric, constant):
+        ...     return dp.m.make_user_measurement(
+        ...         input_domain=input_domain,
+        ...         input_metric=input_metric,
+        ...         output_measure=dp.max_divergence(),
+        ...         function=lambda _: constant,
+        ...         privacy_map=lambda _: 0.0,
+        ...     )
+        ...
+        >>> dp.register(make_anything_constant)
+        >>> context.query().anything_constant("denied").release()
+        'denied'
+
+This modified constructor function doesn't care what the input domain and metric are,
+and will happily build a measurement that will always conform with the previous transformation.
+In practice, the constructor function usually contains checks to ensure that the input domain and metric are meaningful for your function.
+
+Constructors in the OpenDP Library almost always accept the input domain and metric as the first two arguments,
+and we recommend it when building your own plugins.
--- a/docs/source/api/user-guide/plugins/context-api-plugins.rst
+++ b/docs/source/api/user-guide/plugins/context-api-plugins.rst
@@ -2,9 +2,47 @@ Context API Plugins
 ===================
 
 Constructor functions built from plugins can be registered in the Context API.
+This follows up on the :ref:`measurement-plugin` example,
+where we built a measurement that always returns a constant value.
 
-We'll start by briefly copying over the constructor from the :ref:`measurement-plugin` and registering it.
-The constructor function may now be called from the Context API.
+Constructors in the OpenDP Library almost always accept the input domain and metric as the first two arguments,
+and we recommend it when building your own plugins.
+When the first two arguments of the constructor function are the ``input_domain`` and ``input_metric``,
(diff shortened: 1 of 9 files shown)


Shoeboxam:
* **#2343** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2343?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/opendp/opendp/2343?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

docs/source/api/user-guide/plugins/context-api-plugins.rst
```diff
@@ -0,0 +1,66 @@
...
+        >>> dp.register(make_anything_constant)
+        >>> context.query().anything_constant("denied").release()
+        'denied'
+
+This modified constructor function doesn't care what the input domain and metric are,
+and will happily build a measurement that will always conform with the previous transformation.
+In practice, the constructor function usually contains checks to ensure that the input domain and metric are meaningful for your function.
+
+Constructors in the OpenDP Library almost always accept the input domain and metric as the first two arguments,
+and we recommend it when building your own plugins.
```
mccalluc, line 66: How about if you delete the first example, and just keep the second? It's what the user should probably be doing in any case, and I think it can stand on its own.

(That a simpler version is possible, if you don't need the flexibility, could be mentioned in passing at the end.)

python/src/opendp/context.py
```diff
@@ -90,6 +92,40 @@
...
+    name: Optional[str] = None,
+):
+    """Register a constructor function to be used in the Context API.
+
+    If the constructor supports partial application (first two arguments are ``input_domain`` and ``input_metric``),
+    then the input domain and input metric are omitted when called via the Context API.
+
+    Constructor requirements:
+    - The constructor must return a transformation or measurement.
+    - If name is None, the constructor's name must start with ``make_``.
```
mccalluc, line 106:
```suggestion
    * The constructor must return a transformation or measurement.
    * If name is None, the constructor's name must start with ``make_``.
```
RST doesn't convert `-` to bullet list.

python/src/opendp/context.py
```diff
@@ -90,6 +92,40 @@
...
+        if not constructor.__name__.startswith("make_"):
+            raise ValueError(  # pragma: no cover
+                f"constructor.__name__ must start with 'make_', found {constructor.__name__}"
+            )
+        name = constructor.__name__[5:]
+
+    if name in constructors:
+        raise ValueError(  # pragma: no cover
+            f"'{name}' is already registered in the Context API. Please choose a different name."
+        )
```
mccalluc, line 121: These would be easy to test.

python/src/opendp/context.py
```diff
@@ -90,6 +92,40 @@
...
 
 
+def register(
+    constructor: Callable[..., Union[Transformation, Measurement]],
+    name: Optional[str] = None,
+):
+    """Register a constructor function to be used in the Context API.
+
+    If the constructor supports partial application (first two arguments are ``input_domain`` and ``input_metric``),
+    then the input domain and input metric are omitted when called via the Context API.
```
mccalluc, line 102: Having these two parameters is the convention for `then_` functions, I think? Maybe allow either `make_` or `then_` prefix from the user, and our example should use `then_`?

If we get a function named `make_whatever`, but it has the right arguments, I don't think we need to make it a big error.

python/src/opendp/context.py
```diff
@@ -920,8 +1014,11 @@ def _caster(measurement):
...
+        rho = _translate_measure_distance(
+            (epsilon, delta_zCDP), from_measure, zero_concentrated_divergence()
+        )
         return rho, delta_inf
-        
 
-    raise ValueError(f"Unable to translate distance from {from_to[0]} to {from_to[1]}")  # pragma: no cover
+    raise ValueError(
+        f"Unable to translate distance from {from_to[0]} to {from_to[1]}"
+    )  # pragma: no cover
```
mccalluc, line 1024: Slight preference for keeping reformating in separate PR.

python/src/opendp/extras/numpy/_make_np_clamp/__init__.py
```diff
@@ -77,4 +78,5 @@ def _function(arg):
 
 # generate then variant of the constructor
 # TODO: Show this in the API Reference?
-then_np_clamp = register_transformation(make_np_clamp)
+register(make_np_clamp)
+then_np_clamp = to_then(make_np_clamp)
```
mccalluc, line 82: Does `then_np_clamp` need to be registered?

python/src/opendp/extras/sklearn/_make_eigendecomposition/__init__.py
```diff
@@ -64,6 +65,5 @@ def make_private_np_eigendecomposition(
 
 
 # generate then variant of the constructor
-then_private_np_eigendecomposition = register_measurement(
-    make_private_np_eigendecomposition
-)
+then_private_np_eigendecomposition = to_then(make_private_np_eigendecomposition)
+register(make_private_np_eigendecomposition)
```
mccalluc, line 69: Ditto. Does the `then` need registration?

mccalluc (commented):
Some little comments.

I'm wondering about how this intersects with query plan serialization. It seems like the serializations might have references which would not be satisfiable if they are deserialized in a new environment. Not sure we should try to catch this in the code, but maybe mention in the docs? Or I might be thinking about this wrong.

python/src/opendp/context.py
mccalluc: But don't worry about it here!

docs/source/api/user-guide/plugins/context-api-plugins.rst
```diff
@@ -0,0 +1,66 @@
+Context API Plugins
+===================
+
+Constructor functions built from plugins can be registered in the Context API.
+
+We'll start by briefly copying over the constructor from the :ref:`measurement-plugin` and registering it.
```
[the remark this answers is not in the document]
Shoeboxam: I changed the wording a bit.

docs/source/api/user-guide/plugins/context-api-plugins.rst
Shoeboxam: Good suggestion! Thanks.

python/src/opendp/context.py
Shoeboxam: The user will always have a `make_` function, but only sometimes (or rarely) have a then function, if they find the `to_then` function. I'm leaning towards keeping the registration simpler and having less flexibility. It would also introduce an ambiguity if `name` is set, unless I still detect based on the constructor name.

python/src/opendp/extras/numpy/_make_np_clamp/__init__.py
Shoeboxam: Nope, just make. The Context API papers over this distinction.

python/src/opendp/context.py
Shoeboxam: It's a mistake when it happens. Sorry for the noise.

PR #2345, opened 2025-04-11
fix: Polars group by with explicit keys fill value

### LINKED ISSUE #2344: Polars group by with explicit keys broadcasts fill value
Bug shared by Daniel Simmons-Marengo. Thank you!

The issue comes from a Polars performance optimization that I missed. `lit(0)` is kept scalar for as long as possible until it must be broadcast. The broadcasting then happens after noise addition. `repeat(lit(0), len())` fills the column, resulting in the expected behavior.

```python
import opendp.prelude as dp
import polars as pl

dp.enable_features("contrib")

    data=pl.scan_csv(dp.examples.get_france_lfs_path(), ignore_errors=True),
    privacy_loss=dp.loss_of(epsilon=1.0, delta=1e-6),
    split_evenly_over=1,
)


keys_age = pl.DataFrame({"AGE": [float(i) for i in range(100)]})
keys_ilostat = pl.DataFrame({"ILOSTAT": [1, 2, 3, 4, 5, 9]})

query = (
    context.query()
    .filter(pl.col("HWUSUAL") != 99.0)
    .group_by("AGE", "ILOSTAT")
    .agg(dp.len())
    .with_keys(keys_age.join(keys_ilostat, how="cross"))
)

print(query.release().collect().group_by("len").count())
```

```
shape: (6, 2)
┌───────┬───────┐
│ len   ┆ count │
│ ---   ┆ ---   │
│ u32   ┆ u32   │
╞═══════╪═══════╡
│ 73    ┆ 1     │
│ 33786 ┆ 1     │
│ 6431  ┆ 1     │
│ 10578 ┆ 1     │
│ 27709 ┆ 1     │
│ 14    ┆ 595   │
└───────┴───────┘
```

Shoeboxam:
Closes #2344

In addition to fixing the bug, adds distributional tests to ensure we're getting the outputs we expect.

### DIFF
--- a/rust/src/polars/mod.rs
+++ b/rust/src/polars/mod.rs
@@ -280,11 +280,11 @@ impl<TI: 'static> Function<TI, ExprPlan> {
     /// Polars only keeps non-empty partitions in group-by,
     /// so this is used to fill missing values after joining with an explicit key set.
     pub(crate) fn fill_with(self, value: Expr) -> Self {
-        // Without repeat, the expression will be scalar-valued,
+        // Without this repeat, the expression would be scalar-valued,
         // and broadcast later to the required length.
-        // The drawback is that randomized plugins, like noise and report_noisy_max,
-        // will then only be applied to one row,
-        // and the one noisy row will then be broadcast to the entire column.
+        // This would cause randomized plugins, like noise and report_noisy_max,
+        // to only be applied to one row,
+        // and the one noisy row would then be broadcast to the entire column.
         let fill = repeat(value.clone(), len());
 
         Self::new_fallible(move |arg: &TI| {


Shoeboxam:
* **#2345** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2345?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/opendp/opendp/2345?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rust/src/measurements/make_private_expr/expr_report_noisy_max/mod.rs
```diff
@@ -295,6 +296,15 @@ fn report_noisy_max_gumbel_udf(
                         PolarsError::InvalidOperation("input dtype does not match".into())
                     })?;
 
+                // when all scores are same, return a random index
+                // this is a workaround for slow performance of the gumbel mechanism
+                // when all scores are the same
```
mccalluc, line 301:
```suggestion
                // When all scores are same, return a random index.
                // This is a workaround for slow performance of the gumbel mechanism
                // when all scores are the same.
```
Just a little more readable.

rust/src/measurements/make_private_lazyframe/group_by/test.rs
```diff
@@ -97,3 +100,62 @@ fn test_stable_keys_zCDP() -> Fallible<()> {
...
+        .median(Series::new("".into(), candidates.clone()), None)
+        .alias("med");
+    let meas = make_private_lazyframe(
+        lf_domain,
+        SymmetricDistance,
+        MaxDivergence,
+        lf.clone()
+            .group_by(&[col("A")])
+            .agg(&[sum_expr, median_expr])
+            .join(keys, [col("A")], [col("A")], JoinType::Right.into()),
```
mccalluc, line 136: Maybe a comment to explain this part?

rust/src/measurements/make_private_lazyframe/group_by/test.rs
```diff
@@ -97,3 +100,62 @@ fn test_stable_keys_zCDP() -> Fallible<()> {
...
+    kolmogorov_smirnov(gauss_samples, normal_cdf)?;
+
+    // check for uniformity of samples (all scores are matching)
+    let unif_samples: Vec<_> = release.column("med")?.f64()?.iter().flatten().collect();
+    let mut counts = [0.0; N_CANDIDATES];
+    unif_samples.iter().for_each(|&s| counts[s as usize] += 1.0);
+    chisquare(
+        counts,
+        [N_SAMPLES as f64 / (N_CANDIDATES as f64); N_CANDIDATES],
+    )?;
```
mccalluc, line 158: Wouldn't mind names like `assert_kolmogorov_smirnow` and `assert_chi_square`, but not essential.

rust/src/polars/mod.rs
```diff
@@ -286,14 +291,19 @@ impl Function<DslPlan, ExprPlan> {
...
     pub(crate) fn fill_with(self, value: Expr) -> Self {
         Self::new_fallible(move |arg: &DslPlan| {
             let mut plan = self.eval(arg)?;
-            plan.fill = Some(value.clone());
+            // Without repeat, the expression will be scalar-valued,
+            // and broadcast later to the required length.
+            // The drawback is that randomized plugins, like noise and report_noisy_max,
+            // will then only be applied to one row,
+            // and the one noisy row will then be broadcast to the entire column.
+            plan.fill = Some(repeat(value.clone(), len()));
```
mccalluc, line 306: If it were just the code, I wouldn't worry about refactoring it, but with the comment block, maybe find a way to consolidate?

mccalluc (dismissed):
A couple comments, but nothing essential. Approved.

rust/src/measurements/make_private_expr/expr_report_noisy_max/mod.rs
```diff
@@ -295,6 +296,15 @@ fn report_noisy_max_gumbel_udf(
...
+                let min = arr.values_iter().min_by(|a, b| a.partial_cmp(b).unwrap());
+                let max = arr.values_iter().max_by(|a, b| a.partial_cmp(b).unwrap());
+                if min == max {
+                    return sample_uniform_uint_below(arr.len() as u32);
+                }
```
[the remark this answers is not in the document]
Shoeboxam: I found another way to get at the data that allows me to use windows, which is typically how I write this.

rust/src/polars/mod.rs
```diff
@@ -284,16 +271,25 @@ impl Function<DslPlan, ExprPlan> {
...
     ///
     /// Polars only keeps non-empty partitions in group-by,
     /// so this is used to fill missing values after joining with an explicit key set.
     pub(crate) fn fill_with(self, value: Expr) -> Self {
-        Self::new_fallible(move |arg: &DslPlan| {
+        // Without repeat, the expression will be scalar-valued,
+        // and broadcast later to the required length.
+        // The drawback is that randomized plugins, like noise and report_noisy_max,
+        // will then only be applied to one row,
+        // and the one noisy row will then be broadcast to the entire column.
```
mccalluc, line 287: Perhaps more subjunctive to clarify that this is no longer a problem?
```suggestion
        // Without this repeat, the expression would be scalar-valued,
        // and broadcast later to the required length.
        // This would cause randomized plugins, like noise and report_noisy_max,
        // to only be applied to one row,
        // and the one noisy row would then be broadcast to the entire column.
```

mccalluc (commented):
Looks good. Maybe rephrase comment? 

I wonder if the test could be simpler. Instead of the kolmogorov_smirnov check, just make sure a block of values aren't equal?

Or going in a different direction, I wonder if it might be good to have a test in python? Since the bug report was in python, it would be good to confirm that the fix works in python too. (Not that I have any reason to doubt that this works!) 

... On the other hand, everything you do in Rust has implications on the Python side, and we don't want tests in duplicate.

So, whatever you think best.

PR #2366, opened 2025-04-16
feat: show errors when binary search type inference fails

### LINKED ISSUE #2365: Unable to infer type errors from binary search should contain error

Shoeboxam:
Close #2365

This also cleans up the existing code.

### DIFF
--- a/python/src/opendp/mod.py
+++ b/python/src/opendp/mod.py
@@ -1293,24 +1293,8 @@ def exponential_bounds_search(
     """
 
     # try to infer T
-    if T is None:
-        def _check_type(v):
-            try:
-                predicate(v)
-            except TypeError:
-                return False
-            except OpenDPException as e:
-                if "No match for concrete type" in (e.message or ""):
-                    return False # pragma: no cover
-            return True
-        
-        if _check_type(0.):
-            T = float
-        elif _check_type(0):
-            T = int
-        else:
-            raise TypeError("unable to infer type `T`; pass the type `T` or bounds")  # pragma: no cover
-
+    T = T or _infer_type(predicate)
+                
     # core search functionality
     def signed_band_search(center, at_center, sign):
         """Identify which band (of eight) the decision boundary lies in.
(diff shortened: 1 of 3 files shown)


Shoeboxam:
* **#2366** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2366?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/opendp/opendp/2366?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

python/src/opendp/mod.py
```diff
@@ -1293,24 +1293,8 @@ def exponential_bounds_search(
...
-            return True
-        
-        if _check_type(0.):
-            T = float
-        elif _check_type(0):
-            T = int
-        else:
-            raise TypeError("unable to infer type `T`; pass the type `T` or bounds")  # pragma: no cover
-
+    T = T or _infer_type(predicate)
```
mccalluc, line 1296: Perhaps
```suggestion
    T = T if T is not None else _infer_type(predicate)
```
depending on the behavior you want for `0`, `False`, or `""`? But keep it simple if it doesn't matter.

python/src/opendp/mod.py
```diff
@@ -1377,6 +1361,37 @@ def _exception_predicate(v):
     return signed_band_search(center, at_center, sign)
 
 
+def _infer_type(predicate: Callable[[float], bool]) -> Union[Type[float], Type[int]]:
+    def _is_type_error(e):
+        is_match_error = isinstance(e, OpenDPException) and "No match for concrete type" in (e.message or "")
```
mccalluc, line 1366: This worries me: Someone making a tweak to the error message where it is thrown isn't going to expect that the exact value matters.

python/src/opendp/mod.py
```diff
@@ -1377,6 +1361,37 @@ def _exception_predicate(v):
...
+        return int
+    except Exception as e:
+        if not _is_type_error(e):
+            # it didn't reject the type, but did fail with a different error
+            return int
+
+        # enrich the error message if in Python 3.11+.
+        if hasattr(e, "add_note"):
+            e.add_note("Unable to infer type `T`; pass the type `T` or bounds. This exception is raised when the predicate is evaluated at 0")
+        raise
```
mccalluc, line 1392: Would [exception chaining](https://docs.python.org/3/tutorial/errors.html#exception-chaining) be better here? 

We have no idea what exception was thrown by `predicate`, but whatever it was doesn't matter at this point: We want to tell the user there's a `TypeError` or something similar.

mccalluc (commented):
Some comments on the details, but the overall shape seems good.

python/src/opendp/mod.py
mccalluc: If necessary, a new error class, so you can catch it without looking at the message?

python/src/opendp/mod.py
Shoeboxam: If `T` is falsey, then it needs to be inferred.

python/src/opendp/mod.py
Shoeboxam: Yes, plumbing error classes would be a good thing. This is working and non-critical though.

python/src/opendp/mod.py
Shoeboxam: The exception thrown by the predicate is usually much more interesting! I typically don't care about the trace leading to the binary search.

mccalluc (approved):
Thanks!

PR #2368, opened 2025-04-16
fix: exception in PCA constructor

### LINKED ISSUE #2363: Issues Running Example PCA code provided in the documentation
**Describe the bug**
I am trying to get Differentially Private PCA working from the example found in the documents at this link [here](https://docs.opendp.org/en/stable/getting-started/statistical-modeling/pca.html). I ran it both locally and on Google Colab with various package and python versions, but regardless it will not run. Instead, I get the error below. 

  File ".../test.py", line 35, in <module>
    transformed_data = model.transform(example_dataset)
  File ".../lib/python3.10/site-packages/sklearn/utils/_set_output.py", line 319, in wrapped
    data_to_wrap = f(self, X, *args, **kwargs)
  File ".../lib/python3.10/site-packages/sklearn/decomposition/_base.py", line 136, in transform
    check_is_fitted(self)
  File ".../lib/python3.10/site-packages/sklearn/utils/validation.py", line 1756, in check_is_fitted
    if not _is_fitted(estimator, attributes, all_or_any):
  File ".../venv/lib/python3.10/site-packages/sklearn/utils/validation.py", line 1668, in _is_fitted
    v for v in vars(estimator) if v.endswith("_") and not v.startswith("__")
TypeError: descriptor '__dict__' for 'PCA' objects doesn't apply to a 'PCA' object

**To Reproduce**
Code snippet to reproduce the behavior.

import opendp.prelude as dp
import numpy as np

dp.enable_features("contrib", "floating-point", "honest-but-curious")

def sample_microdata(*, num_columns=None, num_rows=None, cov=None):
    cov = cov or np.random.rand(num_columns, num_columns)  # Generate random covariance matrix
    cov = cov @ cov.T  # Make it symmetric positive-definite
    microdata = np.random.multivariate_normal(
        np.zeros(cov.shape[0]), cov, size=num_rows or 100_000
    )
    microdata -= microdata.mean(axis=0)
    return microdata

num_columns = 4
num_rows = 10_000
example_dataset = sample_microdata(num_columns=num_columns, num_rows=num_rows)

# Create a differentially private PCA model
model = dp.sklearn.decomposition.PCA(

    epsilon=1.,

    row_norm=1.,

    n_samples=num_rows,

    n_features=4,

)
# Fit the model to the data
model.fit(example_dataset)

# Transform the data
transformed_data = model.transform(example_dataset)

print(transformed_data)

Shoeboxam:
Closes #2363

Sklearn is pinned in CI, but not in the package. A newer version of sklearn introduced a regression for us that prevents us from using the type trick to inherit from a class in an optional dependency.

### DIFF
--- a/python/src/opendp/extras/sklearn/decomposition/__init__.py
+++ b/python/src/opendp/extras/sklearn/decomposition/__init__.py
@@ -180,16 +180,16 @@ class PCA:
         n_components: int | float | str | None = None,
         n_changes: int = 1,
         whiten: bool = False,
-    ) -> None:
+    ) -> None:  # pragma: no cover
         # Error if constructor called without dependency:
-        import_optional_dependency('sklearn.decomposition')  # pragma: no cover
+        import_optional_dependency('sklearn.decomposition')
         # used for mypy typing
-        self.n_samples_ = None  # pragma: no cover
-        self.components_ = None  # pragma: no cover
-        self.n_components_ = None  # pragma: no cover
-        self.explained_variance_ = None  # pragma: no cover
-        self.explained_variance_ratio_ = None  # pragma: no cover
-        self.singular_values_ = None  # pragma: no cover
+        self.n_samples_ = None
+        self.components_ = None
+        self.n_components_ = None
+        self.explained_variance_ = None
+        self.explained_variance_ratio_ = None
+        self.singular_values_ = None
 
     @property
     def n_features(self):
@@ -230,9 +230,7 @@ class PCA:
 
 _decomposition = import_optional_dependency('sklearn.decomposition', False)
 if _decomposition is not None:
-    PCAStub = PCA
-    
-    class PCA(_decomposition.PCA):  # type: ignore
+    class PCA(_decomposition.PCA):  # type: ignore  # noqa: F811
         def __init__(
             self,
             *,
@@ -244,8 +242,6 @@ if _decomposition is not None:
             n_changes: int = 1,
             whiten: bool = False,
         ) -> None:
-            # Mypy has a lot of complaints because of the class redefinition,
-            # so there's a lot of "type: ignore" below.
             super().__init__(
                 n_components or n_features,
                 whiten=whiten,
--- a/python/src/opendp/extras/sklearn/decomposition/__init__.py
+++ b/python/src/opendp/extras/sklearn/decomposition/__init__.py
@@ -157,7 +157,7 @@ then_private_pca = to_then(make_private_pca)
 register(make_private_pca)
 
 
-class PCA():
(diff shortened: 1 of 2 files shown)


Shoeboxam:
* **#2368** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2368?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/opendp/opendp/2368?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

python/src/opendp/extras/sklearn/decomposition/__init__.py
```diff
@@ -180,139 +180,191 @@ def __init__(
         whiten: bool = False,
     ) -> None:
```
mccalluc, line 181:
```suggestion
    ) -> None:  # pragma: no cover
```
Would this work, instead of putting it on every line inside?

python/src/opendp/extras/sklearn/decomposition/__init__.py
```diff
@@ -180,139 +180,191 @@ def __init__(
...
+            epsilon: float,
+            row_norm: float,
+            n_samples: int,
+            n_features: int,
+            n_components: int | float | str | None = None,
+            n_changes: int = 1,
+            whiten: bool = False,
+        ) -> None:
+            # Mypy has a lot of complaints because of the class redefinition,
+            # so there's a lot of "type: ignore" below.
```
mccalluc, line 246: Is this still accurate?

python/src/opendp/extras/sklearn/decomposition/__init__.py
```diff
@@ -180,139 +180,191 @@ def __init__(
...
-    # overrides an sklearn method
     def _validate_params(*args, **kwargs):
-        pass
+        ...
 
 
 _decomposition = import_optional_dependency('sklearn.decomposition', False)
 if _decomposition is not None:
-    PCA = type('PCA', (_decomposition.PCA,), dict(PCA.__dict__))  # type: ignore[assignment,misc]
+    PCAStub = PCA
```
mccalluc, line 231: `PCAStub` doesn't seem to be referenced? Is this left over from an earlier attempt?

mccalluc (approved):
looking at the original issue, do you want to consider this done, or is there a follow-up for this comment?:
> The complexity here is arising due to needing to inherit from a class in an optional extra. In the long run, I'm considering splitting the code out into a (small) separate package.

I'm going to say "Approved" to unblock this, but it feels like there may be follow-up work?

PR #2369, opened 2025-04-17
feat: radius parameter on additive noise mechanisms

fleetingfound:
Addresses Issue #2375: truncated noise addition

This PR introduces bounded variants of our noise mechanisms by implementing an optional `radius` parameter that limits the magnitude of generated noise through rejection sampling.

## Changes

- Added an optional `radius` parameter to noise mechanisms.
- Implemented rejection sampling to enforce constraints on noise magnitude.
- Implemented `NoisePrivacyMap` with respective type parameters `Approximate<MaxDivergence>` and `Approximate<ZeroConcentratedDivergence>` for Laplacian and Gaussian noise mechanisms when `radius` is provided.

  - These implementations call out to the existing implementations of `NoisePrivacyMap` with type parameters `MaxDivergence` or `ZeroConcentratedDivergence` to compute `rho` and `epsilon` parameters, while computing `delta` parameters separately.
  - For computation of `delta` parameters, new tail lower bounds are introduced for Gaussian and Laplacian distributions, specifically `conservative_discrete_laplacian_tail_to_alpha_lower` and `conservative_discrete_gaussian_tail_to_alpha_lower` and their continuous analogues.
    The bound for the discrete Gaussian is based on ["The Discrete Gaussian for Differential Privacy" by Canonne, Kamath, and Steinke (2020)](https://proceedings.neurips.cc/paper/2020/file/b53b3a3d6ab90ce0268229151c9bde11-Paper.pdf).

### DIFF
--- a/python/test/test_meas.py
+++ b/python/test/test_meas.py
@@ -67,6 +67,11 @@ def test_vector_laplace():
     print("base laplace:", meas([80., 90., 100.]))
     assert meas.check(1., 1.3)
 
+def test_laplace_radius():
+    input_space = dp.atom_domain(T=float, nan=False), dp.absolute_distance(T=float)
+    meas = dp.m.make_laplace(*input_space, 10.5, radius=10.0)
+    print("base laplace:", meas(100.))
+    print("epsilon, delta", meas.map(1.))
 
 def test_gaussian_smoothed_max_divergence():
     input_space = dp.atom_domain(T=float, nan=False), dp.absolute_distance(T=float)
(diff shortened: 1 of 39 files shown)


Shoeboxam:
One thing that strikes me now is that a rejection-sampling approach from a pre-defined acceptance region would satisfy the same privacy guarantee without introducing delta. That's another variation of the mechanism, though, with some practical drawbacks in that you can't release estimates that are guaranteed not to underestimate a true private quantity.

@fleetingfound: Any chance you still have the time to polish off the proofs for this, or are you busy nowadays? Thanks!

fleetingfound:
Yes, definitely. I should be able to get a proof up in the next few weeks.

It probably makes sense to have both options:

1. Pre-determined acceptance region independent of the data, where you can have pure epsilon-DP.
2. Acceptance region depends on the value of the query on the data, but you fall back on (epsilon,delta)-DP.

Shoeboxam:
My earlier statement about rejection satisfying the same privacy guarantee when conditioning on a predefined output space is too broad. I made that statement based on [this paper](https://arxiv.org/pdf/2309.12647), but the actual theorems in that paper seem to be much narrower than what is claimed in the abstract. Typically the additive noise mechanisms don't require inputs be clipped to a range of width equal to the sensitivity, and at least in our setting that totally defeats the purpose. There's a couple papers from Naoise about bounded additive noise mechanisms that are directly relevant for that setting, and where it causes an increase in the privacy parameters.

PR #2372, opened 2025-04-21
Version bump and changelog for 0.13

mccalluc:
Some hiccups on the process to this point:
- `python channel_tool.py changelog` only added a new-line to the changelog. Maybe the pre-existing state wasn't quite what was expected?
- `python changelog.py` added the updates twice. Edited by hand.
- "Update `VERSION` file on the `main` branch to match the version you want to release" -- I think this should be `version-changelog`, not `main`?
- `python channel_tool.py bump_version --position minor` fails: line 82: `with open(path, f"r{b}") as f`: "No such file or directory: 'rust/Cargo.toml'" :point_right: `cd ..` first

(I can make an issue to tidy these up, unless it's in your court.)

Looking at the next step:
> We use a release train with nightly, beta and stable channels. Repeatedly run the Release workflow through each channel (see preceding section).

What is the "see preceding section" about? Can that be dropped, and instead we just link to https://github.com/opendp/opendp/actions/workflows/release.yml?

I hope this wouldn't happen, but with the release chain, is it possible that changes could accidentally be introduced to `beta` or `stable`, and as long as no changes in `main` hit the same file, the accidental changes could persist? Or is there something that would prevent this?


mccalluc:
Error running notebooks:
```
zmq.error.ZMQError: Address already in use (addr='tcp://127.0.0.1:44163')
------------------------------ Captured log call -------------------------------
ERROR    traitlets:client.py:568 Error occurred while starting new kernel client for kernel 216e2d3a-6b99-491c-88a5-f28a23966f35: Kernel died before replying to kernel_info
```
Retry.

Shoeboxam (approved):
Thanks Chuck! If you don't mind making tweaks to the release process, that would be ideal.

Not totally sure what you mean by changes in main hitting the same file. The nightly release checks out from main, the beta release checks out from nightly, and the stable release checks out from beta.

PR #2048, opened 2024-09-16
add `make_odometer_to_filter`

Shoeboxam:
Close #2047

PR #1854, opened 2024-07-27
feat: thread-safe queryables

Shoeboxam:
Closes #1849

The current approach for queryable wrapping is to store a wrapper in a global variable. When multiple wrappers need to be applied, as a query descends through wrapper queryables towards the core queryable, the global wrapper is temporarily swapped with the composition of the prior wrapper and new wrapper. Any new queryable spawned during evaluation of the core queryable is passed through the global wrapper. As the call stack unwinds, and the answer is returned back out through wrapper queryables, the prior wrapper (or none) is swapped back in.

Two drawbacks to this:

**First drawback**: Since the global is thread-local, wrappers won't be applied to queryables spawned in new threads started in the function, or may get wrongly applied if the thread is part of a pool spawned outside the function. 

**Second drawback**: Odometer nesting wants queryables to be thread-safe. When you query an odometer with an odometer, it creates a new privacy map that points to the child odometer. Since privacy maps are thread-safe, queryables must be thread-safe to build this map.

This PR moves wrappers from the thread-local to the call stack.

Now instead of queryables getting their wrapper from the global state automatically, it is up to the implementor to plumb the wrapper through their measurement/odometer function or queryable transition function.

This PR also makes AnyObject thread-safe. Say you are invoking an interactive composition measurement from Python. The function takes in the data as an AnyObject and returns a queryable holding this AnyObject data. Since queryables must now be thread-safe, the compiler now requires that things you store inside it are also thread-safe. Making AnyObject thread-safe requires a fair number of added Send + Sync bounds in FFI code.

PR #745, opened 2023-05-22
odometer chaining

Shoeboxam:
Closes #744

PR #743, opened 2023-05-22
add `make_concurrent_odometer`

Shoeboxam:
Closes #742

TODO: update bounds

PR #2384, opened 2025-05-09
feat: Context API odometer/filter integration

### LINKED ISSUE #2381: Integrate odometers/filters into Context API

Shoeboxam:
Close #2381


Shoeboxam:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/opendp/opendp/2384?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2384** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2384?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/opendp/opendp/2384?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#678** <a href="https://app.graphite.dev/github/pr/opendp/opendp/678?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2 other dependent PRs ([#743](https://github.com/opendp/opendp/pull/743) <a href="https://app.graphite.dev/github/pr/opendp/opendp/743?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#745](https://github.com/opendp/opendp/pull/745) <a href="https://app.graphite.dev/github/pr/opendp/opendp/745?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#2048** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2048?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#676** <a href="https://app.graphite.dev/github/pr/opendp/opendp/676?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2377** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2377?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1854](https://github.com/opendp/opendp/pull/1854) <a href="https://app.graphite.dev/github/pr/opendp/opendp/1854?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

python/src/opendp/context.py
```diff
@@ -413,38 +416,47 @@ class Context(object):
...
         self,
         accountant: Measurement,
         queryable: Queryable,
         d_in: Union[float, Sequence[Bound]],
-        d_mids: Optional[Sequence[float]] = None,
-        d_out: Optional[float] = None,
-        space_override: Optional[
-            tuple[Domain, Metric]
-        ] = None,  # TODO: Document or add leading underscore and explain that is is for internal use only.
+        d_mids: Optional[Sequence[Union[float, tuple[float, float]]]] = None,
```
mccalluc, line 431: Does an empty list behavior differently than `None`? That could be confusing.

python/src/opendp/context.py
```diff
@@ -413,38 +416,47 @@ class Context(object):
...
+        d_out: Optional[Union[float, tuple[float, float]]] = None,
+        _space_override: Optional[tuple[Domain, Metric]] = None,
```
mccalluc, line 433: Still shows up in API docs: Maybe people should infer that it is internal from the underscore, but a docstring might still be good, for developers if not for end users.

python/src/opendp/context.py
```diff
@@ -413,38 +416,47 @@ class Context(object):
...
+        self.d_mids_consumed: list[Union[float, tuple[float, float]]] = []
+        self._space_override = _space_override
 
     def __repr__(self) -> str:
-        return f"""Context(
+        if self.d_mids is not None:
+            return f"""Context(
     accountant = {indent(repr(self.accountant))},
     d_in       = {self.d_in},
     d_mids     = {self.d_mids})"""
```
mccalluc, line 449: Better to keep the repr simple: Showing a `None` for one of these fields seems fine?
```suggestion
            return f"""Context(
    accountant = {indent(repr(self.accountant))},
    d_in       = {self.d_in},
    d_mids     = {self.d_mids},
    d_out      = {self.d_out})"""
```

python/src/opendp/context.py
```diff
@@ -529,13 +541,17 @@ def compositor(
...
             d_mids=d_mids,
+            d_out=d_out,
         )
 
     def __call__(self, query: Measurement):
         """Executes the given query on the context."""
         answer = self.queryable(query)
         if self.d_mids is not None:
+            self.d_mids_consumed.append(self.d_mids[0])
             self.d_mids = self.d_mids[1:]
```
mccalluc, line 552:
```suggestion
            self.d_mids_consumed.append(self.d_mids.pop(0))
```

python/src/opendp/context.py
```diff
@@ -584,9 +601,32 @@ def query(self, **kwargs) -> Union["Query", LazyFrameQuery]:
...
+    
+    def remaining_privacy_loss(self):
+        """Returns the remaining privacy loss."""
+        if isinstance(self.queryable, OdometerQueryable):
+            if self.d_out is None:
+                raise ValueError("The privacy loss is unbounded.")
+            
+            def _sub(a, b):
+                if isinstance(a, tuple):
+                    return a[0] - b[0], a[1] - b[1]
```
mccalluc, line 620: Confirm that len is 2? or `[x - y for x, y in zip(x, y)]`?

python/src/opendp/context.py
```diff
@@ -742,39 +785,74 @@ def __dir__(self):
...
 
-    def resolve(self, allow_transformations: bool = False):
-        """Resolve the query into a measurement.
-
-        :param allow_transformations: If true, allow the response to be a transformation instead of a measurement.
+    def resolve(
+        self, 
+        allow_transformations: bool = False, 
+        bounds: Optional[tuple[float, float]] = None, 
+        T: Optional[Type[float]] = None
```
mccalluc, line 792: I know floats are a supertype of int for typing, but for the `Type` itself, should it be a `Union`? and does it need to be typed `Optional`? Maybe `float` as the default value?

python/src/opendp/context.py
```diff
@@ -742,39 +785,74 @@ def __dir__(self):
...
         if not allow_transformations and isinstance(chain, Transformation):
-            raise ValueError("Query is not yet a measurement")
+            raise ValueError("Query is not yet a measurement or odometer.")
         return _cast_measure(chain, self._output_measure, self._d_out)
 
-    def release(self) -> Any:
-        """Release the query. The query must be part of a context."""
-        # TODO: consider adding an optional `data` parameter for when _context is None
-        answer = self._context(self.resolve())  # type: ignore[misc]
+    def release(self, data=None, bounds: Optional[tuple[float, float]] = None, T: Optional[Type[float]] = None):
```
mccalluc, line 811: Same uncertainty about `Type[float]`

python/src/opendp/context.py
```diff
@@ -812,13 +887,12 @@ def _compositor(chain: Union[tuple[Domain, Metric], Transformation], d_in):
...
             privacy_unit = input_metric, d_in
-            assert d_out is not None
             privacy_loss = output_measure or self._output_measure, d_out
 
-            accountant, d_mids = _adaptive_composition_by_weights(
+            accountant, d_mids, _ = _normalize_compositor(
                 input_domain,
                 privacy_unit,
-                privacy_loss,
+                privacy_loss, # type: ignore[arg-type]
```
mccalluc, line 895: Maybe explain why the type checking is confused, if we can't fix it?

python/src/opendp/context.py
```diff
@@ -866,19 +941,27 @@ def __call__(self, v):
...
+        :param d_in: Upper bound on the distance between adjacent input datasets.
+        :param d_out: Upper bound on the distance between adjacent output distributions.
+        :param output_measure: How to measure distances between output distributions.
+        :param bounds: The bounds for the parameter search.
+        :param T: The type of the parameter to search for.
         """
         # When the output measure corresponds to approx-DP, only optimize the epsilon parameter.
         # The delta parameter should be fixed in _cast_measure, and if not, then the search will be impossible here anyways.
-        if output_measure == fixed_smoothed_max_divergence():
+        if output_measure is not None and isinstance(output_measure.type, RuntimeType) and output_measure.type.origin == "Approximate":
```
mccalluc, line 964: I worry about logic that depends on literal strings: typos can be easy to overlook.

mccalluc (commented):
Several comments, but none of them are necessarily blockers. As the typing gets more complicated, might use a type variable, but not sure if it would actually be any clearer.

Would it be worth having a named tuple or something like that instead of just a pair of floats, just for the semantics?

python/src/opendp/context.py
```diff
@@ -299,7 +302,7 @@ def loss_of(
     """
 
     def _range_warning(name, value, info_level, warn_level):
-        if value > warn_level:
+        if value > warn_level and value != float("inf"):
```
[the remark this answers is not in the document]
Shoeboxam: I guess I could still see throwing the warning, to further incentivize filters.

python/src/opendp/context.py
Shoeboxam: An empty list is an invalid compositor, whereas a None d_mids with some d_out will be a filter.

python/src/opendp/context.py
Shoeboxam: Yeah, optional is necessary. The types are distinct though.

python/src/opendp/_lib.py
```diff
@@ -79,7 +79,7 @@ def import_optional_dependency(name, raise_error=True):
         return None
 
 
-_np_csprng = None
+_np_csprng: Any = None
```
mccalluc, line 85: Not sure what the explicit `Any` gets us?

python/src/opendp/context.py
```diff
@@ -419,52 +424,54 @@ class Context(object):
...
+        query_space: Optional[tuple[Domain, Metric]] = None,
     ):
         self.accountant = accountant
         self.queryable = queryable
         self.d_in = d_in
         self.d_mids = d_mids
         self.d_out = d_out
-        self.space_override = space_override
+
+        self.d_mids_consumed: Union[list[float], list[tuple[float, float]]] = []
```
mccalluc, line 453:
```suggestion
        self._d_mids_consumed: Union[list[float], list[tuple[float, float]]] = []
```

To encourage users to go through `current_privacy_loss`?

python/src/opendp/context.py
```diff
@@ -537,13 +544,16 @@ def compositor(
...
         )
 
     def __call__(self, query: Measurement):
         """Executes the given query on the context."""
         answer = self.queryable(query)
-        if self.d_mids is not None:
-            self.d_mids = self.d_mids[1:]
+
+        self.d_mids_consumed.append(
+            query.map(self.d_in) if self.d_mids is None else self.d_mids.pop(0)  # type: ignore[arg-type]
```
mccalluc, line 556: If we don't need to worry about `d_mids` being an empty list, explain?

python/src/opendp/context.py
```diff
@@ -565,16 +575,21 @@ def query(self, **kwargs) -> Union["Query", LazyFrameQuery]:
...
-        chain = self.space_override or self.accountant.input_space
+        elif kwargs:
+            measure, d_query = loss_of(**kwargs)
+
+            if measure != self.accountant.output_measure:
+                measure_type = self.accountant.output_measure.type
+                msg = f"Expected output measure {self.accountant.output_measure} but got {measure}."
+                if (
+                    isinstance(measure_type, RuntimeType)
+                    and measure_type.origin == "Approximate"
```
mccalluc, line 587: Worry a bit about about hard-coded strings. This is ultimately being pulled out of a type name?

python/src/opendp/context.py
```diff
@@ -593,8 +608,37 @@ def query(self, **kwargs) -> Union["Query", LazyFrameQuery]:
 
         return query
 
+    def current_privacy_loss(self):
+        """When the query losses are static, returns the list of consumed privacy losses,
+        otherwise returns the current privacy loss of the queryable."""
+        if isinstance(self.queryable, OdometerQueryable):
+            return self.queryable.privacy_loss(self.d_in)
+        return self.d_mids_consumed
```
mccalluc: Should we worry that returning a list lets the user mutate state that we consider to be internal? The paranoid alternative is to take a copy with slice before returning.

python/src/opendp/context.py
```diff
@@ -593,8 +608,37 @@ def query(self, **kwargs) -> Union["Query", LazyFrameQuery]:
...
+
+            def _sub(a, b):
+                if (
+                    isinstance(a, tuple)
+                    and isinstance(b, tuple)
+                    and len(a) == len(b) == 2
+                ):
+                    return a[0] - b[0], a[1] - b[1]
+                else:
+                    return a - b
```
mccalluc, line 633: I think one helper function like this is fine, but if there ends up being more conditional logic, could consider making a class for the union type, with OO for the different cases.

python/src/opendp/context.py
```diff
@@ -973,9 +1081,18 @@ def _adaptive_composition_by_weights(
...
+        odometer = make_fully_adaptive_composition(
+            input_domain=domain,
+            input_metric=input_metric,
+            output_measure=output_measure,
+        )
+
+        # While delta = 1.0 is equivalent to no bound on the privacy loss,
+        #     float("inf") carries more semantic meaning.
+        if d_out in {float("inf"), (float("inf"), float("inf"))}:
+            return odometer, None, None
```
mccalluc: But would you want this same behavior if `1.0` were used instead of `float("inf")`? May be misinterpreting, but it sounds like you're establishing a convention about how particular values should come in, and that could be fragile.

python/src/opendp/context.py
```diff
@@ -1100,7 +1217,7 @@ def _caster(measurement):
         rho = _translate_measure_distance(
             (epsilon, delta_zCDP), from_measure, zero_concentrated_divergence()
         )
-        return rho, delta_inf
+        return rho, delta_inf  # type: ignore[return-value]
```
mccalluc, line 1221: This function didn't have a return type before, right? Is there a common reason you need to ignore types, or is each case special? If we're going to ignore it, is the benefit of the return type mostly for documentation?

mccalluc (dismissed):
Several comments, but none of them are blockers.

I worry a bit about having lots of Union types, but I'm not sure what the best alternative is. For the documentation, I think the status quo is more clear than defining our own types. There might be opportunities to make it more OO.

Similar worry-without-concrete-alternative for longer tuple returns, and more `type: ignore`s

python/src/opendp/_lib.py
Shoeboxam: It sates incremental MyPy checks that don't appear on a full run. I didn't feel it important to try to debug this different and sporadic behavior on incremental checks, instead opting to mark it as `Any`.

python/src/opendp/context.py
Shoeboxam: I figured it would be useful for odometers/filters. It works for compositors too just for consistency.

python/src/opendp/context.py
Shoeboxam: I've added a type for this, but it's in a future PR.

python/src/opendp/context.py
Shoeboxam: The recursive call can return different types than the original call. Benefit is for documentation as I'm not sure how to correctly annotate this.

Issue #2285, opened 2025-02-18
Clean up references to `make_split_dataframe` in docs


mccalluc:
With Polars in place `make_split_dataframe` is deprecated.

https://docs.opendp.org/en/stable/search.html?q=make_split_dataframe#

- [Accuracy: Pitfalls and Edge Cases](https://docs.opendp.org/en/stable/theory/accuracy-pitfalls.html)
- [Differencing](https://docs.opendp.org/en/stable/theory/attacks/differencing.html)
- [Example Data Analysis](https://docs.opendp.org/en/stable/getting-started/examples/pums-data-analysis.html) (Trying to do too much!)
- [Privatizing Histograms](https://docs.opendp.org/en/stable/getting-started/examples/histograms.html)

> Assume you have data that is a vector of ints. In production, we recommend doing pre-processing with the library, but to keep this example focused and minimal, we'll do it outside the library:
> ```
> # Small code snip to grab data
> ```

mccalluc:
At this point, besides references in the API docs (which are marked as deprecated), the only reference is in https://docs.opendp.org/en/stable/theory/accuracy-pitfalls.html

PR #2388, opened 2025-05-29
Check that all notebook cells are executed in order

### LINKED ISSUE #2370: Check that notebooks have been executed
All cells should have execution counts, and they should increase by one for each cell.

mccalluc:
- Fix #2370

There's almost certainly a tool out there that already does this, but didn't find it with a a quick search.

I think we're on the same page that this is a goal, but I don't think we have a plan for how to get there. Would you like to just get all the notebooks executed in this PR? Or is it something we don't even want to worry about until release, and we don't want to worry about regressions between releases? I would prefer something simple and consistent, but I haven't been editing the notebooks as often as you.

### DIFF
--- a/python/test/test_docs.py
+++ b/python/test/test_docs.py
@@ -1,3 +1,5 @@
+from pathlib import Path
+from json import loads
 import pytest
 from opendp import measurements, transformations
 
@@ -19,4 +21,13 @@ def test_thens_are_documented(module, function):
 
     assert function.__doc__ is not None, 'missing documentation'
     assert f':py:func:`{m_name}.{make_name}`' in function.__doc__, f'no link to {make_name}'
+
+
+@pytest.mark.parametrize(
+    "nb_path",
+    list((Path(__file__).parent.parent.parent / 'docs' / 'source').glob("**/*.ipynb")),
+    ids=lambda path: path.name
+)
+def test_notebooks_are_executed(nb_path):
+    loads(nb_path.read_text())
 
--- a/python/test/test_docs.py
+++ b/python/test/test_docs.py
@@ -29,5 +29,10 @@ def test_thens_are_documented(module, function):
     ids=lambda path: path.name
 )
 def test_notebooks_are_executed(nb_path):
-    loads(nb_path.read_text())
+    nb = loads(nb_path.read_text())
+    counts_sources = [(cell.get('execution_count'), ''.join(cell.get('source', ''))) for cell in nb['cells'] if cell['cell_type'] == 'code']
+    triples = [(index, count, source) for (index, (count, source)) in enumerate(counts_sources, start=1)]
+    indexes, counts, sources = zip(*triples)
+    bad_sources = [source for (index, count, source) in triples if index != count]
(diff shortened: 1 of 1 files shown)


mccalluc:
(draft until upstream fix to CI is merged.)

mccalluc:
@Shoeboxam - Keeping as draft because tests aren't going to pass in the current state, but flagging you for review to get feedback on the best direction for this... and the best direction might be to just set it aside for a while.

docs/source/api/user-guide/transformations/aggregation-sum.ipynb
```diff
@@ -366,16 +432,27 @@
...
+     "iopub.status.idle": "2025-06-04T18:19:03.092503Z",
+     "shell.execute_reply": "2025-06-04T18:19:03.092296Z"
+    }
+   },
    "outputs": [
     {
      "data": {
       "text/plain": [
-       "<opendp.mod.Transformation at 0x107fd06d0>"
+       "Transformation(\n",
```
mccalluc, line 448: improved repr!

docs/source/getting-started/assessing-utility/noise.ipynb
```diff
@@ -177,9 +198,24 @@
...
+     "iopub.status.idle": "2025-06-04T18:12:55.619679Z",
+     "shell.execute_reply": "2025-06-04T18:12:55.619431Z"
+    }
+   },
    "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "/var/folders/jm/40wztt291610f8pk5w0sw20r0000gq/T/ipykernel_49512/3463634038.py:10: MapWithoutReturnDtypeWarning: Calling `map_elements` without specifying `return_dtype` can lead to unpredictable results. Specify `return_dtype` to silence this warning.\n",
```
mccalluc, line 215: I could file an issue and follow up later?

docs/source/theory/accuracy-pitfalls.ipynb
```diff
@@ -127,20 +133,36 @@
...
     "pycharm": {
      "name": "#%%\n"
     }
    },
    "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "/var/folders/jm/40wztt291610f8pk5w0sw20r0000gq/T/ipykernel_31426/793901930.py:27: DeprecationWarning: Call to deprecated function (or staticmethod) make_split_dataframe. (Use Polars instead) -- Deprecated since version 0.12.0.\n",
```
mccalluc, line 154: Updating the docs is tech debt:
- #2285

mccalluc (commented):
I'd prefer not to have all the new metadata, but that doesn't seem to be an nbconvert option (although there is a [clear metadata preprocessor](https://nbconvert.readthedocs.io/en/latest/api/preprocessors.html?highlight=metadata#nbconvert.preprocessors.ClearMetadataPreprocessor)) and this isn't worth development just to clean up notebook files.

docs/source/getting-started/assessing-utility/noise.ipynb
Shoeboxam: This notebook is going away

PR #2391, opened 2025-05-30
do not generate __all__

### LINKED ISSUE #2390: Codegen simplification: drop underscore function from `__all__`?
~~It looks like all of our generated modules include an `__all__` which lists all members, public and underscore-private. Could we just drop this definition of `__all__`? Unless our code relies on getting private names via an `import *`, it should be safe, and if end-users were relying on exports of private functions, that is something we'd like to discourage.~~

Update: the `__all__` also prevents the `*` imports in `prelude.py` from picking up names that are imported into each of these. What might still be good would be to stop listing underscored, private functions in the generated `__all__`, but not a high priority.

Going to let this sit for at least a day, and then might unilaterally close it unless I can convince myself that it matters.

mccalluc:
- Fix #2390

Small simplification to code generation.

Even if there is no obvious merge conflict, it would be good to check that #2391 and #2392 do not conflict: Merge one to main, and then update the other and make sure tests still pass.

### DIFF
--- a/rust/opendp_derive/src/lib.rs
+++ b/rust/opendp_derive/src/lib.rs
@@ -40,7 +40,7 @@ mod full;
 ///
 /// # Features
 /// You can indicate a list of features that must be enabled by the user for the function to exist.
-/// ```compile_fail
+/// ```no_run
 /// #[bootstrap(
 ///     features("contrib", "floating-point")
 /// )]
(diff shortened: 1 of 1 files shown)


Shoeboxam (changes requested):
Is it intentional that you allow extraneous items from imports to get leaked into the module? An `from opendp.measurements import *` would then bring in the `Transformation` class, for example. Maybe I'm misunderstanding the intent.

PR #2398, opened 2025-06-03
Tidy up release docs + scripts

### LINKED ISSUE #2373: Tidy up release docs
Edits needed on release docs, and perhaps tweaks to scripts. Comes from 
- #2372

---

Some hiccups on the process to this point:
- `python channel_tool.py changelog` only added a new-line to the changelog. Maybe the pre-existing state wasn't quite what was expected?
- `python changelog.py` added the updates twice. Edited by hand.
- "Update `VERSION` file on the `main` branch to match the version you want to release" -- I think this should be `version-changelog`, not `main`?
- `python channel_tool.py bump_version --position minor` fails: line 82: `with open(path, f"r{b}") as f`: "No such file or directory: 'rust/Cargo.toml'" :point_right: `cd ..` first

(I can make an issue to tidy these up, unless it's in your court.)

Looking at the next step:
> We use a release train with nightly, beta and stable channels. Repeatedly run the Release workflow through each channel (see preceding section).

What is the "see preceding section" about? Can that be dropped, and instead we just link to https://github.com/opendp/opendp/actions/workflows/release.yml?

mccalluc:
- Fix #2373 (though there might be more iterations on release process)

There's a lot of complexity here, and although I got a release out last time, I'm worrying now that parts of it could have been misconfigured, since the scripts didn't run completely.

I'm also wondering how the release process works for you. My guess is that you understand what should be happening, and so you don't need to follow every step literally. ... But if you think you _are_ following the steps verbatim, then maybe there is a problem with my environment? Either way, we should make sure that both of us can run the release process without hitting errors.

The riskiest part of this PR is probably dropping `python channel_tool.py changelog` -- On the last release it didn't really make any changes, and I'm not sure what we expect it to do. There is a lot of conditional logic here.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -142,14 +142,12 @@ on github, or with the `gh` command line tool. Parameters:
     1. Update changelog
 
         ```shell
-        python channel_tool.py changelog
-
         git fetch --tags
         python changelog.py
         ```
         Additional manual edits will be useful to arrange individual commits thematically.
 
-    1. Update `VERSION` file on the `main` branch to match the version you want to release.
+    1. Update `VERSION` file to match the version you want to release.
 
         - If it's a patch release, no change to the version number should be needed: The patch should have been incremented after the previous release.
         - If it's a minor release, run `python channel_tool.py bump_version --position minor`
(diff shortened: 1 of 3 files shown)


mccalluc:
Closing this: This isn't what is blocking release right now, and I don't think we should introduce more changes until what we have is reliably passing CI.

PR #2403, opened 2025-06-04
feat: add `dp.sklearn.linear_model.LinearRegression`

### LINKED ISSUE #2395: Move linear regression from documention example to extras
DP Wizard is the primary motivation:
- https://github.com/opendp/dp-wizard/issues/463

mccalluc:
- fix #2395

We'll also want to update the documentation, probably keeping the current example because it is a good example of a plugin, but adding a pointer to this, if people just want to use it.

And I think we'll also want to provide access to this via polars? Maybe the expression syntax will look like that offered by https://github.com/azmyrajab/polars_ols?


python/src/opendp/extras/sklearn/linear_model/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+    The interface is parallel to that offered by sklearn's
+    `LinearRegression <https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LinearRegression.html>`_.
+    The ``fit`` method returns an sklearn ``LinearRegression`` object.
+    """
+
+    def fit(
+        self,
+        X,
+        y,
+        x_bounds: tuple[float, float], # TODO: Wrap this so each feature has its own bound tuple?
```
Shoeboxam: Yeah, each feature should have its own bounds tuple.

python/src/opendp/extras/sklearn/linear_model/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+        ...     scale=1,
+        ... )
+        >>> lin_reg.predict([[10]])
+        array([...])
+        """
+        meas = _make_private_theil_sen(
+            x_bounds=x_bounds, y_bounds=y_bounds, scale=scale, runs=runs
+        )
+        np = import_optional_dependency("numpy")
+        slope, intercept = meas(np.stack([[x[0] for x in X], y], axis=1))
```
Shoeboxam: You can just index like so: `X[:, 0]` to avoid the slow list comprehension.

python/src/opendp/extras/sklearn/linear_model/_make_private_theil_sen/__init__.py
```diff
@@ -0,0 +1,119 @@
...
+    fits a line that passes through each pair of points,
+    and then computes the y values of each line at `x_cuts`.
+    That is, it predicts the y values at `x_cuts`, pairwise.
+    """
+    np = import_optional_dependency("numpy")
+    # Get an even number of rows.
+    data = np.array(data, copy=True)[: len(data) // 2 * 2]
+
+    # Shuffle the data to get random matchings.
+    np.random.shuffle(data)
```
Shoeboxam, line 17: Any permutation of the data is valid, so the CSPRNG isn't necessary here. The shuffling is for utility. Maybe add a comment?

python/src/opendp/extras/sklearn/linear_model/_make_private_theil_sen/__init__.py
```diff
@@ -0,0 +1,119 @@
...
+def make_pairwise_predict(x_cuts, runs: int = 1):
+    """
+    The parameter `runs` controls how many times randomized pairwise predictions are computed.
+    The default is 1. Increasing `runs` can improve the robustness and accuracy of the results;
+    however, it can also increase computational cost and amount of noise needed later in the algorithm.
+    """
+    np = import_optional_dependency("numpy")
+    return dp.t.make_user_transformation(
+        # Outputs are Nx2 non-nan float numpy arrays.
+        input_domain=dp.numpy.array2_domain(num_columns=2, T=float),
```
Shoeboxam: Can T be an argument that defaults to `float`?

Shoeboxam: The comment is wrong, it doesn't need to be non-nan.

python/src/opendp/extras/sklearn/linear_model/_make_private_theil_sen/__init__.py
```diff
@@ -0,0 +1,119 @@
...
+            [pairwise_predict(x, x_cuts) for _ in range(runs)]
+        ),
+        # Each execution of pairwise predict contributes b_in records.
+        stability_map=lambda b_in: b_in * runs,
+    )
+
+
+def make_select_column(j):
+    return dp.t.make_user_transformation(
+        input_domain=dp.numpy.array2_domain(num_columns=2, T=float),
```
Shoeboxam: Similarly pass T in.

python/src/opendp/extras/sklearn/linear_model/_make_private_theil_sen/__init__.py
```diff
@@ -0,0 +1,119 @@
...
+        stability_map=lambda b_in: b_in,
+    )
+
+
+def make_private_percentile_medians(y_bounds, scale):
+    np = import_optional_dependency("numpy")
+    # this median mechanism favors candidates closest to the true median
+    m_median = dp.m.then_private_quantile(
+        # 100 evenly spaced points between y_bounds
+        candidates=np.linspace(*y_bounds, 100),
```
Shoeboxam: Can you make the 100 a default argument and thread it up?

python/src/opendp/extras/sklearn/linear_model/_make_private_theil_sen/__init__.py
```diff
@@ -0,0 +1,119 @@
...
+    and returns a (slope, intercept) tuple.
+
+    >>> import numpy as np
+    >>> meas = make_private_theil_sen((0, 100), (0, 100), scale=1.0)
+    >>> slope, intercept = meas(np.array([[x, x] for x in range(100)]))
+    """
+    np = import_optional_dependency("numpy")
+    # x_cuts are the 25th and 75th percentiles of x_bounds.
+    # We'll predict y's at these x_cuts.
+    x_cuts = x_bounds[0] + (x_bounds[1] - x_bounds[0]) * np.array([0.25, 0.75])
```
Shoeboxam: Can you make the 0.25 and 0.75 an argument and thread it up?

Shoeboxam (changes requested):
A few changes. Looking good!

python/src/opendp/extras/sklearn/linear_model/__init__.py
mccalluc: Looking at the next steps here, I think `make_private_theil_sen` will also need to be generalized?

Currently:
```
    x_cuts = x_bounds[0] + (x_bounds[1] - x_bounds[0]) * np.array([0.25, 0.75])

    # we want coefficients, not y values!
    # Luckily y values are related to coefficients via a linear system
    P_inv = np.linalg.inv(np.vstack([x_cuts, np.ones_like(x_cuts)]).T)
```
So `x_cuts` and `P_inv` would also need to be replaced with Iterables wrapping elements like the current values?

And I'm not sure how to adopt the chain which follows.

python/src/opendp/extras/sklearn/linear_model/__init__.py
mccalluc: done!

python/src/opendp/extras/sklearn/linear_model/_make_private_theil_sen/__init__.py
mccalluc: Sure. I'll make it here, and in the notebook where this is copied from.

python/src/opendp/extras/sklearn/linear_model/_make_private_theil_sen/__init__.py
mccalluc: done. (Just changed the type param here, not in notebook.)

python/src/opendp/extras/sklearn/linear_model/_make_private_theil_sen/__init__.py
mccalluc: done.

python/src/opendp/extras/sklearn/linear_model/_make_private_theil_sen/__init__.py
mccalluc: done.

mccalluc:
The docs build has problems on CI:
```
/home/runner/work/opendp/opendp/docs/source/api/user-guide/plugins/theil-sen-regression.ipynb:13: WARNING: File not found: 'api/python/opendp.extras.sklearn.linear_model.html'
```
It works locally for me, but I wonder if this is an ordering problem. This is the only place I can find where a notebook links to the API docs, and maybe they aren't in place at this phase of the docs build?

mccalluc:
Also getting sporadic notebook failures in CI:
```
_ /home/runner/work/opendp/opendp/docs/source/api/user-guide/polars/expressions/aggregation.ipynb _
[gw1] linux -- Python 3.9.23 /opt/hostedtoolcache/Python/3.9.23/x64/bin/python
NBMAKE INTERNAL ERROR
Kernel died
```
or
```
NoDataError                               Traceback (most recent call last)
Cell In[2], line 1
----> 1 context = dp.Context.compositor(
      2     data=pl.scan_csv(dp.examples.get_france_lfs_path(), ignore_errors=True),
      3     privacy_unit=dp.unit_of(contributions=36),
      4     privacy_loss=dp.loss_of(epsilon=1.0),
      5     split_evenly_over=1,
      6     margins=[
      7         dp.polars.Margin(max_length=150_000 * 36)
      8     ],
      9 )
     11 context.query().select(split_exprs).summarize()
```
I don't think they are related to this PR... maybe there's something that can be fixed in main, and then merged here?

python/src/opendp/extras/sklearn/linear_model/__init__.py
```diff
@@ -0,0 +1,118 @@
...
+    def fit(
+        self,
+        X,
+        y,
+        x_bounds: Iterable[tuple[float, float]],
+        y_bounds: tuple[float, float],
+        scale: float,
+        runs: int = 1,
+        candidates_count: int = 100,
+        fraction_bounds: tuple[float, float] = (0.25, 0.75),
```
Shoeboxam: I've thought more about the APIs and think it would be better to move these into the constructor.

This way arguments that would typically go to the constructor function are still a part of the constructor, but this time for the LinearRegression class. It would also be easier to integrate with the Context API, in that a user could ultimately do something like: `dp.query().sklearn(LinearRegression(x_bounds, y_bounds)).release()`.

Shoeboxam (changes requested):

mccalluc:
@Shoeboxam : Moved parameter up to constructor, and updated the docs. (Thanks, `test_ast`!) This parameter isn't demonstrated in the doctests, but the default value should make the usage clear enough?

python/src/opendp/extras/sklearn/linear_model/__init__.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def fit(
+        self,
+        X,
+        y,
+        x_bounds: Iterable[tuple[float, float]],
+        y_bounds: tuple[float, float],
+        scale: float,
+        runs: int = 1,
+        candidates_count: int = 100,
```
Shoeboxam: All of these arguments too. Basically everything except the data. This way there will be 
1. enough arguments to be able to build the measurement without the data, for use in Context API
2. the fit method will more closely match the sklearn fit method

python/src/opendp/extras/sklearn/linear_model/__init__.py
```diff
@@ -0,0 +1,125 @@
...
+        ...     X=[[1], [2], [3], [4], [5]],
+        ...     y=[1, 2, 3, 4, 5],
+        ...     x_bounds=[(0, 10)],
+        ...     y_bounds=(0, 10),
+        ...     scale=1,
+        ... )
+        >>> lin_reg.predict([[10]])
+        array([...])
+        """
+        assert_features("floating-point")
```
Shoeboxam: If we assert a feature, it should be done in the relevant constructor(s), not in a wrapper API. The constructors should be correct independently of how they are wrapped. 

That being said, this doesn't suffer from floating-point issues, because the transformation is microdata -> microdata, the maps are exact, and the actual mechanism is hardened. So the assertion can be dropped.

Shoeboxam (changes requested):
A few more changes.

python/src/opendp/extras/sklearn/linear_model/__init__.py
mccalluc: Got it. Sorry to misunderstand.

python/src/opendp/extras/sklearn/linear_model/__init__.py
mccalluc: Sorry: Misremembering what the interns were working on again.

python/src/opendp/extras/sklearn/linear_model/__init__.py
Shoeboxam: No worries at all!

PR #2404, opened 2025-06-04
docs: JS redirection for regression

### LINKED ISSUE #2396: With linear regression move, there should also be an update to the JS redirects
Looking at:
- #2360

What was `getting-started/statistical-modeling/regression` is now `api/user-guide/plugins/theil-sen-regression`: Any time we move pages, we need to update 404.html. I should have noted this during review. (The new page is in the nightly docs, but we haven't made a release, so the old one is still in stable.)

I'm also planning to take this code as the basis of a linear regression feature in extras:
- #2395

Maybe the new page is retained as a example of a plugin, even if the code is in the library? And maybe the old page can be brought back to document what will now be a library feature? Several moving pieces.

mccalluc:
- Fix #2396

I expect that another PR
- #2403

will re-establish the `getting-started/statistical-modeling/regression.html` path, but it would be good to point users to the new location until that is merged. (That PR will probably remove this redirect.)


Shoeboxam (approved):
Great, thanks.

PR #2405, opened 2025-06-04
docs: Explain constructor lingo

### LINKED ISSUE #2364: Explain that "constructor" in OpenDP isn't a python constructor
I don't think we should change our naming convention at this point, but I think it would be good to acknowledge that our use of the term "constructor" doesn't align with Python's. Perhaps just a little info box in the getting-started.

mccalluc:
- Fix #2364

I started looking at whether "constructor" could be replaced with "function" across the board, but I don't think that makes sense: We actually do mean something particular when we talk about constructors. For instance, I don't think pure-python code in `extras` is considered "constructors".

"Constructor" doesn't actually come up in `getting-started`, so this seemed like a reasonable place, if it belongs anywhere.

Or we could just close this, if this doesn't seem like a real problem, or the best fix to the problem.


docs/source/theory/dp-with-opendp.rst
```diff
@@ -33,6 +33,14 @@ us if you are interested in proof-writing. Thank you!
...
             >>> dp.enable_features("contrib")
             
+.. note::
+    In this documentation, we'll consistently refer to OpenDP Library functions
+    as "constructors", although they are not "constructors" as that term is used in Python:
+    class names which are called to create instances.
+
+    Instead, in OpenDP these "constructors" are regular functions which provide
+    a common interface across all supported languages,
+    and return wrappers around objects created in the Rust core.
```
Shoeboxam: I think it's ok to swap constructor and function in the places you have, but the definition of a constructor is not quite right: In OpenDP parlance, a "constructor" is a function that returns a transformation or measurement. A function like `laplace_scale_to_accuracy` would not be a constructor, but `make_gaussian` would be.

Shoeboxam (changes requested):
A minor correction. Thanks for the tweaks!

docs/source/theory/dp-with-opendp.rst
mccalluc: Thanks! Just pushed an update:

> Instead, in OpenDP these "constructors" are regular functions
> which return a transformation or measurement,
> and they provide a common interface across all supported languages.

Issue #1343, opened 2024-02-29
Rust check feature combinations


mccalluc:
- split off from #1338

I think the deliverables here are:
- [ ] identify the combos that should work
- [ ] document that in development-environment.html
- [ ] fix the cases that don't work but should.
- [ ] add a test to make sure they keep on working.

here's a script to start with:

```python
import subprocess
import itertools
from utils import run_command


def main():
    features = ['untrusted', 'contrib', 'honest-but-curious', 'floating-point', 'bindings', 'ffi', 'derive', 'use-openssl']
	
    feature_sets = []
    for size in range(0,len(features)+1):
        feature_sets += [','.join(subset) for subset in list(itertools.combinations(features, size))]
    fail_sets = []
    pass_sets = []
    for feature_set in feature_sets:
        try:
            run_command('check features', cmd=f'cd rust; cargo check --no-default-features --features {feature_set}')
            pass_sets.append(feature_set)
        except subprocess.CalledProcessError:
            fail_sets.append(feature_set)
    print(f'pass: {" ".join(pass_sets)}')
    print(f'fail: {" ".join(fail_sets)}')


if __name__ == "__main__":
    main()
```

and here's the output:

<details>

[...]

PR #2419, opened 2025-06-09
chore: update Polars to 0.50.0

### LINKED ISSUE #2418: Upgrade Polars to 0.48.1

Shoeboxam:
Close #2418

### DIFF
--- a/docs/source/contributing/development-environment.rst
+++ b/docs/source/contributing/development-environment.rst
@@ -56,15 +56,6 @@ Now run ``cargo build`` in the ``rust`` subdirectory of the repo:
 This will compile a debug build of the OpenDP shared library, placing it in the directory ``opendp/rust/target/debug``. 
 (The specific name of the library file will vary depending on your platform.)
 
-
-.. note::
-
-  On Mac, you may need to set ``RUSTFLAGS`` before build (`necessary for PyO3 <https://pyo3.rs/v0.25.1/building-and-distribution.html#macos>`_):
-
-  .. code-block:: bash
-
-    export RUSTFLAGS="-C link-arg=-undefined -C link-arg=dynamic_lookup"
-
 Substitute ``cargo build`` with ``cargo test`` to test, or ``cargo check`` to check syntax.
 
 .. note::
@@ -94,6 +85,8 @@ Setting a feature changes how the crate compiles.
 
       * - Name
         - Description
+      * - ``polars``
+        - Enable for functionality related to the Polars dataframe library.
       * - ``contrib``
         - Enable to include constructors that have not passed the vetting process.
       * - ``honest-but-curious``
@@ -105,13 +98,15 @@ Setting a feature changes how the crate compiles.
       * - ``floating-point``
         - Enable to include transformations and measurements with floating-point vulnerabilities.
       * - ``untrusted``
-        - Enables untrusted features ``contrib`` and ``floating-point``.
+        - Enables untrusted features ``contrib``, ``honest-but-curious`` and ``floating-point``.
       * - ``ffi``
         - Enable to include C foreign function interfaces.
       * - ``derive``
         - Enable to support code generation and links to proofs in documentation.
+      * - ``pyo3/extension-module``
+        - Enable creation of a python extension module for use by Python Polars. Must be disabled for Rust tests.
       * - ``bindings``
-        - Enable to generate Python and R source code. Also enables the ``ffi`` and ``derive`` features. 
+        - Enable to generate Python and R source code. Also enables the ``ffi``, ``derive`` and ``pyo3/extension-module`` features. 
       * - ``partials``
         - Enabled by default. When enabled, ``then_*`` functions are generated from ``make_*`` functions. Also enables the ``derive`` feature.
       * - ``use-openssl``
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -176,6 +176,7 @@ opendp_tooling = { path = "opendp_tooling", optional = true, version = "0.13.0-d
 syn = { workspace = true, optional = true }
 proc-macro2 = { workspace = true, optional = true }
 cbindgen = { version = "0.28.0", optional = true }
+pyo3-build-config = { version = "0.25", optional = true }
 
 [features]
 default = ["partials", "use-openssl"]
@@ -203,7 +204,7 @@ ffi = ["lazy_static", "cbindgen"]
 derive = ["opendp_tooling", "opendp_derive/full", "syn", "proc-macro2"]
 
 # generate language bindings
-bindings = ["ffi", "derive", "pyo3/extension-module"]
+bindings = ["ffi", "derive", "pyo3/extension-module", "pyo3-build-config"]
 
 # generate "then_xxx" functions in Rust for "make_xxx" functions whose first two arguments are the metric space
 partials = ["derive", "opendp_derive/partials"]
--- a/rust/src/transformations/make_stable_expr/expr_replace/mod.rs
+++ b/rust/src/transformations/make_stable_expr/expr_replace/mod.rs
@@ -134,26 +134,21 @@ pub(crate) fn literal_len(literal: &LiteralValue) -> Fallible<i128> {
     })
 }
 
+fn to_series<T: PolarsPhysicalType, N>(x: Box<[Option<N>]>) -> Series
+where
+    ChunkedArray<T>: NewChunkedArray<T, N>,
+{
+    ChunkedArray::<T>::from_iter_options("literal".into(), x.into_iter()).into_series()
+}
+
 pub(crate) fn materialize_lit_list(literal: &LiteralValue) -> Fallible<LiteralValue> {
     Ok(match literal {
         // desugar lists
         LiteralValue::Dyn(DynLiteralValue::List(list)) => {
-            LiteralValue::Series(SpecialEq::new(match list {
-                DynListLiteralValue::Int(list) => Int128Chunked::from_iter_options(
-                    PlSmallStr::from_static("literal"),
-                    list.into_iter().cloned(),
-                )
-                .into_series(),
-                DynListLiteralValue::Float(list) => Float64Chunked::from_iter_options(
-                    PlSmallStr::from_static("literal"),
-                    list.into_iter().cloned(),
-                )
-                .into_series(),
-                DynListLiteralValue::Str(list) => StringChunked::from_iter_options(
-                    PlSmallStr::from_static("literal"),
-                    list.into_iter().cloned(),
-                )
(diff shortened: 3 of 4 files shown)


mccalluc:
Now it fails really fast!
```
cargo build --all-features
error: failed to select a version for the requirement `polars-json = "^0.48.1"`
candidate versions found which didn't match: 0.47.1, 0.47.0, 0.46.0, ...
location searched: crates.io index
required by package `polars-io v0.48.1`
    ... which satisfies dependency `polars-io = "^0.48.1"` (locked to 0.48.1) of package `polars v0.48.1`
    ... which satisfies dependency `polars = "=0.48.1"` (locked to 0.48.1) of package `opendp v0.13.0-dev (/Users/chuckmccallum/github/opendp/opendp/rust)`
```

Shoeboxam:
@mccalluc: The issue you encountered was earlier in the compilation, the tricky issue I was running into was at link-time. Try again?

.github/workflows/smoke-test.yml
```diff
@@ -126,7 +126,7 @@ jobs:
         run: cargo test --verbose
 
       - name: Test (all-features)
-        run: cargo test --verbose --all-features
+        run: cargo test --verbose --features=ffi,derive,polars,untrusted
```
mccalluc: Testing all features seems safer, but locally I do see
```
dyld[43840]: symbol not found in flat namespace '_PyBaseObject_Type'
error: test failed, to rerun pass `-p opendp --lib`
```
which is probably the reason for the change?

Perhaps just add a comment here explaining why, unless you have a better idea?

docs/source/getting-started/tabular-data/grouping.ipynb
```diff
@@ -19,7 +19,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 16,
```
mccalluc: We'd like to make sure all the cells are executed in order.
- This will be required with #2388

Can you re-execute this notebook from the top so it doesn't need another pass when that test is added?

python/src/opendp/extras/polars/__init__.py
```diff
@@ -410,6 +410,7 @@ def _discrete_quantile_score(self, alpha: float, candidates: list[float]):
             function_name="discrete_quantile_score",
             args=[self.expr, alpha, Series(candidates)],
             returns_scalar=True,
+            changes_length=True,
```
mccalluc, line 413: Not obvious that an upgrade would require a change here. Double-check that it's not stray, but no particular concerns beyond that.

rust/Cargo.toml
```diff
@@ -46,44 +46,125 @@ readonly = "0.2"
...
-    "serde-lazy",
-    "streaming",
-    "round_series",
-    "rank",
-    "polars-ops",
-    "abs",
+#   If you change the version of Polars, 
+#   be sure to also change the Python package version requirement and test binary compatibility.
+# The features track the features enabled in the Rust build for the Polars Python crate.
+#   Having matching features is necessary for cross-binary DSL compatibility.
```
mccalluc: - Perhaps just a blank line (instead of indenting) to distinguish sections?
- "also change the Python package version requirement": I think there are at least four files that need to be kept in sync? Maybe add a test that just reads these files and checks?
- Having the same boilerplate text in all these files also adds up. Maybe one place has the authoritative description, and the others just point to it?

rust/src/data/ffi/mod.rs
```diff
@@ -246,8 +248,21 @@ pub extern "C" fn opendp_data__slice_as_object(
...
+        )
+    }
+    #[cfg(feature = "polars")]
+    pub fn deserialize_raw_dslplan(
+        raw: &FfiSlice
+    ) -> Fallible<DslPlan> {
+        let slice = unsafe { slice::from_raw_parts(raw.ptr as *const u8, raw.len) };
+        DslPlan::deserialize_versioned(slice)
+            .map_err(
+            |e| err!(FFI, "Error when deserializing 'DslPlan'. This may be because you're using features from Polars that are not currently supported. {}", e)
```
mccalluc, line 265: Many serializations include a version number. Any chance we have that information so we could check and have an error message like this?
> Was expecting polars vX but got polars vY.

(Might even check the version number first.)

rust/src/polars/mod.rs
```diff
@@ -246,11 +251,13 @@ impl ExtractValue for Series {
...
-            LiteralValue::String(string) => Some(string.into_string()),
-            _ => return fallible!(FailedFunction, "expected String, found: {:?}", literal),
-        })
+        if literal.is_null() {
+            return Ok(None);
+        }
+        literal
+            .extract_str()
+            .map(|s| Some(s.to_string()))
+            .ok_or_else(|| err!(FailedFunction, "expected String, found: {:?}", literal))
```
mccalluc, line 260: We're losing the parallelism with `impl ExtractValue for Series`. If this is the best approach, no worries.

rust/src/polars/mod.rs
```diff
@@ -498,7 +505,8 @@ impl From<LazyFrame> for OnceFrame {
             Ok(match query {
                 Query::External(q_external) => Answer::External(match q_external {
                     OnceFrameQuery::Collect => {
-                        let dataframe = lazyframe.collect()?;
+                        // TODO: allow CSE once https://github.com/pola-rs/polars/pull/23032 is released
```
mccalluc: File a tracking issue on our side and reference it here, if we want to be sure not to miss this?

mccalluc (commented):
Does the shell export need to be added to the instructions for developers? Beyond that, I'm mostly concerned about documentation for future devs. Whatever makes the most sense for you, and I can probably approve on the next pass.

python/src/opendp/extras/polars/__init__.py
Shoeboxam: This now segfaults without it. Was a bit tricky to track down.

mccalluc: Sorry about the segfaults! Add a comment reminding future maintainers? (They might want to turn this off for some reason as part of a larger change, hit a segfault, and have no clue that this was the root cause.)

Shoeboxam: The diff would reveal it, and the segfault could go away. I think it's fine.

rust/src/polars/mod.rs
Shoeboxam: The series held in literal should generally be `O(1)` in size, not `O(n)` in size.

rust/src/data/ffi/mod.rs
Shoeboxam: Version mismatch should be caught separately, leaving only unsupported features.

.github/workflows/smoke-test.yml
```diff
@@ -126,7 +126,9 @@ jobs:
         run: cargo test --verbose
 
       - name: Test (all-features)
-        run: cargo test --verbose --all-features
+        # pyo3/extension-module must be disabled to run tests.
+        # Context: https://pyo3.rs/v0.25.1/faq.html#i-cant-run-cargo-test-or-i-cant-build-in-a-cargo-workspace-im-having-linker-issues-like-symbol-not-found-or-undefined-reference-to-_pyexc_systemerror
+        run: cargo test --verbose --features=untrusted,polars,ffi,derive
```
mccalluc, line 131: This is a subset of https://docs.opendp.org/en/stable/contributing/development-environment.html#rust-feature-listing. Maybe the others are redundant with these, or they imply extension-module, or something else? More explanation?

Also worry if this means that some feature-only code will not be tested? Related:
- #2306 
- #1343

docs/source/contributing/development-environment.rst
```diff
@@ -56,8 +56,29 @@ Now run ``cargo build`` in the ``rust`` subdirectory of the repo:
...
+
+  On Mac, you may need to set ``RUSTFLAGS`` before build (`necessary for PyO3 <https://pyo3.rs/v0.25.1/building-and-distribution.html#macos>`_):
+
+  .. code-block:: bash
+
+    export RUSTFLAGS="-C link-arg=-undefined -C link-arg=dynamic_lookup"
+
 Substitute ``cargo build`` with ``cargo test`` to test, or ``cargo check`` to check syntax.
 
+.. note::
```
mccalluc, line 61: Slight preference to combine these two notes, since I think they are talking about the same thing.

Would it do any harm to make the `RUSTFLAGS` the default? The build is complicated enough, and if I'm following it line by line, and something doesn't work, I might get derailed and start looking up the error message, instead of seeing the note.

so something like:

```
.. code-block:: bash

    cd rust
    export RUSTFLAGS="-C link-arg=-undefined -C link-arg=dynamic_lookup"
    cargo build --all-features

This will compile a debug build of the OpenDP shared library, placing it in the directory ``opendp/rust/target/debug``. 
(The specific name of the library file will vary depending on your platform.)

.. note::
   
    ``RUSTFLAGS`` may not be necessary....
```

Or could this go in the TOML? It looks like there is a [`rustflags` option](https://doc.rust-lang.org/cargo/reference/config.html#configuration-format).

rust/src/transformations/make_stable_expr/expr_replace/mod.rs
```diff
@@ -126,12 +134,49 @@ pub(crate) fn literal_len(literal: &LiteralValue) -> Fallible<i64> {
...
+                .into_series(),
+                DynListLiteralValue::Float(list) => Float64Chunked::from_iter_options(
+                    PlSmallStr::from_static("literal"),
+                    list.into_iter().cloned(),
+                )
+                .into_series(),
+                DynListLiteralValue::Str(list) => StringChunked::from_iter_options(
+                    PlSmallStr::from_static("literal"),
+                    list.into_iter().cloned(),
+                )
```
mccalluc: Int128Chunked / Float64Chunked / StringChunked are the only things that are changing, I think. I don't know if there's an easy way to factor this out?

mccalluc (dismissed):
Approving so this isn't held up, but would it be possible to move that envvar into the TOML? Or add it to the main line of the docs, but suggest that for some users it might not be necessary?

.github/workflows/smoke-test.yml
Shoeboxam: More added. 

We should still get full coverage with the current set of features. If more features are added, then this would need to be updated.

docs/source/contributing/development-environment.rst
Shoeboxam: PyO3 seems to have a better recommendation now for the RUSTFLAGS, allowing us to hide this stuff in the build script.

I left the note for cargo test.

PR #2422, opened 2025-06-27
fixing comparison in _make_eigenvector

rionides:
What it says on the tin. Instead of accepting with probability 1 we now compare `if np_csprng.random() < ...:`


Shoeboxam (approved):
Thanks for finding this issue and fixing it!

PR #2436, opened 2025-07-28
docs: Bounds page in API user guide

### LINKED ISSUE #2432: Identifier truncation doc tweaks
Chuck McCallum:
> Looking at the "Identifier Truncation" docs now, and these are some first impressions:
> High level:
> - Ok to translate to RST, for the usual reasons?
> - Is this in the right place in the Tabular Data pages? It feels like it's related, or easily confused with grouping, so maybe it should come third, and the "Preparing Microdata" page goes last?
> 
> Low level:
> - https://docs.opendp.org/en/nightly/api/python/opendp.context.html#opendp.context.unit_of : "Greatest number of records or identifiers a privacy unit may contribute to microdata" -- I wonder if there's a way to explain this so it doesn't seem like two separate cases... Like we image there's a unique row ID by default, or you can select a different identifier?
> - Same page: "Greatest number of records a privacy unit may change in microdata": Should "or identifiers" also be added here?
> - https://docs.opendp.org/en/nightly/getting-started/tabular-data/identifier-truncation.html#Truncating-Per-Group-Contributions: "In order to make differentially private releases on this data": But the "data" is the same as in the previous pages. Maybe "context"?
> - Bad rendering: "See the API documentation for `truncate_per_group <../../api/python" -- I'll try to fix.
> - The discussion of keep and sorting might be getting into the weeds? Maybe move this to the API docs? The explanation is clear, I'm just not sure we need to get do to that level of detail here. Not a strong preference!
> - "To release queries that involve grouping, it is also necessary to bound the number of groups an individual may contribute to, not just the number of contributions per-group." Diagram might help here!
> - "you can also set keep to first or last." Just style: Quotes around "first" and "last" , assuming these are strings and not keywords?
> - "This means individuals could contribute data under two user identifiers" -- Not always a concern? "And further assume that identifiers are not shared between sources, even if an individual is included in multiple sources"

Michael Shoemate
> High Level:
> Yes!
> Sure!
> Low Level:
> Sure. Maybe something like, "If no identifier is set, each record is considered to correspond to a different identifier". It takes us further away from that nice records explanation though. Another option-- "Greatest number of records a privacy unit may contribute to microdata. If identifiers is set, corresponds to the greatest number of identifiers a privacy unit may contribute to the data."
> Yes, it should match
> Yes. Or maybe "with this context"?
> Thanks!
> Yes, it could be moved.
> Yes

Make a quick PR to translate to RST and fix link and change order, and then make a separate PR for the more involved changes.

We can start a new API user-guide page (to follow `api/user-guide/polars/gaussian-noise.html`) where we start to talk about bounds and norms.

mccalluc:
- Fix #2432?

Cut out some of the details and move to a new page. Should we also bring in a discussion of L0 and L1 here?


docs/source/getting-started/tabular-data/essential-statistics.rst
```diff
@@ -561,11 +561,6 @@ implementation details.
 Since work hours tend to be concentrated a little less than 40, this
 release seems reasonable.
 
-Throughout this notebook, all ``.dp`` expressions take an optional scale
```
mccalluc, line 604: Propose striking this unless we want to add an example right here.

docs/source/getting-started/tabular-data/identifier-truncation.rst
```diff
@@ -83,95 +83,27 @@ contributions to ten.
...
 Previous examples with this dataset assumed the worst-case of 36
 contributed records per individual (one contribution per quarter for
-nine years) which resulted in a 36-time increase in the amount of noise.
+nine years) which resulted in a 36-fold increase in the amount of noise.
 By truncating to at most ten records, there is only a 10-fold increase
 in the amount of noise. This statistical estimator is introducing some
 bias by dropping records from individuals who contributed more than ten
-records, at the benefit of attaining a much lower variance.
-
-By default, ``truncate_per_group`` takes a random sample of records
```
mccalluc, line 104: This block is moved to the API User Guide.

docs/source/getting-started/tabular-data/identifier-truncation.rst
```diff
@@ -83,95 +83,27 @@ contributions to ten.
             │ HWUSUAL ┆ Length    ┆ Integer Laplace ┆ 80.0   │
             └─────────┴───────────┴─────────────────┴────────┘
 
+The ``.truncate_per_group(10)`` is equivalent to ``.filter(pl.int_range(pl.len()).over("PIDENT") < 10)``
+and returns the same scale parameters,
+but trucate easier to read, and keeps track of the identifier columns automatically.
```
Shoeboxam:
```suggestion
but `truncate_per_group` is easier to read and write.
```

docs/source/api/user-guide/polars/bounds.rst
```diff
@@ -0,0 +1,177 @@
+.. _bounds-user-guide:
+
+Identifier Truncation and Bounds
+================================
+
+It's also important to be mindful of the structure of our data
+when thinking about identifier truncation and bounds.
+This is another area where there are opportunities to lower the
+sensitivity of the analysis to individual contributions, and hence the noise required.
```
Shoeboxam:
```suggestion
sensitivity of the analysis to individual contributions, 
and hence reduce the amount of noise required to meet a given privacy guarantee.
```

Shoeboxam (approved):

PR #2448, opened 2025-08-13
docs: format code samples

### LINKED ISSUE #2446: Check width and formatting of rst code sample
At a minimum, check line width: Shouldn't need to horizontal scroll to read docs. 

Is it possible/easy to do more formatting/linting of doctests, either in python docstrings or RST?

### LINKED ISSUE #2447: Remove `testsetup::` and `doctest::`
RST gives us a lot of different ways to do pretty much the same thing. If I'm trying to do some cleaning across the docs, these variations can get in the way. 

For users, I think it also helps to have any required setup explicit in the docs.

Suggest scanning the docs for `testsetup::` and `doctest::` (and any others) and replacing with plain `code::`.

Prompted by
- #2446

mccalluc:
- Fix #2446
- Fix #2447 (Doesn't add a check, but I think things are standardized now.)

Installed https://pypi.org/project/blacken-docs/ locally, and it seems to work well. (Change from `python` to `pycon` required by `blacken-docs`.)

Assuming it doesn't break any tests, this much seems harmless. Chose 59 as the linewidth for black because that seems to be the worst case with the responsive docs site, until you make the window much narrower.

Regex to find some comment lines which are too long, though this isn't everything.

If this much is useful, would it be good to do anything more here? We decided against precommits earlier, as it adds one more step to get started, but the line-length could be checked in CI, and the error there could suggest running blacken-docs.


docs/source/api/user-guide/plugins/measurement.rst
```diff
@@ -24,42 +24,47 @@ This example mocks the typical API of the OpenDP library to make the *most priva
...
         ...     >> dp.t.then_sum()
         ...     >> make_base_constant("denied")
         ... )
-        ...
         >>> meas([2, 3, 4])
         'denied'
-        >>> meas.map(1) # computes epsilon, because the output measure is max divergence
+        >>> meas.map(
+        ...     1
+        ... )  # computes epsilon, because the output measure is max divergence
```
Shoeboxam: Stuff like this is unfortunate, and could be cleaned up by some manual work splitting lines.

Shoeboxam (dismissed):
Agreed with the later comments. Thanks for the continual continual integration improvements.

Shoeboxam (approved):

PR #2451, opened 2025-08-15
dev: cache rust build

### LINKED ISSUE #2435: Cache rust build in CI?
Most of my work doesn't touch the rust codebase, but we need a rust build on every run of smoketest. 

What if we hash the rust directory, and if there are no changes, reuse a previously built binary?

It looks like each branch has its own cache space:
https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#restrictions-for-accessing-a-cache

Syntax would be something like:
```
with:
    path: rust/target
    key: ${{ hashFiles('rust/src/**') }}
```

mccalluc:
- Fix #2435

It seems like we could use github actions cache instead of artifacts between jobs, and skip successive build steps if the artifact already exists.
- Faster CI (particularly for someone mostly working on the Python codebase)
- Less config? instead of a name for the artifact, we only need to think about paths.

Questions:
- Does this actually work?
- Is there a way to avoid the boilerplate?
- Should the successive build steps be combined, rather than checking for each one?
- Is there any problem with using the existence check, rather than checking for cache hit?

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -42,8 +42,21 @@ jobs:
             ~/.cargo/registry
             ~/.cargo/git
             rust/target
-          key: ${{ runner.os }}-cargo-${{ hashFiles('rust/Cargo.toml') }}
+          key: dependencies-${{ runner.os }}-cargo-${{ hashFiles('rust/Cargo.toml') }}
+          # Even if there are some changes to Cargo.toml, the cache is a useful head start:
           restore-keys: ${{ runner.os }}-cargo-
+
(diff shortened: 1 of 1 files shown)


mccalluc:
(Changed the key structure, so this run didn't get a cache hit for rust artifacts.)

mccalluc:
This could work, but switching to cache and complicating the logic at the same time is making this hard to work with. Going to start over, and make fewer changes, but keep this around until we get it done, one way or the other.

Shoeboxam:
Closing for now, can reopen if useful!

PR #2452, opened 2025-08-15
If envvar is set, show more info

### LINKED ISSUE #2430: confusing error message if `OPENDP_LIB_DIR` is set to ""
Not sure how or why I set it to an empty string, but it is:
```
>>> os.environ.get("OPENDP_LIB_DIR")
''
```

The empty string becomes `.` when wrapped in a path:
```
>>> Path('')
PosixPath('.')
```

This causes a confusing error message from pytest:
```
    raise Exception(f"Expected exactly one binary to be present in {lib_dir}. Got: {lib_dir_file_names}")  # pragma: no cover
E   Exception: Expected exactly one binary to be present in .. Got: []
```

`unset OPENDP_LIB_DIR` fixes it for me, but we could also make this a little more robust.

mccalluc:
- Fix #2430?

### DIFF
--- a/python/src/opendp/_lib.py
+++ b/python/src/opendp/_lib.py
@@ -20,7 +20,8 @@ ATOM_EQUIVALENCE_CLASSES: MutableMapping[str, Sequence[str]] = {
 
 def _load_library():
     default_lib_dir = Path(__file__).absolute().parent / "lib"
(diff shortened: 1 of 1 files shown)


Shoeboxam (approved):
Why not? Seems useful. Thanks.

PR #2458, opened 2025-08-18
feat: extrinsic metric descriptor; lifeline simplification

### LINKED ISSUE #2443: metric descriptor for extrinsic distance
Similar to how extrinsic domains can carry data, it would be helpful if extrinsic metrics could similarly carry data.

Shoeboxam:
Closes #2443


docs/source/conf.py
```diff
@@ -106,9 +106,10 @@ def setup(app):
     ('py:class', 'opendp.mod.LP_AnyTransformation'),
 
     # I think the problem is that Sphinx is making parameter list documentation,
-    # and it doesn't understand that `M` and `T` are type parameters, not actual types.
+    # and it doesn't understand that `M`, `T` and `D` are type parameters, not actual types.
```
```suggestion
    # and it doesn't understand that `M`, `T`, and `D` are type parameters, not actual types.
```
very slight preference.

python/src/opendp/_convert.py
```diff
@@ -243,8 +246,20 @@ def c_to_py(value: Any) -> Any:
                 value.__class__ = SymmetricIdDistance
             elif rt_type == "ChangeOneIdDistance":
                 value.__class__ = ChangeOneIdDistance
+            elif rt_type == "ExtrinsicDistance":
+                value.__class__ = ExtrinsicDistance
```
mccalluc: Here and above, could we minimize the use of strings? I always worry about having a typo that would be hard to debug.

```suggestion
            elif rt_type == ExtrinsicDistance.__name__:
                value.__class__ = ExtrinsicDistance
```

python/src/opendp/_lib.py
```diff
@@ -57,6 +57,54 @@ def _load_library():
...
+    try:
+        # compare equality first in case order not supported
+        if left == right:
+            cmp = 1
+        elif left < right:
+            cmp = 0
+        elif left > right:
+            cmp = 2
+        else:
+            raise ValueError("left and right are not comparable")
```
mccalluc, line 70: I think you'd probably get a `TypeError` before you reach this line, but it's good to handle the fall-through.

python/src/opendp/mod.py
```diff
@@ -967,6 +978,30 @@ def __hash__(self) -> int:
...
         raise ValueError("Metric does not support iteration")
+    
+    def cast(self, type_: Type[D]) -> D:
+        """Retrieve the descriptor as the prescribed type, or error."""
+        if not (
+            isinstance(self, ExtrinsicDistance)
+            and isinstance(descriptor := self.descriptor, type_)
+        ):
+            raise ValueError(f"metric descriptor must be a {type_.__name__}")
+        return descriptor
```
mccalluc, line 990: A little bit of duplication here, except for the error message. If there are a lot of other parallels, might try to make it OO, or pull out shared util functions, but not something that needs to block this PR.

python/test/test_metrics.py
```diff
@@ -50,3 +49,48 @@ def test_group_bound():
...
+def test_user_metric_total_cmp():
+    m_comp = dp.c.make_adaptive_composition(
+        input_domain=dp.atom_domain(T=bool),
+        input_metric=dp.user_distance("user distance"),
+        output_measure=dp.max_divergence(),
+        d_in=1,
+        d_mids=[1.0],
+    )
+    m_comp.map(0)
+    m_comp.map(1)
```
mccalluc: make assertions?

python/test/test_metrics.py
```diff
@@ -50,3 +49,48 @@ def test_group_bound():
...
+    with pytest.raises(
+        dp.OpenDPException,
+        match="d_in from the privacy map must be no greater than the d_in",
+    ):
+        m_comp.map(2)
+    with pytest.raises(dp.OpenDPException, match="not comparable"):
+        m_comp.map(float("nan"))
+
+    class Dist:
```
mccalluc, line 73: since `m_comp` is redefined after this point, maybe pull into separate test? It seems like the rest doesn't depend on what came before.

rust/src/ffi/util.rs
```diff
@@ -54,30 +58,51 @@ use chrono::{NaiveDate, NaiveDateTime, NaiveTime};
...
+
+impl ProductOrd for ExtrinsicObject {
+    fn total_cmp(&self, other: &Self) -> Fallible<std::cmp::Ordering> {
+        let obj = Fallible::from(into_owned(TOTAL_CMP.get().unwrap()(self.0, other.0))?)?;
+        match obj.downcast_ref::<u8>()? {
+            0 => Ok(Ordering::Less),
+            1 => Ok(Ordering::Equal),
+            2 => Ok(Ordering::Greater),
+            _ => fallible!(FailedFunction, "failed to compare extrinsic objects"),
+        }
```
mccalluc: If this needs to stay a `u8`, or there are other constraints, ignore this, but it might be more typical to get {-1, 0, 1} back from a cmp-style function.

mccalluc (dismissed):
A couple comments, but nothing that is essential. Approved!

(I'll keep on worrying about hard-coded strings, but if it gets old, let me know if a reminder isn't needed in every PR.)

python/src/opendp/_lib.py
Shoeboxam: It depends on how comparison is implemented for the type. I was thinking people could create custom classes for distance types if they want to customize comparisons, which allows us to avoid the FFI complexity of closing over comparison functions in the AnyObject.

PR #2459, opened 2025-08-18
feat: add make_noise constructors

### LINKED ISSUE #2444: feat: add `make_noise` constructor
For consistency with Polars, and simpler switching between mechanisms in higher-level APIs.

Shoeboxam:
Close #2444

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -158,6 +158,8 @@ polars = { version = "=0.50.0", features = [
 ], optional = true }    
 polars-plan = { version = "=0.50.0", features = ["ffi_plugin", "python"], optional = true }
 polars-utils = { version = "=0.50.0" }
+# pinned due to: https://github.com/pola-rs/polars/issues/24208
+planus = { version = "=1.1.1" }
 
 # matches the version specified in Polars: 
 #   https://github.com/pola-rs/polars/blob/main/Cargo.toml


python/test/test_meas.py
```diff
@@ -208,6 +222,7 @@ def test_gaussian():
 ])
 def test_noisy_max(measure, d_out):
     input_domain = dp.vector_domain(dp.atom_domain(T=dp.usize))
+    input_metric = dp.linf_distance(T=dp.usize)
 
     input_metric = dp.linf_distance(T=dp.usize)
```
mccalluc:
```suggestion
    input_metric = dp.linf_distance(T=dp.usize)
```

python/test/test_meas.py
```diff
@@ -312,10 +327,18 @@ def test_gaussian_threshold_int():
...
     assert meas.map((1, 10, 10)) == (12.5, 1.1102230246251565e-16)
 
-def test_gaussian_threshold_float():
+def make_noise_threshold(domain, metric, scale, threshold):
+    measure = dp.approximate(dp.zero_concentrated_divergence())
+    return dp.m.make_noise_threshold(domain, metric, measure, scale, threshold)
+
+@pytest.mark.parametrize("constructor", [
+    dp.m.make_gaussian_threshold,
+    make_noise_threshold
```
mccalluc:
```suggestion
def new_make_noise_threshold(domain, metric, scale, threshold):
    measure = dp.approximate(dp.zero_concentrated_divergence())
    return dp.m.make_noise_threshold(domain, metric, measure, scale, threshold)

@pytest.mark.parametrize("constructor", [
    dp.m.make_gaussian_threshold,
    new_make_noise_threshold
```

The pattern of `new_make_noise` seems useful here, just to avoid ambiguity.

rust/src/measurements/noise/distribution/mod.rs
```diff
@@ -24,3 +34,77 @@ impl<T: 'static + CheckAtom> NoiseDomain for AtomDomain<T> {
...
+    (DI, MI): MetricSpace,
+{
+    output_measure
+        .new_distribution(scale, k)
+        .make_noise((input_domain, input_metric))
+}
+
+pub trait NoiseMeasure: Measure + 'static {
+    #[cfg(feature = "polars")]
+    const DISTRIBUTION: NoiseDistribution;
```
mccalluc: I'm a little surprised that there's a connection between this and the polars feature. I don't know the big picture, but if it's useful in that one case, maybe it should be on by default? You know better than me.

rust/src/measurements/noise_threshold/distribution/test.rs
```diff
@@ -0,0 +1,3 @@
+fn test_noise_threshold() -> Fallible<()> {
+
```
mccalluc: Fill in a test!

mccalluc (commented):
Big thing is just to fill in the testing, but looks good overall.

python/test/test_meas.py
mccalluc: Huh: It really looks like the lines are duplicated, and it's not obvious to me why both are needed. Add a comment?

mccalluc (commented):
One questions carried over. Sorry if I'm missing something.

python/test/test_meas.py
Shoeboxam: Whoops! Fixed now.

mccalluc (dismissed):
Thanks. Changes look good. CI just needs to be happy.

PR #2460, opened 2025-08-18
docs: contingency tables and synthetic data

### LINKED ISSUE #2442: docs: contingency tables

Shoeboxam:
Close #2442

### DIFF
--- a/docs/source/getting-started/statistical-modeling/synthetic-data.rst
+++ b/docs/source/getting-started/statistical-modeling/synthetic-data.rst
@@ -1,0 +1,148 @@
+.. _synthetic-data:
+
+Synthetic Data
+==============
+
+Synthetic data mechanisms in OpenDP attempt to model the relationships between columns in your data via a contingency table,
+and then generate synthetic data with the same relationships.
+
+OpenDP uses `Private-PGM <https://private-pgm.readthedocs.io/en/latest/introduction.html>`_ 
+to build a contingency table that is consistent with a set of marginal queries via a process called model-based inference (``mbi``).
+This functionality is not enabled by default.
+
+.. tab-set::
+
+    .. tab-item:: Python
+        :sync: python
+
+        Add ``[mbi]`` to the package installation to enable model-based-inference functionality.
+
+        .. prompt:: bash
+
+            pip install 'opendp[mbi]'
+
+        This installs the specific version of the Private-PGM package that is compatible with the OpenDP Library.
+
+    .. tab-item:: R
+        :sync: r
+
+        ``mbi`` (Private-PGM) is only available in Python. 
+
+    .. tab-item:: Rust
+        :sync: rust
+
+        ``mbi`` (Private-PGM) is only available in Python. 
+
+Let's get started by setting up the context for the Labor Force dataset.
+
+.. code:: pycon
+
+    >>> import opendp.prelude as dp
+    >>> import polars as pl
+
+    >>> dp.enable_features("contrib")
+
+    >>> context = dp.Context.compositor(
+    ...     data=pl.scan_csv(dp.examples.get_france_lfs_path(), ignore_errors=True),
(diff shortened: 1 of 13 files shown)


docs/source/getting-started/statistical-modeling/synthetic-data.rst
```diff
@@ -0,0 +1,148 @@
...
+    >>> dp.enable_features("contrib")
+
+    >>> context = dp.Context.compositor(
+    ...     data=pl.scan_csv(dp.examples.get_france_lfs_path(), ignore_errors=True),
+    ...     privacy_unit=dp.unit_of(contributions=36),
+    ...     privacy_loss=dp.loss_of(rho=0.2, delta=2e-7),
+    ... )
+
+
+The :ref:`contingency table <contingency-tables>` documentation is recommended reading
```
mccalluc: Preference to include everything in the getting-started that is necessary for a basic understanding. If extra reading could help, introduce it at the end.

docs/source/getting-started/statistical-modeling/synthetic-data.rst
```diff
@@ -0,0 +1,148 @@
...
+The :ref:`contingency table <contingency-tables>` documentation is recommended reading 
+and gives an introduction to the problem,
+but differs in that the query workload is fixed (using the :py:class:`opendp.extras.mbi.Fixed` algorithm).
+All algorithms for differentially private contingency table estimation inherit from :py:class:`opendp.extras.mbi.Algorithm`:
+these being :py:class:`opendp.extras.mbi.Fixed`, :py:class:`opendp.extras.mbi.AIM`, :py:class:`opendp.extras.mbi.MST` and :py:class:`opendp.extras.mbi.Sequential`.
+
+The algorithms for synthetic data differ from the fixed algorithm
+in that they first estimate all unknown first-order marginals, not just the marginals with missing keys.
+This behavior can be controlled by switching between ``.oneway="all"`` and ``.oneway="unkeyed"``
+in the algorithm of choice.
```
mccalluc: For getting started, I think we should just pick one, and leave the rest, and the comparison between the choices, to the API User Guide.

docs/source/getting-started/tabular-data/contingency-tables.rst
```diff
@@ -0,0 +1,238 @@
...
+In this setting, the scale and threshold are reasonably small.
+To make the threshold smaller, consider adding more key-sets 
+or bounding the number of groups a user may contribute.
+
+When you release, all marginals are estimated and stored inside a :py:class:`opendp.extras.mbi.ContingencyTable`.
+The contingency table supports counting queries over arbitrary sets of grouping columns.
+
+.. code:: pycon
+
+    >>> table1: dp.mbi.ContingencyTable = query.release()
```
mccalluc: more descriptive name than `table1`?

mccalluc (approved):
Approve: This is fine, but I also think a lot of this would be better in the the API User Guide. There are different ways this could be done, but I'd suggest:

- Keep `tabular-data/contingency-tables.rst`, but trim it down, and make sure we're not referencing ideas that haven't already been introduced.
- At the end of that page, demonstrate synthetic data
- But for all the details, point to the API User Guide.

This could be follow-up work after the release.

docs/source/getting-started/tabular-data/contingency-tables.rst
Shoeboxam: I just renamed table1 to table. table2 then makes sense as the updated table.

docs/source/getting-started/statistical-modeling/synthetic-data.rst
Shoeboxam: That ref is to another "getting-started" section.

mccalluc (approved):
Changes look good. Moving details into API docstrings makes sense.

PR #2461, opened 2025-08-19
feat: Save time by caching results of rust build

### LINKED ISSUE #2435: Cache rust build in CI?

mccalluc:
- fix #2435

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -58,7 +58,7 @@ jobs:
       - name: Upload (dynamic w/ polars)
         uses: actions/upload-artifact@v4
         with:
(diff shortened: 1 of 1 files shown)


Shoeboxam:
Closing for now, can reopen if useful!

PR #2463, opened 2025-08-19
docs: Simplify "assessing utility"

### LINKED ISSUE #2454: More docs on "Assessing Utility"?
Currently we just say

> Finding an acceptable tradeoff between privacy and utility can be a challenge. In this section of the documentation, we will demonstrate tools to help you navigate this tradeoff.
> 
> - Accuracy

I don't see an issue to remind us to actually do this. I don't like long-standing TODOs in the docs, so maybe simplify this if we don't have more to add right now?

mccalluc:
- Fix #2454... unless we want to keep the issue open to remind us to add more content here?

With just one child in this section, it seems like it can be flattened. Is this a place where we'd like to mention the registry? Other ideas for expanding it?

### DIFF
--- a/docs/source/getting-started/index.rst
+++ b/docs/source/getting-started/index.rst
@@ -14,4 +14,4 @@ This is an overview of OpenDP's Python and R APIs.
   typical-workflow
   tabular-data/index
   statistical-modeling/index
-  utility/index
+  utility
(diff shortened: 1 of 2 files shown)


Shoeboxam (approved):
Can always add to it later, and it improves the current state.

PR #2464, opened 2025-08-19
docs: Combinators and compositors docs reorganization

### LINKED ISSUE #2453: Tidy up Combinators / Compositors in API User Guide
With incoming work
- #678

the Compositors page (`api/user-guide/combinators/compositors.html`) is getting pretty long. Compositors is also the only child page of Combinators. Suggestions:
- Split the sections of `compositors.html` out as separate pages.
- Move some of the information in `combinators` down to a child page.
- "Chaining" is described on both `combinators` and `compositors`. Maybe consolidate?

mccalluc:
- Fix #2453

This is almost entirely just splitting up the big pages into smaller pages and adding navigation. One thing that's a little different here is that I've made three ToCs on the page: 
- Chaining
- Composition
- Other topics

It does seem worth while to keep these sections separate, but I also think it's useful to have one index page that surveys everything, rather than pushing these down a level.

Odometers as a concept didn't have their own page. I've made it a little more prominent, and maybe this is enough?

### DIFF
--- a/docs/source/api/user-guide/combinators/compositor-chaining-and-nesting.rst
+++ b/docs/source/api/user-guide/combinators/compositor-chaining-and-nesting.rst
@@ -14,9 +14,25 @@ they also support chaining.
 
         .. code:: python
 
+            >>> meas_adaptive_comp = dp.c.make_adaptive_composition(
+            ...     input_domain=dp.vector_domain(dp.atom_domain(T=int)),
+            ...     input_metric=dp.symmetric_distance(),
+            ...     output_measure=dp.max_divergence(),
+            ...     d_in=1,
+            ...     d_mids=[2., 1.]
+            ... )
+
             >>> str_space = dp.vector_domain(dp.atom_domain(T=str)), dp.symmetric_distance()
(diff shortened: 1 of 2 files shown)


docs/source/api/user-guide/combinators/non-adaptive-composition.rst
```diff
@@ -0,0 +1,136 @@
...
+
+.. tab-set::
+
+    .. tab-item:: Python
+        :sync: python
+
+        .. code:: python
+
+            >>> import opendp.prelude as dp
+            >>> dp.enable_features("contrib")
```
Shoeboxam, line 16: If we carry state over from prior pages, can we drop this boilerplate?

Shoeboxam (approved):
Minor optional comment. This is clearly an improvement.

docs/source/api/user-guide/combinators/non-adaptive-composition.rst
mccalluc: I've filed:
- #2468 

We could have a single file which is referenced from each of these RSTs, but I kind of like being able to run each page from the top. We can come back to this.

PR #2473, opened 2025-08-25
Polars: add plugins for DP aggregators

### LINKED ISSUE #2472: Polars: add plugins for DP aggregators
* for use in DP SQL
* plan-dependent insertion of imputation, clipping, noise scales

Shoeboxam:
Close #2472


.github/workflows/smoke-test.yml
```diff
@@ -195,20 +195,20 @@ jobs:
           esac
 
       - name: Test replace-binary-path
-        if: ${{ matrix.extra-install == 'polars' }}
+        if: matrix.extra-install == 'polars'
```
mccalluc, line 198: I've definitely been confused about this. TLDR: [braces are optional, unless there's a `!`.](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idif)

> You must always use the `${{ }}` expression syntax or escape with '', "", or () when the expression starts with !, since ! is reserved notation in YAML format.

So these are all safe.

docs/source/getting-started/tabular-data/essential-statistics.rst
```diff
@@ -232,11 +232,7 @@ work hours across responses.
             ...     # 99 represents "Not applicable"
             ...     context.query().filter(pl.col("HWUSUAL") != 99.0)
             ...     # compute the DP sum
-            ...     .select(
-            ...         pl.col.HWUSUAL.cast(int)
-            ...         .fill_null(35)
-            ...         .dp.sum(bounds=(0, 80))
-            ...     )
+            ...     .select(pl.col.HWUSUAL.cast(int).dp.sum(bounds=(0, 80)))
```
mccalluc: Wouldn't you want to keep this, since the fill is non-zero?

docs/source/getting-started/tabular-data/essential-statistics.rst
```diff
@@ -325,9 +321,7 @@ estimates.
             ...     context.query().filter(pl.col.HWUSUAL != 99.0)
             ...     # release both the sum and length in one query
             ...     .select(
-            ...         pl.col.HWUSUAL.cast(int)
-            ...         .fill_null(35)
-            ...         .dp.sum(bounds=(0, 80)),
+            ...         pl.col.HWUSUAL.cast(int).dp.sum(bounds=(0, 80)),
```
mccalluc, line 330: ditto

docs/source/getting-started/tabular-data/essential-statistics.rst
```diff
@@ -425,9 +419,7 @@ filtered data, as shown above.
         .. code:: pycon
 
             >>> query_mean_work_hours = context_bounded_dp.query().select(
-            ...     pl.col.HWUSUAL.cast(int)
-            ...     .fill_null(35)
-            ...     .dp.mean(bounds=(0, 80))
+            ...     pl.col.HWUSUAL.cast(int).dp.mean(bounds=(0, 80))
```
mccalluc, line 428: ditto

docs/source/getting-started/tabular-data/essential-statistics.rst
```diff
@@ -491,9 +483,7 @@ set candidates to whole numbers between 20 and 60:
             ...     context.query()
             ...     .filter(pl.col.HWUSUAL != 99.0)
             ...     .select(
-            ...         pl.col.HWUSUAL.cast(int)
-            ...         .fill_null(35)
-            ...         .dp.median(candidates)
+            ...         pl.col.HWUSUAL.cast(int).dp.median(candidates)
```
mccalluc, line 492: ditto

docs/source/getting-started/tabular-data/essential-statistics.rst
```diff
@@ -556,7 +546,6 @@ hours:
             ...     .filter(pl.col.HWUSUAL != 99.0)
             ...     .select(
             ...         pl.col.HWUSUAL.cast(int)
-            ...         .fill_null(35)
```
mccalluc, line 559: ditto

docs/source/theory/attacks/differencing.ipynb
```diff
@@ -258,7 +258,7 @@
     {
      "data": {
       "text/plain": [
-       "32893"
+       "86414"
```
mccalluc, line 261: big change is surprising, but not impossible?

python/src/opendp/extras/polars/__init__.py
```diff
@@ -110,72 +108,33 @@ def noise(
...
 
         return register_plugin_function(
             plugin_path=os.environ.get("OPENDP_POLARS_LIB_PATH", lib_path),
             function_name="noise",
-            args=(self.expr, lit(distribution), scale),
+            args=(self.expr, scale),
             is_elementwise=True,
         )
 
+    @deprecated(reason="Use .noise instead.")
```
mccalluc: Add a `since`... but do you know yet whether this will be 0.14.1 or 0.15.0? I had back-filled others where the `since` was missing.

mccalluc: Could someone have been using the wrong kind of noise before, and since `distribution` is no longer a param, it would be doing something different? If the behavior is changing, I wonder if it would be safer to raise an exception, rather than just an error?

python/src/opendp/extras/polars/__init__.py
```diff
@@ -211,7 +170,14 @@ def len(self, scale: float | None = None):
         It can differ from frame length (``.select(dp.len())``) if the expression uses transformations that change the number of rows,
         like filtering.
         """
-        return self.expr.len().dp.noise(scale)
+        from polars.plugins import register_plugin_function  # type: ignore[import-not-found]
+
+        return register_plugin_function(
+            plugin_path=os.environ.get("OPENDP_POLARS_LIB_PATH", lib_path),
```
mccalluc: This is repeated a few times. Make a tiny helper function?
```
def _get_lib_path():
```
and then here:
```suggestion
            plugin_path=_get_lib_path(),
```

python/src/opendp/extras/polars/__init__.py
```diff
@@ -246,7 +212,14 @@ def count(self, scale: float | None = None):
...
         """
-        return self.expr.count().dp.noise(scale)
+        from polars.plugins import register_plugin_function  # type: ignore[import-not-found]
+
+        return register_plugin_function(
+            plugin_path=os.environ.get("OPENDP_POLARS_LIB_PATH", lib_path),
+            function_name="dp_count",
+            args=(self.expr, scale),
+            returns_scalar=True,
+        )
```
mccalluc: And since this is also repeated, another helper function?
```
def _register_polars_plugin_function(function_name):
    from polars.plugins import register_plugin_function  # type: ignore[import-not-found]

    return register_plugin_function(
        plugin_path=_get_lib_path(),
        function_name=function_name,
        args=(self.expr, scale),
        returns_scalar=True,
    )
```
and here:
```suggestion
        return _register_polars_plugin_function("dp_count")
```

python/src/opendp/extras/polars/__init__.py
```diff
@@ -285,7 +258,14 @@ def null_count(self, scale: float | None = None):
...
         """
-        return self.expr.null_count().dp.noise(scale)
+        from polars.plugins import register_plugin_function  # type: ignore[import-not-found]
+
+        return register_plugin_function(
+            plugin_path=os.environ.get("OPENDP_POLARS_LIB_PATH", lib_path),
+            function_name="dp_null_count",
+            args=(self.expr, scale),
+            returns_scalar=True,
+        )
```
mccalluc: ditto

python/src/opendp/extras/polars/__init__.py
```diff
@@ -320,7 +300,14 @@ def n_unique(self, scale: float | None = None):
...
         """
-        return self.expr.n_unique().dp.noise(scale)
+        from polars.plugins import register_plugin_function  # type: ignore[import-not-found]
+
+        return register_plugin_function(
+            plugin_path=os.environ.get("OPENDP_POLARS_LIB_PATH", lib_path),
+            function_name="dp_n_unique",
+            args=(self.expr, scale),
+            returns_scalar=True,
+        )
```
mccalluc: ditto

python/src/opendp/extras/polars/__init__.py
```diff
@@ -359,20 +346,29 @@ def sum(self, bounds: tuple[float, float], scale: float | None = None):
...
         Output is noise added to two due to each value being clipped to (0, 1).
         """
-        return self.expr.clip(*bounds).sum().dp.noise(scale)
+        from polars.plugins import register_plugin_function  # type: ignore[import-not-found]
+        from polars import lit  # type: ignore[import-not-found]
+
+        return register_plugin_function(
+            plugin_path=os.environ.get("OPENDP_POLARS_LIB_PATH", lib_path),
+            function_name="dp_sum",
+            args=(self.expr, lit(bounds[0]), lit(bounds[1]), scale),
```
mccalluc: args make this one different. Your call.

python/src/opendp/extras/polars/__init__.py
```diff
@@ -398,68 +394,21 @@ def mean(
...
-        :param alpha: a value in [0, 1]. Choose 0.5 for median
-        :param candidates: Set of possible quantiles to evaluate the utility of.
-        """
         from polars.plugins import register_plugin_function  # type: ignore[import-not-found]
-        from polars import Series  # type: ignore[import-not-found]
+        from polars import lit  # type: ignore[import-not-found]
+
+        if isinstance(scale, tuple):  # pragma: no cover
+            raise ValueError("OpenDP 0.15 adjusts the scale to only consist of a single float. "
+                             "Individually estimate sum and len to tune budget distribution.")
```
mccalluc: Just to make the deprecation path easier, would it make any sense to calculate the value for them, but raise a deprecation warning?

(This also answers my question about whether we expect these changes to go out in 0.14.1 or 0.15.0. If we're not 100% sure about that, could imagine using a `NEXT_VERSION` global, and then filling in the right value before release. Probably overkill.)

python/test/test_polars.py
```diff
@@ -309,24 +310,6 @@ def test_mechanisms(measure):
...
-    plan = seed(lf.collect_schema()).select(pl.len().dp.gaussian(0.0))
-    with pytest.raises(dp.OpenDPException) as err:
-        dp.m.make_private_lazyframe(
-            lf_domain,
-            dp.symmetric_distance(),
-            dp.max_divergence(),
-            plan,
-            0.0,
-        )
-    assert "expected Laplace distribution, found Gaussian" in (err.value.message or "")
```
mccalluc, line 327: I'd keep this around: A user could still try to do this, but we should check that our warnings/errors are right.

rust/src/measurements/make_private_expr/expr_dp_counting_query/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+        #[derive(Clone, Serialize, Deserialize)]
+        pub(crate) struct $plugin;
+        impl ColumnsUdf for $plugin {
+            // makes it possible to downcast the AnonymousFunction trait object back to Self
+            fn as_any(&self) -> &dyn std::any::Any {
+                self
+            }
+
+            fn call_udf(&self, _: &mut [Column]) -> PolarsResult<Option<Column>> {
+                polars_bail!(InvalidOperation: "OpenDP expressions must be passed through make_private_lazyframe to be executed.")
```
mccalluc, line 38: Here and below, wouldn't mind a python test of error handling.

rust/src/polars/mod.rs
```diff
@@ -137,6 +143,35 @@ where
     }))
 }
 
+// TODO: move the `const V` into `P` once generic parameters may be used in const operations
```
mccalluc: Is there an issue for this that could be referenced here? Unless this is really the next thing you'll tackle, it would be good to make sure we don't lose track of it.

.github/workflows/smoke-test.yml
Shoeboxam: Thanks for the pointer. I noticed it worked without, so I simplified, but the ! gotcha is annoying.

docs/source/theory/attacks/differencing.ipynb
Shoeboxam: This is querying for a single person, so it should be a very poor-utility estimate that will differ by a lot on each run. In this case, it's the sum of a dataset of size one, with bounds of 0 and 70k-- so a relatively large amount of noise will be added compared to the data size.

python/src/opendp/extras/polars/__init__.py
Shoeboxam: Beforehand, it would have errored. Now it might use a different noise distribution.

Shoeboxam: I added a bit more info to the deprecation warning.

python/src/opendp/extras/polars/__init__.py
Shoeboxam: This case still has a fair amount of diversity- different numbers of args, names, may or may not return scalar.

python/src/opendp/extras/polars/__init__.py
Shoeboxam: This is a case where the user passes more data than is needed, but the library would need to discard it. I think it would be better for the user to know that they can't do that anymore, and need to take a different approach to manually control budget distribution amongst the plugin queries.

python/test/test_polars.py
Shoeboxam: This is structurally no longer possible to check, as the new methods just call to noise, and don't know what the user-requested noise distribution was. I'd have to somehow route that data down into Rust just to make a better deprecation warning.

rust/src/measurements/make_private_expr/expr_dp_counting_query/mod.rs
Shoeboxam: Added!

rust/src/polars/mod.rs
Shoeboxam: I changed to a NOTE, and added more explanation. I couldn't find the relevant Rust compiler issue that is blocking that feature.

PR #2475, opened 2025-08-27
Debug release build

Shoeboxam:
Workflow now only fails due to lack of space on PyPy (the next step in the journey!):
https://github.com/opendp/opendp/actions/runs/17463677750

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -42,7 +42,7 @@ on:
         required: false
         default: false
 env:
-  FEATURES: untrusted,ffi,polars
+  FEATURES: untrusted,ffi,polars,extension-module
 
 jobs:
   credential-check:
@@ -143,20 +143,21 @@ jobs:
     
       - uses: actions/setup-python@v5
         with:
-          python-version: '3.10'
+          # necessary to use cibuildwheel 3.x
+          python-version: '3.11'
           cache: pip
 
       - name: Install cibuildwheel
         run: |
           python -m pip install --upgrade pip
-          python -m pip install cibuildwheel==2.19.2
+          python -m pip install cibuildwheel==3.1.4
     
       - name: Build wheels
         working-directory: python
         run: python -m cibuildwheel --output-dir dist
         env:
-          # don't build separate wheels for pypy
-          CIBW_SKIP: "pp*"
+          # don't build separate wheels for free-threaded cpython
+          CIBW_SKIP: "cp31?t-*"
           CIBW_ENVIRONMENT: 'PATH="$HOME/.cargo/bin:$PATH"'
           CIBW_ENVIRONMENT_WINDOWS: 'PATH="$UserProfile\.cargo\bin;$PATH"'
           # first find the xcode version for the mac images:
@@ -196,7 +197,7 @@ jobs:
     needs: bindings
     runs-on: ubuntu-22.04
     env:
-      DOCKER_IMAGE: messense/manylinux_2_24-cross:aarch64
+      DOCKER_IMAGE: messense/manylinux_2_28-cross:aarch64
       TARGET: aarch64-unknown-linux-gnu
     steps:
       - name: Free disk space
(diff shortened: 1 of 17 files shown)


Shoeboxam:
* **#2475** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2475?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/opendp/opendp/2475?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.github/workflows/build.yml
```diff
@@ -143,22 +143,24 @@ jobs:
...
+          python -m pip install cibuildwheel==3.1.4
     
       - name: Build wheels
         working-directory: python
         run: python -m cibuildwheel --output-dir dist
         env:
-          # don't build separate wheels for pypy
-          CIBW_SKIP: "pp*"
+          # don't build separate wheels for free-threaded cpython
+          CIBW_SKIP: "cp31?t-*"
```
Shoeboxam, line 160: pypy wheels are no longer default in cibuildwheel 3.x, so `pp*` is redundant.

.github/workflows/smoke-test.yml
```diff
@@ -116,7 +116,7 @@ jobs:
       - name: Test (all-features)
         # pyo3/extension-module must be disabled to run tests.
         # Context: https://pyo3.rs/v0.25.1/faq.html#i-cant-run-cargo-test-or-i-cant-build-in-a-cargo-workspace-im-having-linker-issues-like-symbol-not-found-or-undefined-reference-to-_pyexc_systemerror
-        run: cargo test --verbose --features=untrusted,polars,ffi,derive
+        run: cargo test --verbose --no-default-features --features=partials,use-openssl,untrusted,polars,ffi,derive
```
mccalluc, line 119: Do subsequent actions depend on artifacts from smoke-test, or something like that? No objection to this per se, but wondering if it is necessary for the full CI build?

docs/source/contributing/development-environment.rst
```diff
@@ -60,13 +60,13 @@ Substitute ``cargo build`` with ``cargo test`` to test, or ``cargo check`` to ch
 
 .. note::
 
-  If linking fails when runnning ``cargo test``, you may need to build without the ``pyo3/extension-module`` feature (`context here <https://pyo3.rs/v0.25.1/faq.html#i-cant-run-cargo-test-or-i-cant-build-in-a-cargo-workspace-im-having-linker-issues-like-symbol-not-found-or-undefined-reference-to-_pyexc_systemerror>`_):
+  If linking fails when runnning ``cargo test``, you may need to build without the ``extension-module`` feature (`context here <https://pyo3.rs/v0.25.1/faq.html#i-cant-run-cargo-test-or-i-cant-build-in-a-cargo-workspace-im-having-linker-issues-like-symbol-not-found-or-undefined-reference-to-_pyexc_systemerror>`_):
 
   .. code-block:: bash
 
-    cargo test --features=untrusted,polars,ffi,derive
+    cargo test --no-default-features --features=partials,use-openssl,untrusted,polars,ffi,derive
```
mccalluc, line 67: If it's important to keep this in sync, would it make any sense to pull this into a shell script, and reference the script here and in CI?

rust/src/core/mod.rs
```diff
@@ -281,6 +281,7 @@ where
         })
     }
 
+    #[allow(dead_code)]
```
mccalluc, line 284: Add a note to explain? Or maybe this is something to clean up in the future?

mccalluc: (same for the `allow(dead_code)` below.)

mccalluc: Or never mind: While I don't really understand this, these aren't new either.

mccalluc (approved):
approved, with some comments.

rust/src/core/mod.rs
Shoeboxam: It's just that certain functions may not be exercised under certain combinations of build flags. Instead of scraping out the exact conditional compilation flags from its call sites, I just blanket ignore the dead code, which the compiler will eliminate anyways. The risk is that if all call sites are removed, we may not realize these functions are truly dead code.

.github/workflows/smoke-test.yml
Shoeboxam: This is changed because I added `extension-module` to the default features. Part of the reasoning for doing that was because setting the extension-module feature is a no-op when pyo3 isn't enabled. Since cargo test can't have that feature enabled, I then had to adjust commands elsewhere to ensure the extension module feature isn't enabled when `polars`, `ffi` and `test` are jointly enabled.

docs/source/contributing/development-environment.rst
Shoeboxam: I feel like "test with all features but extension-module" is clearer than "test via some other shell script", even if it's a bit longer. I can see the argument for the script though.

Shoeboxam: Looks like we'll get some reprieve from this madness in the next Polars release, because it will likely allow us to use this: https://github.com/PyO3/pyo3/issues/5202

mccalluc: status quo is fine!

PR #2479, opened 2025-08-28
reformat mbi code blocks

### LINKED ISSUE #2477: Prettier doctests in `opendp.extras.mbi.html`
Each empty line is giving us a new cell. One solution is to add empty `>>>` lines (and I think we might actually check for that in the generated API docs?) but maybe it's possible to add more RST so that this is just one code block?

Image

Also, is there way to hide the pytest import? What if we put that code inside an RST comment? It will still be run by doctest.

mccalluc:
- Fix #2477
Screenshot 2025-08-28 at 3 15 02 PM

There's a handful of places where we've added bare `>>>` to avoid this. Will file a small follow-up issue to clean up those.


Shoeboxam (approved):
Thanks for the cleanup!

PR #2487, opened 2025-09-03
docs: Add missing docstring boilerplate in extras

### LINKED ISSUE #2486: Missing extra install info: `api/python/opendp.extras.polars.contingency_table.html`
Missing our boilerplate about installing extras.

Add a check that all the extra API pages have this?

mccalluc:
- Fix #2486

This got a little more complicated than I would have liked, but it should make sure the text gets added to any new modules.

### DIFF
--- a/python/test/test_docs.py
+++ b/python/test/test_docs.py
@@ -73,15 +73,15 @@ First cell with missing or misordered execution:\n{bad_sources[0].splitlines()[0
 
 def is_public_and_not_top(path: Path):
     '''
-    >>> is_public(Path('_parent/child.py'))
+    >>> is_public_and_not_top(Path('_parent/child.py'))
     False
-    >>> is_public(Path('_parent/__init__.py'))
+    >>> is_public_and_not_top(Path('_parent/__init__.py'))
     False
-    >>> is_public(Path('parent/__init__.py'))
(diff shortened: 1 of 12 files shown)


python/test/test_docs.py
```diff
@@ -70,3 +70,96 @@ def test_notebooks_are_executed(nb_path):
...
+    actual = get_docstring(py_path)
+    namespace = get_namespace(py_path)
+    if 'sklearn' in namespace:
+        extra = 'scikit-learn'
+    elif 'numpy' in namespace:
+        extra = 'numpy'
+    elif 'polars' in namespace:
+        extra = 'polars'
+    elif 'mbi' in namespace:
+        extra = 'mbi'
```
Shoeboxam, line 159: I was going to say, isn't there a dict that could be used? But module names != feature names.

mccalluc: I thought so too, but in `_lib.py` we just have this:
```
# Key: Optional module name
# Value: The opendp extra that provides this module,
#        with the appropriate version pin, if needed
install_names = {
    'sklearn': 'scikit-learn',
    'randomgen': 'numpy',
}
```
but it's not quite the same.

The test tooling is not pretty, but my sense is that it is good enough.

PR #2492, opened 2025-09-03
docs: Fill in docstrings for NamedTuple params

### LINKED ISSUE #2485: `opendp.extras.numpy.html`: `Alias for field number 0`?
A lot of "Alias for field number" is being rendered on this page: Is there a way to suppress this, or get a more useful message to render?

> T: str | RuntimeType
> 
>     Alias for field number 7
> 
> cardinalities: numpy.ndarray | None
> 
>     Alias for field number 6

etc.

mccalluc:
- Fix #2485

### DIFF
--- a/python/src/opendp/extras/sklearn/decomposition/__init__.py
+++ b/python/src/opendp/extras/sklearn/decomposition/__init__.py
@@ -37,9 +37,19 @@ class PCAEpsilons(NamedTuple):
     mean: Optional[float]
 
 
-PCAEpsilons.eigvals.__doc__ = 'ε-expenditure per changed record in the input data'
-PCAEpsilons.eigvecs.__doc__ = 'ε-expenditures per changed record in the input data'
-PCAEpsilons.mean.__doc__ = ''  # TODO
+PCAEpsilons.eigvals.__doc__ = 'ε-expenditure to estimate the eigenvalues'
+PCAEpsilons.eigvecs.__doc__ = 'ε-expenditure to estimate the eigenvectors'
+PCAEpsilons.mean.__doc__ = ''  """ε-expenditure to estimate the mean.
+
+A portion of the budget is used to estimate the mean because the OpenDP PCA algorithm 
+releases an eigendecomposition of the sum of squares and cross-products matrix (SSCP), 
+not of the covariance matrix. 
+If the data is centered beforehand (either by a prior from the user or by privately estimating the mean and then centering), 
+then PCA will correspond to the covariance matrix, as expected, 
+because the SSCP matrix of centered data is equivalent to a scaled covariance matrix.
+
+If the data is not centered (or the mean is poorly estimated), then the first eigenvector will be dominated by the true mean.
+"""
 
 
 def make_private_pca(


python/src/opendp/extras/numpy/__init__.py
```diff
@@ -205,6 +205,16 @@ class NPArray2Domain(NamedTuple):
...
 
+# Without these we get: "Alias for field number ..."
+NPArray2Domain.origin.__doc__ = 'center of the norm region'
+NPArray2Domain.norm.__doc__ = 'each row in x is bounded by the norm'
+NPArray2Domain.p.__doc__ = 'designates L`p` norm'
+NPArray2Domain.size.__doc__ = 'number of rows in data'
+NPArray2Domain.num_columns.__doc__ = 'number of columns in the data'
+NPArray2Domain.nan.__doc__ = 'whether NaN values are allowed'
+NPArray2Domain.cardinalities.__doc__ = 'cardinalities of the categorical columns'
+NPArray2Domain.T.__doc__ = 'atom type'
```
Shoeboxam, line 216: Does typing.NamedTuple subclassing allow you to put docstrings on the attributes themselves, like how dataclasses work? Maybe we should consider switching everything to dataclasses, for consistency.


```suggestion

class NPArray2Domain(NamedTuple):
    origin: numpy.ndarray | None
    """center of the L`p` ball enclosing each row"""
    norm: float | None
    """radius of the L`p` ball enclosing each row"""
    p: Literal[1, 2, None]
    """designates L`p` norm"""
    size: int | None
    """number of rows in the data"""
    num_columns: int | None
    """number of columns in the data"""
    nan: bool
    """whether NaN values are allowed"""
    cardinalities: numpy.ndarray | None
    """cardinalities of categorical columns"""
    T: str | RuntimeType
    """atom type"""
```

python/src/opendp/extras/sklearn/decomposition/__init__.py
```diff
@@ -37,6 +37,11 @@ class PCAEpsilons(NamedTuple):
     mean: Optional[float]
 
 
+PCAEpsilons.eigvals.__doc__ = 'ε-expenditure per changed record in the input data'
+PCAEpsilons.eigvecs.__doc__ = 'ε-expenditures per changed record in the input data'
+PCAEpsilons.mean.__doc__ = ''  # TODO
```
Shoeboxam:
```suggestion
PCAEpsilons.mean.__doc__ = ''  """ε-expenditure to estimate the mean.

A portion of the budget is used to estimate the mean because the OpenDP PCA algorithm 
releases an eigendecomposition of the sum of squares and cross-products matrix (SSCP), 
not of the covariance matrix. 
If the data is centered beforehand (either by a prior from the user or by privately estimating the mean and then centering), 
then PCA will correspond to the covariance matrix, as expected, 
because the SSCP matrix of centered data is equivalent to a scaled covariance matrix.

If the data is not centered (or the mean is poorly estimated), then the first eigenvector will be dominated by the true mean.
"""
```

python/src/opendp/extras/sklearn/decomposition/__init__.py
```diff
@@ -37,6 +37,11 @@ class PCAEpsilons(NamedTuple):
     mean: Optional[float]
 
 
+PCAEpsilons.eigvals.__doc__ = 'ε-expenditure per changed record in the input data'
```
Shoeboxam:
```suggestion
PCAEpsilons.eigvals.__doc__ = 'ε-expenditure to estimate the eigenvalues'
```

python/src/opendp/extras/sklearn/decomposition/__init__.py
```diff
@@ -37,6 +37,11 @@ class PCAEpsilons(NamedTuple):
...
+PCAEpsilons.eigvecs.__doc__ = 'ε-expenditures per changed record in the input data'
```
Shoeboxam:
```suggestion
PCAEpsilons.eigvecs.__doc__ = 'ε-expenditure to estimate the eigenvectors'
```

Shoeboxam (changes requested):
See comments. Thanks for cleaning this up!

python/src/opendp/extras/numpy/__init__.py
mccalluc: That doesn't seem to work: I'm still getting 
```
 |  norm
 |      Alias for field number 1
```

I'll try switching to dataclasses...

mccalluc:
- Filed #2495

... but I wouldn't let a bigger refactor get in the way of a small docs improvement.

PR #2501, opened 2025-09-04
Make the "nightly" build once a week

### LINKED ISSUE #2498: Weekly build instead of nightly?
- This would fill up PyPI 1/7th as fast
- Development pace in the fall may be pretty slow, so nightly would be redundant most of the time...
- ... and we haven't been able to respond to nightly failures promptly, in any case.

mccalluc:
- Fix #2498

This is only a proposal. Even if it goes forward, I'm not sure it's worth changing names to match.


Shoeboxam:
I surely wouldn't change the name. The benefit of this being that we'd have to clear old versions 1/7th as often? I'm ok with the current state. I also have the cleanup procedure ironed out.

PR #2504, opened 2025-09-05
Version changelog: v0.14.0

### LINKED ISSUE #2503: Make sure that CHANGELOG header after release works with script
I'm not sure if there's a problem with the script run now, or a problem with the state of the changelog, and if there's a problem with the changelog, whether it was a human problem or a software problem from the last release.

Going in to the v0.14 release, the top of the changelog looks like this:

```
# OpenDP Changelog

This file documents the version history of OpenDP. The links on each version number will take you to a comparison
showing the source changes from the previous version.


## [0.13.0-dev](https://github.com/opendp/opendp/compare/v0.12.1...HEAD) - TBD
```
- Why `-dev`?
- Why `TBD` instead of a date?

Following the instructions from this point:
```
$ python channel_tool.py changelog


# *** UPDATING CHANGELOG ***
# Reading CHANGELOG
# Updating heading to 0.13.0-dev, v0.12.1...HEAD, TBD
$ git diff
diff --git a/CHANGELOG.md b/CHANGELOG.md
index 270551f6a..e439edebe 100644
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ showing the source changes from the previous version.
 ## [0.13.0-dev](https://github.com/opendp/opendp/compare/v0.12.1...HEAD) - TBD


+
 ### Feat

 - Polars
 ```

ie, it just added an empty line. Let me try fixing the changelog by hand:
```diff
-## [0.13.0-dev](https://github.com/opendp/opendp/compare/v0.12.1...HEAD) - TBD
+## [0.13.0](https://github.com/opendp/opendp/compare/v0.12.1...HEAD) - 2025-04-22
```
... but the script just reverts this change:
```diff
-## [0.13.0](https://github.com/opendp/opendp/compare/v0.12.1...HEAD) - 2025-04-22
+## [0.13.0-dev](https://github.com/opendp/opendp/compare/v0.12.1...HEAD) - TBD
```

That doesn't seem right: the v0.13.0 release date needs to be in the log, I'd think? Discarding this change, and moving to the next step:
```
$ python changelog.py
```
That's a mess: the update is inserted twice into the log, and it looks like it is also including commits from the previous release?

---

Back up: Let's assume `channel_tool.py changelog` is working correctly, even if the result seems weird to me.
```
-## [0.13.0](https://github.com/opendp/opendp/compare/v0.12.1...HEAD) - 2025-04-22
+## [0.13.0-dev](https://github.com/opendp/opendp/compare/v0.12.1...HEAD) - TBD
```

No difference: updates are still inserted twice.

---

New plan:
- Get the changelog into pre-update state that makes sense to me.
- Update the script so it does what I think it should do.
- Update the instructions to match.

mccalluc:
- Fix #2503

The instructions didn't work for me out of the box, so I've updated the script and instructions: Not sure if this is the right direction.

One big question for me is what is `python channel_tool.py changelog` supposed to do? It's a complicated script that has different behavior depending on the environment, but it's not doing anything except adding an empty line when I run it. I've dropped it from the instructions, but not sure if that's best.

I'm going to open this PR, then do a self review and explain some of the changes. Before merging this to main, I think it would make sense to merge the couple last doc changes to main, and then merge main to this branch.


.github/workflows/README.md
```diff
@@ -142,18 +142,19 @@ on github, or with the `gh` command line tool. Parameters:
     1. Update changelog
 
         ```shell
-        python channel_tool.py changelog
```
mccalluc, line 145: Doesn't do anything except add an empty line for me.

.github/workflows/README.md
```diff
@@ -142,18 +142,19 @@ on github, or with the `gh` command line tool. Parameters:
...
-
-        git fetch --tags
```
mccalluc, line 147: Incorporated into the script, so there's one less manual step.

.github/workflows/README.md
```diff
@@ -142,18 +142,19 @@ on github, or with the `gh` command line tool. Parameters:
...
         python changelog.py
         ```
         Additional manual edits will be useful to arrange individual commits thematically.
 
     1. Update `VERSION` file on the `main` branch to match the version you want to release.
 
+        First `cd ..`, so CWD is the top level of the repo.
```
mccalluc, line 151: `channel_tool.py` assumes it is being run from the top of the repo:
```
update_file("rust/Cargo.toml"...)
```
Could file an issue to make this relative to `__file__`, so it's less fragile?

.github/workflows/README.md
```diff
@@ -142,18 +142,19 @@ on github, or with the `gh` command line tool. Parameters:
...
+
         - If it's a patch release, no change to the version number should be needed: The patch should have been incremented after the previous release.
-        - If it's a minor release, run `python channel_tool.py bump_version --position minor`
-        - Similarly, if it's a major release, run `python channel_tool.py bump_version --position major`
+        - If it's a minor release, run `python tools/channel_tool.py bump_version --position minor`
+        - Similarly, if it's a major release, run `python tools/channel_tool.py bump_version --position major`
+
+    1. Add a header for the current version to the CHANGELOG.md.
```
mccalluc, line 157: Maybe if `CHANGELOG.md` has the right shape, `tools/channel_tool.py bump_version` would re-write it? Or maybe `changelog.py` should be run last, so it can look at `VERSION` to see what the new version should be?

CHANGELOG.md
```diff
@@ -4,7 +4,95 @@ This file documents the version history of OpenDP. The links on each version num
 showing the source changes from the previous version.
 
 
-## [0.13.0-dev](https://github.com/opendp/opendp/compare/v0.12.1...HEAD) - TBD
+## [0.14.0](https://github.com/opendp/opendp/compare/v0.13.0...HEAD) - TBD
```
mccalluc, line 7: This is the line I added by hand.

tools/changelog.py
```diff
@@ -3,27 +3,44 @@
 import re
 from collections import defaultdict
 from pathlib import Path
-from channel_tool import match_first_changelog_header, get_changelog_lines
```
mccalluc, line 6: Drop import, so this can be self-contained.

tools/changelog.py
```diff
@@ -3,27 +3,44 @@
...
+    subprocess.check_output(['git', 'fetch', '--tags']) # To make sure we have tags locally.
     return subprocess.check_output(['git', 'log', f"{tag}..HEAD", '--oneline'], text=True).splitlines()
     
 
 
-def get_changelog_update(lines):
+def reformat_log(lines):
     '''
-    >>> print(parse_log([
+    >>> print('\n'.join(reformat_log([
```
mccalluc: Doctests weren't being run here: the name changed at some point and the test wasn't updated to match.

tools/changelog.py
```diff
@@ -52,27 +69,32 @@ def get_changelog_update(lines):
...
     parser.parse_args()
 
-    old_changelog_lines = (Path(__file__).parent.parent / 'CHANGELOG.md').read_text().splitlines()
-    new_changelog_lines = []
+    old_lines = get_changelog_lines()
 
-    prev_version = get_prev_version()
-    log_lines = log_until(prev_version.replace('-dev', ''))
-    changelog_update = get_changelog_update(log_lines)
+    (i, prev_version) = get_prev_version(old_lines)
```
mccalluc, line 91: Alternatively, we could get read `VERSION`.

.github/workflows/README.md
Shoeboxam: There should be a patch update to VERSION and corresponding new header in the changelog after making a release. When making the next release, you can bump minor if necessary, or otherwise cut a patch without bumping.

Shoeboxam (approved):
`python channel_tool.py changelog` should write the new header, and it won't do anything (or maybe just leave a newline), if the changelog version is up-to-date with the current VERSION. Maybe the changelog should be moved to after the version bump? Not sure.

These changes look good, but it would be good to look back into simplifying the changelog update code in follow-up work.

PR #2506, opened 2025-09-05
Fix github CI syntax: only run the tests we need

### LINKED ISSUE #2505: In test matrix `[mbi]` might not be running the right tests
https://github.com/opendp/opendp/actions/runs/17498584764/job/49706586928

This should be a test of `3.10, .[mbi]`, but the pytest run is skipping a lot!

```
pytest -v -k "not lazyframe and not polars and not pca and not rst and not contingency and not synthetic"
```

The `mbi` run should have no exclusions: It's the most all-encompassing extra.

mccalluc:
- Fix #2505
- We were running extra tests, I think, because string expressions just evaluated to true
- But rather than using a lot of CI `if` checks, just put everything in one case statement: The closer we can keep it to plain shell syntax, the easier it will be to maintain.
- Use `==` instead of `contains`: Less likely to be surprised by a partial match.
- Pull out just the name of the extra, and use a dummy `PLACEHOLDER` when we don't want any extras: Less clutter.
- Add `numpy` to the test matrix: I'm not sure the extras need to be this fine grained, but if we offer an extra, I think we're on the hook for testing it.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -185,13 +185,13 @@ jobs:
         # (Even in unittests we should use "importskip" rather than "skipif".)
         run: |
           case "${{ matrix.extra-install }}" in
-            PLACEHOLDER)          pytest -v -k "not contingency and not synthetic and not lazyframe and not polars and not pca and not rst"
-            numpy)                pytest -v -k "not contingency and not synthetic and not lazyframe and not polars and not pca and not rst"
-            scikit-learn)         pytest -v -k "not contingency and not synthetic and not lazyframe and not polars and not pca and not rst"
-            polars)               pytest -v -k "not contingency and not synthetic"
-            mbi)  coverage run -m pytest -v
+            PLACEHOLDER)          pytest -v -k "not contingency and not synthetic and not lazyframe and not polars and not pca and not rst";;
+            numpy)                pytest -v -k "not contingency and not synthetic and not lazyframe and not polars and not pca and not rst";;
+            scikit-learn)         pytest -v -k "not contingency and not synthetic and not lazyframe and not polars and not pca and not rst";;
+            polars)               pytest -v -k "not contingency and not synthetic";;
+            mbi)  coverage run -m pytest -v;;
 
-            *)    print "unexpected extra-install"; exit 1
+            *)    print "unexpected extra-install"; exit 1;;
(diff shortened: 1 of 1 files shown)


.github/workflows/smoke-test.yml
```diff
@@ -177,37 +178,37 @@ jobs:
...
         # There's not a "skipif" for doctests, so use "-k" to exclude based on name patterns.
         # (Even in unittests we should use "importskip" rather than "skipif".)
-        run: pytest -v -k "not lazyframe and not polars and not pca and not rst and not contingency and not synthetic"
-
-      - name: Test Polars
-        if: contains(matrix.extra-install, 'polars')
-        run: pytest -v -k "not contingency and not synthetic"
+        run: |
+          case "${{ matrix.extra-install }}" in
+            PLACEHOLDER)          pytest -v -k "not contingency and not synthetic and not lazyframe and not polars and not pca and not rst";;
```
Shoeboxam, line 188: I think you can do `""` here and use an empty string above too.
```suggestion
            "")          pytest -v -k "not contingency and not synthetic and not lazyframe and not polars and not pca and not rst";;
```

Shoeboxam (changes requested):
Do my suggestions work? I think an empty string would be a bit clearer.

.github/workflows/smoke-test.yml
mccalluc: The pip install fails if this is just an empty string:
```
$ pip install -e '.[]'
...
ERROR: .[] is not a valid editable requirement. It should either be a path to a local project or a VCS URL (beginning with bzr+http, bzr+https, bzr+ssh, bzr+sftp, bzr+ftp, bzr+lp, bzr+file, git+http, git+https, git+ssh, git+git, git+file, hg+file, hg+http, hg+https, hg+ssh, hg+static-http, svn+ssh, svn+http, svn+https, svn+svn, svn+file).
```
vs.
```
$ pip install -e '.[not_actually_an_extra]'
...
WARNING: opendp 0.14.0.dev0 does not provide the extra 'not-actually-an-extra'
...
Successfully installed opendp-0.14.0.dev0
```

but I'll grant that `PLACEHOLDER` is weird. I thought the `case` might be easier to work with, but happy to just get the `${{...}}` working correctly, if you prefer.

PR #2513, opened 2025-09-09
last steps on v0.14 release

mccalluc:
This is the result of the last steps of the documented release process. Hopefully this is what it should look like at this point?


Shoeboxam (approved):
Yep, now the changelog reads properly. To cut a minor patch, then just increment minor, which resets patch to zero.

PR #2529, opened 2025-09-13
docs: update Typical Workflow

### LINKED ISSUE #2528: Update Typical Workflow

Shoeboxam:
Closes #2528


Shoeboxam:
* **#2529** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2529?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/opendp/opendp/2529?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

docs/source/getting-started/quickstart.rst
```diff
@@ -45,7 +45,7 @@ Enable ``contrib`` globally with the following snippet:
 
 .. tab-set::
 
-    .. tab-item:: Python
+    .. tab-item:: Python (Context API)
```
mccalluc, line 48: Not absolutely blocking this, but we need to think about what this looks like for an absolutely new user: They don't know anything about OpenDP at this point, and the first thing we do is throw two different APIs at them. Are they supposed to look at both?

If the framework API is something they don't even need to worry about, I'm fine replacing it with a Polars example; I'm just really not sure about showing both.

docs/source/getting-started/typical-workflow.rst
```diff
@@ -236,27 +248,37 @@ Here's a differentially private mean:
...
+            :start-after: sum
+            :end-before: /sum
+
+Altogether, the data flow of this analysis looks like:
+
+.. Diagram source: https://docs.google.com/drawings/d/1W4l9x3UM3hbVLWlC0nzijqgaQ31wY5ERebp8jkYy1yc/edit
+
+.. image:: code/typical-workflow-diagram.svg
+    :width: 100%
+    :alt: Diagram representing typical data flow with OpenDP, raw data to differentially private releases.
```
mccalluc, line 277: Not an absolute, but I liked a diagram up front, as an outline of what is to follow. You may be giving it a different role?

docs/source/getting-started/typical-workflow.rst
```diff
@@ -5,15 +5,8 @@ A differentially private analysis in OpenDP typically has the following steps:
...
-4. Mediate access to data
-5. Submit DP queries
-
-.. Diagram source: https://docs.google.com/drawings/d/1W4l9x3UM3hbVLWlC0nzijqgaQ31wY5ERebp8jkYy1yc/edit
-
-.. image:: code/typical-workflow-diagram.svg
-    :width: 60%
-    :alt: Diagram representing typical data flow with OpenDP, from a raw CSV to a differentially private release. 
+3. Mediate access to data
+4. Submit DP queries (examples: count and sum)
```
mccalluc, line 9: At this point in the docs, I think we would do well to keep it simple. What is the benefit of doing count and sum? Could it be left to a later section?

docs/source/getting-started/code/typical-workflow-diagram.svg
mccalluc, line 1: With the old one, the round-rect identified what happened inside OpenDP. Not sure how the rects on the right relate to that on the left.

(But before spending too much time here, let's make sure we're on the same page about the advantages of showing both mean and count here.)

mccalluc (commented):
Before commenting more, could we talk about the direction here? If we add complexity here, I want to make sure it's something that will improve user understanding.

Shoeboxam:
We're taking this documentation in a different direction anyways!

PR #2533, opened 2025-09-14
docs: randomized response bitvec

### LINKED ISSUE #2532: Document bit-vector randomized response

Shoeboxam:
Close #2532


R/opendp/NAMESPACE
```diff
@@ -30,7 +30,6 @@ export(bitvector_domain)
 export(bool)
 export(change_one_distance)
 export(choose_branching_factor)
-export(debias_randomized_response_bitvec)
```
mccalluc, line 33: There's a note at the top saying:
```
# Generated by roxygen2: do not edit by hand
```
If this is getting out of date, maybe it should be gitignored?

docs/source/api/user-guide/measurements/randomized-response.rst
```diff
@@ -354,3 +352,61 @@ Since the dataset size is known, simply post-process the mean estimates:
...
+
+    .. tab-item:: Python
+        :sync: python
+
+        .. code:: pycon
+
+            >>> frequencies = dp.m.debias_randomized_response_bitvec(
+            ...     [m_rr(data) for _ in range(10_000)],
+            ...     f=0.8,
+            ... )
```
mccalluc, line 412: Doctest to at least show the shape of `frequencies`, even if we don't want to show the actual, random, results?

mccalluc: Would it be good to have some conclusion on this in page in English after the code sample? Not a requirement.

R/opendp/NAMESPACE
Shoeboxam: Unfortunately it can't be gitignored because it needs to exist to bootstrap the build process.

docs/source/api/user-guide/measurements/randomized-response.rst
Shoeboxam: I'm not sure what else to add that would be useful to the reader.

PR #2534, opened 2025-09-14
fix: Python/R -> Rust API docs links

### LINKED ISSUE #1025: Python API Measures: Links from Python to Rust are broken

Shoeboxam:
Close #1025

This introduces a mechanism to manually set the destination in generated links. We'll still need to be vigilant not to introduce breaking links in the future.

### DIFF
--- a/rust/opendp_derive/src/lib.rs
+++ b/rust/opendp_derive/src/lib.rs
@@ -65,6 +65,16 @@ mod full;
 ///     pass
 /// ```
 ///
+/// # Rust Link
+/// Set to the expected path of Rust documentation, if it can't be correctly inferred.
+///
+/// ```compile_fail
+/// #[bootstrap(rust_path = "domains/struct.AtomDomain")]
+/// fn my_func() {}
+/// ```
+///
+/// This gives a manual way to fix dead links to rust documentation in bindings language documentation.
+///
 /// # Arguments, Generics and Return
 /// You can pass additional metadata that is specific to each argument or generic.
 ///
--- a/rust/opendp_tooling/src/bootstrap/arguments.rs
+++ b/rust/opendp_tooling/src/bootstrap/arguments.rs
@@ -14,6 +14,8 @@ pub struct BootstrapArguments {
     pub name: Option<String>,
     pub proof_path: Option<String>,
     #[darling(default)]
+    pub rust_path: Option<String>,
+    #[darling(default)]
     pub has_ffi: Option<bool>,
     #[darling(default)]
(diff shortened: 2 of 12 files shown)


Shoeboxam:
* **#2534** <a href="https://app.graphite.dev/github/pr/opendp/opendp/2534?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/opendp/opendp/2534?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mccalluc (commented):
I'm uncomfortable with this. Maybe if we can pin down what "vigilance" means:
- Should reviewers be checking that links work? In that case, it would almost be better to make the path a required argument, so they'll be reminded to check it.
- Could we get automated link checking working? Then CI would do the vigilance for us. I could try to debug it: cargo.io is currently in the ignore list in `docs/linkchecker-external.cfg`:
```
ignore=
  # External:
  # crates.io returns 404 to curl and wget, but works in browser.
  # JS sniffing is involved? Easiest just to leave this ignore in place.
  https://crates.io/crates/opendp
```
- There don't seem to be that many special cases: could a couple fixes be made in the code generation?

Not an absolute block, but several concerns.

Shoeboxam:
One reason why this doesn't currently work is because links for new APIs won't exist yet on the docs website. There's no way to distinguish between not-yet-there links and broken links. I could see this working for the weekly link-check, though, which could test nightly links.

At the same time, the heuristic used to make links works in 99% of cases, so manually adding links and maintaining them is a lot of extra work. 

In the cases where manual fixes are necessary, I'm linking to completely different APIs. Alternatively, I could suppress the link altogether, but I think it's more meaningful to send EG `atom_domain` to the `AtomDomain` struct.

PR #2541, opened 2025-09-17
docs: Fill in migration notes back to v12

### LINKED ISSUE #2512: Document upgrade path for v0.14
Taking DP Wizard as an example of a downstream application, [CI is failing](https://github.com/opendp/dp-wizard/actions/runs/17563676795/job/49885811790) because it is using `public_info`. The CHANGELOG here doesn't mention that field. We need to make it easy for current users to migrate, and we need to be able to summarize the migrations necessary without depending too much on our fallible memories.

There is a stub for `public_info` in the [nightly build docs](https://docs.opendp.org/en/nightly/api/python/opendp.extras.polars.html#opendp.extras.polars.Margin.public_info) -- Should there be a deprecation note there?

mccalluc:
- Copy notes from blog
- Fix #2512
- v11 does have migration notes!
- Even older versions don't have migration notes, but I don't think we need to go so far back.


CHANGELOG.md
```diff
@@ -123,6 +142,10 @@ showing the source changes from the previous version.
 
 ## [0.13.0](https://github.com/opendp/opendp/compare/v0.12.1...v0.13.0) - 2025-04-22
 
+### Migration
+
+- We now default to the assumption that NaN values exist in float data in all settings. You’ll notice that you now need to specify nan=Falsenan=False in atom domains, [even when building the Laplace or Gaussian measurements](https://docs.opendp.org/en/v0.13.0/api/user-guide/measurements/additive-noise-mechanisms.html#Distribution:-Laplace-vs.-Gaussian).
```
Shoeboxam: You've got duplicate `nan=False` here.

CHANGELOG.md
```diff
@@ -224,6 +247,14 @@ showing the source changes from the previous version.
 
 ## [0.12.0](https://github.com/opendp/opendp/compare/v0.11.1...v0.12.0) - 2024-12-19
 
+### Migration
+
+Previously, OpenDP allowed arbitrary functions to be used as postprocessors. Unfortunately this could be abused to build measurements whose postprocessors introduce vulnerabilities. For instance, a post-processor that reads the system time could be used to add timing information to the output of a mechanism. OpenDP now requires `honest-but-curious` to be enabled in order to build postprocessors from Python, and states necessary conditions for valid postprocessors.
+
+Also:
+- Rename `SMDCurve` to `PrivacyProfile`
+- Remove argument for distance type on privacy measures: `max_divergence(T=float)` is now just `max_divergence()`
```
Shoeboxam, line 308: - `SMDCurve` represented the function `ε(δ)`, and the new `PrivacyProfile` represents the function `δ(ε)`.

Shoeboxam (changes requested):

CHANGELOG.md
mccalluc:
```suggestion
- We now default to the assumption that NaN values exist in float data in all settings. You’ll notice that you now need to specify `nan=False` in atom domains, [even when building the Laplace or Gaussian measurements](https://docs.opendp.org/en/v0.13.0/api/user-guide/measurements/additive-noise-mechanisms.html#Distribution:-Laplace-vs.-Gaussian).
```

CHANGELOG.md
```diff
@@ -224,6 +247,14 @@ showing the source changes from the previous version.
 
 ## [0.12.0](https://github.com/opendp/opendp/compare/v0.11.1...v0.12.0) - 2024-12-19
 
+### Migration
+
+Previously, OpenDP allowed arbitrary functions to be used as postprocessors. Unfortunately this could be abused to build measurements whose postprocessors introduce vulnerabilities. For instance, a post-processor that reads the system time could be used to add timing information to the output of a mechanism. OpenDP now requires `honest-but-curious` to be enabled in order to build postprocessors from Python, and states necessary conditions for valid postprocessors.
+
+Also:
+- Rename `SMDCurve` to `PrivacyProfile`
```
mccalluc:
```suggestion
- Replace `SMDCurve`, which represented the function ε(δ), with `PrivacyProfile`, which represents the function δ(ε).
```

CHANGELOG.md
```diff
@@ -123,6 +142,10 @@ showing the source changes from the previous version.
 
 ## [0.13.0](https://github.com/opendp/opendp/compare/v0.12.1...v0.13.0) - 2025-04-22
 
+### Migration
+
+- We now default to the assumption that NaN values exist in float data in all settings. You’ll notice that you now need to specify `nan=False` in atom domains, [even when building the Laplace or Gaussian measurements](https://docs.opendp.org/en/v0.13.0/api/user-guide/measurements/additive-noise-mechanisms.html#Distribution:-Laplace-vs.-Gaussian).
+- The recommended format for margins in the Context API is now a list of margins containing group by keys, instead of a dictionary of group by keys and margins. An example of the new syntax [can be found here](https://docs.opendp.org/en/v0.13.0/api/user-guide/polars/gaussian-noise.html#Max-Partition-Contributions). This allows group by keys to be arbitrary expressions.
```
TedTed: the tiniest nit: "group by keys" parses weirdly to me, I would go with "group-by keys" so it's clear it's not a verb

CHANGELOG.md
```diff
@@ -224,6 +247,14 @@ showing the source changes from the previous version.
 
 ## [0.12.0](https://github.com/opendp/opendp/compare/v0.11.1...v0.12.0) - 2024-12-19
 
+### Migration
+
+Previously, OpenDP allowed arbitrary functions to be used as postprocessors. Unfortunately this could be abused to build measurements whose postprocessors introduce vulnerabilities. For instance, a post-processor that reads the system time could be used to add timing information to the output of a mechanism. OpenDP now requires `honest-but-curious` to be enabled in order to build postprocessors from Python, and states necessary conditions for valid postprocessors.
```
TedTed: you spell it "postprocessor" everywhere except once where it's "post-processor"

as a side-note, marking postprocessors as potentially unsafe is a bit of a weird choice to me. what's the difference with executing the thing and then figuring out how much time it took from the outside of the framework, compared to putting that logic in the postprocessor?

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ showing the source changes from the previous version.
 
 ## [0.14.1](https://github.com/opendp/opendp/compare/v0.14.0...v0.14.1) - 2025-09-16
 
+### Migration
+
+- `fill_null` and `fill_nan` are no longer needed, unless you need non-default behavior.
```
TedTed: I don't understand what this means

CHANGELOG.md
mccalluc: Will do, but I think there are other instances where "group by" might be switched "group-by", so it reads as a noun.

- "group by truncation" -> "group-by truncation"
- "via a group by" -> "via a group-by"
- "group by in logical plan" -> "group-by in logical plan"

Don't want to get into those changes here, but raise an issue from this comment if that sounds like the right direction? (I don't care strongly myself.)

CHANGELOG.md
mccalluc: Took out the dash here and in a few other places.

Timing attacks seem contrived, but I think it does make sense to keep `honest-but-curious`. We always flag user code: If an adversary can run arbitrary postprocessing, they could just read from the filesystem and release information in the clear! But timing attacks are the least of our worries.

I'll revise the explanation.

mccalluc:
- #2589 should fix CI

CHANGELOG.md
mccalluc: I'll revise. The midpoint of the upper and lower bounds is now used for imputation by default.

CHANGELOG.md
mccalluc: - Filed #2601

Shoeboxam (dismissed):
Thanks!

Shoeboxam (approved):

PR #2557, opened 2025-10-08
fix: support building stable truncation transformations from Python

Shoeboxam:


python/test/polars/test_truncation.py
```diff
@@ -0,0 +1,213 @@
...
+        context.query()
+        .truncate_per_group(2)
+        .truncate_num_groups(1, keep=keep, by=["alpha"])
+        .group_by("alpha")
+        .agg(dp.len())
+    )
+    assert query.summarize()["scale"][0] == 2.0000000000000004  # type: ignore[index]
+
+
+def test_truncation_contingency():
```
Shoeboxam, line 193: Only new test.

python/src/opendp/context.py
```diff
@@ -1077,7 +1077,7 @@ def _normalize_compositor(
     elif split_by_weights is not None:
         # TODO: Wrap d_out in an object that defines __mul__,
         # and fix the typing on the signature.
-        if isinstance(d_out, float):
+        if isinstance(d_out, (int, float)):
```
Shoeboxam, line 1080: Now you can use int privacy losses with weights...

rust/src/transformations/make_stable_lazyframe/group_by/mod.rs
```diff
@@ -304,6 +304,7 @@ fn check_infallible_function(
         FunctionExpr::Boolean(bool_expr) => match bool_expr {
             BooleanFunction::Any { .. } => check_inputs!(aggregate),
             BooleanFunction::All { .. } => check_inputs!(aggregate),
+            BooleanFunction::IsClose { .. } => check_inputs!(),
```
Shoeboxam, line 307: Added in Polars 0.50. This was in pursuit of fixing an unrelated json schema check in my dev tooling.

rust/src/transformations/make_stable_lazyframe/h_stack/mod.rs
```diff
@@ -59,10 +59,11 @@ where
         let names = (t_exprs.iter())
             .map(|t| t.output_domain.column.name.clone())
             .collect::<HashSet<_>>();
-        if !names.is_disjoint(&HashSet::from_iter(identifier.meta().root_names())) {
+        let root_names = HashSet::from_iter(identifier.meta().root_names());
+        if !names.is_disjoint(&root_names) {
             return fallible!(
                 MakeTransformation,
-                "identifiers ({names:?}) may not be modified"
+                "identifiers ({root_names:?}) may not be modified"
```
Shoeboxam, line 66: The error message need not include non-identifier column names.

rust/Cargo.toml
```diff
@@ -157,9 +158,7 @@ polars = { version = "=0.50.0", features = [
     "string_normalize",
 ], optional = true }    
 polars-plan = { version = "=0.50.0", features = ["ffi_plugin", "python"], optional = true }
-polars-utils = { version = "=0.50.0" }
-# pinned due to: https://github.com/pola-rs/polars/issues/24208
-planus = { version = "=1.1.1" }
+polars-utils = { version = "=0.50.0", optional = true }
```
Shoeboxam, line 161: Marking this as optional so that Polars crates aren't pulled in when building without Polars.

rust/Cargo.toml
```diff
@@ -157,9 +158,7 @@ polars = { version = "=0.50.0", features = [
     "string_normalize",
 ], optional = true }    
 polars-plan = { version = "=0.50.0", features = ["ffi_plugin", "python"], optional = true }
-polars-utils = { version = "=0.50.0" }
-# pinned due to: https://github.com/pola-rs/polars/issues/24208
-planus = { version = "=1.1.1" }
```
Shoeboxam, line 162: Issue is closed now, fixed by Polars itself by pushing a patch release on a transitive dependency.

python/src/opendp/extras/mbi/_table/__init__.py
```diff
@@ -345,6 +345,7 @@ def postprocess(qbl: Queryable) -> ContingencyTable:
                 .clip(0, len(stable_keys[c]) - 1)
                 for c in input_domain.columns
             ),
+            MO="FrameDistance<SymmetricDistance>"
```
Shoeboxam, line 348: The API always assumed MO := MI. Now you can state otherwise.

python/test/polars/test_truncation.py
```diff
@@ -0,0 +1,213 @@
+import opendp.prelude as dp
+import polars as pl
+import pytest
+import re
+
+
+def test_lazyframe_bounded_dp_truncation():
```
Shoeboxam, line 6: These tests (except for the last) are moved from the (too big) Polars test file.

rust/Cargo.toml
```diff
@@ -84,6 +84,7 @@ polars = { version = "=0.50.0", features = [
     "is_last_distinct",
     "is_unique",
     "is_between",
+    "is_close",
```
Shoeboxam, line 87: For completeness with Polars 0.50's Python feature list.

python/src/opendp/context.py
mccalluc: Add a test with ints?

python/src/opendp/typing.py
```diff
@@ -494,21 +494,21 @@ def get_carrier_type(domain: Domain) -> Union[RuntimeType, str]:
...
+    Returns the type for a supporting element.
 
-    :param domain: A domain
+    :param element: A supporting element
     '''
-    return domain.type
+    return element.type
 
-def get_value_type(value: Union[Metric, Measure]):
+def get_value_type(type_descriptor: RuntimeTypeDescriptor):
```
mccalluc, line 505: The behavior isn't changing, but is this type signature over-promising? Just looking at what `parse()` could return:
```
       >>> dp.RuntimeType.parse(int)
        'i32'
        >>> dp.RuntimeType.parse("i32")
        'i32'
        >>> dp.RuntimeType.parse(L1Distance[int])
        L1Distance<i32>
        >>> dp.RuntimeType.parse(L1Distance["f32"])
        L1Distance<f32>
```
If this returned a str, then `.args[1]` doesn't make sense.

Not a hard block, and not sure what the alternative would be.

python/test/polars/test_truncation.py
```diff
@@ -0,0 +1,213 @@
...
+        split_evenly_over=1,
+    )
+
+    synth_query = (
+        synth_context.query()
+        .truncate_per_group(1)
+        .select("ILOSTAT")
+        .contingency_table(
+            keys={"ILOSTAT": []},
+            algorithm=dp.mbi.Fixed(queries=[dp.mbi.Count(("ILOSTAT",))]),
```
mccalluc, line 211: Just to keep the test simpler, could this be left off? In the dp-wizard example I was just specifying `cuts`.

rust/src/transformations/make_stable_lazyframe/ffi.rs
```diff
@@ -37,49 +44,67 @@ pub extern "C" fn opendp_transformations__make_stable_lazyframe(
...
-            DslPlan: StableDslPlan<MI, FrameDistance<MI>>,
-        {
-            let input_metric = input_metric.downcast_ref::<MI>()?.clone();
-            make_stable_lazyframe(input_domain, input_metric, lazyframe).into_any()
+        // FrameDistance<SymmetricIdDistance> -> FrameDistance<SymmetricDistance>
+        } else if MI == Type::of::<FrameDistance<SymmetricIdDistance>>() && MO == Type::of::<FrameDistance<SymmetricDistance>>() {
+            let input_metric = try_!(input_metric.downcast_ref::<FrameDistance<SymmetricIdDistance>>()).clone();
+            make_stable_lazyframe::<_, FrameDistance<SymmetricDistance>>(input_domain, input_metric, lazyframe).into_any()
+        } else {
+            err!(FFI, "MI ({:?}) must match MO ({:?}), or FrameDistance<SymmetricIdDistance> -> FrameDistance<SymmetricDistance>", MI.descriptor, MO.descriptor).into()
```
mccalluc, line 76: If we wanted to be scrupulous about test coverage, we could make sure we can reach this error message and the similar one below.

python/src/opendp/typing.py
Shoeboxam: The input needs to be generic (like a dict type descriptor mentioned in the docstring).

python/test/polars/test_truncation.py
Shoeboxam: It makes things run fast! I left a comment to that effect.

PR #2579, opened 2025-12-01
feat: Allow empty with_keys

### LINKED ISSUE #2573: Could/should empty `with_keys` LazyFrame still work?
For example:
```python
import opendp.prelude as dp
import polars as pl
dp.enable_features("contrib")

privacy_unit = dp.unit_of(contributions=1)

privacy_loss = dp.loss_of(
    epsilon=1,
    delta=1 / max(1e7, 100000),
)

stats_context = dp.Context.compositor(
    data=pl.scan_csv('/Users/chuckmccallum/github/opendp/dp-wizard/tests/fixtures/abc.csv', encoding="utf8-lossy")
        .with_columns(
            pl.col('B')
            .cut([1,2,3]) 
            .alias('b_bin')
            .cast(pl.String)),
    privacy_unit=privacy_unit,
    privacy_loss=privacy_loss,
    split_by_weights=[1],
)

groups = ['b_bin']
b_query = (
    stats_context.query()
    .group_by(groups)
    .agg(pl.len().dp.noise().alias("count"))
    .with_keys(pl.LazyFrame({}))
)
b_query.release().collect()
```
abc.csv:
```
A,B,C,D,E,F,G
fake,2,3,4,5,6,7
```
Error:
```
OpenDPException: 
  FailedFunction("InvalidOperation(ErrString("expected join keys/predicates\n\nResolved plan until failure:\n\n\t---> FAILED HERE RESOLVING 'sink' <---\nAGGREGATE[maintain_order: false]\n  [len().lib/python3.10/site-packages/opendp/lib/opendp.abi3.so:noise_plugin().alias(\"count\")] BY [col(\"b_bin\")]\n  FROM\n   WITH_COLUMNS:\n   [col(\"B\").cut().strict_cast(String).alias(\"b_bin\")] \n    Csv SCAN [/Users/chuckmccallum/github/opendp/dp-wizard/tests/fixtures/abc.csv]\n    PROJECT */7 COLUMNS"))")
```

Comment out `.with_keys(pl.LazyFrame({}))` and it runs without error.

Naively, I was hoping that supplying an empty frame would be the same as not supplying `with_keys` at all. Motivated by DP Wizard, but maybe more generally useful?

Glancing at the definition of `with_keys`, if this isn't just a bad idea, I think I could make the PR.

mccalluc:
- Fix #2573
- although the preferred solution would be upstream in rust.

### DIFF
--- a/python/src/opendp/extras/polars/__init__.py
+++ b/python/src/opendp/extras/polars/__init__.py
@@ -854,6 +854,9 @@ class LazyFrameQuery:
         if on is None:
             on = keys.collect_schema().names()
 
+        if len(on) == 0:
+            return self
+
         return LazyFrameQuery(
             keys.join(self.polars_plan, how="left", on=on, nulls_equal=True),
             self._query,
--- a/python/test/test_polars.py
+++ b/python/test/test_polars.py
@@ -1002,7 +1002,7 @@ def test_no_grouping_keys_context_error():
     )):
         query.release().collect()
 
-def test_no_grouping_keys_context_success():
+def test_no_grouping_keys_context():
     # Based on generated code from DP Wizard:
     pl = pytest.importorskip("polars")
     privacy_unit = dp.unit_of(contributions=1)
@@ -1012,7 +1012,8 @@ def test_no_grouping_keys_context_success():
         delta=1 / max(1e7, 100000),
     )
 
-    lf = pl.LazyFrame({"B": [1, 2, 3]})
+    _, series = example_series()
+    lf = pl.LazyFrame(series)
 
     stats_context = dp.Context.compositor(
         data=lf.with_columns(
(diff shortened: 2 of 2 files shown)


Shoeboxam (commented):
I feel more doubtful about this PR now that I've thought about it more. If on is empty, shouldn't an error be raised instead of silently eating delta? And in the DP-Wizard output, isn't providing an empty key-set to with_keys a bit misleading/hard to interpret?

mccalluc:
@Shoeboxam :

>  If on is empty, shouldn't an error be raised instead of silently eating delta?

I might be misunderstanding, but I'd prefer no error, though a warning would be welcome. If the user provides delta for a calculation that doesn't require delta, it's not great, but it doesn't seem that different from all the other ways that the user can do something inefficient. A silent inefficiency seems different than a silent error.

I'm also thinking of opendp as a library that might be near the bottom of the stack. If the inputs are typed correctly, I'd like to avoid runtime errors, so that downstream users (particularly end users of products that build on opendp) don't have to parse errors that relate to internals they don't control.

> And in the DP-Wizard output, isn't providing an empty key-set to with_keys a bit misleading/hard to interpret?

Rather than thinking of the generated notebook as the optimal solution for any particular problem, is it helpful to think of it as an example that is showing the options which are available, even if not all of them are being used?

That said, I don't care strongly about this particular solution, and it might be better addressed in Rust. If you want to close this PR and move discussion back to the issue, that's fine.

mccalluc:
Not the right direction: just an opt for dp-wizard. Closing!

PR #2581, opened 2025-12-04
docs: explain trade-offs in data loading

### LINKED ISSUE #2580: Justify `ignore_errors=True`
Should we add a little comment to https://docs.opendp.org/en/stable/getting-started/tabular-data/essential-statistics.html?

This is not the default, so justification would be good. (DP Wizard didn't have `ignore_errors=True`, but I'm adding it now, and I want to explain why.)

suggestion:
> We'll consistently use `ignore_errors=True`: Polars infers types from the first rows of a CSV, but if a later value can not be parsed as the inferred type the default behavior is to error. That's not the behavior we want: Runtime errors that depend on a single value would leak information and violate the DP guarantee.

If a large number of values are dropped, that would bias the results, but not sure we want to get into that?

mccalluc:
- Towards #2580

... but we don't actually want to encourage people to use `ignore_errors=True`, not so much because it is wrong, but because providing any kwargs sets a precedent. Users need to think about what data loading regime makes the most sense for each particular case. So: keep that issue open, but retitle and instead try to remove usage of `ignore_errors` across the docs.


docs/source/getting-started/tabular-data/essential-statistics.rst
```diff
@@ -56,10 +56,11 @@ introduction <index.rst>`__.
...
 .. note::
-    In practice, it is recommended to only ever create one Context that
-    spans all queries you may make on your data. However, to more clearly
-    explain the functionality of the library, the following examples do
-    not follow this recommendation.
+    We'll consistently use ``ignore_errors=True``:
+    Polars infers types from the first rows of a CSV, but without this setting,
+    if a later value can not be parsed as the inferred type it will error.
+    That's not the behavior we want:
+    Runtime errors would leak information and violate the DP guarantee.
```
Maegereg: Per the discussion this morning, does this really make sense? First, all of these examples are written in a context where the reader has direct access to the private data. And second, we can't really extend the DP guarantee to cover reading in data.

I'd suggest re-phrasing this to say that the process of reading in data is not covered by the DP guarantee, and then we can suggest `ignore_errors=True` if we still think it has benefits. I will note that I think `ignore_errors=True` is not always the right choice: imagine that I'm a trusted curator using OpenDP to release tabular statistics to the public. I'm reading in data from a CSV file, and it has a string column, but where the first `infer_scheme_length` entries can be parsed as floats. If I set `ignore_errors=True`, it'll incorrectly determine the column type, read in the data, and I may never notice that something is wrong. Then, if I release the results, at best I'll have wasted budget on invalid data, and a worse I'll have leaked information about the first set of rows. I would much rather get an error when reading in the data so I can correct it before sending data out the door.

I might suggest instead noting that:
- Inferring the schema sometimes gets it wrong, and can produce a schema that doesn't match the dataset. This is data-dependent.
- It's worth thinking about how you want to handle rows that don't match the schema - do you want errors, so any problems can be corrected, or do you want to ignore malformed rows? If you inferred the schema, rows that don't match it may be a sign it was inferred incorrectly.
- This whole process leaks information about the contents of the dataset and should be done by a trusted curator. If you are concerned about data leakage, you should manually specify your schema and ignore rows that don't match it.

Also, one other suggestion: it might be useful to separate out the `scan_csv` call from the context initialization in the example, just to highlight that this is not a privacy-preserving operation.

mccalluc: All good points: I had made this before the dev meeting, but I should revisit it now. Thanks!

mccalluc:
Revised! 
- Going to ping Mike, to make sure this seems good to him.
- I'm going to change this back to draft, with the idea that if this is the right thing for the tutorial, we should remove `ignore_errors=True` everywhere in the docs, and make those changes in this PR as well.

(For DP Wizard on the other hand, I think [we should keep `ignore_errors=True`](https://github.com/opendp/dp-wizard/pull/737/commits/fd40985acc19da0afffaefb5c9d5976508a55668), because that is exactly a case where we don't want to bother the user for a schema... but if the inferred schema is wrong, we also don't want a run-time error, but the code can explain the risks of that approach... but if that is not the right call, please comment in the dp-wizard repo.)

This is in flux, and if more thought makes us change our approach here, that's not a bad thing.

docs/source/getting-started/tabular-data/essential-statistics.rst
Maegereg: I think this looks much better now!

Maegereg:
One other suggestion: in cases where we're inferring the schema, it might be a good idea to also explicitly set `infer_schema_length` to something reasonably large. It doesn't prevent DP violations, but it does prevent a single bad record from messing up the schema inference.

docs/source/getting-started/tabular-data/essential-statistics.rst
```diff
@@ -45,22 +45,29 @@ introduction <index.rst>`__.
...
+    Typically, data loading will be outside the scope of the privacy guarantee.
+    While it's possible for a CSV's column names and data types to contain
+    private information (perhaps because of a one-hot encoding, or pivot operation),
+    those aren't scenarios we'll worry about here.
+
+    By default, polars will infer the schema from the first rows of a CSV.
+    Other options include:
+
+    * If you know the column types ahead of time, you may use the ``schema`` kwarg;
+    * Alternatively, if you need to avoid data dependent errors, the ``ignore_errors`` kwarg can be used.
```
mccalluc: mention `infer_schema_length=None`?

Shoeboxam: I think it would be useful to mention this option as relevant for stable data loading.

Shoeboxam (changes requested):
There are other options that could arguably be tied to the stability properties of data loading, but it would be better to keep the docs sparser and more approachable than it would be to technically exact.

docs/source/getting-started/tabular-data/essential-statistics.rst
mccalluc: @Shoeboxam : Perhaps this?
```suggestion
    * If you know all column types ahead of time, you may use the ``schema`` kwarg;
    * Alternatively, ``infer_schema=False`` will treat all columns as strings, and numeric columns will need to be explicitly cast.
    * Using ``infer_schema_length=None`` will read the entire file before inferring types.
    * Finally, default schema inference plus ``ignore_errors=True`` will avoid run time errors, but values will be missing if they can not be parsed as the inferred type.
```

I'm wary of describing any one of these as a "best-practice". It depends on context.
- Explicit casts add complexity to the downstream expressions
- For hundred column CSVs, providing a schema is a nuisance.
- `infer_schema_length=None` could be slow for large files.
- `ignore_errors=True` drops data without warning.
- and the default possibility of data-dependent errors is not strictly DP.

Shoeboxam:
How about a note:

```
.. note::
Loading data is not covered by OpenDP's privacy guarantee and should be
performed by a trusted curator. CSV schema inference is data-dependent:
private values can affect both the inferred types and whether parsing succeeds. 
In a trusted environment, however, a parsing error can be useful,
since it may reveal malformed data or an incorrect schema 
before incorrect statistics are released.

If loading success or failure may be observed outside the trusted environment, 
prefer a schema-bearing source such as Parquet or a database table, 
or load columns as strings via ``infer_schema=False`` and cast them explicitly. 
``ignore_errors=True`` avoids some parsing failures, 
but may silently change the loaded data and reduce utility, 
so it should be used deliberately rather than by default. 
See the `Polars scan_csv documentation <https://docs.pola.rs/api/python/stable/reference/api/polars.scan_csv.html>`_ 
for the available schema and error-handling options.
```

mccalluc:
@Shoeboxam , please see the description at the very top to make sure you agree about the next steps. (At one point I had wanted to cleanup kwargs across the codebase in this PR, but I don't want to drag it out any longer.)

Shoeboxam:
Removing the extra arg now makes sense with this framing.

If we want to model good behavior and speed up our tests, we could swap the data from csv to parquet. I'd bet this is where most of our polars query execution time in notebook-based docs goes to. I'm ok with either.

Shoeboxam (approved):
Of course I'll approve it because I suggested the text. 📦 

@SalilVadhan Do you agree with the framing?

mccalluc:
With this approval, I'll merge.
- Added to core dev agenda to confirm with Salil.
-  #2580 remains open for more docs clean up.
- I'm not sure about using parquet everywhere, but adding at least one example would be good. #2776

docs/source/getting-started/tabular-data/essential-statistics.rst
```diff
@@ -45,22 +45,32 @@ introduction <index.rst>`__.
...
-            
+
+    Loading data is not covered by OpenDP's privacy guarantee and should be performed by a trusted curator.
+    CSV schema inference is data-dependent:
+    private values can affect both the inferred types and whether parsing succeeds. 
+    In a trusted environment, however, a parsing error can be useful,
+    since it may reveal malformed data or an incorrect schema before incorrect statistics are released.
+
+    If loading success or failure may be observed outside the trusted environment, 
+    prefer a schema-bearing source such as Parquet or a database table,
```
mccalluc, line 67: ... which could be a potential leak, then ...

Make the connection more explicit.

PR #2586, opened 2025-12-04
docs: fix ordering of docs sections by adding indent

### LINKED ISSUE #2521: Ordering of docs sections: Empty "Example" before "Return Type"
https://docs.opendp.org/en/stable/api/python/opendp.measurements.html#opendp.measurements.make_laplace

> ### Example:
> 
> ### Return type:
> 
> [Measurement](https://docs.opendp.org/en/stable/api/python/opendp.mod.html#opendp.mod.Measurement)
> 
> ```
> >>> import opendp.prelude as dp
> >>> dp.enable_features("contrib")
> >>> input_space = dp.atom_domain(T=float, nan=False), dp.absolute_distance(T=float)
> >>> laplace = dp.m.make_laplace(*input_space, scale=1.0)
> >>> print('100?', laplace(100.0))
> 100? ...
> ```

Instead of being under the "Example" header, the example is at the end, following the "Return type". This feels familiar, and I think there is something related to this in the `test_ast`.

mccalluc:
- Fix #2521: Just need to indent the lines that follow `:example:`.
- Don't include code samples for `then_*` functions: The example doesn't really feature the `then_*`, so better not to include it? (I don't feel strongly, but this seemed like it might have been an accident.)

### DIFF
--- a/rust/opendp_tooling/src/codegen/python.rs
+++ b/rust/opendp_tooling/src/codegen/python.rs
@@ -180,12 +180,6 @@ pub(crate) fn generate_function(
     let docstring = tab_py(generate_docstring(module_name, func, hierarchy));
     let body = tab_py(generate_body(module_name, func, typemap));
 
-    let example_path = format!("src/{}/code/{}.rst", &module_name, func.name);
-    let example = match fs::read_to_string(example_path) {
-        Ok(string) => tab_py(format!("\n\n:example:\n\n{string}\n")),
-        Err(_) => "".to_string(),
-    };
-
     let then_name = func.name.replacen("make_", "then_", 1);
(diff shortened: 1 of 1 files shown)


rust/opendp_tooling/src/codegen/python.rs
```diff
@@ -201,7 +195,7 @@ def {then_name}(
     .. seealso:: 
       Delays application of ``input_domain`` and ``input_metric`` in :py:func:`opendp.{module_name}.{func_name}`
 
-{doc_params}{example}
```
Shoeboxam, line 204: Makes sense to drop examples on then.

Shoeboxam (approved):
Good fix!

PR #2589, opened 2025-12-12
Fix: swap in numpy cumsum for deprecated sklearn function

### LINKED ISSUE #2588: `FutureWarning: Function stable_cumsum is deprecated`
Deprecation warning in CI. Aren't versions pinned? Will try rerunning.
- On #2581: https://github.com/opendp/opendp/actions/runs/20144560100/job/57822294389
```
FAILED test/extras/test_pca.py::test_pca_skl - opendp.mod.OpenDPException: 
  FFI("Continued stack trace from Exception in user-defined function:
  File "/home/runner/work/opendp/opendp/python/src/opendp/_convert.py", line 878, in wrapper_func
    py_out = func(py_arg)
  File "/home/runner/work/opendp/opendp/python/src/opendp/extras/sklearn/decomposition/__init__.py", line 353, in _postprocess
    ratio_cumsum = stable_cumsum(explained_variance_ratio_np)
  File "/opt/hostedtoolcache/Python/3.13.11/x64/lib/python3.13/site-packages/sklearn/utils/deprecation.py", line 95, in wrapped
    warnings.warn(msg, category=FutureWarning)
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FutureWarning: Function stable_cumsum is deprecated; `sklearn.utils.extmath.stable_cumsum` is deprecated in version 1.8 and will be removed in 1.10. Use `np.cumulative_sum` with the desired dtype directly instead.
```

mccalluc:
- Fix #2588, hopefully

Sklearn had added their function originally because these _don't_ behave exactly the same, but I'm not sure if the difference in behavior matters for us, or if this was taken by chance. I think we could get exactly the same behavior by adding dtype, but if we don't need it, wouldn't want to add complexity.

Another option is [`numpy.cumulative_sum`](https://numpy.org/doc/stable/reference/generated/numpy.cumulative_sum.html#numpy.cumulative_sum).

Discussion which lead up to deprecation and removal of method, with example of behavior:
- https://github.com/scikit-learn/scikit-learn/issues/31533

### DIFF
--- a/python/src/opendp/extras/sklearn/decomposition/__init__.py
+++ b/python/src/opendp/extras/sklearn/decomposition/__init__.py
@@ -320,7 +320,7 @@ if _decomposition is not None:
         def _postprocess(self, values):
             """A function that applies a release of the mean and eigendecomposition to self"""
             np = import_optional_dependency('numpy')
-            from sklearn.utils.extmath import stable_cumsum, svd_flip # type: ignore[import]
+            from sklearn.utils.extmath import svd_flip # type: ignore[import]
             from sklearn.decomposition._pca import _infer_dimension # type: ignore[import]
 
             self.mean_, S, Vt = values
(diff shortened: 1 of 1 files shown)


Shoeboxam (approved):
LGTM!

PR #2593, opened 2026-01-07
docs: include job and internship announcement

### LINKED ISSUE #2592: Link to job posting from docs?
Proposal: we should make a release soon. We could link to the job posting from the docs site... and once the hire has been made, make patch release to remove the notice.

mccalluc:
The message will not appear until this is pushed to main, making the reference correct. The first commits used a separate repo, and that did work.

- Fix #2592


Shoeboxam (approved):
Great!

PR #2597, opened 2026-01-15
maintenance: Swap MacOS 15 for 13

### LINKED ISSUE #2596: Drop macOS-13 (and add macOS-15?)
Nightly is failing because "The macOS-13 based runner images are now retired". Should also try adding 15 while we're at it.

mccalluc:
- Fix #2596

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -111,9 +111,11 @@ jobs:
       # it is helpful to try to build all platforms to completion to catch any platform-specific issues
       fail-fast: false
       matrix:
-        # macos-13 is an intel runner, macos-14 is apple silicon
+        # macos-14 is apple silicon;
+        # macos-15-intel is intel, and will be available until August 2027.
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -111,9 +111,11 @@ jobs:
       # it is helpful to try to build all platforms to completion to catch any platform-specific issues
       fail-fast: false
       matrix:
-        # macos-13 is an intel runner, macos-14 is apple silicon
+        # macos-14 is apple silicon;
+        # macos-15-intel is intel, and will be available until August 2027.
+        # https://github.com/actions/runner-images/issues/13046
         # https://cibuildwheel.pypa.io/en/stable/faq/#how
-        os: [ubuntu-latest, windows-latest, macos-13, macos-14]
+        os: [ubuntu-latest, windows-latest, macos-14, macos-15-intel]
```
Shoeboxam, line 118: Great that there are macos-15 intel images.

PR #2604, opened 2026-01-20
Proof: sample_standard_bernoulli

### LINKED ISSUE #2029: Prove `sample_standard_bernoulli`
This should be a relatively simple proof.

Pranav-2509:
This PR adds a formal proof of soundness for sample_standard_bernoulli, following the OpenDP proof format.

Closes opendp/opendp#2029

### DIFF
--- a/rust/src/traits/samplers/bernoulli/sample_standard_bernoulli.tex
+++ b/rust/src/traits/samplers/bernoulli/sample_standard_bernoulli.tex
@@ -1,0 +1,83 @@
+\documentclass{article} % necessary for Overleaf to recognize the file
+\input{../lib.sty} % "rust/src/lib.sty" contains boilerplate and macros
+
+\title{\texttt{fn sample\_standard\_bernoulli}}
+\author{Pranav S Ramanujam}\date{18/01/2026}
+
+\begin{document}
+\maketitle\contrib
+Proves soundness of \rustdoc{traits/samplers/bernoulli/fn}{sample\_standard\_bernoulli}
+in \asOfCommit{rust/src/traits/samplers/bernoulli/mod.rs}{655696c}.
+
+
+\subsection*{Vetting history}
+\begin{itemize}
+    \item \vettingPR{XXXX}
+\end{itemize}
+
+\section{Hoare Triple}
+
+\subsection*{Preconditions}
+\begin{itemize}
+    \item The function \rustdoc{traits/samplers/uniform/fn}{fill\_bytes} either
+    \begin{itemize}
+        \item returns \texttt{Err(e)} if there is insufficient system entropy, or
+        \item fills the provided buffer with uniformly random bytes.
+    \end{itemize}
+\end{itemize}
+
+\subsection*{Pseudocode}
+\begin{lstlisting}[language = Python, escapechar=|]
+def sample_standard_bernoulli():
+    buffer = sample_uniform_byte()
+    return (buffer & 1) == 1   |\label{line:out}|
+\end{lstlisting}
+
+\subsection*{Postcondition}
+For any setting of the input parameters for which the preconditions hold,
+\texttt{sample\_standard\_bernoulli} either:
+\begin{itemize}
+    \item returns \texttt{Err(e)} if there is insufficient system entropy, or
+    \item returns \texttt{Ok(sample)} where \texttt{sample} is a draw from
+    $\texttt{Bernoulli}(0.5)$.
+\end{itemize}
+
+\section{Proof}
+
+Assume the preconditions are met.
+
+We first consider the behavior of the function
+\rustdoc{traits/samplers/uniform/fn}{fill\_bytes}. By the precondition, if
+\texttt{fill\_bytes} returns \texttt{Err(e)}, then
+\texttt{sample\_standard\_bernoulli} immediately returns \texttt{Err(e)} via
+the error-propagation operator \texttt{?}. This satisfies the postcondition.
+
+We now consider the case where \texttt{fill\_bytes} succeeds. In this case,
+the function fills the single-byte buffer \texttt{buffer} with a uniformly
+random value in the set $\{0,1,\dots,255\}$.
+
+The function then evaluates the expression
+\texttt{buffer[0] \& 1 == 1}. The bitwise AND operation \texttt{\& 1} extracts
+the least significant bit of \texttt{buffer[0]}. Therefore, the returned value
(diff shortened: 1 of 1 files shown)


rust/src/traits/samplers/bernoulli/sample_standard_bernoulli.tex
```diff
@@ -0,0 +1,83 @@
...
+\begin{document}
+\maketitle\contrib
+Proves soundness of \rustdoc{traits/samplers/bernoulli/fn}{sample\_standard\_bernoulli}
+in \asOfCommit{rust/src/traits/samplers/bernoulli/mod.rs}{655696c}.
+
+
+\subsection*{Vetting history}
+\begin{itemize}
+    \item \vettingPR{XXXX}
+\end{itemize}
```
Shoeboxam: You can just omit this!

rust/src/traits/samplers/bernoulli/sample_standard_bernoulli.tex
```diff
@@ -0,0 +1,83 @@
...
+\section{Hoare Triple}
+
+\subsection*{Preconditions}
+\begin{itemize}
+    \item The function \rustdoc{traits/samplers/uniform/fn}{fill\_bytes} either
+    \begin{itemize}
+        \item returns \texttt{Err(e)} if there is insufficient system entropy, or
+        \item fills the provided buffer with uniformly random bytes.
+    \end{itemize}
+\end{itemize}
```
Shoeboxam: There are actually zero preconditions for this function. A precondition for the function would be something like-- a given argument is non-negative, or well-formed in some way that is not guaranteed by the type system. For this function, you have no arguments, so the function will automatically be valid for all choices of input arguments.

Without any preconditions, you can still argue that the preconditions for fill_bytes are met, because... fill_bytes also doesn't have any preconditions. 

Therefore you can, in turn, claim the postcondition of fill_bytes is satisfied (which is this highlighted text) and then use it in the proof for sample_standard_bernoulli.

Shoeboxam: Since it is possible to remove this precondition, it will also give a nice benefit in that proofs for functions that call sample_standard_bernoulli will not need to fulfill these transitive obligations on fill_bytes (and so on, for deeper proof chains).

Shoeboxam (changes requested):
A small change. Thanks for opening this PR!

rust/src/traits/samplers/bernoulli/sample_standard_bernoulli.tex
Pranav-2509: Thank you very much for the feedback, Michael! I will go ahead and make the changes.

Pranav-2509: Fixed in a96ff57e — removed the preconditions and updated the proof to rely directly on the postcondition of fill_bytes. Thanks for the helpful explanation!

rust/src/traits/samplers/bernoulli/sample_standard_bernoulli.tex
```diff
@@ -0,0 +1,76 @@
...
+in \asOfCommit{rust/src/traits/samplers/bernoulli/mod.rs}{655696c}.
+
+\section{Hoare Triple}
+
+\subsection*{Preconditions}
+\noindent
+There are no preconditions for \texttt{sample\_standard\_bernoulli}, since the function takes no arguments and is therefore well-defined for all inputs.
+
+\noindent
+Similarly, \rustdoc{traits/samplers/uniform/fn}{fill\_bytes} has no preconditions. As a result, its postcondition—that it either returns \texttt{Err(e)} if there is insufficient system entropy or fills the provided buffer with uniformly random bytes—may be used directly in the proof below.
```
Shoeboxam: Sorry to be picky, but you don't have to justify your preconditions, and it un-necessary to talk about preconditions for other functions in the precondition. This can be reduced just:

```suggestion
There are no preconditions for \texttt{sample\_standard\_bernoulli}.
```

rust/src/traits/samplers/bernoulli/sample_standard_bernoulli.tex
```diff
@@ -0,0 +1,76 @@
...
+    \item returns \texttt{Ok(sample)} where \texttt{sample} is a draw from
+    $\texttt{Bernoulli}(0.5)$.
+\end{itemize}
+
+\section{Proof}
+
+Assume the preconditions are met.
+
+We first consider the behavior of the function
+\rustdoc{traits/samplers/uniform/fn}{fill\_bytes}. By the precondition, if
```
Shoeboxam: The argument here is more that-- given the precondition for fill_bytes is met (since it is trivial/always met), then we can claim fill_bytes' postcondition. By \texttt{fill\_bytes}' postcondition, if \texttt{fill\_bytes} returns \texttt{Err(e)}, then it is due to insufficient system entropy. Therefore if an error is subsequently returned via the try operator, the postcondition is satisfied.

The general flow being-
1. you call a sub-function
2. you argue the precondition of that sub-function is satisfied
3. therefore you can claim its postcondition/use it in your proof

Shoeboxam (changes requested):
I'll be happy with this on the next round. Thanks for sticking with it!

rust/src/traits/samplers/bernoulli/sample_standard_bernoulli.tex
Pranav-2509: Thanks for the comments, Michael! I will go ahead and make the necessary changes shortly.

rust/src/traits/samplers/bernoulli/sample_standard_bernoulli.tex
Pranav-2509: I’ve made the requested changes, Michael. Please let me know if anything needs further adjustment.

rust/src/traits/samplers/bernoulli/sample_standard_bernoulli.tex
Pranav-2509: I’ve made the requested changes, Michael. Please let me know if anything needs further adjustment.

Shoeboxam:
@Pranav-2509 I rebased, fixed the build, and sent you an invite to the OpenDP organization. If you accept and then sign the CLA, then I'll be ready to merge.

The CI failure for latex build is due to the PR coming from an external branch. I've got it building locally.

Issue #2232, opened 2025-01-15
Warn when passing a large inline keys dataframe to `with_keys`


Shoeboxam:
If the keys dataframe is a scan of a large in-memory dataframe, then the parameter search will be slow due to repeated serialization of a large query plan.

PR #2609, opened 2026-02-06
Add a warning in with_keys if the size of the input is large

eddiestudies:
Initial attempt on https://github.com/opendp/opendp/issues/2232


python/src/opendp/extras/polars/__init__.py
```diff
@@ -53,6 +55,25 @@
...
 
+def _get_key_size_threshold() -> int:  # in mb
+    try:
+        return int(os.environ["OPENDP_POLARS_KEY_SIZE_THRESHOLD_MB"])
+    except (KeyError, ValueError):
+        return 1 << 10
+
+def _size_warning(keys):
+    mb_factor = 1024**2  # For scaling to mb to shorten warnings
+    est_size: int = sys.getsizeof(keys) / mb_factor # A rough underestimate of size, does not include referenced objects sizes
```
mccalluc:
```suggestion
    est_size: int = sys.getsizeof(keys) // mb_factor # A rough underestimate of size, does not include referenced objects sizes
```
Fix:

> src/opendp/extras/polars/__init__.py:66: error: Incompatible types in assignment (expression has type "float", variable has type "int")

python/test/test_polars.py
```diff
@@ -950,6 +950,38 @@ def test_explicit_grouping_keys():
     pl_testing.assert_frame_equal(df_act.sort("B"), df_exp)
 
 
+def test_large_keys_warns(monkeypatch, recwarn):
+    pl = pytest.importorskip("polars")
+    pl_testing = pytest.importorskip("polars.testing")
```
mccalluc:
```suggestion
```
fix
> ./python/test/test_polars.py:955:5: F841 local variable 'pl_testing' is assigned to but never used

python/src/opendp/extras/polars/__init__.py
```diff
@@ -53,6 +55,25 @@
 def _get_opendp_polars_lib_path():
     return os.environ.get("OPENDP_POLARS_LIB_PATH", lib_path)
 
+def _get_key_size_threshold() -> int:  # in mb
+    try:
+        return int(os.environ["OPENDP_POLARS_KEY_SIZE_THRESHOLD_MB"])
```
mccalluc: Not sure what this envvar is for:
- If this is part of the public interface, needs to be documented.
- If this is just for testing, that's not great. Is there another approach?

python/src/opendp/extras/polars/__init__.py
```diff
@@ -53,6 +55,25 @@
...
+    except (KeyError, ValueError):
+        return 1 << 10
```
mccalluc:
```suggestion
    return int(os.environ.get("OPENDP_POLARS_KEY_SIZE_THRESHOLD_MB", 2 ** 10))
```

python/src/opendp/extras/polars/__init__.py
mccalluc: or leave as float. Either way!

python/src/opendp/extras/polars/__init__.py
```diff
@@ -53,6 +55,25 @@
...
+
+def _size_warning(keys):
+    mb_factor = 1024**2  # For scaling to mb to shorten warnings
+    est_size: int = sys.getsizeof(keys) / mb_factor # A rough underestimate of size, does not include referenced objects sizes
+    if hasattr(keys, "estimated_size"):
```
mccalluc, line 63: Mike should comment on the general approach, but if `estimated_size` is part of the interface, it should be documented. (I've sometimes marked PRs a draft, while also asking for review, when I know there is work to be done, but I also want feedback.)

python/src/opendp/extras/polars/__init__.py
```diff
@@ -53,6 +55,25 @@
...
+        est_size = keys.estimated_size("mb")
+    elif hasattr(keys, "width"):
+        lower_bound_col_name = "key"  # Assume columns have > 3 characters on average for lowerbound
+        width = keys.width  # number of column names
+        est_size = (sys.getsizeof(lower_bound_col_name) * width) / mb_factor
+
+    if est_size > _get_key_size_threshold():
+        warn(f"Large (key-set: ~{est_size}mb) loaded into memory. Consider writing it to disk for the plan to read it in via scan_parquet.")
```
mccalluc:
```suggestion
    if est_size > _get_key_size_threshold():
        warn(f"Large key-set (~{est_size}mb) loaded into memory. Consider writing it to disk for the plan to read it in via scan_parquet.")
```
Phrasing. Also, I'd include the `_get_key_size_threshold` value in the message. Maybe it's set lower than normal, and the user might miss that.

python/test/test_polars.py
```diff
@@ -950,6 +950,38 @@ def test_explicit_grouping_keys():
...
+        domain=lf_domain,
+    )
+
+    keys_small = pl.DataFrame(pl.Series("B", [2, 3, 4, 5, 6], dtype=pl.Int32))
+    before_warns = len(recwarn)
+    query = (
+        context.query().group_by("B").agg(pl.col("D").dp.sum((0, 10))).with_keys(keys_small)
+    )
+    after_warns = len(recwarn)
+    assert before_warns == after_warns, "There should be no warning if keys are below the threshold"
```
mccalluc: By default, tests will fail on warnings, so why introduce recwarn here?

mccalluc (changes requested):

python/test/test_polars.py
eddiestudies: Makes sense. I was not aware they would fail on warning. Given the feature raises, I wanted to cover the negative case behavior. I can remove the negative check if it seems redundant.

It would be covered by other uses of with_text that use small key-sets.

python/src/opendp/extras/polars/__init__.py
eddiestudies: Within the build different python versions lead to different behaviors. For python 3.11 it used width, and for 3.14 estimated_size. 

LazyFrame does not have the same API. 

I tried to make the function not leak the polars df internals to the user. This is at the expense of correctness if the objects are custom/have a different size api.

This logic was meant to calculate size and warn in O(1) to limit the performance hit. Hence the lowerbound string size vs expanding out the keys for a more accurate size calculation.

python/src/opendp/extras/polars/__init__.py
eddiestudies: This raises if the key is set incorrectly by the user. I think removing the key, given the other comment might be the desired end result.

python/src/opendp/extras/polars/__init__.py
eddiestudies: It is a habit I have of making limits configurable. It helps with libraries run in remote environments.

An example is a build or a remote job. If the warning causes an issue in a test, it can be configured out with the env var.

Given the concerns I will remove the configuration. This reduces the code and avoids the try except altogether.

Tests can easily be monkeypatched to avoid this.

python/test/test_polars.py
mccalluc: Negative checks are a good impulse! I double-checked, and warnings should be test errors... but if that doesn't seem to be the case, let us know. Here's the configuration from `.pytest.ini`:
```
# Treat warnings as errors,
# but ignore deprecation errors for old dataframe functions: They still need to be tested!
filterwarnings =
	error
	ignore:Call to deprecated function.*Use Polars instead:DeprecationWarning
```

eddiestudies: Ok, it seems like a warning is breaking the mbi builds.

I will remove recwarn but leave the call. The test not failing is validation that it did not warn. I should wait until the mbi fix since there is no way for the build to pass until then. It looks like mbi is calling a deprecated function for norm

mccalluc:
(Merging latest changes from main should fix CI.)

eddiestudies:
> Cool: All the tests are passing, just missing test coverage on this branch:
> 
> ```
>     elif hasattr(keys, "width"):
>         lower_bound_col_name = "key"  # Assume columns have > 3 characters on average for lowerbound
>         width = keys.width  # number of column names
>         est_size = (sys.getsizeof(lower_bound_col_name) * width) / mb_factor
> ```
> 
> ... but feel free to wait on that until @Shoeboxam can comment on the overall approach.

Thanks, I tried to find modern instances of df without estimated_size but the new code paths are mostly different across python versions. I added checks for graceful degradation and updated the checks to include shape.

The overall approach seems over engineered but given the multi version support, some fallback behavior is required.

Shoeboxam (changes requested):
I was surprised that the existence of these methods was based on the Python version, and not the Polars version, so I dug a bit into it. It is Polars-version-specific, which is nice, because the repo explicitly pins a single Polars version, meaning you won't need to special-case multiple python or polars versions. My best guess is you have two interpreters with different polars versions installed?

If you rebase and update your deps, you'll get a newer Polars version to work with. 

**For DataFrames**: On the current pinned polars version, estimated_size is guaranteed to exist.

**For LazyFrames**: There's no easy way, as you've seen, and width will throw performance warnings on the current polars version because it has to collect the schema. Maybe we keep things simple here and only run a check for DataFrames.

Shoeboxam:
Actually, I have a better approach. We should only run the check based on `len(keys.serialize())` and the warning should be specific to the keys being large in memory, which is the property that causes the slowdown. It can then suggest a fix where you offload the in-memory keyset (assuming the plan has one) into parquet. We don't even need `estimated_size` at all.

PR #2623, opened 2026-02-18
cleanup: Convert from namedtuples to dataclasses

### LINKED ISSUE #2495: Switch `NamedTuple` to `dataclass` for better prop docs
> Maybe we should consider switching everything to dataclasses, for consistency.

_Originally posted by @Shoeboxam in https://github.com/opendp/opendp/pull/2492#discussion_r2320085109_

mccalluc:
- Fix #2495
- The (only?) motivation for this is to make the documentation more idiomatic.
- Some of these instances are not public, but seemed good to be consistent.
- Could add an style check for this, but that seems excessive.
- Left one in-line usage:
```
ExprPlan = namedtuple("ExprPlan", ["plan", "expr", "fill"])
    return ExprPlan(plan, expr, fill)
```
- and one usage in `test_ast.py`:
```
class CodeObj(NamedTuple):
...
```

### DIFF
--- a/python/src/opendp/extras/numpy/__init__.py
+++ b/python/src/opendp/extras/numpy/__init__.py
@@ -51,7 +51,7 @@ def _check_nonnegative_int(v: int | None, name: str):
             raise ValueError(f"{name} must be non-negative")  # pragma: no cover
 
 
-def _fmt_attrs(attrs: dataclass) -> str:
+def _fmt_attrs(attrs) -> str:
     return ", ".join(f"{k}={v}" for k, v in asdict(attrs).items() if v is not None)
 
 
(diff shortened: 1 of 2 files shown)


Shoeboxam (dismissed):
Thanks for making this adjustment.

python/src/opendp/extras/numpy/__init__.py
```diff
@@ -49,8 +51,8 @@ def _check_nonnegative_int(v: int | None, name: str):
             raise ValueError(f"{name} must be non-negative")  # pragma: no cover
 
 
-def _fmt_attrs(attrs: NamedTuple) -> str:
-    return ", ".join(f"{k}={v}" for k, v in attrs._asdict().items() if v is not None)
+def _fmt_attrs(attrs) -> str:
```
mccalluc, line 54: I tried adding an annotation for all the exact types this gets used with, but I thought it just added clutter.

PR #2642, opened 2026-03-04
chore: upgrade Polars to 0.52

### LINKED ISSUE #2641: Polars: support 0.52
I think this will also resolve reports of failure to run getting-started, although I can't reproduce it.

Shoeboxam:
I took on the complexity of introducing a patch into our packaging because the newer Polars release pairs aren't working out.

### DIFF
--- a/docs/requirements_notebooks.txt
+++ b/docs/requirements_notebooks.txt
@@ -4,6 +4,6 @@ pandas
 seaborn
 scipy
 scikit-learn
-polars==1.32.0
+polars==1.36.1
 pyarrow
 hvplot
--- a/docs/source/contributing/development-environment.rst
+++ b/docs/source/contributing/development-environment.rst
@@ -59,18 +59,6 @@ This will compile a debug build of the OpenDP shared library, placing it in the 
 
 Substitute ``cargo build`` with ``cargo test`` to test, or ``cargo check`` to check syntax.
 
-.. note::
-
-  If linking fails when runnning ``cargo test``, you may need to build without the ``extension-module`` feature (`context here <https://pyo3.rs/v0.25.1/faq.html#i-cant-run-cargo-test-or-i-cant-build-in-a-cargo-workspace-im-having-linker-issues-like-symbol-not-found-or-undefined-reference-to-_pyexc_systemerror>`_):
-
-  .. code-block:: bash
-
-    cargo test --no-default-features --features=partials,use-openssl,untrusted,polars,ffi,derive
-  
-  Since ``extension-module`` is a default feature, 
-  the other remaining features are enabled in the above command.
-
-
 Note that Python and R require builds with different features.
 Details are in the :ref:`python-setup` and :ref:`r-setup` sections below.
 Setting a feature changes how the crate compiles.
@@ -102,6 +90,8 @@ Setting a feature changes how the crate compiles.
         - Enables untrusted features ``contrib``, ``honest-but-curious`` and ``floating-point``.
       * - ``ffi``
         - Enable to include C foreign function interfaces.
+      * - ``polars-ffi``
+        - Enable both ``polars`` and ``ffi``.
       * - ``derive``
         - Enable to support code generation and links to proofs in documentation.
       * - ``extension-module``
--- a/docs/source/contributing/development-environment.rst
+++ b/docs/source/contributing/development-environment.rst
@@ -91,7 +91,7 @@ Setting a feature changes how the crate compiles.
       * - ``ffi``
         - Enable to include C foreign function interfaces.
       * - ``polars-ffi``
-        - Enable both ``polars`` and ``ffi``.
+        - Build Polars with linking against Python. Enables both ``polars`` and ``ffi``.
       * - ``derive``
         - Enable to support code generation and links to proofs in documentation.
       * - ``extension-module``
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -51,7 +51,7 @@ bitvec = "1.0"
 #
 # The features track the features enabled in the Rust build for the Polars Python crate.
 # Having matching features is necessary for cross-binary DSL compatibility.
-polars = { version = "=0.52.0", features = ["full", "meta", "cutqcut", "is_unique", "row_hash"], optional = true }
+polars = { version = "=0.52.0", features = ["full", "meta", "cutqcut", "row_hash"], optional = true }
 polars-python = { version = "=0.52.0", optional = true }
 
 # these crates are implicit in the tree behind polars, 
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -6,7 +6,7 @@ homepage = "https://opendp.org/"
 repository = "https://github.com/opendp/opendp"
 authors = ["The OpenDP Project <info@opendp.org>"]
 edition = "2024"
-rust-version = "1.85.0" # MSRV via `cargo install cargo-msrv && cargo msrv --min 1.49.0 -- cargo check --all-features`
+rust-version = "1.88.0" # MSRV via `cargo install cargo-msrv && cargo msrv --min 1.49.0 -- cargo check --all-features`
 
 [workspace]
 members = [".", "opendp_derive", "opendp_tooling"]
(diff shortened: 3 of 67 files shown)


mccalluc (commented):
Better review coming soon, but do you see this as essential for the release? We are a quarter behind on the release cadence, and even if something should be simple and low risk, that doesn't mean that it is. At this point, I think we should be focused on the necessary changes so we can get this out the door.

I'll get started on the actual review now!

docs/source/contributing/development-environment.rst
```diff
@@ -102,6 +90,8 @@ Setting a feature changes how the crate compiles.
         - Enables untrusted features ``contrib``, ``honest-but-curious`` and ``floating-point``.
       * - ``ffi``
         - Enable to include C foreign function interfaces.
+      * - ``polars-ffi``
+        - Enable both ``polars`` and ``ffi``.
```
mccalluc: I'm guessing this isn't just a shortcut, and that it does something more than `polars` and `ffi` independently, but this doesn't explain.

Beyond that, we know that unexpected combinations of features can already be a problem: I'd be reluctant to add new features, particularly when the semantics seem to already be covered.

docs/requirements_notebooks.txt
```diff
@@ -4,6 +4,6 @@ pandas
 seaborn
 scipy
 scikit-learn
-polars==1.32.0
+polars==1.36.1
```
mccalluc, line 7: Current version is up to 1.38.1: If we're upgrading, maybe upgrade all the way?

You might be aiming to get a particular feature in 1.36.1, and that's enough, but in that case it would be good to explain in the PR what that particular feature is: It would help the review, and it would help someone in the future trying to understand project history.

rust/build/main.rs
```diff
@@ -2,6 +2,19 @@
 mod derive;
 
 fn main() {
+    let polars = std::env::var_os("CARGO_FEATURE_POLARS").is_some();
+    let ffi = std::env::var_os("CARGO_FEATURE_FFI").is_some();
+    let polars_ffi = std::env::var_os("CARGO_FEATURE_POLARS_FFI").is_some();
+
+    if polars && ffi && !polars_ffi {
```
mccalluc, line 9: Good catch. Would you also want a warning if `(polars || ffi) && polars_ffi`?

rust/src/accuracy/mod.rs
```diff
@@ -3,7 +3,7 @@
 #[cfg(feature = "ffi")]
 mod ffi;
 
-#[cfg(feature = "polars")]
+#[cfg(all(feature = "polars", feature = "contrib"))]
```
mccalluc, line 6: Is `contrib` something that always should have been here? I don't see changes in this PR that by themselves would require this.

rust/src/measurements/make_private_expr/expr_dp_counting_query/mod.rs
```diff
@@ -34,21 +32,32 @@ macro_rules! new_make_expr_counting_query {
...
+            fn as_column_udf(self: Arc<Self>) -> Arc<dyn ColumnsUdf> {
+                self
+            }
+
+            fn deep_clone(self: Arc<Self>) -> Arc<dyn AnonymousColumnsUdf> {
+                Arc::new(Arc::unwrap_or_clone(self))
+            }
+
+            fn get_field(&self, _: &polars::prelude::Schema, fields: &[polars::prelude::Field]) -> PolarsResult<polars::prelude::Field> {
+                fields.first().cloned().ok_or_else(|| polars_err!(InvalidOperation: "{} expects one column", Self::NAME))
```
mccalluc:
```suggestion
                fields.first().cloned().ok_or_else(|| polars_err!(InvalidOperation: "{} expects at least one column", Self::NAME))
```
I assume more than one column isn't a problem?

This same logic seems to be copy-and-pasted several times, though in other places it's split across more lines.
```

        fields
            .first()
            .cloned()
            .ok_or_else(|| polars_err!(InvalidOperation: "{} expects one column", Self::NAME))
    }
```

Would it make sense to pull this up so there's less duplication? Or if there's a subtle difference in one of these, highlight it?

rust/src/measurements/make_private_expr/expr_index_candidates/mod.rs
```diff
@@ -127,10 +148,26 @@ impl OpenDPPlugin for IndexCandidatesShim {
...
+    }
 
-    fn get_output(&self) -> Option<GetOutput> {
-        // dtype is unknown
-        Some(GetOutput::from_type(DataType::Null))
+    fn deep_clone(self: Arc<Self>) -> Arc<dyn AnonymousColumnsUdf> {
+        Arc::new(Arc::unwrap_or_clone(self))
+    }
+
+    fn get_field(
```
mccalluc, line 162: Ok, this one is different!

rust/src/measurements/make_private_lazyframe/group_by/matching.rs
```diff
@@ -125,6 +126,13 @@ pub(crate) fn match_group_by(mut plan: DslPlan) -> Fallible<Option<MatchGroupBy>
         return fallible!(MakeMeasurement, "Apply is not supported in logical plan");
     }
 
+    if !predicates.is_empty() {
+        // Support could be added by composing over the measurements and then composing with the aggs.
+        // Not a priority to add because we'd typically rather keep groups in the DP release,
+        // especially if we've spent budget on it.
+        return fallible!(MakeMeasurement, "having is not supported in logical plan");
```
mccalluc:
```suggestion
        return fallible!(MakeMeasurement, "Having is not supported in logical plan");
```
Consider filing a placeholder issue and include in message? If someone hit this, I don't think we should assume they'll look in the source and see the longer message with more context.

rust/src/transformations/make_stable_lazyframe/group_by/mod.rs
```diff
@@ -43,6 +44,15 @@ pub fn make_stable_group_by<M: UnboundedMetric>(
         return fallible!(MakeTransformation, "apply is not currently supported");
     }
 
+    if !predicates.is_empty() {
+        // This is equivalent to running a filter after, as far as I can tell.
+        // Possibly useful to support.
+        return fallible!(
+            MakeTransformation,
+            "having/predicates are not currently supported"
+        );
```
mccalluc, line 56: Ditto: Consider moving more info into error message, and since we'd consider implementing, file a placeholder so that users can give it a thumbs-up.

rust/src/transformations/make_stable_lazyframe/group_by/mod.rs
```diff
@@ -153,6 +164,25 @@ pub(crate) enum Resize {
...
+        }
+        Expr::Element => return fallible!(MakeTransformation, "eval is not supported"),
+        Expr::Over { .. } => {
+            return fallible!(MakeTransformation, "rolling is not a stable transformation");
+        }
+        // will be useful for Rust Polars 0.53+
+        // Expr::StructEval { .. } => {
+        //     return fallible!(MakeTransformation, "eval is not supported");
+        // }
+        Expr::Eval { .. } => return fallible!(MakeTransformation, "eval is not supported"),
```
mccalluc: Everywhere we say something is not supported, we could have an invitation to let us know if it could be useful.

rust/Cargo.toml
```diff
@@ -177,10 +76,26 @@ opendp_tooling = { path = "opendp_tooling", optional = true, version = "0.14.1-d
 syn = { workspace = true, optional = true }
 proc-macro2 = { workspace = true, optional = true }
 cbindgen = { version = "0.28.0", optional = true }
-pyo3-build-config = { version = "0.25", optional = true }
+pyo3-build-config = { version = "*", features = ["resolve-config"], optional = true }
```
mccalluc, line 79: Worries me a little to see this without a version pin, but you're probably loosening this for a reason? I don't know the best pinning philosophy for Rust.

mccalluc (commented):
Lots of little comments. Top concerns are:
- copy and paste of `AnonymousColumnsUdf`
- introduction of new feature that overlaps with existing

... but for these, and everything else, if the current solution is the best, I wouldn't block.

(If you do think anything here could use more explanation, suggest putting comments in source code, rather than just in github, but your call.)

rust/build/main.rs
Shoeboxam: IMO it is better to continue without warning. Less finicky.

docs/requirements_notebooks.txt
Shoeboxam: I had issues finding a suitable DSL alignment in later versions.

docs/source/contributing/development-environment.rst
Shoeboxam: New crates need to be added when both polars and ffi are enabled, and neither of those features should add them independently. I agree we'd rather not have more features, and I tried to avoid it, but I don't have any other choice unfortunately.

rust/src/accuracy/mod.rs
Shoeboxam: Yep, I'm guarding against cfg compile failures.

rust/Cargo.toml
Shoeboxam: Cargo picks different versions depending on whether the patch is applied. It gets picked based on the requirements on pyo3/up the tree.

rust/src/measurements/make_private_expr/expr_dp_counting_query/mod.rs
Shoeboxam: The base case of passing through the input happens a number of times, but there's also a fair amount of variation.

rust/src/measurements/make_private_lazyframe/group_by/matching.rs
Shoeboxam: I expanded the allowlist.

rust/Cargo.toml
mccalluc: Cool: Might be useful info for someone maintaining this in the future.
```suggestion
# Normally we'd pin, but Cargo picks different versions depending on whether the patch is applied.
# It gets picked based on the requirements on pyo3/up the tree.
pyo3-build-config = { version = "*", features = ["resolve-config"], optional = true }
```

rust/src/measurements/make_private_lazyframe/group_by/matching.rs
mccalluc: I don't follow your comment, but it looks like the message is now
> Having is not currently supported in logical plan. Please open an issue if this would be useful to you.

which addresses my concern. Thanks!

mccalluc (dismissed):
Looks good. I think sometimes explanations might work well as comments in the code, but it depends on the case: No more edits required. Thanks!

PR #2645, opened 2026-03-04
chore: upgrade MBI to GitHub hash for l2 loss and warning fixes

### LINKED ISSUE #2643: Update MBI dependency
PyPi doesn't have the updated MBI package since August. 
* PyPi package uses deprecated APIs.
* GitHub package has seen significant improvements.

Shoeboxam:
Fixes #2643 
Fixes https://github.com/ryan112358/mbi/pull/101
Fixes budget being increased too aggressively for l1/laplace/pure-DP AIM/MWEM

### DIFF
--- a/python/setup.cfg
+++ b/python/setup.cfg
@@ -55,6 +55,9 @@ polars =
 	# to just have a single axis of increasing functionality.
 	%(scikit-learn)s
 mbi = 
+    # Fix merged upstream, but release hasn't been made.
+    # https://github.com/ryan112358/mbi/pull/101
+    # TODO: Pin to release, when available.
 	mbi @ git+https://github.com/ryan112358/mbi.git@a34eefa9f8d479bbdf209aad31e5bfe1c1987e61
 	%(polars)s
 
--- a/python/src/opendp/extras/mbi/_utilities/__init__.py
+++ b/python/src/opendp/extras/mbi/_utilities/__init__.py
@@ -298,6 +298,7 @@ def make_noise_marginal(
 ) -> Measurement:
     """Make a measurement that releases a DP marginal"""
     from opendp.extras.numpy import NPArrayDDomain
+    # use jax arrays because lms will be used in optimization
     import jax.numpy as np  # type: ignore[import-not-found]
     from mbi import LinearMeasurement  # type: ignore[import-untyped,import-not-found]
 
--- a/python/test/mbi/test_table.py
+++ b/python/test/mbi/test_table.py
@@ -100,7 +100,7 @@ def test_fit_effectiveness(algorithm, privacy_loss, approximate):
         i, j = idx[clique[0]], idx[clique[-1]]
         
         assert abs(cov[i, j] - cov_syn[i, j]) < 0.2, (
-            f"{clique} cov drifted: true={cov[i, j]:.3f}, syn={cov_syn[i, j]:.3f}"
+            f"{clique} cov drifted: {cov[i, j]:.3f=}, {cov_syn[i, j]:.3f=}"
         )
 
     # in an MRF, a missing 2-way marginal does not imply independence


python/setup.cfg
```diff
@@ -55,7 +55,7 @@ polars =
 	# to just have a single axis of increasing functionality.
 	%(scikit-learn)s
 mbi = 
-	mbi==1.0.0
+	mbi @ git+https://github.com/ryan112358/mbi.git@d916733542ebe9d3adfde44fd3971b24816afb81
```
mccalluc: This might be getting ahead of things, but do you know if Ryan plans to make a release with this? General preference to pin to formal releases on pypi:
- Semantics are more clear: Is one version older or younger than another?
- We might trust this, but a maintainer has probably done more QA on a release than on an arbitrary commit.

Shoeboxam: Seems to be a good patch. I poked Ryan about a release. The hash should be fine though, given we don't want to get hung on this.

python/src/opendp/extras/mbi/_aim/__init__.py
```diff
@@ -252,7 +252,7 @@ def _make_aim_marginal(
 
     # factors to convert stddev -> scale and scale -> half-distribution expectation
     if output_measure == max_divergence():
-        to_scale, to_mu = sqrt(2), 1.0
+        to_scale, to_mu = 1 / sqrt(2), 1.0
```
mccalluc, line 255: Was this a bug? Is there a test of the new correct behavior?

python/src/opendp/extras/mbi/_utilities/__init__.py
```diff
@@ -64,12 +64,12 @@ def mirror_descent(
     potentials=None,  # CliqueVector | None
 ):  # MarkovRandomField
     """Fit a MarkovRandomField over the domain and loss function using mirror descent.
+    
+    If you want to use a custom estimator, consider this a contract/example.
+    Your function can then close over configuration for any MBI estimator."""
```
mccalluc, line 69: Who is the audience for this comment? It feels like it is addressed to an end user, but since it's in a private module `_utilities`, I'm not sure when they would see it.

python/src/opendp/extras/mbi/_utilities/__init__.py
```diff
@@ -298,7 +298,7 @@ def make_noise_marginal(
 ) -> Measurement:
     """Make a measurement that releases a DP marginal"""
     from opendp.extras.numpy import NPArrayDDomain
-    import numpy as np  # type: ignore[import-not-found]
+    import jax.numpy as np  # type: ignore[import-not-found]
```
mccalluc, line 302: Worth a comment to explain why regular numpy isn't sufficient?

python/test/mbi/test_table.py
```diff
@@ -93,17 +93,29 @@ def test_fit_effectiveness(algorithm, privacy_loss, approximate):
...
-    synthetic_cov = np.cov(np.stack(synthetic_list))
-    assert np.linalg.norm(cov - synthetic_cov) < 0.5, f"\n{synthetic_cov}"
+    cov_syn = np.cov(np.stack(synthetic_list))
+
+    def test_cov(clique):
+        idx = {"A": 0, "B": 1, "C": 2}
+        i, j = idx[clique[0]], idx[clique[-1]]
+        
+        assert abs(cov[i, j] - cov_syn[i, j]) < 0.2, (
+            f"{clique} cov drifted: true={cov[i, j]:.3f}, syn={cov_syn[i, j]:.3f}"
```
mccalluc:
```suggestion
            f"{clique} cov drifted: {cov[i, j]:.3f=}, {cov_syn[i, j]:.3f=}"
```
Completely optional, but, the `=` syntax in f-strings is nice.

python/setup.cfg
```diff
@@ -55,7 +55,7 @@ polars =
 	# to just have a single axis of increasing functionality.
 	%(scikit-learn)s
 mbi = 
-	mbi==1.0.0
+	mbi @ git+https://github.com/ryan112358/mbi.git@a34eefa9f8d479bbdf209aad31e5bfe1c1987e61
```
mccalluc, line 61:
```suggestion
    # Fix merged upstream, but release hasn't been made.
    # https://github.com/ryan112358/mbi/pull/101
    # TODO: Pin to release, when available.
	mbi @ git+https://github.com/ryan112358/mbi.git@a34eefa9f8d479bbdf209aad31e5bfe1c1987e61
```

python/src/opendp/extras/mbi/_aim/__init__.py
Shoeboxam: Yep, this was a bug. The bug leads to budget being increased twice more aggressively in the l1/laplace setting than it should.

python/src/opendp/extras/mbi/_utilities/__init__.py
Shoeboxam: It is public-- pulled up into the mbi namespace:

https://docs.opendp.org/en/stable/api/python/opendp.extras.mbi.html#opendp.extras.mbi.mirror_descent

mccalluc: Had to remind myself how this works:
```
from ._utilities import Algorithm, Count, mirror_descent

__all__ = [
    ...
    "mirror_descent",
]
```
That makes sense: 
- For the developer, we want smaller files with few functions to make it more navigable.
- For the end user, having a flat, stable namespace is more important.

If anything, having more explicit exports would not be a bad thing, but don't think it's a priority. Thanks!

PR #2651, opened 2026-03-16
maintenance: Clean up release process

### LINKED ISSUE #2649: Release process notes, from v0.14.2 release
> If it's a patch release, no change to the version number should be needed: The patch should have been incremented after the previous release.

Not true for me: `VERSION` has
```
0.14.1-dev
```

I probably made the last release, so this is probably on me, but after the release train is complete, there are two more steps:

> 3. Increment patch version in main.
> ```
> python tools/channel_tool.py bump_version --position patch
> ```
> 4. Increment changelog.
> ```
> python tools/channel_tool.py changelog --prepend
> ```

It's going to be hard to remember these, because once the release is out on pypi, it's easy to forget that we have additional steps. I'd suggest moving them to the start of the next release instead.

Closing earlier release process issues:
- #2373
- #2507
- #2539

mccalluc:
- Fix #2649
- Extra steps after the "release" are easy to miss.
- Cleaner to just do one version bump, rather than bumping the patch number at the end of the previous release, and then doing the real bump at the start of the next release.
- Make the script a little more robust by explicitly setting directory.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -140,7 +140,7 @@ on github, or with the `gh` command line tool. Parameters:
     1. Increment patch version in `main`.
 
         ```shell
-        python channel_tool.py bump_version --position patch
+        python channel_tool.py bump_version --position patch # or minor or major
         ```
 
     1. Increment changelog.


Shoeboxam (approved):
I see, it looks like it's inviting someone to edit the file by hand. Also, the order matches better with what we've been doing.

PR #2661, opened 2026-03-25
docs: OpenDP Commons commitments

mccalluc:
The "OpenDP Commons" is an idea which goes back to the start of the project, but we haven't been really clear about what it entails. This is a proposal: If accepted, we would also update the text on the [project site](https://sites.harvard.edu/opendp/tools/#opendp-commons).

Feedback welcome!:
- Do these commitments seem like the right sorts of things to promise?
- Will we be able to follow through on these commitments?
- Would other projects also be interested in making these commitments?
  - Maybe the response time can be left up to the repo, but not longer than a month? Would that work for **Tumult**?
  - We're not expecting all maintainers to know the entire codebase for a project, but do we need to be more explicit? Would it be possible to get a second maintainer for **DP Wizard**?
- Would it be good to set a start date and/or a due date for the audit, perhaps in the fall? 
- Salil suggested that it not be a "privacy audit" -- That might both imply something more technical than we need, and it might also be too narrow in scope. ok?
- The opendp library has a number of [older PRs with unclear status](https://docs.google.com/document/d/15H4p5pkUWB5QOidK6ds8pWPzF_-aHh0vgl7F2TN5oJ4/edit?tab=t.0). This proposal doesn't touch them, though they need attention at some point.
- Should we mention things like having a test suite, CI, and PR reviews?

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,9 +19,11 @@ OpenDP is implemented in Rust, with bindings for easy use from Python and R.
 The architecture of the OpenDP Library is based on a conceptual framework for expressing privacy-aware computations.
 This framework is described in the paper [A Programming Framework for OpenDP](https://opendp.org/files/2025/11/opendp_programming_framework_11may2020_1_01.pdf).
 
-The OpenDP Library is part of the larger [OpenDP Project](https://opendp.org), a community effort to build trustworthy,
-open source software tools for analysis of private data.
-(For simplicity in these docs, when we refer to “OpenDP,” we mean just the library, not the entire project.)
+> [!NOTE]
+> This software is part of the [**OpenDP Commons**](https://sites.harvard.edu/opendp/tools/#opendp-commons). As such, the OpenDP Executive Committee commits to:
+> - Releasing this software under an [OSI approved licence](https://opensource.org/licenses), in this case the [MIT License](https://github.com/opendp/opendp/blob/main/LICENSE).
+> - Ensuring there are at least two maintainers, in this case Michael Shoemate (`Shoeboxam`) and Chuck McCallum (`mccalluc`), who will respond within a week to new issues and PRs.
+> - On an annual basis, recruiting one or more volunteers (not active contributors) to conduct an audit, and publishing the results of the audit.
 
 ## Status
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -19,9 +19,11 @@ OpenDP is implemented in Rust, with bindings for easy use from Python and R.
 The architecture of the OpenDP Library is based on a conceptual framework for expressing privacy-aware computations.
 This framework is described in the paper [A Programming Framework for OpenDP](https://opendp.org/files/2025/11/opendp_programming_framework_11may2020_1_01.pdf).
 
-The OpenDP Library is part of the larger [OpenDP Project](https://opendp.org), a community effort to build trustworthy,
-open source software tools for analysis of private data.
-(For simplicity in these docs, when we refer to “OpenDP,” we mean just the library, not the entire project.)
+> [!NOTE]
+> This software is part of the [**OpenDP Commons**](https://sites.harvard.edu/opendp/tools/#opendp-commons). As such, the OpenDP Executive Committee commits to:
+> - Releasing this software under an [OSI approved licence](https://opensource.org/licenses), in this case the [MIT License](https://github.com/opendp/opendp/blob/main/LICENSE).
+> - Ensuring there are at least two maintainers, in this case Michael Shoemate (`Shoeboxam`) and Chuck McCallum (`mccalluc`), who will respond within a week to new issues and PRs.
```
cbaycity, line 25: Is it better to hyperlink your names to your github profiles instead of posting your usernames here? You could also put parenthesis after your names and hyperlink to your github profiles and OpenDP slack profiles - which seems like a good way to display how to get in touch if someone wants to.

README.md
```diff
@@ -19,9 +19,11 @@ OpenDP is implemented in Rust, with bindings for easy use from Python and R.
...
+> - On an annual basis, recruiting one or more volunteers (not active contributors) to conduct an audit, and publishing the results of the audit.
```
cbaycity: Is there a page that we can link which has the past audits? If not, can we link to the most recent one?

README.md
mccalluc: I had a link to https://github.com/mccalluc originally, but it didn't feel like it really added anything, and was a little distracting, and more to maintain... but could add that back.

I was thinking that the maintainers don't need to volunteer for general inquiries: it's really just limited to responding to PRs and issues, and it's up the maintainers to decide who responds (though it's probably just whoever gets to it first.)

README.md
mccalluc: Short answer, no. All we have in the docs is this line:
> OpenDP has performed well in [independent security audits](https://www.sri.inf.ethz.ch/publications/lokna2023groupandattack).

Just filed:
- https://github.com/opendp/opendp/issues/2662

This is a new thing. Maybe the feedback will be that the README isn't clear, or the PR backlog needs work. It doesn't just need to be about algorithms, or at least that's what I was thinking.

cbaycity: I think this issue is good. Consider my comment resolved.

mccalluc:
@Shoeboxam , in your court. https://opendp.org/tools/ has already been updated.

README.md
Shoeboxam: Consider linking to the CODEOWNERS instead? Single source of truth.

```suggestion
> - Ensuring there are at least two maintainers, in this case designated in the [default CODEOWNERS](.github/CODEOWNERS), who will respond within a week to new issues and PRs.
```

Shoeboxam (approved):
I'm fine with that is written here, but have an optional suggestion.

PR #2670, opened 2026-03-31
fix: skip invalid wald tests

Shoeboxam:
When scale=1/2 and t=8, then prob of landing in tail is 2e-7. We run 120_000 samples, so the expected number of tail draws is np = .0238. We'll almost always get zero draws, and only about 2.3% of the time more. But the wald test is poorly calibrated in this region, with just one draw in the tail being outside of the acceptance region.


mccalluc (approved):
Probably too small for it to make sense to factor out a helper function? Otherwise looks good.

PR #2676, opened 2026-04-01
fix: patch-free publish-able polars build

### LINKED ISSUE #2675: crates.io disallows wildcards in published dependencies
I had left the pyo3 dependency's version a wildcard, because it is totally determined by upstream crates, and only exists in our explicit dependency list so that I can enable Python 3.10 ABI 3. I wanted it to remain free so that our patch could then be more concise. 

However, an explicit version must be specified to publish on crates.io.

Shoeboxam:
I found build regressions in the non-patched polars version while testing the crates.io publish pipeline.

Ideally the crate publish path is checked as part of CI, but patches can only be disabled by simulating an entire publish (quite heavy) or doctoring the Cargo.toml. 

To be unmarked from draft once we see a successful release build push to crates.io:
https://github.com/opendp/opendp/actions/workflows/release.yml

### DIFF
--- a/rust/build/main.rs
+++ b/rust/build/main.rs
@@ -1,10 +1,22 @@
 #[cfg(feature = "derive")]
 mod derive;
 
+use std::process::Command;
+
 fn main() {
+    // causes patch_polars to update if patch is commented out
+    println!("cargo:rerun-if-changed=Cargo.toml");
+    println!("cargo:rerun-if-changed=Cargo.lock");
+
+    autocfg::emit_possibility("patch_polars");
+
     let polars = std::env::var_os("CARGO_FEATURE_POLARS").is_some();
     let ffi = std::env::var_os("CARGO_FEATURE_FFI").is_some();
     let polars_ffi = std::env::var_os("CARGO_FEATURE_POLARS_FFI").is_some();
+
+    if polars && resolved_polars_plan_uses_git_patch() {
+        println!("cargo:rustc-cfg=patch_polars");
+    }
 
     if polars && ffi && !polars_ffi {
         eprintln!(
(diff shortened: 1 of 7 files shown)


rust/build/main.rs
```diff
@@ -1,11 +1,23 @@
 #[cfg(feature = "derive")]
 mod derive;
 
+use std::process::Command;
```
mccalluc, line 4: Starting with the original issue:
> However, an explicit version must be specified to publish on crates.io

Is this something you're 100% sure about? If you have a link to docs that explained this for you, that could be useful to someone picking up this code later.

rust/build/main.rs
```diff
@@ -1,11 +1,23 @@
...
+
 fn main() {
+    // causes patch_polars to update if patch is commented out
+    println!("cargo:rerun-if-changed=Cargo.toml");
+    println!("cargo:rerun-if-changed=Cargo.lock");
+
+    autocfg::emit_possibility("patch_polars");
```
mccalluc, line 15: Is this the "patch" you mention in the comment above? I'm a little confused about cause and effect, since this is below.

mccalluc (commented):
Just skimming this.

Complexity worries me. If the root problem is that cargo requires dependencies to be pinned, I was just expecting to see the dependencies being pinned? Is this trying to preserve more flexibility? (Not so much interested in an explanation for me, but information for future maintainers, and maybe a consideration of alternatives.)

I think we want to list all the features in the documentation? https://docs.opendp.org/en/stable/contributing/development-environment.html#rust-feature-listing. If this is the right direction, the explanation there might help me understand this.

rust/build/main.rs
Shoeboxam: Yes, this is the patch. It isn't a user-facing feature/something a user can configure.

rust/build/main.rs
Shoeboxam: Yes, the CI was failing due to this.

Shoeboxam:
This is ensuring that both the local build with the patch enabled, and the cargo publish build (with opendp as a dependency from crates.io, without the patch) builds. I didn't realize the publish version would not build.

Shoeboxam:
Also, this is not actually a feature a user has control over, and as such shouldn't be documented on the features page. It's a build-time environment detection to adjust the crate to the correct Polars logic.

mccalluc (approved):
Thanks for explanation. I think I should take more time to understand this better, but that shouldn't block. Filed follow-up issue:
- #2705

PR #2681, opened 2026-04-02
feat: R plugin APIs

### LINKED ISSUE #833: R: user-defined callbacks
Needed to support `make_user_transformation` `make_user_measurement` `make_postprocess`

Shoeboxam:
Closes #833


Shoeboxam:
* **#2681** <a href="https://app.graphite.com/github/pr/opendp/opendp/2681?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opendp/opendp/2681?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2678** <a href="https://app.graphite.com/github/pr/opendp/opendp/2678?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2676** <a href="https://app.graphite.com/github/pr/opendp/opendp/2676?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

mccalluc (commented):
Before getting deep into this:
- [CI fails on](https://github.com/opendp/opendp/actions/runs/24701933437/job/72246802748?pr=2681#step:3:1) `files=$(git --no-pager diff --diff-filter=d --name-only origin/graphite-base/2681 -- '*.tex')`
- Why now? Is this necessary to fix the r-universe build, or is there a collaborator who needs this?
- Why does this touch python? If this work pointed out gaps in the python API, maybe split that into a separate PR, and only then do the work to align the R API with the Python API? I might not be seeing this correctly.

Shoeboxam:
* Didn't catch that CI failure- it is intermittent due to my dev tooling, unrelated to the PR content, and didn't matter anyways because there's no .tex documents. The normalize tex PR adjusted how this resolution worked to make it robust for the future.
* I had another local line of work for eliminating globals for queryable wrapping, and wanted to handle wrapper propagation uniformly across bindings languages. I haven't pushed that wrapper PR yet, but figured I could get this in first.
* It touches Python because R needed a little extra help keeping data alive that is attached to closures by including userdata in the struct. This userdata changes the unified API that both R and Python depends on by replacing the more barebones lifeline.

docs/source/api/user-guide/plugins/code/context-api-plugins.py
```diff
@@ -0,0 +1,41 @@
+import opendp.prelude as dp
```
mccalluc: Preference for python code samples in docs to be doctests, so we can be sure that they run.

docs/source/api/user-guide/plugins/code/measurement.py
```diff
@@ -0,0 +1,40 @@
...
+        input_domain=dp.atom_domain(T=int),
+        input_metric=dp.absolute_distance(T=int),
+        output_measure=dp.max_divergence(),
+        function=function,
+        privacy_map=privacy_map,
+        TO=type(constant),
+    )
+
+
+# /make-base-constant
```
mccalluc:
```suggestion
    )
# /make-base-constant
```

docs/source/api/user-guide/plugins/code/measurement.py
```diff
@@ -0,0 +1,40 @@
+import opendp.prelude as dp
```
mccalluc: Ditto: keep as doctest so we can be sure it runs, and so the output is shown.

docs/source/api/user-guide/plugins/code/selecting-grouping-columns.py
```diff
@@ -0,0 +1,81 @@
...
+def make_select_grouping_cols(candidates, min_bin_size, scale):
+    """Create a measurement that selects a set of grouping columns from `candidates`."""
+    return (
+        make_grouping_cols_score(candidates, min_bin_size)
+        >> dp.m.then_noisy_max(dp.max_divergence(), scale)
+        >> (lambda idx: candidates[idx])
+    )
+
+
+# /plugins
```
mccalluc:
```suggestion
    )
# /plugins
```

docs/source/api/user-guide/plugins/code/selecting-grouping-columns.py
```diff
@@ -0,0 +1,81 @@
+import opendp.prelude as dp
```
mccalluc: doctest?

docs/source/api/user-guide/plugins/code/theil-sen-regression.py
```diff
@@ -0,0 +1,170 @@
+import opendp.prelude as dp
```
mccalluc: doctest?

docs/source/api/user-guide/plugins/code/theil-sen-regression.py
```diff
@@ -0,0 +1,170 @@
...
+        input_domain=dp.numpy.array2_domain(num_columns=2, T=float),
+        input_metric=dp.symmetric_distance(),
+        output_domain=dp.numpy.array2_domain(num_columns=2, T=float),
+        output_metric=dp.symmetric_distance(),
+        function=lambda x: np.vstack([pairwise_predict(x, x_cuts) for _ in range(runs)]),
+        stability_map=lambda d_in: d_in * runs,
+    )
+
+
+# /pairwise-predict
```
mccalluc:
```suggestion
    )
# /pairwise-predict
```

docs/source/api/user-guide/plugins/theil-sen-regression.rst
```diff
@@ -0,0 +1,138 @@
...
+            :language: python
+            :start-after: import opendp.prelude as dp
+            :end-before: # pairwise-predict
+
+    .. tab-item:: R
+
+        .. literalinclude:: code/theil-sen-regression.R
+            :language: r
+            :start-after: library(opendp)
+            :end-before: # pairwise-predict
```
mccalluc: It's a substring match, so `#` isn't necessary. 

I'd encourage us to stick to the convention of using comments as anchors, so that someone editing the example won't inadvertently remove an anchor.

docs/source/api/user-guide/plugins/theil-sen-regression.rst
```diff
@@ -0,0 +1,138 @@
+Thiel-Sen Regression
+====================
+
+This example demonstrates how transformation and measurement plugins can be combined to build a differentially private Theil-Sen-style regression estimator.
+
```
mccalluc, line 5: Would prefer to keep this line from the notebook:

> NOTE: If you actually want to use Theil-Sen, don’t copy-and-paste this code! It is already implemented by the OpenDP library: see the [API Reference](https://docs.opendp.org/en/stable/api/python/opendp.extras.sklearn.linear_model.html).

but with one addition:
```diff
- NOTE: If you actually want to use Theil-Sen, don’t copy-and-paste this code!
+ NOTE: If you actually want to use Theil-Sen in Python, don’t copy-and-paste this code!
```

python/src/opendp/_convert.py
```diff
@@ -928,98 +931,104 @@ def _wrap_in_slice(ptr, len_: int) -> FfiSlicePtr:
...
-                ctypes.c_char_p(traceback.format_exc().encode()),
-            )
+_CALLBACK_DISPATCH = CallbackFnValue(_invoke_py_callback)
 
-    c_wrapper_func = CallbackFnValue(wrapper_func)
-    lifeline = ExtrinsicObject(ctypes.py_object(c_wrapper_func))
 
-    return ctypes.pointer(CallbackFn(c_wrapper_func, lifeline))
+def _wrap_py_func(func, TO):
+    userdata = ExtrinsicObject(ctypes.py_object((func, TO)))
```
mccalluc, line 974: Is `userdata` just a better name, or is there more I should know about this bit?

mccalluc (commented):
Strong preference to keep the python in the docs as doctests, so we can ensure that it runs. Possible exception might be the code to generate the plots.

Beyond that, really just skimming: Let me know if something should get a closer look.

python/src/opendp/_convert.py
Shoeboxam: userdata is different from lifeline because it can contain arbitrary data. Since it can now contain data, I moved the user's Python function and expected output type TO into it. The actual C function is now statically defined with arbitrary userdata (function/TO), whereas before it was dynamically defined for each new function/TO.

docs/source/api/user-guide/plugins/context-api-plugins.rst
```diff
@@ -11,8 +11,11 @@ When the first two arguments of the constructor function are the ``input_domain`
 then they can be omitted when you call the function from the Context API. 
 The Context API will fill them in from the compositor's input space or from the output space of the previous transformation.
 
-.. literalinclude:: code/context-api-plugins.py
-    :language: python
+.. literalinclude:: code/context-api-plugins.rst
+    :start-after: # enable-features
```
mccalluc, line 17: Do we need to retail `language` here? If not, great!

mccalluc (dismissed):
Thanks. Haven't tried rendering locally, but if python syntax highlighting still works without `language`, great!

mccalluc (approved):
Thanks!

PR #2695, opened 2026-04-13
config: Add cla-bot

### LINKED ISSUE #2694: Use cla-bot
We should be able to check CLA signature automatically!

https://github.com/opendp/clabot-config/?tab=readme-ov-file#configuration

> The configuration here is shared across any opendp repos that opt into the CLA requirement. This way, contributors that sign the CLA in the scope of one repo will be verified for other repos that have opted in.

Add it here, and if it seems to work, add it to the other repos under active development by the team:

- [ ] [opendp](https://github.com/opendp/opendp): The OpenDP Library
- [ ] [polars-to-ibis](https://github.com/opendp/polars-to-ibis)
- [ ] [dp-wizard](https://github.com/opendp/dp-wizard): DP Wizard
- [ ] [dp-wizard-templates](https://github.com/opendp/dp-wizard-templates)

mccalluc:
- Fix #2694

(I've already added this to the three other repos, and along the way updated the template text and the install instructions.)


Shoeboxam (approved):
Thanks! EZ check.

PR #2696, opened 2026-04-16
fix: re-normalize distance bounds before thresholding

### LINKED ISSUE #2640: No grouped medians released without `invariant="keys"`
- prompted by DP Wizard work: https://github.com/opendp/dp-wizard/pull/885

Walking into this, I'm assuming that Polars API methods are swapable, and if one can't be exchanged for another, there is some explanation in an error message or in the docs. Right now I'm having trouble replacing grouped means with grouped medians: the latter seems to fail silently without `invariant="keys"`. Details:

```
import opendp.prelude as dp
import polars as pl

dp.enable_features("contrib")

def example(median: bool, invariant: bool):
    grade_expr = (
        pl.col("grade").dp.quantile(0.5, [50, 60, 70, 80, 90, 100])
        if median else
        pl.col("grade").dp.mean((50, 100))
    )
    privacy_unit = dp.unit_of(contributions=10)
    privacy_loss = dp.loss_of(
        epsilon=5.0,
        delta=1 / max(1e7, 10000),
    )

    stats_context = dp.Context.compositor(
        data=pl.scan_csv(
            "/Users/chuckmccallum/.dp-wizard/demo.csv",
        ).with_columns(),
        privacy_unit=privacy_unit,
        privacy_loss=privacy_loss,
        split_by_weights=[1],
        margins=[
            dp.polars.Margin(
                by=list({"class_year_str": []}.keys()),
                invariant="keys" if invariant else None,
                max_length=10000,
                max_groups=100,
            ),
        ],
    )
    confidence = 0.95  # 95% confidence interval
    groups = ["class_year_str"]
    grade_query = (
        stats_context.query().group_by(groups).agg(dp.len(), grade_expr)
    )
    return grade_query.release().collect()
```
Means, invariant keys: low count groups are released:
```
print(example(False, True))

shape: (4, 3)
┌────────────────┬─────┬────────────┐
│ class_year_str ┆ len ┆ grade      │
│ ---            ┆ --- ┆ ---        │
│ str            ┆ u32 ┆ f64        │
╞════════════════╪═════╪════════════╡
│ junior         ┆ 85  ┆ 104.059524 │
│ sophomore      ┆ 361 ┆ 89.090652  │
│ senior         ┆ 67  ┆ 85.483871  │
│ first year     ┆ 485 ┆ 89.766667  │
└────────────────┴─────┴────────────┘
```
Means, no invariant keys: low count groups not released. (This is good!)
```
print(example(False, True))

shape: (2, 3)
┌────────────────┬─────┬───────────┐
│ class_year_str ┆ len ┆ grade     │
│ ---            ┆ --- ┆ ---       │
│ str            ┆ u32 ┆ f64       │
╞════════════════╪═════╪═══════════╡
│ sophomore      ┆ 360 ┆ 90.520661 │
│ first year     ┆ 505 ┆ 88.03055  │
└────────────────┴─────┴───────────┘
```
Medians, invariant keys has reasonable results.
```
print(example(True, True))

shape: (4, 3)
┌────────────────┬─────┬───────┐
│ class_year_str ┆ len ┆ grade │
│ ---            ┆ --- ┆ ---   │
│ str            ┆ u32 ┆ i64   │
╞════════════════╪═════╪═══════╡
│ sophomore      ┆ 442 ┆ 60    │
│ senior         ┆ 926 ┆ 60    │
│ junior         ┆ 0   ┆ 80    │
│ first year     ┆ 462 ┆ 80    │
└────────────────┴─────┴───────┘
```
But medians without keys is empty:
```
print(example(True, False))

shape: (0, 3)
┌────────────────┬─────┬───────┐
│ class_year_str ┆ len ┆ grade │
│ ---            ┆ --- ┆ ---   │
│ str            ┆ u32 ┆ i64   │
╞════════════════╪═════╪═══════╡
└────────────────┴─────┴───────┘
```
My guess is that the candidates for medians are similar to the public keys, but I'm not sure if that's on the right path, and even if it is, what the next step is.

- Is there an OpenDP bug?
- Assuming not, is there something else would work better for grouped medians?

Shoeboxam:
This adjustment simply improves utility!

### DIFF
--- a/python/test/test_polars.py
+++ b/python/test/test_polars.py
@@ -600,6 +600,62 @@ def test_polars_threshold_rho():
     assert len(release) == 2
 
 
+def test_polars_grouped_quantile_max_groups_contribution_bound():
+    # Regression test for https://github.com/opendp/opendp/issues/2640
+    # A loose explicit max_groups should not dominate a tighter total contribution bound.
+    pl = pytest.importorskip("polars")
+    pl_testing = pytest.importorskip("polars.testing")
+
+    counts = {"first year": 485, "sophomore": 361, "junior": 85, "senior": 67}
+    rows = []
+    for group, n in counts.items():
+        for i in range(n):
+            rows.append(
+                {
+                    "class_year_str": group,
+                    "grade": [50, 60, 70, 80, 90, 100][i % 6],
+                }
+            )
+
+    lf = pl.DataFrame(rows).lazy()
+
+    context = dp.Context.compositor(
+        data=lf,
+        privacy_unit=dp.unit_of(contributions=10),
+        privacy_loss=dp.loss_of(epsilon=50.0, delta=1e-7),
+        split_by_weights=[1],
+        margins=[
+            dp.polars.Margin(
+                by=["class_year_str"],
+                max_length=10_000,
+                max_groups=100,
+            )
+        ],
+    )
+
+    query = context.query().group_by(["class_year_str"]).agg(
+        dp.len(),
+        pl.col("grade").dp.quantile(0.5, [50, 60, 70, 80, 90, 100]),
+    )
+
+    expected = pl.DataFrame(
+        {
+            "column": ["len", "grade"],
+            "aggregate": ["Frame Length", "0.5-Quantile"],
+            "distribution": ["Integer Laplace", "ExponentialMin"],
+            "scale": [4.2, 4.2],
+            "threshold": [84, None],
+        },
+        schema_overrides={"threshold": pl.UInt32},
+    )
+
+    pl_testing.assert_frame_equal(query.summarize(), expected)
+
+    release = query.release().collect()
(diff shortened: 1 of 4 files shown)


Shoeboxam:
* **#2696** <a href="https://app.graphite.com/github/pr/opendp/opendp/2696?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opendp/opendp/2696?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mccalluc (commented):
The original issue
- #2640

had a lot of code samples. It would be good to have a test here that demonstrates that the problems described there are fixed: You're probably seeing a root problem below the level of those examples, but it would be good to have something here that demonstrates what is being improved, and makes sure it doesn't regress.

mccalluc (approved):
Thanks!

PR #2697, opened 2026-04-16
feat: optimize numpy vector FFI

### LINKED ISSUE #2709: fast-path for numpy FFI
Currently, numpy arrays are converted into python lists before being passed to Rust. Similarly, there's no way to unload the contiguous Rust FfiSlice vec data into a contiguous numpy array. I expect this to significantly improve performance for typical mechanism invocation.

Shoeboxam:
Direct FFI conversion benchmarks, mean ms per call:

* n=1_000 input: 0.013 current vs 0.423 baseline, about 32x faster
* n=100_000 input: 0.053 current vs 40.829 baseline, about 770x faster
* n=1_000 output: 0.018 current vs 0.034 baseline, about 1.9x faster
* n=100_000 output: 0.023 current vs 2.514 baseline, about 109x faster

### DIFF
--- a/python/src/opendp/_convert.py
+++ b/python/src/opendp/_convert.py
@@ -48,7 +48,7 @@ _NUMPY_COMPATIBLE_ATOM_TYPES = frozenset(ATOM_MAP) - {'AnyMeasurementPtr', 'AnyT
 def _numpy_dtype_for_rust_type(type_name: str) -> Any:
     np = import_optional_dependency('numpy')
     if type_name not in _NUMPY_COMPATIBLE_ATOM_TYPES:
-        raise ValueError("unrecognized numpy dtype")
+        raise ValueError(f"unrecognized numpy dtype: {type_name}")
     return np.dtype(ATOM_MAP[type_name])
 
 
@@ -183,7 +183,6 @@ def c_to_py(value: Any) -> Any:
         from opendp._data import object_type, object_as_slice, slice_free
 
         obj_type = object_type(value)
-        obj_rt_type = RuntimeType.parse(obj_type)
 
         if obj_type == PrivacyProfile.__name__:
             return PrivacyProfile(value)
@@ -205,7 +204,7 @@ def c_to_py(value: Any) -> Any:
 
         ffi_slice = object_as_slice(value)
         try:
-            return _slice_to_py(ffi_slice, obj_rt_type)
+            return _slice_to_py(ffi_slice, RuntimeType.parse(obj_type))
         finally:
             slice_free(ffi_slice)
 
@@ -560,7 +559,7 @@ def _slice_to_vector(raw: FfiSlicePtr, type_name: RuntimeType) -> Sequence[Any]:
 def _numpy_to_slice(val, type_name: RuntimeType) -> FfiSlicePtr:
     np = import_optional_dependency("numpy")
     if type_name.origin != 'NDArray' or len(type_name.args) != 1:
-        raise ValueError("type_name must be NDArray<_>")  # pragma: no cover
+        raise ValueError(f"type_name must be NDArray<T> with one type argument, found {type_name}")  # pragma: no cover
 
     inner_type_name = type_name.args[0]
     if not isinstance(inner_type_name, str):
@@ -570,8 +569,10 @@ def _numpy_to_slice(val, type_name: RuntimeType) -> FfiSlicePtr:
     if not isinstance(val, np.ndarray):
         raise TypeError(f"Expected type is {type_name}.")
 
-    if val.ndim != 1 or val.dtype != np.dtype(np_dtype):
+    if val.ndim != 1:
         raise TypeError("Only 1d arrays are currently supported. Flatten first.")
+    if val.dtype != np.dtype(np_dtype):
+        raise TypeError(f"Expected dtype {np.dtype(np_dtype)}, got {val.dtype}.")
 
     contiguous = np.ascontiguousarray(val)
     array = np.ctypeslib.as_ctypes(contiguous)
@@ -583,13 +584,13 @@ def _numpy_to_slice(val, type_name: RuntimeType) -> FfiSlicePtr:
 def _slice_to_numpy(raw: FfiSlicePtr, type_name: RuntimeType):
     np = import_optional_dependency("numpy")
     if type_name.origin != 'NDArray' or len(type_name.args) != 1:
-        raise ValueError("type_name must be NDArray<_>")  # pragma: no cover
+        raise ValueError(f"type_name must be NDArray<T> with one type argument, found {type_name}")  # pragma: no cover
 
     inner_type_name = type_name.args[0]
     if not isinstance(inner_type_name, str):
         raise ValueError(f"inner type must be atomic, found {inner_type_name}")  # pragma: no cover
     
-    _numpy_dtype_for_rust_type(inner_type_name)
+    _numpy_dtype_for_rust_type(inner_type_name)  # validate numpy compatibility before casting
 
     array_ptr: Any = ctypes.cast(raw.contents.ptr, ctypes.POINTER(ATOM_MAP[inner_type_name]))
     return np.ctypeslib.as_array(array_ptr, shape=(raw.contents.len,)).copy()
--- a/python/src/opendp/_convert.py
+++ b/python/src/opendp/_convert.py
@@ -1,4 +1,4 @@
-from typing import Sequence, Union, cast, MutableMapping
+from typing import Any, Sequence, Union, cast, MutableMapping
 from inspect import signature
 
 from opendp._lib import *
@@ -43,6 +43,14 @@ ATOM_MAP = {
     'AnyMeasurementPtr': Measurement,
     'AnyTransformationPtr': Transformation,
 }
+
+_NUMPY_COMPATIBLE_ATOM_TYPES = frozenset(ATOM_MAP) - {'AnyMeasurementPtr', 'AnyTransformationPtr'}
+def _numpy_dtype_for_rust_type(type_name: str) -> Any:
+    np = import_optional_dependency('numpy')
+    if type_name not in _NUMPY_COMPATIBLE_ATOM_TYPES:
+        raise ValueError("unrecognized numpy dtype")
+    return np.dtype(ATOM_MAP[type_name])
+
 
 def c_int_limits(type_name):
     c_int_type = ATOM_MAP[type_name]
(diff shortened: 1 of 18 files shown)


Shoeboxam:
* **#2697** <a href="https://app.graphite.com/github/pr/opendp/opendp/2697?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opendp/opendp/2697?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

python/src/opendp/extras/sklearn/decomposition/__init__.py
```diff
@@ -114,11 +114,13 @@ def _make_eigdecomp(norm, origin):
             >> then_np_clamp(norm, p=2, origin=origin)
             >> then_center()
             >> then_private_np_eigendecomposition(unit_epsilon.eigvals, unit_epsilon.eigvecs)
-            >> (lambda out: PCAResult(
-                mean=origin,
-                S=np.sqrt(np.maximum(out[0], 0))[::-1],
-                Vt=out[1].T
-            ))
+            >> _new_pure_function(
```
Shoeboxam, line 117: This should also drop the honest-but-curious requirement that snuck in for decomposition.

python/src/opendp/extras/sklearn/linear_model/_make_private_theil_sen/__init__.py
```diff
@@ -122,5 +123,5 @@ def make_private_theil_sen(
             output_measure, y_bounds, scale, candidates_count=candidates_count
         )
         # transform median y values to coefficients (slope and intercept)
-        >> (lambda ys: P_inv @ ys)
+        >> _new_pure_function(lambda ys: P_inv @ np.asarray(ys))
```
Shoeboxam, line 126: This fixes an honest-but-curious requirement that snuck in for theil-sen. It's easy to mistakenly introduce that requirement via these implicit function postprocessors.

python/src/opendp/_convert.py
```diff
@@ -44,6 +44,14 @@
     'AnyTransformationPtr': Transformation,
 }
 
+_NUMPY_COMPATIBLE_ATOM_TYPES = frozenset(ATOM_MAP) - {'AnyMeasurementPtr', 'AnyTransformationPtr'}
+def _numpy_dtype_for_rust_type(type_name: str) -> Any:
+    np = import_optional_dependency('numpy')
+    if type_name not in _NUMPY_COMPATIBLE_ATOM_TYPES:
+        raise ValueError("unrecognized numpy dtype")
```
mccalluc: Preference for a little more info in the error message.
```suggestion
        raise ValueError(f"unrecognized numpy dtype: {type_name}")
```

python/src/opendp/_convert.py
```diff
@@ -454,18 +469,26 @@ def _slice_to_bitvector(raw: FfiSlicePtr) -> bytes:
 
 def _vector_to_slice(val: Sequence[Any], type_name: RuntimeType) -> FfiSlicePtr:
     if type_name.origin != 'Vec' or len(type_name.args) != 1:
-        raise ValueError("type_name must be a Vec<_>")  # pragma: no cover
+        raise ValueError("type_name must be Vec<_>")  # pragma: no cover
```
mccalluc, line 471: Did an earlier PR move us away from rust-style types with `<>` in the python context?

python/src/opendp/_convert.py
```diff
@@ -175,6 +183,7 @@ def c_to_py(value: Any) -> Any:
         from opendp._data import object_type, object_as_slice, slice_free
 
         obj_type = object_type(value)
+        obj_rt_type = RuntimeType.parse(obj_type)
```
mccalluc: Usually prefer to save definitions till just before use, but not a strong preference.

python/src/opendp/_convert.py
```diff
@@ -528,10 +551,50 @@ def _slice_to_vector(raw: FfiSlicePtr, type_name: RuntimeType) -> Sequence[Any]:
         array = ctypes.cast(raw.contents.ptr, ctypes.POINTER(ctypes.c_char_p))[0:raw.contents.len]
         return list(map(lambda v: v.decode(), array))
 
-    assert isinstance(inner_type_name, str), f"inner type must be string, found {inner_type_name} of type {type(inner_type_name)}"
+    if not isinstance(inner_type_name, str):
+        raise ValueError(f"inner type must be atomic, found {inner_type_name}")  # pragma: no cover
```
mccalluc, line 554: If we don't have a path that can exercise this, keeping it as `assert` seems ok to me, but not a strong preference.

python/src/opendp/_convert.py
```diff
@@ -528,10 +551,50 @@ def _slice_to_vector(raw: FfiSlicePtr, type_name: RuntimeType) -> Sequence[Any]:
...
+    
     return ctypes.cast(raw.contents.ptr, ctypes.POINTER(ATOM_MAP[inner_type_name]))[0:raw.contents.len]
 
 
+def _numpy_to_slice(val, type_name: RuntimeType) -> FfiSlicePtr:
+    np = import_optional_dependency("numpy")
+    if type_name.origin != 'NDArray' or len(type_name.args) != 1:
+        raise ValueError("type_name must be NDArray<_>")  # pragma: no cover
```
mccalluc: If `len(type_name.args) != 1`, would the error message really describe the problem?

python/src/opendp/_convert.py
```diff
@@ -528,10 +551,50 @@ def _slice_to_vector(raw: FfiSlicePtr, type_name: RuntimeType) -> Sequence[Any]:
...
+
+    inner_type_name = type_name.args[0]
+    if not isinstance(inner_type_name, str):
+        raise ValueError(f"inner type must be atomic, found {inner_type_name}")  # pragma: no cover
+
+    np_dtype = _numpy_dtype_for_rust_type(inner_type_name)
+    if not isinstance(val, np.ndarray):
+        raise TypeError(f"Expected type is {type_name}.")
```
mccalluc, line 570: Error message doesn't obviously align with test?

python/src/opendp/_convert.py
```diff
@@ -528,10 +551,50 @@ def _slice_to_vector(raw: FfiSlicePtr, type_name: RuntimeType) -> Sequence[Any]:
...
+
+    if val.ndim != 1 or val.dtype != np.dtype(np_dtype):
+        raise TypeError("Only 1d arrays are currently supported. Flatten first.")
```
mccalluc: Split into two checks?

python/src/opendp/_convert.py
```diff
@@ -528,10 +551,50 @@ def _slice_to_vector(raw: FfiSlicePtr, type_name: RuntimeType) -> Sequence[Any]:
...
+def _slice_to_numpy(raw: FfiSlicePtr, type_name: RuntimeType):
+    np = import_optional_dependency("numpy")
+    if type_name.origin != 'NDArray' or len(type_name.args) != 1:
+        raise ValueError("type_name must be NDArray<_>")  # pragma: no cover
+
+    inner_type_name = type_name.args[0]
+    if not isinstance(inner_type_name, str):
+        raise ValueError(f"inner type must be atomic, found {inner_type_name}")  # pragma: no cover
+    
+    _numpy_dtype_for_rust_type(inner_type_name)
```
mccalluc: Since we're not using the return value, comment that this is just for validation?

python/test/test_convert.py
```diff
@@ -115,14 +116,63 @@ def test_hashmap():
...
+    roundtrip_vector([1, 2], "Vec<i32>", np.int32)
+    roundtrip_scalar(1, "i32", dtype=np.int32)
+    roundtrip_vector([1., 2.], "Vec<f64>", np.float64)
+    roundtrip_scalar(1., "f64")
+    assert c_to_py(py_to_c(np.array(["A", "B"]), AnyObjectPtr, type_name="Vec<String>")) == ["A", "B"]
+    roundtrip_scalar("A", "String")
+
+    result = dp.as_array(T="i32")(np.array([1, 2], dtype=np.int32))
+    assert isinstance(result, np.ndarray)
+    assert np.array_equal(result, np.array([1, 2], dtype=np.int32))
```
mccalluc: This test is getting harder to read.

Preference for parameterized tests instead of assertion functions.

The asserts seem ad-hoc: Add an explanation of why list of string can't be handled the same way?

python/test/test_convert.py
```diff
@@ -134,6 +184,24 @@ def test_numpy_trans():
...
+        dp.symmetric_distance(),
+        bounds=(0, 10),
+    )
+
+    assert trans(np.array([-1, 2, 11], dtype=np.int32)) == [0, 2, 10]
+
+    result = dp.as_array(T="i32")(trans(np.array([-1, 2, 11], dtype=np.int32)))
+
+    assert isinstance(result, np.ndarray)
+    assert np.array_equal(result, np.array([0, 2, 10], dtype=np.int32))
```
mccalluc, line 208: Could you drop the first assert, here and below?

python/test/test_user_transform.py
```diff
@@ -43,6 +43,32 @@ def test_make_custom_transformation_error():
         make_duplicate(2, raises=True)([1, 2, 3])
 
 
+def test_numpy_user_transformation_roundtrip():
+    np = pytest.importorskip("numpy")
+
+    seen = {}
```
mccalluc: Not sure why this is a dict, if only one key is used.

mccalluc: Maybe?
``` 
input_is_ndarray |= isinstance(arg, np.ndarray)
```

mccalluc (commented):
All my comments are superficial. Confirming my understanding that these are all internal changes, and while they improve the user experience, there's no change in the interface that would need to be documented?

I do wonder about the `honest-but-curious` being required where it's not necessary: Are tests enabling that feature where it shouldn't be needed? Do you think there are other places that may need a similar fix? Might be a follow-up issue.

Tweak what you feel like tweaking, and I'll approve on the next pass.

python/src/opendp/_convert.py
Shoeboxam: At some point in the past, I thought you had pointed out (and we agreed) to not use assert in library code, because of some Python technicalities. Like optimized mode?

mccalluc: Yeah: 
> The current code generator emits no code for an [assert](https://docs.python.org/3/reference/simple_stmts.html#assert) statement when optimization is requested at compile time.

If we didn't have any asserts, that would be fine, but right now I wouldn't aim for 100%
- They are useful for clarifying the invariants of an algorithm.
- They can also help with typing, where we know something about the type, but it has trouble with inference.

mccalluc (approved):
Approving on re-review, as promised.

PR #2698, opened 2026-04-16
fix: add missing proven annotations

Shoeboxam:

PR #2699, opened 2026-04-16
chore: normalize .tex proof template

### LINKED ISSUE #2704: Standardize LaTeX proofs

Shoeboxam:
* Use GitHub handles in authors
* Standardize naming of sections in Hoare triples
* Standardize use of /theorem outside of /validX macros
* Split preconditions consistently into caller and compiler verified
* Remove Vetting History
* Don't use Proof sections
* fixes a transient actions issue related to my use of Graphite where .tex change detection fails

### DIFF
--- a/.github/workflows/latex.yml
+++ b/.github/workflows/latex.yml
@@ -15,8 +15,10 @@ jobs:
       - name: Collect diffs
         id: "diffs"
         run: |
-          files=$(git --no-pager diff --diff-filter=d --name-only origin/${{ github.event.pull_request.base.ref }} -- '*.tex')
-          
+          set -euo pipefail
+          mb="$(git merge-base HEAD^1 HEAD^2)"
+          files="$(git --no-pager diff --diff-filter=d --name-only "$mb" HEAD^2 -- '*.tex')"
+
           # using syntax from: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings
           # construct list of .tex files
           echo "tex<<EOF" >> $GITHUB_OUTPUT
@@ -25,7 +27,9 @@ jobs:
     
           # construct list of .pdf files
           echo "pdf<<EOF" >> $GITHUB_OUTPUT
-          echo "${files//$'.tex'/.pdf}" >> $GITHUB_OUTPUT
+          # Command substitution strips the trailing newline from git diff output,
+          # so re-add one before rewriting the extension on the final path.
+          printf '%s\n' "$files" | sed 's/\.tex$/.pdf/' >> $GITHUB_OUTPUT
           echo "EOF" >> $GITHUB_OUTPUT
 
           echo "$files"
--- a/rust/src/lib.sty
+++ b/rust/src/lib.sty
@@ -31,8 +31,6 @@
 \newcommand{\contrib}{{\begin{warn_box}This proof resides in \textbf{``contrib''} because it has not completed the vetting process.\end{warn_box}}} 
 \newcommand{\honestButCurious}{{\begin{warn_box}This proof relies on caller-verified preconditions, so the corresponding API belongs behind the \textbf{``honest-but-curious''} feature flag.\end{warn_box}}}
 \newcommand{\idealizedNumerics}{{\begin{warn_box}This proof relies on an idealized model of numeric computation that ignores finite-machine limitations such as non-closure and overflow, so the corresponding API belongs behind the \textbf{``idealized-numerics''} feature flag.\end{warn_box}}}
-% Deprecated alias. Prefer \idealizedNumerics.
-\newcommand{\floatingPoint}{\idealizedNumerics}
 
 % asOfCommit macro to version a code dependency. Arguments:
 %    #1: relative path to file you are dependent on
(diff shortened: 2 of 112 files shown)


Shoeboxam:
* **#2699** <a href="https://app.graphite.com/github/pr/opendp/opendp/2699?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#2700](https://github.com/opendp/opendp/pull/2700) <a href="https://app.graphite.com/github/pr/opendp/opendp/2700?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2701](https://github.com/opendp/opendp/pull/2701) <a href="https://app.graphite.com/github/pr/opendp/opendp/2701?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/opendp/opendp/2699?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2698** <a href="https://app.graphite.com/github/pr/opendp/opendp/2698?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

.github/workflows/latex.yml
```diff
@@ -25,7 +27,7 @@ jobs:
     
           # construct list of .pdf files
           echo "pdf<<EOF" >> $GITHUB_OUTPUT
-          echo "${files//$'.tex'/.pdf}" >> $GITHUB_OUTPUT
+          printf '%s\n' "$files" | sed 's/\.tex$/.pdf/' >> $GITHUB_OUTPUT
```
mccalluc, line 32: sed is more readable, but is appending `\n` necessary? No so much that I'm doubting, but I'm a little surprised, and an explanation might be good.

rust/src/lib.sty
```diff
@@ -29,7 +29,10 @@
 \usepackage{tcolorbox}
 \newtcolorbox{warn_box}{colback=red!5!white,colframe=red!75!black}
 \newcommand{\contrib}{{\begin{warn_box}This proof resides in \textbf{``contrib''} because it has not completed the vetting process.\end{warn_box}}} 
-\newcommand{\floatingPoint}{{\begin{warn_box}This implementation is susceptible to floating-point vulnerabilities.\end{warn_box}}} 
+\newcommand{\honestButCurious}{{\begin{warn_box}This proof relies on caller-verified preconditions, so the corresponding API belongs behind the \textbf{``honest-but-curious''} feature flag.\end{warn_box}}}
+\newcommand{\idealizedNumerics}{{\begin{warn_box}This proof relies on an idealized model of numeric computation that ignores finite-machine limitations such as non-closure and overflow, so the corresponding API belongs behind the \textbf{``idealized-numerics''} feature flag.\end{warn_box}}}
+% Deprecated alias. Prefer \idealizedNumerics.
+\newcommand{\floatingPoint}{\idealizedNumerics}
```
mccalluc:
```suggestion
```
Same comment on another PR: I don't think these commands are a public interface, so we can just drop `\floatingPoint`, instead of deprecating.

rust/src/lib.sty
```diff
@@ -132,9 +135,24 @@
...
 \newtheorem{note}{Note}
 
-\newcommand{\vicki}[1]{{ {\color{olive}{(vicki)~#1}}}}
-\newcommand{\hanwen}[1]{{ {\color{purple}{(hanwen)~#1}}}}
-\newcommand{\zach}[1]{{ {\color{red}{(zach)~#1}}}}
+\newcommand{\proofAuthor}[2]{%
+  #1%
+  \if\relax\detokenize{#2}\relax\else~(\href{https://github.com/#2}{\texttt{@#2}})\fi%
+}
+\newcommand{\AbigailGentleAuthor}{\proofAuthor{Abigail Gentle}{abigail-gentle}}
```
mccalluc, line 140: We're expecting that new proof authors will find this on their own? Not a big hurdle, obviously, but one more step.

.github/workflows/latex.yml
```diff
@@ -15,8 +15,10 @@ jobs:
       - name: Collect diffs
         id: "diffs"
         run: |
-          files=$(git --no-pager diff --diff-filter=d --name-only origin/${{ github.event.pull_request.base.ref }} -- '*.tex')
-          
+          set -euo pipefail
```
mccalluc, line 18: Would we ever run this locally? Big blocks of code that are only in CI worry me, but if it works, it's also not worth overdoing this.

mccalluc (commented):
A couple comments, but nothing essential. If these sort of formatting changes matter, and things have slipped before, is it worth having some CI checks? Wouldn't need to be in this PR, but you could file a follow-up issue.

I could give this a rubber-stamp approval, even without changes, but is there anything particular you would like me to look into here?

mccalluc (commented):
One more comment: With longer author strings, it looks like line wrapping is an issue. See https://raw.githubusercontent.com/opendp/artifacts/refs/heads/main/PR/2699/rust/src/measurements/canonical_noise/approximate_to_tradeoff.pdf

Image

.github/workflows/latex.yml
Shoeboxam: No, IMO we wouldn't run it locally. Locally you just look at the LaTeX in your editor and see the document. Don't typically build a batch of "just stuff for this PR".

rust/src/lib.sty
Shoeboxam: Added to the documentation.

Shoeboxam:
I adjusted the macro to allow line breaks between name and handle. I'd rather not add more CI checks. IMO this PR is not very substantive, doesn't need a deep analysis, as it's just formatting.

mccalluc (approved):
> I adjusted the macro to allow line breaks between name and handle.

I see that the line in shorter, because of the smaller font, but it still extends into the right margin. Anyway, it all fits on the page now, and if it's a problem in the future we can revisit.

Approved.

Shoeboxam:
Typical poor formatting from LaTeX... Good eye for pointing it out. It it better now, but can tweak it later if it bothers.

PR #2700, opened 2026-04-16
feat: github handles in .tex proofs

Shoeboxam:
A suggestion from @michaelghay. Ties names to GitHub handles when possible.

PR #2701, opened 2026-04-16
fix: rename floating-point to idealized-numerics

Shoeboxam:
"floating-point" is too narrow for what we need. Widen to any numeric dtype.

### DIFF
--- a/docs/source/api/user-guide/index.rst
+++ b/docs/source/api/user-guide/index.rst
@@ -57,8 +57,12 @@ Features that are available from Python and R:
        That is, if a user/adversary is 'honest' in specifying the constructor arguments,
        then even if they later become 'curious' and try to learn something from the measurement outputs,
        they will not be able to violate the differential privacy promises of the measurement.
+   * - ``idealized-numerics``
+     - Enable to assume an idealized numeric model
+       that ignores finite-precision limitations such as non-closure and overflow,
+       resulting in an underestimation of the actual privacy loss.
    * - ``floating-point``
-     - Enable to include transformations and measurements with floating-point vulnerabilities.
+     - Deprecated alias for ``idealized-numerics``.
    * - ``rust-stack-trace``
      - Enable to allow stack traces to include stack frames from Rust.
 
--- a/docs/source/contributing/proof-initiation.rst
+++ b/docs/source/contributing/proof-initiation.rst
@@ -147,7 +147,7 @@ proof:
    an idealized numeric model that ignores finite-machine limitations
    such as non-closure and overflow. Use this when the proof holds in
    idealized numerics but the implementation is not privately sound on
-   finite computers. ``\floatingPoint`` is a deprecated alias.
+   finite computers.
 
 -  ``\YourNameAuthor`` and similar macros expand to an author's
    name and GitHub handle together. Use these in ``\author{...}`` so
--- a/docs/source/contributing/proof-initiation.rst
+++ b/docs/source/contributing/proof-initiation.rst
@@ -18,6 +18,8 @@ Each document should at least have these components:
 1. Hoare Triple
 
    1. Precondition
+      1. Compiler-verified
+      2. Caller-verified
    2. Pseudocode
    3. Postcondition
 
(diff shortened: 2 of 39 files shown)


Shoeboxam:
* **#2701** <a href="https://app.graphite.com/github/pr/opendp/opendp/2701?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opendp/opendp/2701?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2699** <a href="https://app.graphite.com/github/pr/opendp/opendp/2699?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#2700](https://github.com/opendp/opendp/pull/2700) <a href="https://app.graphite.com/github/pr/opendp/opendp/2700?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#2698** <a href="https://app.graphite.com/github/pr/opendp/opendp/2698?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

docs/source/contributing/proof-initiation.rst
```diff
@@ -100,27 +102,52 @@ Here’s a template for ``absolute_distance.tex``:
...
+
+-  ``\honestButCurious`` adds a header indicating the proof relies on
+   caller-verified preconditions. Use this whenever the
+   ``Caller-verified`` subsection contains any substantive assumptions.
+
+-  ``\idealizedNumerics`` adds a header indicating the proof relies on
+   an idealized numeric model that ignores finite-machine limitations
+   such as non-closure and overflow. Use this when the proof holds in
+   idealized numerics but the implementation is not privately sound on
+   finite computers. ``\floatingPoint`` is a deprecated alias.
```
mccalluc: User code might use the feature, but I don't think user code is using the latex macros, so I don't think we need to deprecate it: We can just remove it.

Also update `lib.sty`?
```
\newcommand{\floatingPoint}{{\begin{warn_box}This implementation is susceptible to floating-point vulnerabilities.\end{warn_box}}} 
```

docs/source/api/user-guide/index.rst
```diff
@@ -57,8 +57,12 @@ Features that are available from Python and R:
        That is, if a user/adversary is 'honest' in specifying the constructor arguments,
        then even if they later become 'curious' and try to learn something from the measurement outputs,
        they will not be able to violate the differential privacy promises of the measurement.
+   * - ``idealized-numerics``
+     - Enable to assume an idealized numeric model
+       that ignores finite-precision limitations such as non-closure and overflow,
+       resulting in an underestimation of the actual privacy loss.
    * - ``floating-point``
-     - Enable to include transformations and measurements with floating-point vulnerabilities.
+     - Deprecated alias for ``idealized-numerics``.
```
mccalluc, line 65: This is fine, but I'd also be ok with expanding the comment on the old `floating-point`.

If we do switch to `idealized-numerics`, slight preference just to drop `floating-point` from docs.

Is there a deprecation warning if `floating-point` is used? Sorry if I'm missing it.

mccalluc (commented):
I'd be ok with staying with `floating-point`, even if it's not perfectly accurate, but I won't block.

I think there is a little more work necessary for the migration, if it does go forward. See comments.

Do we have a policy on deprecation, and how long deprecated features are kept around? Maybe that's a follow-up issue.

docs/source/api/user-guide/index.rst
Shoeboxam: Good observation on the floating-point feature flag deprecation. I had misunderstood your comment on it when we were chatting.

Shoeboxam:
I actually think it's bad for two reasons:
1. It's too narrow, as mentioned in the PR
2. floating-point isn't necessarily bad. There's actually a ton of floating-point functionality that's totally safe, even in mechanisms, so long as it's handled correctly


Updated to forward floating-point assert/enable/disable to idealized-numerics and warn. Deprecation policy can be handled separately.

python/test/test_typing.py
```diff
@@ -101,6 +102,26 @@ def test_set_feature():
     assert "A" not in GLOBAL_FEATURES
 
 
+def test_deprecated_floating_point():
+    with warnings.catch_warnings():
```
mccalluc, line 106: Slight preference to define some variables when strings are repeated like this: It would be easy to miss a typo, and not test what you intended to test.

mccalluc (dismissed):
Thanks!

PR #2708, opened 2026-04-19
fix: RR bitvec randomization is off-by-one

### LINKED ISSUE #2707: randomized response bitvec off-by-one error
The randomized_response_bitvec function only samples n - 1 bits because the iterator is (1..n), not (1..=n) or (0..n). The mismatched length then passes through silently because bitvec XOR is implicitly zero-extended. This results in the last bit not being protected.

Shoeboxam:


Shoeboxam:
* **#2708** <a href="https://app.graphite.com/github/pr/opendp/opendp/2708?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opendp/opendp/2708?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rust/src/measurements/randomized_response_bitvec/mod.rs
```diff
@@ -137,12 +136,39 @@ mod test {
...
+            seen_false |= !last_bit;
+            if seen_true && seen_false {
+                break;
+            }
+        }
+
+        assert!(
+            seen_true && seen_false,
+            "expected last bit to be randomized"
+        );
```
mccalluc, line 165:
```rust
        for _ in 0..64 {
            let release = m_rr.invoke(&input)?;
            let last_bit = release[release.len() - 1];
            seen_true |= last_bit;
            seen_false |= !last_bit;
            if seen_true && seen_false {
                break; // Could you exit with Ok(()) here?
            }
        }

        // Could you unconditionally fail here?
        assert!(
            seen_true && seen_false,
            "expected last bit to be randomized"
        );
```

mccalluc (approved):
Seems like a good catch. Test might be simpler, but whatever you think best.

rust/src/measurements/randomized_response_bitvec/mod.rs
Shoeboxam: You're right, you could return in the loop and panic instead of assert.

PR #2717, opened 2026-04-30
chore: use RAND_priv_bytes

### LINKED ISSUE #2716: Use RAND_priv_bytes instead of RAND_bytes in OpenSSL
I wouldn't classify this as a vulnerability, but a nice additional layer of protection against compromise against OpenSSL. If bytes are drawn from RAND_priv_bytes then compromise of the public OpenSSL RNG won't affect the security of OpenDP's randomness.

Shoeboxam:
See #2716

### DIFF
--- a/rust/src/traits/samplers/mod.rs
+++ b/rust/src/traits/samplers/mod.rs
@@ -30,6 +30,8 @@ use crate::error::Fallible;
 /// Return `Err(e)` if there is insufficient system entropy, otherwise return `Ok(())`.
 #[cfg(feature = "use-openssl")]
 pub fn fill_bytes(buffer: &mut [u8]) -> Fallible<()> {
+    // "RAND_priv_bytes() has the same semantics as RAND_bytes(). It is intended to be used for generating values that should remain private."
+    // https://docs.openssl.org/3.0/man3/RAND_bytes/
     use openssl::rand::rand_priv_bytes;
     if let Err(e) = rand_priv_bytes(buffer) {
         fallible!(FailedFunction, "OpenSSL error: {:?}", e)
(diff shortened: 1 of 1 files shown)


Shoeboxam:
* **#2717** <a href="https://app.graphite.com/github/pr/opendp/opendp/2717?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opendp/opendp/2717?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rust/src/traits/samplers/mod.rs
```diff
@@ -30,8 +30,8 @@ use crate::error::Fallible;
 /// Return `Err(e)` if there is insufficient system entropy, otherwise return `Ok(())`.
 #[cfg(feature = "use-openssl")]
 pub fn fill_bytes(buffer: &mut [u8]) -> Fallible<()> {
-    use openssl::rand::rand_bytes;
-    if let Err(e) = rand_bytes(buffer) {
+    use openssl::rand::rand_priv_bytes;
```
mccalluc, line 35:
```suggestion
    // "RAND_priv_bytes() has the same semantics as RAND_bytes(). It is intended to be used for generating values that should remain private."
    // https://docs.openssl.org/3.0/man3/RAND_bytes/
    use openssl::rand::rand_priv_bytes;
```

mccalluc (dismissed):
Sure: approved.

One little concern: I think we would like to move to a pure-rust PRNG, right? I wouldn't want a change like this to make an eventual migration harder, by raising the bar on what any replacement would need to provide.

Shoeboxam:
I looked back into the pure-rust PRNG, and still felt that the options weren't mature enough. This doesn't IMO raise the bar, it's just a nice, small hardening.

PR #2726, opened 2026-05-18
Maintenance: Just run nightly once a week

### LINKED ISSUE #2635: Tweaks so we don't get stuck running out of space on PyPI as often
It’s annoying to run out of PyPI space. If we ignore nightly failures, we can accrue build problems, and it's a pain to dig out. Options
- CI to automatically clean up older nightlies? Downside: More code, and I'd worry about the risk of deleting too much.
- Nightly -> weekly? We haven't been using the information the nightly provides.
- Look for bloat?
- Ask for increase? https://docs.pypi.org/project-management/storage-limits/ 
- Give Chuck privs?

mccalluc:
- Fix #2635

This is a proposal. Alternatives:
- Also rename to `weekly` instead of `nightly`.
- Invest, plan to automate PyPI cleaning.
- Accept the status quo, and close #2635 as won't-fix.

Since we usually haven't responded to failures within a week, I don't think we're losing any important information by just running less frequently.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -3,7 +3,8 @@ name: Nightly
 on:
   schedule:
     # 03:45 UTC = 22:45 EST/23:45 EDT (i.e., always before midnight), so nightlies will have a sensible date
-    - cron: '45 3 * * *'
+    # 0 = Sunday: Run just once a week.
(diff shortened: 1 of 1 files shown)


Shoeboxam (approved):
Ok, this is more realistic.

PR #2728, opened 2026-05-19
Docs: Add r universe badge to readme

### LINKED ISSUE #2710: Add badges for r-universe builds?
Follow up to
- #2677
- #2678

I was not actually aware of 
- https://github.com/r-universe/opendp/
- https://github.com/r-universe/opendp-dev/

or that they have been failing. Would it make sense to add badges for them to the top of the readme, just so we can see more of the failures in one place?

mccalluc:
- Fix #2710


Shoeboxam (changes requested):
It would be better to have it be green if all but WASM passed, which as admittedly quite hard to get a badge for. Otherwise this will just always be red.

mccalluc:
Good point. Closing this, but keeping the issue open.

PR #2731, opened 2026-05-21
fix: multiversion conf.py references invalid git state

Shoeboxam:
In #2679 I adjusted the CI to build docs from only the last two years. In the multiversion build, conf.py is run once, creating a refs list. Then for each ref, sphinx multiversion prepares a folder with the contents of that ref, and builds, thus loading that ref's conf.py. 0.15 was the first version to have a newer conf.py that tries to read git state to determine which versions to build. 0.15 (and similarly stable/beta/nightly) fail because the temporary folder provisioned by multiversion isn't a git repository.

This PR:
* moves the ref selection logic out of conf.py, instead passing it in as an env var
* adds a check to avoid a bogus docs publish again

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -75,8 +75,18 @@ jobs:
           python -m pip install --upgrade pip
           pip install -r requirements.txt
 
+          echo "Select recent version tags."
+          export OPENDP_SMV_TAG_WHITELIST="$(python tools/select_smv_versions.py)"
+          echo "Using tag whitelist: ${OPENDP_SMV_TAG_WHITELIST}"
+
           echo "Create docs."
           make versions
+
+          if [ ! -f build/html/en/stable/index.html ]; then
+            echo "Missing build/html/en/stable/index.html; refusing to deploy partial docs build."
+            exit 1
+          fi
+
           # Sphinx caches parsed sources in .doctrees, but the published site does not need them.
           rm -rf build/html/en/*/.doctrees
           cp -r build /tmp
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -75,17 +75,8 @@ jobs:
           python -m pip install --upgrade pip
           pip install -r requirements.txt
 
-          echo "Select recent version tags."
-          export OPENDP_SMV_TAG_WHITELIST="$(python tools/select_smv_versions.py)"
-          echo "Using tag whitelist: ${OPENDP_SMV_TAG_WHITELIST}"
-
           echo "Create docs."
           make versions
-
-          if [ ! -f build/html/en/stable/index.html ]; then
-            echo "Missing build/html/en/stable/index.html; refusing to deploy partial docs build."
-            exit 1
-          fi
 
           # Sphinx caches parsed sources in .doctrees, but the published site does not need them.
           rm -rf build/html/en/*/.doctrees
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -14,6 +14,7 @@ SPHINXMULTI   = sphinx-multiversion
 PAPER         =
 BUILDDIR      = build
 VERSION       = $(shell cat ../VERSION)
+SMV_TAG_WHITELIST ?= $(shell python3 tools/select_smv_versions.py)
 
 # User-friendly check for sphinx-build
 ifeq ($(shell which $(SPHINXBUILD) >/dev/null 2>&1; echo $$?), 1)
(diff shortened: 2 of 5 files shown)


Shoeboxam:
* **#2731** <a href="https://app.graphite.com/github/pr/opendp/opendp/2731?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opendp/opendp/2731?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

docs/source/conf.py
```diff
@@ -190,46 +188,8 @@ def setup(app):
...
-    selected_major_minors = sorted(latest_patch_for_major_minor, reverse=True)[:number_to_keep]
-    selected_tags = sorted(
-        latest_patch_for_major_minor[major_minor][1] for major_minor in selected_major_minors
-    )
-    return r"^(%s)$" % "|".join(re.escape(tag) for tag in selected_tags)
-
-
 # Whitelist pattern for tags (set to None to ignore all tags)
-smv_tag_whitelist = _select_recent_release_tags(_list_release_tags())
+smv_tag_whitelist = os.environ.get("OPENDP_SMV_TAG_WHITELIST", r"^$")
```
mccalluc, line 192: Would we ever actually want all tags, ie `r"^$"`? Simplify?

```suggestion
smv_tag_whitelist = os.environ["OPENDP_SMV_TAG_WHITELIST"]
```

Or would the envvar not be set when just one version is being built locally?

docs/source/conf.py
```diff
@@ -190,46 +188,8 @@ def setup(app):
...
 # # keep all released versions, as well as prereleases for the stable version. Doesn't work, because version != stable
 # import re
 # smv_tag_whitelist = rf'(^v\d+\.\d+\.\d+$)|(^v{re.escape(version.split("-")[0])}.+$)'
```
```suggestion
```

Clean-up, if the regex construction is being moved out of this file?

docs/tools/select_smv_versions.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def select_recent_release_tags(tags: list[str]) -> str:
+    """
+    Return a regex matching the latest patch release from the most recent release lines.
+
+    >>> select_recent_release_tags(
+    ...     ["v0.11.0", "v0.11.1", "v0.12.0", "v0.12.0-rc.1", "not-a-tag"],
+    ... )
+    '^(v0\\\\.11\\\\.1|v0\\\\.12\\\\.0)$'
```
mccalluc: I know this is all copy-paste, but a little more readable:

```suggestion
    >>> print(select_recent_release_tags(
    ...     ["v0.11.0", "v0.11.1", "v0.12.0", "v0.12.0-rc.1", "not-a-tag"],
    ... ))
    ^(v0\\.11\\.1|v0\\.12\\.0)$
```

docs/tools/select_smv_versions.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    pareto_versions: dict[tuple[int, int], tuple[int, str]] = {}
```
mccalluc, line 27: Not sure this is more clear than latest_patch_for_major_minor. Maybe add an explanation, if you want to keep the shorter name.

docs/Makefile
```diff
@@ -14,6 +14,7 @@ SPHINXMULTI   = sphinx-multiversion
 PAPER         =
 BUILDDIR      = build
 VERSION       = $(shell cat ../VERSION)
+SMV_TAG_WHITELIST ?= $(shell python3 tools/select_smv_versions.py)
```
mccalluc, line 17: In docs.yml we also have
```
          echo "Select recent version tags."
          export OPENDP_SMV_TAG_WHITELIST="$(python tools/select_smv_versions.py)"
          echo "Using tag whitelist: ${OPENDP_SMV_TAG_WHITELIST}"

          echo "Create docs."
          make versions
```
Seems like we should call `tools/select_smv_versions.py` either inside the Makefile, or outside, but not both?

(I think inside is better, to minimize the logic that is only in CI YAML.)

.github/workflows/docs.yml
```diff
@@ -75,8 +75,18 @@ jobs:
...
+          echo "Select recent version tags."
+          export OPENDP_SMV_TAG_WHITELIST="$(python tools/select_smv_versions.py)"
+          echo "Using tag whitelist: ${OPENDP_SMV_TAG_WHITELIST}"
+
           echo "Create docs."
           make versions
+
+          if [ ! -f build/html/en/stable/index.html ]; then
+            echo "Missing build/html/en/stable/index.html; refusing to deploy partial docs build."
+            exit 1
```
mccalluc: What is the recovery path if we hit this in the future? Or for that matter, from where we are now? It might entail adding `workflow_dispatch` to `docs.yaml`, so that we can just re-run the docs, if new versions have already been pushed successfully, and suggesting that to the user in the error message.

mccalluc (commented):
Great. Some comments. I do worry about the complexity here, but I don't have any concrete suggestions.

docs/source/conf.py
Shoeboxam: All tags seems like a reasonable fallback.

.github/workflows/docs.yml
Shoeboxam: Removed it for now. A good change would be to build the docs first, and only then publish all packages and docs site.

mccalluc (approved):
Looks good. In the earlier draft I wondered if there should be a message to describe next steps if it fails, but I think now it makes sense not to be too prescriptive. We expect this to pass, and if it doesn't I don't think we want to try to guess what the problem is.

PR #2732, opened 2026-05-26
Remove WIP warning from README

### LINKED ISSUE #2727: Still true? "not yet been a stable, usable release suitable for the public"
README.md:
> ```
> [Project Status: WIP – Initial development is in progress, but there has not yet been a stable, usable release suitable for the public.](https://www.repostatus.org/#wip)
> ```

Since we certainly are trying to get the public to use it, is this still necessary? While the API is still evolving, I'm not sure if there are particular things that are missing, or which we expect to change.

mccalluc:
- Fix #2727
- Change from "passed" to "completed the vetting process".

It's a minor change, but if something hasn't "passed" it might seemed like it has failed. Changing to "completed" emphasizes that the process itself gives us something, even if we haven't gotten over the finish line.

(Another direction would be to have a separate flag for proved but not-yet-vetted functions: I think that would add process complexity that we can't afford.)


Shoeboxam (approved):
Thanks, agreed that we're at least *a bit* past WIP!

Issue #2741, opened 2026-05-28
changelog.py not working on fresh checkout


mccalluc:
Trying to make the 0.15.1 release, and hitting speedbumps, even on a fresh checkout. In a fresh checkout, the top of CHANGELOG.md is
```
## [0.15.0](https://github.com/opendp/opendp/compare/v1.14.2...v0.15.0) - 2026-05-08
```
Just before running changelog.py, we have
```
## [0.15.1-dev](https://github.com/opendp/opendp/compare/v0.15.0...HEAD) - TBD

## [0.15.0](https://github.com/opendp/opendp/compare/v1.14.2...v0.15.0) - 2026-05-28
```
Error from changelog.py:
```
$ python changelog.py
fatal: ambiguous argument 'v0.15.1..HEAD': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
Traceback (most recent call last):
  File "/private/tmp/opendp/tools/changelog.py", line 102, in <module>
    main()
  File "/private/tmp/opendp/tools/changelog.py", line 92, in main
    raw_new_lines = log_until(f'v{prev_version}'.replace('-dev', ''))
  File "/private/tmp/opendp/tools/changelog.py", line 33, in log_until
    return subprocess.check_output(['git', 'log', f"{tag}..HEAD", '--oneline'], text=True).splitlines()
  File "/opt/homebrew/Cellar/python@3.10/3.10.19_1/Frameworks/Python.framework/Versions/3.10/lib/python3.10/subprocess.py", line 421, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/opt/homebrew/Cellar/python@3.10/3.10.19_1/Frameworks/Python.framework/Versions/3.10/lib/python3.10/subprocess.py", line 526, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['git', 'log', 'v0.15.1..HEAD', '--oneline']' returned non-zero exit status 128.
```
Manually edit CHANGELOG to remove `v0.15.0...HEAD` and rerun, and it works.

- See why the steps before changelog.py are adding an extra line to the CHANGELOG?
- "Add a header for the current version to the CHANGELOG.md" -- seems to be done by changelog.py?

---------

Additionally, after the hand-tweaked script run, the CHANGELOG has this:
```
[...]

PR #2743, opened 2026-05-29
docs: update banner / fix dates in changelog

mccalluc:
- Update the docs banner.
- Related to #2741, fix recent dates in the changelog.

I think part of the problem is that the date is set when the changelog is generated, which can come some time before the release hits PyPI. I'm not entirely sure about the expected state of the changelog, but I don't think I want to try to dig into it right now, either.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ This file documents the version history of OpenDP. The links on each version num
 
 
 
-## [0.15.0](https://github.com/opendp/opendp/compare/v1.14.2...v0.15.0) - 2026-05-19
+## [0.15.0](https://github.com/opendp/opendp/compare/v0.14.2...v0.15.0) - 2026-05-19
 
 
 ### Migration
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -25,7 +25,7 @@ This file documents the version history of OpenDP. The links on each version num
 
 
 
-## [0.15.0](https://github.com/opendp/opendp/compare/v1.14.2...v0.15.0) - 2026-05-28
+## [0.15.0](https://github.com/opendp/opendp/compare/v1.14.2...v0.15.0) - 2026-05-19
```
eddiestudies:
```suggestion
## [0.15.0](https://github.com/opendp/opendp/compare/v0.14.2...v0.15.0) - 2026-05-19
```

eddiestudies: Looks like the old link had v1 instead of v0

eddiestudies (changes requested):
This looks good after the v0 fix

eddiestudies (approved):

PR #2745, opened 2026-05-29
docs: add tab wrapping where missing

### LINKED ISSUE #2481: Docs consistency? Wrap `contingency-tables.html` examples, or unwrap others?
Generally we've been wrapping the code examples, but there are some exceptions.
- At this point I've used to seeing them everywhere... but they aren't really essential.
- Or, maybe they should be stripped in most places, if we don't need them?

mccalluc:
- Fix #2481

Not essential, but
- Doing it consistently avoids having to sort out whether to add at a particular point.
- Consistent style in the documentation also looks better to readers
- A negative reason, but it highlights what's missing, and invites us to fill in the other tabs (if we had infinite time.)


eddiestudies (approved):
lgtm

Issue #2746, opened 2026-06-01
d_in/out -> b_in/out


mccalluc:
Pulling this out to a separate issue

> @SalilVadhan While we're talking about terminology, I've seen confusion about `d_in` and `d_out` because they overload the `d_M(x, x')` distance function. How would you feel about renaming them? Either:
> 
> *  `p_in` and `p_out`, where `p` is the parameter-- like privacy loss parameter. This is what HoDP is currently using.
> 
> *  `b_in` and `b_out`, where `b` is the bound

> ...
> Your reasoning has convinced me b is better than p. About longer names, a one-variable name is nice for latex formatting, and it comes up a lot so it would be more convenient to keep it short.

 _Originally posted by @Shoeboxam in [#947](https://github.com/opendp/opendp/issues/947#issuecomment-1979629180)_

SalilVadhan:
Let's discuss this when working on the revision of the relevant parts of the Programming Framework paper.

Shoeboxam:
I did a friendly edit of the issue to b_in/out. b is less overloaded (compared to p, a probability) and has a nice interpretation as "bound." It's also what we went with in HoDP, FWIW.

PR #2715, opened 2026-04-29
refactor: rename privacy measures

Shoeboxam:
* MaxDivergence -> PureDP
* Approximate<MaxDivergence> -> ApproxDP
* SmoothedMaxDivergence -> ProfileDP
* RenyiDivergence -> RenyiDP
* ZeroConcentratedDivergence -> zCDP
* FDP -> FDP (unchanged)

PR #2747, opened 2026-06-01
output_measure to privacy_measure

### LINKED ISSUE #947: rename `output_measure` to `privacy_measure`
There's never a distinction between input vs output measure, so "output" is unnecessary. But there is a distinction between a measure (general mathematics) and a privacy measure (DP).

mccalluc:
- Fix #947 (but split #2746 to new issue)
- Just a search and replace
- Should there also be a change from `privacy_measure` to `privacy_metric`? 
  - If "yes!", let's just do it here.
  - If "no!", then it's fine to close the issue.
  - If "maybe?", let's open a new issue.


rust/src/measurements/noisy_top_k/ffi.rs
eddiestudies, line 1: This file is the only one that changed formatting. Should be fine though

eddiestudies (approved):
PR looks good. This will cause an exception in all user code that sets output_measuer= when using python.

Not sure if that is a common use case but the release notes should be very clear on this breaking change. If we have other breaking changes we want to do, it might be best to group them since this change will already affect many users.

I am approving the PR but merging should be timed since it will create conflicts with all existing PRs.

Shoeboxam:
We'd want to time this together with #2715, if we do merge it. 

And I agree with Eddie, in that we'd want to deprecate. 

1. Deprecation will be more work, because we'd need some machinery for single-arg deprecation, and introduce a privacy_measure arg before removing the deprecated output_measure arg sometime in the future. 
2. The new name is IMO better, but only marginally better. 
3. More churn for users.

Given these three reasons, I'm gently leaning towards closing this PR and issue.

mccalluc:
Happy to close, but #2715 doesn't have an issue, so I'll open one.

PR #2748, opened 2026-06-05
Docs: convert polars expression docs to RST

mccalluc:
- Towards #2205

Running `nb-to-rst.py` does most of the work, though the script doesn't fill in the output in the doctests. Not a high priority, but thought it would be a good introduction to documentation conventions, and the direction we'd like to go here.


docs/source/api/user-guide/polars/expressions/aggregation.rst
```diff
@@ -0,0 +1,265 @@
...
+            │ ---     │
+            │ u32     │
+            ╞═════════╡
+            ...
+            └─────────┘
+
+
+Noise added to a count can make the count go negative, but since the
+output data type is an unsigned integer, the library may return zero.
+This is more likely to happen with the true value is small.
```
eddiestudies:
```suggestion
This is more likely to happen when the true value is small.
```

docs/source/api/user-guide/polars/expressions/manipulation.rst
```diff
@@ -0,0 +1,656 @@
...
+            │ column  ┆ aggregate ┆ distribution  ┆ scale        │
+            │ ---     ┆ ---       ┆ ---           ┆ ---          │
+            │ str     ┆ str       ┆ str           ┆ f64          │
+            ╞═════════╪═══════════╪═══════════════╪══════════════╡
+            │ HWUSUAL ┆ Sum       ┆ Float Laplace ┆ 14405.179857 │
+            └─────────┴───────────┴───────────────┴──────────────┘
+
+
+If the fill expression is not a scalar literal, then the input
+expression and fill expression must both by row-by-row. This is to
```
eddiestudies: both be
```suggestion
expression and fill expression must both be row-by-row. This is to
```

docs/source/api/user-guide/polars/expressions/manipulation.rst
```diff
@@ -0,0 +1,656 @@
...
+            │ column  ┆ aggregate ┆ distribution    ┆ scale   │
+            │ ---     ┆ ---       ┆ ---             ┆ ---     │
+            │ str     ┆ str       ┆ str             ┆ f64     │
+            ╞═════════╪═══════════╪═════════════════╪═════════╡
+            │ HWUSUAL ┆ Sum       ┆ Integer Laplace ┆ 14400.0 │
+            └─────────┴───────────┴─────────────────┴─────────┘
+
+
+If the fill expression is not a scalar literal, then the input
+expression and fill expression must both by row-by-row. This is to
```
eddiestudies:
```suggestion
expression and fill expression must both be row-by-row. This is to
```

docs/source/api/user-guide/polars/expressions/manipulation.rst
```diff
@@ -0,0 +1,656 @@
...
+            ┌─────────┬───────────┬───────────────┬──────────────┐
+            │ column  ┆ aggregate ┆ distribution  ┆ scale        │
+            │ ---     ┆ ---       ┆ ---           ┆ ---          │
+            │ str     ┆ str       ┆ str           ┆ f64          │
+            ╞═════════╪═══════════╪═══════════════╪══════════════╡
+            │ HWUSUAL ┆ Sum       ┆ Float Laplace ┆ 14405.179857 │
+            └─────────┴───────────┴───────────────┴──────────────┘
+
+
+If the fill expression is not a scalar literal, then the input
```
eddiestudies, line 342: Is this section duplicate?

docs/source/api/user-guide/polars/expressions/manipulation.rst
```diff
@@ -0,0 +1,656 @@
...
+            ┌─────────┬───────────┬─────────────────┬─────────┐
+            │ column  ┆ aggregate ┆ distribution    ┆ scale   │
+            │ ---     ┆ ---       ┆ ---             ┆ ---     │
+            │ str     ┆ str       ┆ str             ┆ f64     │
+            ╞═════════╪═══════════╪═════════════════╪═════════╡
+            │ HWUSUAL ┆ Sum       ┆ Integer Laplace ┆ 14400.0 │
+            └─────────┴───────────┴─────────────────┴─────────┘
+
+
+If the fill expression is not a scalar literal, then the input
```
eddiestudies, line 394: I realized it may be on purpose but this is an exact line copy from the fill_nan section

docs/source/api/user-guide/polars/expressions/manipulation.rst
```diff
@@ -0,0 +1,656 @@
...
+            ...     .with_keys(pl.LazyFrame([labels]))
+            ... )
+            >>> query.release().collect().sort("HWUSUAL")
+            shape: (7, 2)
+            ┌─────────┬─────────┐
+            │ HWUSUAL ┆ len     │
+            │ ---     ┆ ---     │
+            │ u32     ┆ u32     │
+            ╞═════════╪═════════╡
+            │ 0       ┆ ... │
```
eddiestudies, line 644: Are the ... replaced by the program code later? In the notebook they are written out

docs/source/api/user-guide/polars/expressions/operators.rst
```diff
@@ -0,0 +1,69 @@
...
+            ...     .agg(dp.len())
+            ... )
+            >>> query.release().collect().sort("SEX", "OVER_40")
+            shape: (4, 3)
+            ┌─────┬─────────┬────────┐
+            │ SEX ┆ OVER_40 ┆ len    │
+            │ --- ┆ ---     ┆ ---    │
+            │ i64 ┆ bool    ┆ u32    │
+            ╞═════╪═════════╪════════╡
+            │ 1   ┆ false   ┆ ... │
```
eddiestudies, line 65: Same here, if this was on purpose that works as well just wanted to catch discrepancies.

mccalluc: Intentional, but maybe not ideal. The original notebook reflects the result of one run. In CI, we make sure the notebooks run without errors, but we don't make any another checks. The RST provides a much stronger check... but this means that where the results will vary, we use the doctest `...` as a wildcard. 

It is possible to skip the check entirely, for example if we want to say something about a particular output, but in general I'd like to avoid that, precisely because the user won't see the same thing when they run.

A subtle potential problem is changes in column width: If a particular run produced values that all ended with a zero in the last visible decimal point, the whole column may just be one character more narrow.

docs/source/api/user-guide/polars/expressions/manipulation.rst
mccalluc: I think it's true for both functions, so it makes sense. (Also, just not touching the text in this PR.)

docs/source/api/user-guide/polars/expressions/manipulation.rst
mccalluc: Not replaced: it's a wildcard for doctest matching. 

https://docs.python.org/3/library/doctest.html#doctest.ELLIPSIS

docs/source/api/user-guide/polars/expressions/operators.rst
eddiestudies: makes sense

eddiestudies (approved):
lgtm

PR #2750, opened 2026-06-09
Maintenance: Separate CI helpers

### LINKED ISSUE #2283: Separate top level workflows from helpers
We have a lot of github actions. The mermaid helps show how things are related, but we could also rearrange the files. Inspired by py-shiny, what if we only keep the top-level workflows in `workflows/`, and make a separate `helpers/` directory?

Example from py-shiny:
```
├── .github
│   ├── py-shiny
│   │   ├── check
│   │   │   └── action.yaml
│   │   ├── pytest-browsers
│   │   │   └── action.yaml
│   │   └── setup
│   │       └── action.yaml
│   └── workflows
│       ├── build-docs.yaml
│       ├── deploy-tests.yaml
│       ├── pytest.yaml
│       ├── triage-issues.yaml
│       └── verify-js-built.yaml
```
```
      - name: Setup py-shiny
        id: install
        uses: ./.github/py-shiny/setup
```

mccalluc:
- Fix #2283 

The remaining top-level workflows can either be called by the user, or by cron, or on PRs:

```
nightly.yml
smoke-test.yml
codeql.yml
latex.yml
release.yml
weekly-doc-check.yml
```

Annoyingly, VS Code has some built-in validation which is [stricter than github itself](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-using-an-action-in-the-same-repository-as-the-).
> Invalid workflow reference './.github/workflows/helpers/prepare.yml': workflows must be defined at the top level of the .github/workflows/ directory

I don't think we want to let IDE bugs get in the way of a change that otherwise makes sense, but feel free to push back.


.github/workflows/helpers/docs.yml
```diff
@@ -14,7 +14,7 @@ on:
 
 jobs:
   credential-check:
-    uses: ./.github/workflows/credential-check.yml
+    uses: ./.github/workflow/helpers/credential-check.yml
```
eddiestudies: workflows became workflow, this should cause an issue in builds. 
```suggestion
    uses: ./.github/workflows/helpers/credential-check.yml
```

.github/workflows/release.yml
```diff
@@ -107,7 +107,7 @@ jobs:
 
   sanity-test-post:
     needs: publish
-    uses: ./.github/workflows/sanity-test.yml
+    uses: ./.github/workflows/helpers/sanity-test.yml
```
eddiestudies: This feature may not be supported. Can we make a small two file PR to validate this change before changing the rest. I found https://github.com/orgs/community/discussions/15935.

mccalluc:
Ok, the github docs seem to be out of date. It looks like it does [require all workflows to be top level](https://github.com/opendp/opendp/actions/runs/27226716405):
```
 Invalid workflow file: .github/workflows/smoke-test.yml#L16
invalid value workflow reference: workflows must be defined at the top level of the .github/workflows/ directory
```
(And I guess this interferes with workflow discovery, so it's not even reported as a failure on this page?)

Closing PR, and issue.

PR #2751, opened 2026-06-09
docs: Prune lists-of-links

### LINKED ISSUE #2471: Add an education link, and prune older ones?
- add a link to [Differential Privacy in Artificial Intelligence](https://www.nowpublishers.com/article/BookDetails/9781638284765) on [resources page](https://docs.opendp.org/en/stable/theory/resources.html)?
- Prune back list of links: If we have multiple lists that are all partially out of date, we're not serving our users well.
  - https://docs.opendp.org/en/stable/index.html#what-is-differential-privacy
  - https://opendp.github.io/learning/
  - https://docs.opendp.org/en/stable/theory/resources.html

mccalluc:
- Fix #2471

Sphinx will warn if unlinked pages are included in the source, unless explicitly marked as `:orphan:`. I think we do want to keep the orphan page around, to keep any inbound links from breaking.

The education page doesn't have all of these links, but I don't think we should be trying to create a bibliography.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -10,13 +10,10 @@ The algorithm is differentially private if by looking at the output,
 you cannot tell whether any individual's data was included in the original dataset or not.
 Differential privacy achieves this by carefully injecting random noise into the released statistics to hide the effects of each individual. 
 
-For more background on differential privacy and its applications:
-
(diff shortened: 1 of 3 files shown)


eddiestudies (approved):
lgtm

Issue #2752, opened 2026-06-09
fix and upgrade weekly docs check


mccalluc:
- Fix failures (either add ignores or fix links)
- Upgrade: `Node.js 20 actions are deprecated`

PR #2753, opened 2026-06-09
Maintenance: upgrade github actions

mccalluc:
- Towards #2752

Looked at the release notes, and I don't think there are any breaking changes.


eddiestudies (approved):
Looks good.

We will need to run through each build soon to catch any surprise bugs from this change. 

Github actions are hard to test before the edit, best to check them before they are needed.

PR #2755, opened 2026-06-09
Maintenance: Prefix helper workflows with ".", to distinguish them from the main ones

### LINKED ISSUE #2283: Separate top level workflows from helpers

mccalluc:
- Fix #2283
- Subdirectories don't work: see discussion on #2750.
- No strong feelings about the prefix character. Sorting to the end would be nice, but the only ASCII characters after `z` are `{|}~`, none of which I would put in a filename.


eddiestudies (dismissed):
lgtm

eddiestudies (approved):
lgtm

PR #2757, opened 2026-06-11
docs: Add fix for python linking errors on mac

eddiestudies:
When installing OpenDP on a new Mac with Apple Silicon python linking errors can occur.

During developer setup the linking error occurs after cargo build due to a "missing" python3.9.

The reference to python 3.9 comes for PYO3 however, the fix should install python 3.12 since that is the latest supported python version for OpenDP.

In this PR I add a minor doc update to resolve the possible linking issue.


docs/source/contributing/development-environment.rst
```diff
@@ -57,6 +57,13 @@ Now run ``cargo build`` in the ``rust`` subdirectory of the repo.
 This will compile a debug build of the OpenDP shared library, placing it in the directory ``opendp/rust/target/debug``. 
 (The specific name of the library file will vary depending on your platform.)
 
+For Apple Silicon, if there were linking errors for python such as: library 'python3.*' not found; run the following commands.
+.. code-block:: bash
+
+    brew install python@3.12
+    PYO3_PYTHON=$(brew --prefix python@3.12)/bin/python3.12
+    cargo build --all-features
```
mccalluc: Huh: I have Apple M1, but haven't need to set this envvar.

Is there a reason you choose 3.12? The build should work down to 3.10, and when developing we generally want to use the oldest supported version, so make sure we don't accidentally use new language features, and then hit failures in CI. (I think our commitment is to support Python versions till end-of-life, which will be a few more months for 3.10.)

RST is more line-oriented than markdown. Could be wrong, but I think you'll want a blank line above `.. code-block:: bash`

Maybe use `which`?

mccalluc (commented):
I'm not sure about the diagnosis of the problem.

I'd suggest putting this on the agenda for the dev meeting so we can sort it out with Mike in the room.

docs/source/contributing/development-environment.rst
eddiestudies: No particular reason, I can set it to 3.10 if we decide to add the note. 

If we decide to add the note I will add the empty line above.

My m3 out of the box raised this error. I brew installed python and it still raised the error. I found a forum which led me to this fix but it may have been a nuance in this particular machine and not worth adding the note.

My intel based mac did not run into this issue.

mccalluc: Sorry to slow things down: you're right, let's get this in. I think it would be good to give Mike a head's up, but that doesn't need to block the PR: I put it on the agenda for next Wednesday afternoon.

Some formatting ideas. Thanks!

```suggestion
For Apple Silicon, if there were linking errors for python such as ``library 'python3.*' not found``, try this instead:

.. code-block:: bash

    brew install python@3.10
    PYO3_PYTHON=$(which python3.10)
    cargo build --all-features
```

mccalluc (changes requested):

Shoeboxam (dismissed):
So to recap, an out-of-box M3 came with python 3.9, which is EOL and below our ABI floor for polars pyo3 ffi. Then once you install a newer python version, pyo3 still might need help finding it via PYO3_PYTHON. 

This will be an issue on any machine where pyo3 resolves to an env below our minimum supported version. Since pyo3 resolves the env via virtualenv -> python -> python3, then if `python` resolves to an old version you'll still get explosions. In Eddie's case, I'd guess homebrew didn't shadow `python` upon brew install. 

I think I didn't notice the dependency on PYO3_PYTHON because at some point I set it while experimenting with fixing a build thrashing issue that would cause the crate tree to recompile starting from psm after every edit. (If you notice that, then it is likely due to rust-analyzer's env var state being slightly different from a cargo build's env var state, triggering the psm crate's build script to signal that the build cache is stale).

Given this context, it would probably be better to reword to generalize the note to any platform, and give Apple silicon as an example of how to set PYO3_PYTHON. Might be good to mention that any python version that satisfies our minimum version will do, but minimum is preferred for dev work. Thanks for digging through this!

mccalluc:
Regarding the linkchecker, the fix is upstream.
- Missing a blank line after `code-block:: bash`, this is not seen as a directive.
- Since it's parsed as plain text instead of code, Sphinx is making this a mailto link.
```
brew install <a class="reference external" href="mailto:python%403.12">python<span>@</span>3<span>.</span>12</a>
```

Screenshot 2026-06-12 at 10 16 17 AM

Changing the RST, there's no link:

Screenshot 2026-06-12 at 10 27 58 AM

I'll push this tweak.

mccalluc (approved):
Incorporates feedback from Mike, and explained the RST quirk to Eddie. Ok!

PR #2758, opened 2026-06-12
fix: one_sided_pvalue arange for scipy.stats.binom.pmf off by one bug

eddiestudies:


python/test/extras/test_canonical_noise.py
```diff
@@ -152,21 +152,21 @@ def test_oneside_pvalue():
     pytest.importorskip("scipy")
     tulap = BinomialCND(5.0, d_in=1.0, d_out=(0.1, 1e-6), size=10)
     assert str(tulap) == "BinomialCND(estimate=5.0, d_in=1.0, d_out=(0.1, 1e-06), size=10)"
-    
+
     # should be approximately equal to 0.5
-    assert np.allclose(tulap.p_value(theta=0.5, tail="right"), 0.4993195913951362)
-    assert np.allclose(tulap.p_value(theta=0.5, tail="left"), 0.49970384610486424)
+    assert np.allclose(tulap.p_value(theta=0.5, tail="right"), 0.5000000000000002)
+    assert np.allclose(tulap.p_value(theta=0.5, tail="left"), 0.5000000000000002)
```
mccalluc:
```suggestion
    assert np.allclose(tulap.p_value(theta=0.5, tail="right"), 0.5)
    assert np.allclose(tulap.p_value(theta=0.5, tail="left"), 0.5)
```
With the default tolerances this should pass.

python/test/extras/test_canonical_noise.py
```diff
@@ -152,21 +152,21 @@ def test_oneside_pvalue():
     pytest.importorskip("scipy")
     tulap = BinomialCND(5.0, d_in=1.0, d_out=(0.1, 1e-6), size=10)
     assert str(tulap) == "BinomialCND(estimate=5.0, d_in=1.0, d_out=(0.1, 1e-06), size=10)"
-    
+
     # should be approximately equal to 0.5
```
mccalluc:
```suggestion
```
And since the math is right now, doesn't need an explanation.

python/test/extras/test_canonical_noise.py
```diff
@@ -152,21 +152,21 @@ def test_oneside_pvalue():
...
-    assert np.allclose(tulap.p_value(theta=0.5, tail="right"), 0.4993195913951362)
-    assert np.allclose(tulap.p_value(theta=0.5, tail="left"), 0.49970384610486424)
+    assert np.allclose(tulap.p_value(theta=0.5, tail="right"), 0.5000000000000002)
+    assert np.allclose(tulap.p_value(theta=0.5, tail="left"), 0.5000000000000002)
 
     pvalue = one_sided_pvalue(
         Z=3, d_in=1.0, d_out=(0.5, 1e-8), size=10, theta=0.5, tail="right"
     )
-    assert np.allclose(pvalue, 0.7606589354450621)
+    assert np.allclose(pvalue, 0.7616207531435369)
```
mccalluc, line 162: Would it be possible to use an exact value here, too? Not sure about the underlying math.

python/test/extras/test_canonical_noise.py
```diff
@@ -152,21 +152,21 @@ def test_oneside_pvalue():
...
 
 
 def test_twoside_pvalue():
     pytest.importorskip("scipy")
     tulap = BinomialCND(1.0, d_in=1.0, d_out=(0.5, 1e-8), size=10)
-    assert np.allclose(tulap.p_value(theta=0.5), 0.18443148204450355)
+    assert np.allclose(tulap.p_value(theta=0.5), 0.1841407488015684)
```
mccalluc, line 168: ditto.

Shoeboxam (dismissed):
Yes, this fix makes sense, thanks for finding it! Can you adjust the PR to `fix:` before merging, so that it gets sorted nicely by the changelog script?

python/src/opendp/extras/numpy/canonical.py
```diff
@@ -245,7 +245,7 @@ def one_sided_pvalue(
     stats = import_optional_dependency("scipy.stats")
     np = import_optional_dependency("numpy")
 
-    values = np.arange(size)
+    values = np.arange(0, size + 1)
```
mccalluc:
```suggestion
    values = np.arange(size + 1)
```
Start is 0 by default, I think.

> arange(stop): Values are generated within the half-open interval [0, stop) (in other words, the interval including start but excluding stop).
> 
> arange(start, stop): Values are generated within the half-open interval [start, stop).

Maybe quote the docs, if we were confused? Your call.

```suggestion
    # Values are generated within the half-open interval [0, stop)
    values = np.arange(size + 1)
```

mccalluc (commented):
Good catch. Some tweaks. Thanks!

mccalluc (approved):
Having a tidy expression for the other expressions is an ideal, not a requirement. Thanks!

PR #2760, opened 2026-06-12
Extract convert code maps from convert

eddiestudies:
### Proposing a split of convert code.

The initial goal split was all to_slice, from_slice functions into convert_maps. 
However, convert maps for collections depend on py_to_c as a recursive function call from an internal function.

The current PR is organizational. There is a name to function map for to_slice and from_slice maps introduced. This follows the paradigm of ATOM_DATA, and removes a category of conditionals. Purely an ergonomic change.

Caveats, performance and test coverage are not suitable. I will expand test coverage to cover each type and run some performance focused evaluations to understand the tradeoffs of this change.

### Next steps
I am looking into making the internal call as a passthrough instead of a reference to py_to_c

Simplified overview:
def _to_slice_collection(...):
    for internal object in collection:
            py_to_c(...)
    return out
def py_to_c(py_obj):
      match  py_obj
          case collection:
               return _to_slice_collection(...)

### Proposal:
def _to_slice_collection(..., converter):
    for internal object in collection:
            converter(...)
    return out
def py_to_c(py_obj):
      match  py_obj
          case collection:
               return _to_slice_collection(..., py_to_c)

Linter and * imports changes are for code clarity.


python/src/opendp/_convert.py
```diff
@@ -1,100 +1,58 @@
...
 from opendp.typing import RuntimeType, RuntimeTypeDescriptor
+from opendp._convert_maps import (
+    ATOM_MAP,
+    INT_SIZES,
+    _NUMPY_COMPATIBLE_ATOM_TYPES,
+    check_c_int_cast,
+    _ConvertMaps,
+    _wrap_in_slice,
+    _check_polars_by,
+    _numpy_to_slice,
```
mccalluc, line 23: If we need to import these, then maybe they shouldn't be private? Or maybe we shouldn't touch them? Not sure which.

python/src/opendp/_convert.py
```diff
@@ -1,100 +1,58 @@
...
-    ty: c_int_limits(ty) for ty in (
-        'u8', 'u16', 'u32', 'u64', 'i8', 'i16', 'i32', 'i64', 'usize',
-    )
-}
 _ERROR_URL_298 = "https://github.com/opendp/opendp/discussions/298"
 
 
 def _check_and_cast_scalar(expected, value):
-    '''
+    """
```
mccalluc, line 30: I'm using black on other projects, but we don't enforce that here, and with all the other changes, might be better not to reformat more than necessary? (Don't revert this, but if you're going to redo this work in another PR, something to consider.)

python/src/opendp/_convert.py
```diff
@@ -295,60 +246,19 @@ def _slice_to_py(raw: FfiSlicePtr, type_name: Union[RuntimeType, str]) -> Any:
...
-
-        if type_name.origin == "Function":
-            return _slice_to_function(raw)
-
-        if type_name.origin == "HashMap":
-            return _slice_to_hashmap(raw)
+        if type_name in _ConvertMaps.FROM_SLICE_STR:
+            return _ConvertMaps.FROM_SLICE_STR[type_name](raw)
+        elif type_name in _NestedConvertMaps.FROM_SLICE_STR:
+            return _NestedConvertMaps.FROM_SLICE_STR[type_name](raw)
```
mccalluc, line 253:
```suggestion
        if type_name in _NestedConvertMaps.FROM_SLICE_STR:
            return _NestedConvertMaps.FROM_SLICE_STR[type_name](raw)
        raise UnknownTypeException("some message", type_name)  # pragma: no cover
```

This goes back to the original state, but consider handling the fall-through earlier, and providing a more informative error message.

python/src/opendp/_convert.py
```diff
@@ -922,17 +628,42 @@ def _slice_to_anyobject(raw: FfiSlicePtr):
...
-    obj.__class__ = ctypes.POINTER(AnyObject) # type: ignore[assignment]
-    
+    obj.__class__ = ctypes.POINTER(AnyObject)  # type: ignore[assignment]
+
     return ret
 
 
-def _wrap_in_slice(ptr, len_: int) -> FfiSlicePtr:
-    return FfiSlicePtr(FfiSlice(ctypes.cast(ptr, ctypes.c_void_p), len_))
+class _NestedConvertMaps:
```
mccalluc, line 636: I think there's a parallelism between this and `_ConvertMaps`? Maybe bring them closer together? If I'm working with one of these classes, do I want to think about the other?

python/src/opendp/_convert_maps.py
```diff
@@ -0,0 +1,421 @@
+from typing import Any, Sequence
```
mccalluc, line 1: Assuming this is copy-paste; Not looking at it closely.

mccalluc (commented):
Main comment is just to separate formatting changes from the code changes. I'd be happy to give a quick thumbs up to a formatting-only PR.

Shoeboxam:
@eddiestudies Why close?

Issue #2749, opened 2026-06-08
Enable DP Count with Negative Values


cbaycity:
In the following groupby query, the `dp.len()` method only returns count values greater than or equal to zero. This means that the noise is not centered at zero but above zero - which is a source of unwanted bias. Can we add an arg to the `dp.len` method to enable zero centered noise?

pseudo code:
Assume that `sample_df`, `privacy_unit`, and `privacy_budget` are correctly defined. Also assume that `buckets` are the possible values for the `bucket_col` that is grouped on. 

```python
context = dp.Context.compositor(
    data=sample_df.lazy(),
    privacy_unit=privacy_unit,
    privacy_loss=privacy_budget,
    split_evenly_over=1,
)
# Count how many rows map to each bucket (with DP noise)
bound_counts = (
    context.query()
    .group_by(bucket_col)
    .agg(dp.len())
    .with_keys(pl.DataFrame(buckets))
    .release()
    .collect()
    .sort(bounds_col)
)
```

Shoeboxam:
Could do something like `dp.len(signed=True)`, and adjust `def dp_len` to pass it through as an arg (not a kwarg, to avoid pickling). Then in `fn make_expr_dp_frame_len`, match through to `Some([signed, scale])` and use `len().cast(...)` instead of `len()` if signed is a true bool? Other count helpers could use similar adjustments.

Will also need to add support for a cast expr transformation. Since polars [doesn't support casting with saturation](https://github.com/pola-rs/polars/issues/21471), I think you'd cast with null and impute with max, and match/reason about stability for the cast+impute together as one op. Or maybe simpler, just cast into in Int64, to avoid the overflow handling.

This is the shape that comes to mind at first, but feel free to adjust as you get in.

PR #2761, opened 2026-06-12
feat: support signed Polars frame len via cast transformation

cbaycity:
Adds a cast measurement and adjusts the DPLen measurement to take in a signed arg to cast the len output as a float. This allows for DPLen calculations which don't truncate noise to zero. Truncating to zero introduces bias which users may or may not want. 

Issue: https://github.com/opendp/opendp/issues/2749


python/src/opendp/extras/polars/__init__.py
```diff
@@ -544,7 +545,7 @@ def dp_len(scale: float | None = None):
     return register_plugin_function(
         plugin_path=_get_opendp_polars_lib_path(),
         function_name="dp_frame_len",
-        args=(scale,),
+        args=(scale,allow_negative),
```
eddiestudies:
```suggestion
        args=(scale, allow_negative),
```

rust/src/measurements/make_private_expr/expr_len/mod.rs
```diff
@@ -41,9 +41,28 @@ pub fn make_expr_private_len<MI: 'static + UnboundedMetric, MO: 'static + Measur
...
+            options: _,
+        } => {
+            let len_expr = len().cast(DataType::Int64);
+            let output_type = 0i64;
+        }
+
+        _ => {
+            return fallible!(MakeMeasurement, "Expected len() or len().cast() expression");
+        }
+    }
```
Shoeboxam: This cheaply handles an edge case, but it should match specifically the `len().cast()` fragment, not just a `.cast()`, because a user could currently write `any_expr.cast()` and this constructor would rewrite it into a `len().cast()`.

rust/src/polars/mod.rs
```diff
@@ -477,9 +479,11 @@ impl DPExpr {
 ///
 /// # Arguments
 /// * `scale` - parameter for the noise distribution
-pub fn dp_len(scale: Option<f64>) -> Expr {
+/// * `allow_negative` - parameter to flip len output from unsigned int to signed int.
+pub fn dp_len(scale: Option<f64>, allow_negative: bool) -> Expr {
```
Shoeboxam: `signed: bool` strikes me as a cleaner API, but am open to other opinions.

python/test/test_polars.py
```diff
@@ -3,6 +3,7 @@
 import pytest
 import re
 import warnings
+import polars as pl
```
eddiestudies:
```suggestion
```

python/test/test_polars.py
```diff
@@ -1300,3 +1319,34 @@ def test_zero_budget():
         split_evenly_over=1,
     )
     context.query().select(pl.len()).release().collect()
+
+
+def test_unbiased_groupby_len():
+    """Tests that dp.len() can return counts centered on zero."""
+
```
eddiestudies:
```suggestion
    pl = pytest.importorskip("polars")
```

Shoeboxam (changes requested):
I'd introduce a separate constructor for the cast transformation. Like so (left inside a spoiler in case you want to solve it yourself :D)

<details>

```rust
pub fn make_expr_cast_u32_to_i64<MI, const P: usize>(
    input_domain: WildExprDomain,
    input_metric: L01InfDistance<MI>,
    expr: Expr,
) -> Fallible<
    Transformation<
        WildExprDomain,
        L01InfDistance<MI>,
        ExprDomain,
        LpDistance<P, f64>,
    >,
>
where
    MI: 'static + UnboundedMetric,
    (WildExprDomain, L01InfDistance<MI>): MetricSpace,
    (ExprDomain, LpDistance<P, f64>): MetricSpace,
    Expr: StableExpr<L01InfDistance<MI>, LpDistance<P, f64>>,
{
    let Expr::Cast {
        expr: input,
        dtype: to_type,
        mut options,
    } = expr
    else {
        return fallible!(MakeTransformation, "expected cast expression");
    };

    let to_type_dtype = to_type
        .as_literal()
        .ok_or_else(|| {
            err!(
                MakeTransformation,
                "u32-to-i64 cast only supports literal dtype"
            )
        })?
        .clone();

    if to_type_dtype != DataType::Int64 {
        return fallible!(
            MakeTransformation,
            "u32-to-i64 cast expects target dtype Int64, found {}",
            to_type_dtype
        );
    }

    if matches!(options, CastOptions::Strict) {
        options = CastOptions::NonStrict;
    }

    // This recursively stabilizes len/count/n_unique/count_null.
    // The child emits ExprDomain under LpDistance<P, f64>.
    let t_prior = input
        .as_ref()
        .clone()
        .make_stable(input_domain.clone(), input_metric.clone())?;

    let (middle_domain, middle_metric) = t_prior.output_space();

    let mut output_domain = middle_domain.clone();
    let active_series = &mut output_domain.column;

    let in_dtype = active_series.dtype().clone();
    if in_dtype != DataType::UInt32 {
        return fallible!(
            MakeTransformation,
            "u32-to-i64 cast expects UInt32 input, found {}",
            in_dtype
        );
    }

    active_series.set_element_domain(AtomDomain::<i64>::default());

    t_prior
        >> Transformation::new(
            middle_domain.clone(),
            middle_metric.clone(),
            output_domain,
            middle_metric,
            Function::then_expr(move |expr| Expr::Cast {
                expr: Arc::new(expr),
                dtype: to_type.clone(),
                options,
            }),
            StabilityMap::new(Clone::clone),
        )?
}
```

</details>

From writing this, I realize my earlier suggestion of a dedicated trait impl was off-base.

You basically just need to add:
```rust
            #[cfg(feature = "contrib")]
            Cast { .. } => expr_cast::make_expr_cast_u32_to_i64(
                input_domain,
                input_metric,
                self,
            ),
```
...to the `StableExpr<L01InfDistance<MI>, LpDistance<P, f64>>` dispatch/impl.

I guess in the case of a sum, the user can just cast the input dtype first to be signed. Otherwise if you wanted cast to work on int sums as well, you'd have to be a bit more general about the dtypes and maybe saturate/clip or disallow u64 input.

Looking good, thanks for digging on this!

Shoeboxam:
The reason for having the cast be separate is that it would work for all counting transformations (count, n_unique, count_null, expr len and frame len), not just frame len. I guess it would also work for u32 sum, as-is.

rust/src/polars/mod.rs
cbaycity: Ya that makes sense. Change made for this.

cbaycity:
> I'd introduce a separate constructor for the cast transformation. Like so (left inside a spoiler in case you want to solve it yourself :D)
> 
> Details
> ```rust
> pub fn make_expr_cast_u32_to_i64<MI, const P: usize>(
>     input_domain: WildExprDomain,
>     input_metric: L01InfDistance<MI>,
>     expr: Expr,
> ) -> Fallible<
>     Transformation<
>         WildExprDomain,
>         L01InfDistance<MI>,
>         ExprDomain,
>         LpDistance<P, f64>,
>     >,
> >
> where
>     MI: 'static + UnboundedMetric,
>     (WildExprDomain, L01InfDistance<MI>): MetricSpace,
>     (ExprDomain, LpDistance<P, f64>): MetricSpace,
>     Expr: StableExpr<L01InfDistance<MI>, LpDistance<P, f64>>,
> {
>     let Expr::Cast {
>         expr: input,
>         dtype: to_type,
>         mut options,
>     } = expr
>     else {
>         return fallible!(MakeTransformation, "expected cast expression");
>     };
> 
>     let to_type_dtype = to_type
>         .as_literal()
>         .ok_or_else(|| {
>             err!(
>                 MakeTransformation,
>                 "u32-to-i64 cast only supports literal dtype"
>             )
>         })?
>         .clone();
> 
>     if to_type_dtype != DataType::Int64 {
>         return fallible!(
>             MakeTransformation,
>             "u32-to-i64 cast expects target dtype Int64, found {}",
>             to_type_dtype
>         );
>     }
> 
>     if matches!(options, CastOptions::Strict) {
>         options = CastOptions::NonStrict;
>     }
> 
>     // This recursively stabilizes len/count/n_unique/count_null.
>     // The child emits ExprDomain under LpDistance<P, f64>.
>     let t_prior = input
>         .as_ref()
>         .clone()
>         .make_stable(input_domain.clone(), input_metric.clone())?;
> 
>     let (middle_domain, middle_metric) = t_prior.output_space();
> 
>     let mut output_domain = middle_domain.clone();
>     let active_series = &mut output_domain.column;
> 
>     let in_dtype = active_series.dtype().clone();
>     if in_dtype != DataType::UInt32 {
>         return fallible!(
>             MakeTransformation,
>             "u32-to-i64 cast expects UInt32 input, found {}",
>             in_dtype
>         );
>     }
> 
>     active_series.set_element_domain(AtomDomain::<i64>::default());
> 
>     t_prior
>         >> Transformation::new(
>             middle_domain.clone(),
>             middle_metric.clone(),
>             output_domain,
>             middle_metric,
>             Function::then_expr(move |expr| Expr::Cast {
>                 expr: Arc::new(expr),
>                 dtype: to_type.clone(),
>                 options,
>             }),
>             StabilityMap::new(Clone::clone),
>         )?
> }
> ```
> 
> From writing this, I realize my earlier suggestion of a dedicated trait impl was off-base.
> 
> You basically just need to add:
> 
> ```rust
>             #[cfg(feature = "contrib")]
>             Cast { .. } => expr_cast::make_expr_cast_u32_to_i64(
>                 input_domain,
>                 input_metric,
>                 self,
>             ),
> ```
> 
> ...to the `StableExpr<L01InfDistance<MI>, LpDistance<P, f64>>` dispatch/impl.
> 
> I guess in the case of a sum, the user can just cast the input dtype first to be signed. Otherwise if you wanted cast to work on int sums as well, you'd have to be a bit more general about the dtypes and maybe saturate/clip or disallow u64 input.
> 
> Looking good, thanks for digging on this!

This makes sense to me but I want to confirm some things that I'm slightly confused about. 
1. To create a constructor for make_expr_cast_u32_to_i64, would I be adding a module in rust/src/measurements/make_private_expr? If not, where should this be added?
2. There is a cast transformation in rust/src/transformations/cast, I assume that the difference is that transformations must happen before a private measurement. This means that we need to implement a version of the cast method you've described. I'm still trying to figure out where it goes. 
3. Assuming that this is adding a private measurement, is the thinking that we need a cast measurement because noise is added after a measurement, and this allows the input expr to be parsed and the end cast statement being supported.

cbaycity:
@Shoeboxam, One additional question. I am pretty sure that the query I'm trying to update in python, which is shown in the Python test that I've added, is creating a measurement from rust/src/measurements/make_private_expr/expr_dp_frame_len and not rust/src/measurements/make_private_expr/expr_len. That's because the dp.len() call creates a DPFrameLenShim. Is that correct?

Shoeboxam:
> To create a constructor for make_expr_cast_u32_to_i64, would I be adding a module in rust/src/measurements/make_private_expr? If not, where should this be added?

I'd add it in rust/src/transformations/make_stable_expr, since it is just a transformation. The dispatch is also off the StableExpr trait. No measurements anywhere here.

> There is a cast transformation in rust/src/transformations/cast, I assume that the difference is that transformations must happen before a private measurement. This means that we need to implement a version of the cast method you've described. I'm still trying to figure out where it goes.

That cast transformation is a row-by-row cast, whereas this is a cast on an aggregate. They are separate because they operate on different metric spaces, and thus have different privacy analyses, even though they share the same function.

> Assuming that this is adding a private measurement, is the thinking that we need a cast measurement because noise is added after a measurement, and this allows the input expr to be parsed and the end cast statement being supported.

Nope, a cast on an aggregate will just be a transformation, it will go like:

```
some expression -> cast -> noise
T-------------- -> T--- -> M---- (initial parse)
T---------------------- -> M---- (then chained)
M------------------------------- (then chained)
```

There's ultimately only ever one measurement in a chain and it consumes all. This aggregate cast transformation lives before the noise measurement.

Shoeboxam:
> @Shoeboxam, One additional question. I am pretty sure that the query I'm trying to update in python, which is shown in the Python test that I've added, is creating a measurement from rust/src/measurements/make_private_expr/expr_dp_frame_len and not rust/src/measurements/make_private_expr/expr_len. That's because the dp.len() call creates a DPFrameLenShim. Is that correct?

Yep, you're right on the money! Maybe a better name would be "Stub" not "Shim" if that makes it easier to think about.

Shoeboxam:
You'd end up in the `expr_len` measurement if your python were `pl.len()`. No noise added. Given your changes to that file, a user could write `pl.len().cast(...)` and it would still be recognized as a measurement. Really it could be registered as a postprocessor for maximum flexibility, but this is fine too.

cbaycity:
@shoeboxam Thank you for the good review. I am now passing the two tests that I added and I think that this PR is in good shape. Can you take another look?

python/src/opendp/extras/polars/__init__.py
```diff
@@ -510,12 +510,14 @@ def median(self, candidates: list[float], scale: float | None = None):
     pl.api.register_expr_namespace("dp")(DPExpr)
 
 
-def dp_len(scale: float | None = None):
+def dp_len(scale: float | None = None, signed: bool = False):
```
mccalluc, line 513: Is the `False` mostly for backward compatibility? Without that concern, would we prefer to make the default `True`?

python/test/test_polars.py
```diff
@@ -103,7 +103,8 @@ def test_lazyframe_ffi():
 
 
 @pytest.mark.parametrize(
-    "measure", [dp.max_divergence(), dp.zero_concentrated_divergence()],
+    "measure",
+    [dp.max_divergence(), dp.zero_concentrated_divergence()],
```
mccalluc: Prefer to avoid format changes in feature PRs. (Makes it that much harder to spot the real diff, or read the history.) But not worth reverting!

(If it's hard to work on this code without reformatting, that's a good thing for us to know! Would a more prescriptive style have helped you?)

mccalluc (commented):
This is a very cursory review, but I wonder if it would be better for the default to be not to clip? Avoiding bias seems like a good thing. But not a strong opinion.

(Leaving the Rust review to others.)

rust/src/measurements/make_private_expr/expr_len/mod.rs
cbaycity: I think I match this better by checking that a len().cast() is required. I don't know if there is another edge case where this can be broken but it seems like a reasonable check.

cbaycity:
> This is a very cursory review, but I wonder if it would be better for the default to be not to clip? Avoiding bias seems like a good thing. But not a strong opinion.
> 
> (Leaving the Rust review to others.)

I've discussed this with @Maegereg many years ago at Tumult where I argued that clipping would be a good default for entry level users who are confused by a negative count. But we agreed that avoiding bias is the best option, which is the default for Tumult Analytics. It is ironic that I'm now hitting a use case where I want negative counts, after finding it hard to believe I wouldn't want truncation. 

I think that the correct answer for the user experience is hard to determine. As it stands now though, it would be a breaking change to automatically adjust the return type to avoid bias - and I think that breaking changes should be avoided unless necessary. I'm therefore in favor of adding the signed variable and keeping the base logic the same for now - even though it makes this PR harder by requiring an arbitrary return type.

python/test/test_polars.py
cbaycity: This is a good catch and I can revert it, I generally agree with you. I think this got picked up when I ran black on the file. Do we enforce linting and formatting in the PR pipeline? If not maybe we should, and also provide a sample command in the developer tooling doc so that this is standardized.

rust/src/transformations/make_stable_expr/expr_cast_measurement/test.rs
```diff
@@ -0,0 +1 @@
+
```
eddiestudies: leftover file from an attempted test?

python/src/opendp/extras/polars/__init__.py
cbaycity: Yes this is for backward compatibility. I still think that it is worth not making a breaking change here. I'm going to resolve this comment in favor of discussing this point so that there aren't two comments discussing the same thing. [Link](https://github.com/opendp/opendp/pull/2761#issuecomment-4898941659) to the other comment.

Shoeboxam (changes requested):
We have two different positions for cast now:
* cast row-by-row: input and output metric are both dataset metrics, like symmetric distance
* cast aggregates: input and output metric are both lp-distance

The logic for casting an aggregate vector needs to be different and much pickier than the logic for casting a dataset column vector! 

In the case of casting vectors with stability wrt symmetric distance, casting can fail, so long as it doesn't raise an error. 
* Take for example two datasets: `x = [45, 12]; x' = [45, 12, 128]`. Their symmetric distance is 1.
* When casted to i8 with nonstrict, you get: `cast(x) = [45, 12]; cast(x') = [45, 12, null]`. Their symmetric distance is still just 1. More generally, we say the cast is 1-stable because for any choice of `x, x'` that differ by `d_in`, we have that the outputs differ by at most `d_out = 1 * d_in`.

In the case of casting vectors with stability wrt lp-distance, casts can never fail. 
* Take for example two datasets: `x = [45, 12, 127]; x' = [45, 12, 128]`. Their l1-distance is 1.
* When casted to i8 with nonstrict, you get: `cast(x) = [45, 12, 127]; cast(x') = [45, 12, null]`. Their l1-distance is now undefined! Infinite sensitivity.

So to be stable, you either you need the output dtype to totally enclose the input dtype:
* Allowed: `u8->i16`, `u32 -> u64`, `u8->i64`
* Disallowed: `u8->i8`, `u16->i8`

...or you need a 2-expression transformation that clips inputs to the output dtype range before casting, which is infallible and contractive. I think we can just stick to carefully chosen dtype pairs.

I left an earlier review on this PR with some model code. It wasn't as flexible as this, but it structurally eliminates cast failure by controlling in/out dtypes more tightly.

cbaycity:
> We have two different positions for cast now:
> 
> * cast row-by-row: input and output metric are both dataset metrics, like symmetric distance
> * cast aggregates: input and output metric are both lp-distance
> 
> The logic for casting an aggregate vector needs to be different and much pickier than the logic for casting a dataset column vector!
> 
> In the case of casting vectors with stability wrt symmetric distance, casting can fail, so long as it doesn't raise an error.
> 
> * Take for example two datasets: `x = [45, 12]; x' = [45, 12, 128]`. Their symmetric distance is 1.
> * When casted to i8 with nonstrict, you get: `cast(x) = [45, 12]; cast(x') = [45, 12, null]`. Their symmetric distance is still just 1. More generally, we say the cast is 1-stable because for any choice of `x, x'` that differ by `d_in`, we have that the outputs differ by at most `d_out = 1 * d_in`.
> 
> In the case of casting vectors with stability wrt lp-distance, casts can never fail.
> 
> * Take for example two datasets: `x = [45, 12, 127]; x' = [45, 12, 128]`. Their l1-distance is 1.
> * When casted to i8 with nonstrict, you get: `cast(x) = [45, 12, 127]; cast(x') = [45, 12, null]`. Their l1-distance is now undefined! Infinite sensitivity.
> 
> So to be stable, you either you need the output dtype to totally enclose the input dtype:
> 
> * Allowed: `u8->i16`, `u32 -> u64`, `u8->i64`
> * Disallowed: `u8->i8`, `u16->i8`
> 
> ...or you need a 2-expression transformation that clips inputs to the output dtype range before casting, which is infallible and contractive. I think we can just stick to carefully chosen dtype pairs.
> 
> I left an earlier review on this PR with some model code. It wasn't as flexible as this, but it structurally eliminates cast failure by controlling in/out dtypes more tightly.

This is an excellent description of the current problem with my PR. Thank you for writing it so clearly. I will add changes to deal with this tomorrow.

rust/src/transformations/make_stable_expr/expr_cast_aggregation/mod.rs
```diff
@@ -14,6 +16,29 @@ use crate::transformations::traits::UnboundedMetric;
 #[cfg(test)]
 mod test;
 
+// Constant hashmap of types to their allowed cast types.
+// This is necessary to prevent downcasting.
+static ALLOWED_TRANSFORMATIONS: LazyLock<HashMap<DataType, Vec<DataType>>> = LazyLock::new(|| {
```
Shoeboxam, line 21: This is a little over-engineered, but does the job. Another approach could be to match on the dtype to get i128 ranges, and check if the input range is contained in the output range. But this works too!

rust/src/transformations/make_stable_expr/expr_cast_aggregation/mod.rs
```diff
@@ -85,10 +110,21 @@ where
         DataType::Int16 => active_column.set_element_domain(AtomDomain::<i16>::default()),
         DataType::Int32 => active_column.set_element_domain(AtomDomain::<i32>::default()),
         DataType::Int64 => active_column.set_element_domain(AtomDomain::<i64>::default()),
-
         _ => unreachable!(),
```
Shoeboxam: This unreachable is not actually unreachable, it will panic if provided an unexpected output dtype. That's kind of intuitive for Rust, but in Python it kills the entire process, which feels much more unfriendly/surprising. Reusing the above helper, you could do something like this:

```rust
let Some(input_bounds) = integer_bounds(&input_dtype) else {
    return fallible!(
        MakeTransformation,
        "make_cast_aggregation expects an integer input dtype, found {}",
        input_dtype
    )
})?;
```
This handles the input dtype part of the check you had on line 116.

Then the output dtype check:
```rust
let input_is_contained =
    output_bounds.0 <= input_bounds.0 && input_bounds.1 <= output_bounds.1;

if !input_is_contained {
    return fallible!(
        MakeTransformation,
        "cast from {} to {} cannot preserve all possible values",
        input_dtype,
        to_type_dtype
    );
}
```

Then you could use a helper to set the output domain and avoid another match:
```rust
output_domain.column.set_dtype(to_type_dtype.clone())?;
```

rust/src/transformations/make_stable_expr/expr_cast_aggregation/mod.rs
Shoeboxam: Like so:
```rust
fn integer_bounds(dtype: &DataType) -> Option<(i128, i128)> {
    use DataType::*;

    Some(match dtype {
        UInt32 => (0, u32::MAX.into()),
        UInt64 => (0, u64::MAX.into()),
        Int8 => (i8::MIN.into(), i8::MAX.into()),
        Int16 => (i16::MIN.into(), i16::MAX.into()),
        Int32 => (i32::MIN.into(), i32::MAX.into()),
        Int64 => (i64::MIN.into(), i64::MAX.into()),
        _ => return None,
    })
}
```

Shoeboxam (dismissed):
For future reference, elsewhere the dtype promised by `DPLenShim` et al should technically change as well. Approving but leaving comments for future edits. 

Thanks for the contribution Baird!

rust/src/transformations/make_stable_expr/expr_cast_aggregation/mod.rs
Shoeboxam: Generally in Rust if you have to write unreachable!(), it's a hint that there's likely a cleaner code factoring hiding somewhere. :P

Shoeboxam (commented):
How about a note:

```
.. note::
Loading data is not covered by OpenDP's privacy guarantee and should be
performed by a trusted curator. CSV schema inference is data-dependent:
private values can affect both the inferred types and whether parsing succeeds. 
In a trusted environment, however, a parsing error can be useful,
since it may reveal malformed data or an incorrect schema 
before incorrect statistics are released.

If loading success or failure may be observed outside the trusted environment, 
prefer a schema-bearing source such as Parquet or a database table, 
or load columns as strings and cast them explicitly. 
``ignore_errors=True`` avoids some parsing failures, 
but may silently change the loaded data and reduce utility, 
so it should be used deliberately rather than by default. 
See the `Polars scan_csv documentation <https://docs.pola.rs/api/python/stable/reference/api/polars.scan_csv.html>`_ 
for the available schema and error-handling options.
```

rust/src/transformations/make_stable_expr/expr_cast_aggregation/mod.rs
cbaycity: This is a great comment, let me add this now.

rust/src/transformations/make_stable_expr/expr_cast_aggregation/mod.rs
cbaycity: I'm not sure I understand what you're proposed solution is. Are you saying that there should be a method that checks if the bounds of the new type contain the min and max values of the prior type? That makes sense and could be a much better change, do you think it would be this simple to implement?

rust/src/transformations/make_stable_expr/expr_cast_aggregation/mod.rs
cbaycity: I think I managed this well. Let me know if you agree?

rust/src/transformations/make_stable_expr/expr_cast_aggregation/mod.rs
cbaycity: And this sample method does the check by casting the input min and max to i128, casting the output dtype min and max to i128, and then ensuring that output min and max contain the input min and max?

IMO that isn't more simple. It is extendible to more type theoretically, but it also prevents us from adding support for larger integers. 

If we want this to be more extendible, should we support casting to any arbitrary type and have a better check? For instance, should we allow users to cast to a float? That seems valuable... Maybe we merge this as is and refactor to generalize to more types?

Shoeboxam (approved):
No more panics! Hooray!

PR #2767, opened 2026-06-17
testing: Expand unit test coverage in test_convert and test_convert_extras

eddiestudies:
This PR makes progress on converters being tested directly within unit tests.

### DIFF
--- a/python/src/opendp/_convert.py
+++ b/python/src/opendp/_convert.py
@@ -82,7 +82,7 @@ def _check_and_cast_scalar(expected, value):
 
     if expected not in ATOM_EQUIVALENCE_CLASSES.get(inferred, [inferred]):
         raise TypeError(f"inferred type is {inferred}, expected {expected}. See {_ERROR_URL_298}")
-    
+
     if expected in INT_SIZES:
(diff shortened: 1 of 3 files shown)


python/src/opendp/_convert.py
```diff
@@ -299,6 +298,7 @@ def _slice_to_py(raw: FfiSlicePtr, type_name: Union[RuntimeType, str]) -> Any:
             return _slice_to_bitvector(raw)
 
         if type_name == "ExtrinsicObject":
+            # TODO: Add test_convert.py unit test
```
mccalluc, line 301: I have preference to include an issue URL in TODOs.
- Not critical here, but the issue might provide more info than you want to add inline.
- Having the issue makes sure the problem is being tracked externally, and not just in source.

... but my sense is that you'll be staying on this track, so not essential in this case.

mccalluc (approved):
Looks good: I'll prefix the title with "testing".

python/src/opendp/_convert.py
eddiestudies: If any are leftover after my next PR I will create the issue and link them

PR #2768, opened 2026-06-18
CI: cache rust dependencies in CI

eddiestudies:


.github/workflows/smoke-test.yml
```diff
@@ -47,6 +47,14 @@ jobs:
         # fix formatting with: cargo fmt --manifest-path=rust/Cargo.toml --all
         run: cargo fmt --manifest-path=Cargo.toml --all --check
 
+      - uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
```
mccalluc, line 60: This might be useful, but it doesn't seem to line up with the title of the PR. The rust build would depend on _all_ the rust source files, and not just the `Cargo.lock`, right?

mccalluc (commented):
Maybe just retitle to "Feature: cache rust dependencies in CI", if that's what you mean?

(We have a one-word category at the start of PRs: These go into the git log, and then the changelog script splits off the first word to help us make categories.)

.github/workflows/smoke-test.yml
```diff
@@ -47,6 +47,14 @@ jobs:
         # fix formatting with: cargo fmt --manifest-path=rust/Cargo.toml --all
         run: cargo fmt --manifest-path=Cargo.toml --all --check
 
+      - uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
```
mccalluc: And in a comment, can you add a link to docs explaining how we know that this is all and only what is needed?

eddiestudies: Locally I can see files in these locations. However, build wise I will need to verify changes in future PRs once the cache is being used. 

My plan is to add dependency caching based off of the lock file in this PR. Then I will verify how that affects the build.

This is the safer caching change since it does not expose us to cache hits across PRs that touch rust code.

A follow up PR will cover the rust/ directory and be evaluated more thoroughly given the hashing needs to be more robust.

.github/workflows/smoke-test.yml
eddiestudies: Yes this only caches dependencies. I will be looking into its affect on builds after merging. If the speed up is not enough I will look into rust file caching with the drawback of a worse failure mode for caching bugs.

.github/workflows/smoke-test.yml
```diff
@@ -47,6 +47,18 @@ jobs:
         # fix formatting with: cargo fmt --manifest-path=rust/Cargo.toml --all
         run: cargo fmt --manifest-path=Cargo.toml --all --check
 
+        # Add rust caching following example:
+        #   https://github.com/actions/cache/blob/main/examples.md#rust---cargo
+      - uses: actions/cache@v4
```
mccalluc, line 52: bump to v5? but that can be a follow-up.

```suggestion
      - uses: actions/cache@v5
```

mccalluc (approved):
Great! 

(I tweaked the title, again, since this isn't really a user-facing feature. Sorry for the bad suggestions. We don't have to get the category exactly right, but it helps with the changelog and release notes.)

.github/workflows/smoke-test.yml
eddiestudies: Follow up while I check to see if it actually speeds up the builds. It suggested 3 or 4 in the initial failure, didn't catch that 5 was an option

PR #2770, opened 2026-06-18
CI: Enable compiler caching with sccache

eddiestudies:
Verified local builds are sped up significantly with sccache. 

Previous cache only saved 40 seconds off of the builds.

Reduces rust build to ~3 minutes from 6.5. ~50% improvement.

Follow up run ran in < 2 minutes


mccalluc (approved):
Looks good.

PR #2775, opened 2026-06-23
ux: add warning for large keys in opendp.extras.polars.with_keys

### LINKED ISSUE #2232: Warn when passing a large inline keys dataframe to `with_keys`

eddiestudies:
Resolves https://github.com/opendp/opendp/issues/2232

In repo replacement to: https://github.com/opendp/opendp/pull/2609

### DIFF
--- a/python/src/opendp/extras/polars/__init__.py
+++ b/python/src/opendp/extras/polars/__init__.py
@@ -52,7 +52,8 @@ if TYPE_CHECKING:  # pragma: no cover
     from opendp.context import Query
     from opendp.extras.polars.contingency_table import ContingencyTableQuery
 
-_KEY_SIZE_THRESHOLD_MB = 1 << 10
+# Exposed as global for testing.
+_KEY_SIZE_THRESHOLD_MB = 2 ** 10
 
 def _get_opendp_polars_lib_path():
     return os.environ.get("OPENDP_POLARS_LIB_PATH", lib_path)
--- a/python/src/opendp/extras/polars/__init__.py
+++ b/python/src/opendp/extras/polars/__init__.py
@@ -16,7 +16,6 @@ The members of this module will then be accessible at ``dp.polars``.
 from __future__ import annotations
 
 import os
-import sys
 
 from dataclasses import asdict, dataclass, field, replace
 from deprecated import deprecated
@@ -60,7 +59,7 @@ def _get_opendp_polars_lib_path():
 
 def _size_warning(keys):
     mb_factor = 1024**2  # For scaling to mb to shorten warnings
-    est_size: int = len(keys.serialize()) / mb_factor
+    est_size: float = len(keys.serialize()) / mb_factor
 
     if est_size > _KEY_SIZE_THRESHOLD_MB:
         warn(f"Large key-set (~{est_size}mb > {_KEY_SIZE_THRESHOLD_MB}mb) loaded into memory. Consider writing it to disk for the plan to read it in via scan_parquet.")
--- a/python/src/opendp/extras/polars/__init__.py
+++ b/python/src/opendp/extras/polars/__init__.py
@@ -58,11 +58,17 @@ def _get_opendp_polars_lib_path():
     return os.environ.get("OPENDP_POLARS_LIB_PATH", lib_path)
 
 def _size_warning(keys):
-    mb_factor = 1024**2  # For scaling to mb to shorten warnings
+    mb_factor = 1024**2  # bytes per MB
+
     est_size: float = len(keys.serialize()) / mb_factor
 
     if est_size > _KEY_SIZE_THRESHOLD_MB:
-        warn(f"Large key-set (~{est_size}mb > {_KEY_SIZE_THRESHOLD_MB}mb) loaded into memory. Consider writing it to disk for the plan to read it in via scan_parquet.")
+        warn(
+            f"Large key-set (~{est_size:.2f}MB > {_KEY_SIZE_THRESHOLD_MB}MB) "
+            "loaded into memory. Consider writing it to disk for the plan to "
+            "read it in via scan_parquet.",
+            stacklevel=3,
+        )
 
 class DPExpr(object):
     """
--- a/python/src/opendp/extras/polars/__init__.py
+++ b/python/src/opendp/extras/polars/__init__.py
@@ -13,45 +13,57 @@ We suggest importing under the conventional name ``dp``:
 
 The members of this module will then be accessible at ``dp.polars``.
 """
-
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


python/src/opendp/extras/polars/__init__.py
```diff
@@ -13,46 +13,57 @@
...
 from __future__ import annotations
-from dataclasses import asdict, dataclass, field, replace
+
 import os
-from typing import Any, Literal, Mapping, Optional, Sequence, Union, cast
+
+from dataclasses import asdict, dataclass, field, replace
+from deprecated import deprecated
+from typing import cast, Any, Literal, Mapping, Optional, Sequence, Union, TYPE_CHECKING
+from warnings import warn
```
mccalluc, line 23: typing, warnings, and dataclasses are all built-ins, so I'm not sure why this is separated from os? (This really, really doesn't matter, but if there's a preferred practice you have in mind, I'd like to make sure I'm not pulling against that.)

If this is something we care about, I'd suggest that we go with the ordering from isort. Our guidelines might be something like:
> We don't prescribe an order for imports, but if you're finding it untidy, please use the default order from isort, so we're not pulling in different directions.

python/src/opendp/extras/polars/__init__.py
```diff
@@ -13,46 +13,57 @@
...
-from opendp.measurements import make_private_lazyframe
-from deprecated import deprecated
 from opendp.transformations import make_stable_lazyframe
-from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:  # pragma: no cover
     from opendp.context import Query
     from opendp.extras.polars.contingency_table import ContingencyTableQuery
 
+_KEY_SIZE_THRESHOLD_MB = 1 << 10
```
mccalluc:
```suggestion
# Exposed as global for testing.
_KEY_SIZE_THRESHOLD_MB = 2 ** 10
```
- Explain why global (if my understanding is correct?)
- very slight preference for exponent for readability.

python/src/opendp/extras/polars/__init__.py
```diff
@@ -13,46 +13,57 @@
...
+
 def _get_opendp_polars_lib_path():
     return os.environ.get("OPENDP_POLARS_LIB_PATH", lib_path)
 
+def _size_warning(keys):
+    mb_factor = 1024**2  # For scaling to mb to shorten warnings
+    est_size: float = len(keys.serialize()) / mb_factor
```
mccalluc, line 64:
```suggestion
    est_size: float = len(keys.serialize()) // mb_factor
```
Integers better for the warning message.

python/src/opendp/extras/polars/__init__.py
```diff
@@ -13,46 +13,57 @@
...
+
+    if est_size > _KEY_SIZE_THRESHOLD_MB:
+        warn(f"Large key-set (~{est_size}mb > {_KEY_SIZE_THRESHOLD_MB}mb) loaded into memory. Consider writing it to disk for the plan to read it in via scan_parquet.")
```
mccalluc: We don't have any examples for how to use `scan_parquet`. I think this is a part of the analysis setup that we'd like users to know about, rather than just hitting the warning at run time. I'll file a docs issue.

python/test/test_polars.py
```diff
@@ -1054,6 +1054,31 @@ def test_explicit_grouping_keys_context():
...
+        split_evenly_over=1,
+        domain=lf_domain,
+    )
+
+    # Negative test to verify small keys does not fail
+    keys_small = pl.DataFrame(pl.Series("B", [2, 3, 4, 5, 6], dtype=pl.Int32))
+    context.query().group_by("B").agg(pl.col("D").dp.sum((0, 10))).with_keys(keys_small)
+
+    keys_large = pl.DataFrame([pl.Series(str(x), [2, 3, 4, 5, 6], dtype=pl.Int32) for x in range(local_scale_factor)])
+    with pytest.warns(UserWarning):
```
mccalluc: Slight preference to also check the message.

mccalluc (commented):
None of these comments are essential. I think you still want an approval from Mike, and that will be enough to merge; I don't need to re-review.

python/src/opendp/extras/polars/__init__.py
eddiestudies: I used to follow something akin to https://realpython.com/ref/best-practices/imports/ .

import os  # does not use from or as

from os import environ. # separate from imports

from mystuff import things. # our library imports.

There is a internal vs external convention that I dont have instinctively memorized but I suspect internal  last. Not a global thing but I internalized it

python/src/opendp/extras/polars/__init__.py
eddiestudies: This would no longer be a float. Tests made it 0 if this was int casted

python/src/opendp/extras/polars/__init__.py
eddiestudies: sounds good. Definitely necessary

python/test/test_polars.py
eddiestudies: good point. I'll look into adding that

python/src/opendp/extras/polars/__init__.py
Shoeboxam: `mb` -> `MiB` (sorry for the nit!)

Could also use stacklevel=3 to cut out the warning prep cruft from the error message.

Shoeboxam (changes requested):
Chuck left good comments, and I found a couple little things. Lets clean this up and then merge. Approach looks well executed!

python/src/opendp/extras/polars/__init__.py
eddiestudies: Added both. No worries on the nit(I had no idea MiB was the correct term there)

python/src/opendp/extras/polars/__init__.py
eddiestudies: Formatted the string to keep the string cleaner as per the spirit of the feedback. Keeping the float to maintain fractional errors for testing speed

python/src/opendp/extras/polars/__init__.py
mccalluc: @eddiestudies , Ok to add the comment? I'm imagining someone inherits this, sees this as something that could be simplified, and then hits the test. Not a big deal if you disagree.

mccalluc (dismissed):
Looks good!

Shoeboxam (approved):
MiB technically, but whatever let's ship it. :P

PR #2779, opened 2026-06-29
refactor: Move polars accuracy implementation under polars/ + stub

eddiestudies:
In an effort to document the existing polars code within opendp/rust/, this PR migrates the accuracy code into polars/accuracy/.

This is the first of a series of PRs grouping together the polars code and expanding rust based test coverage.


rust/src/accuracy/polars/mod.rs
```diff
@@ -1,35 +1,16 @@
+// Staging move of polars code into polars/.
+// Keeping stub to maintain generated python as is
+// TODO: Remove once migration and derive.rs updates are completed
 use opendp_derive::bootstrap;
-use polars::{
-    datatypes::{AnyValue, DataType, Field},
```
eddiestudies, line 3: moved to polars/accuracy/mod.rs.

Code left in this file is the stub that calls the original implementation

rust/src/polars/accuracy/mod.rs
```diff
@@ -0,0 +1,282 @@
+use polars::{
```
eddiestudies, line 1: missing use opendp_derive::bootstrap; because it is handled in the stub.

rust/src/polars/accuracy/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+
+/// Summarize the statistics to be released from a measurement that returns a OnceFrame.
+///
+/// If a threshold is configured for censoring small/sensitive groups,
+/// a threshold column will be included,
+/// containing the cutoff for the respective count query being thresholded.
+///
+/// # Arguments
+/// * `measurement` - computation from which you want to read noise scale parameters from
+/// * `alpha` - optional statistical significance to use to compute accuracy estimates
```
eddiestudies, line 40: Missing #[bootstrap( ...) ] because it is handled in the stub

rust/src/accuracy/test.rs
```diff
@@ -314,6 +314,19 @@ pub fn test_empirical_discrete_gaussian_accuracy() -> Fallible<()> {
     Ok(())
 }
 
+#[test]
+pub fn test_discrete_gaussian_scale_to_accuracy_rejects_invalid() {
+    // alpha == 1 is on the inclusive bound, so it stays valid.
+    assert!(discrete_gaussian_scale_to_accuracy(1.0, 1.0).is_ok());
+
+    // out-of-range inputs used to silently misbehave; now they error cleanly.
```
Shoeboxam, line 322: Good to add these checks, thanks!

Shoeboxam (approved):
I thought it might be better to adjust bootstrap to allow saying something like module = "accuracy" so that you don't need the boilerplate stubs. Then you'd make this pub(crate) and then just pub use the function from accuracy.

But that doesn't actually work because you can't pub use an item with only pub(crate) visibility. So if you want to keep the API stable, you either need a stub, or you need to have two API surfaces. We'll have to decide wether we want to introduce a backwards incompat, or if we want to keep a bunch of stubs around forever.

Maybe the best approach is to commit to publicizing the new polars module as the source of truth, double up the bootstraps temporarily, and add #[deprecated] to the stubs. Open to opinions. This PR doesn't commit us to any one approach yet.

PR #2780, opened 2026-07-01
dev: add opt-in pre-commit hooks

eddiestudies:


mccalluc (commented):
I would rather not add new bash scripts to the repo. Would https://github.com/ry-sun/rust-pre-commit-hooks#code-formatting work? With the other repos that are using githooks, I've been happy with https://pre-commit.com/, and generally reusing tools rather than writing new wrappers.

(I'll put some comments on these files that can be taken as style preferences, but I think it might be good to start a new smaller PR.)

.hooks-suggested/pre-commit
```diff
@@ -0,0 +1,33 @@
+#!/bin/sh
+#
+# Suggested pre-commit hook: checks Rust formatting before a commit.
+#
+# This is opt-in. Enable it with:
+#   git config core.hooksPath .githooks
+# Disable it again with:
+#   git config --unset core.hooksPath
```
mccalluc, line 8: (Take this and all that follows as style preferences, but not relavent if you don't go this direction.)

The situation seems analogous to `.vscode-suggested`, so maybe it could be `hooks-suggested/pre-commit.sh`.

.githooks/README.md
```diff
@@ -0,0 +1,34 @@
...
+```
+git config --unset core.hooksPath
+```
+
+Hooks are added in .githooks/ to make them opt in.
+
+## `pre-commit`
+
+Runs `cargo fmt --all -- --check` on the opendp/rust/
+Fix with `cargo fmt --manifest-path rust/Cargo.toml --all`.
```
mccalluc: If a description is necessary, keep it at a high level. This gives us another thing that can get out of sync.

.githooks/README.md
```diff
@@ -0,0 +1,34 @@
...
+```
+git restore --staged rust/src/...
+```
+
+Git can only run the pre-commit hook if the file allows for execution.
+The following command allows the file to be used as an executable.
+
+```
+chmod +x .githooks/pre-commit
+```
```
mccalluc: Is it likely to lose the x bit? Preference for simpler.

docs/source/contributing/development-environment.rst
```diff
@@ -417,3 +417,14 @@ A few notes on Intellij IDEA:
...
+
+Git Hooks
+
+A suggested (opt-in) pre-commit hook lives in ``.githooks``.
+Enable it with ``git config core.hooksPath .githooks``; see ``.githooks/README.md`` for details.
+
+Git only runs the hook if it is executable. A fresh clone is fine, but if the executable
+bit was stripped, git will skip the hook or complain it is not executable.
+
+Restore it with ``chmod +x .githooks/pre-commit``
```
mccalluc: duplicate. Once at most?

docs/source/contributing/development-environment.rst
```diff
@@ -417,3 +417,14 @@ A few notes on Intellij IDEA:
 * Be sure to install the Python and Rust plugins for interactivity
 * Be sure to "attach" the Cargo.toml in the red banner the first time you open a Rust source file
 * Use run configurations to `build the Rust library <https://plugins.jetbrains.com/plugin/8182-rust/docs/cargo-command-configuration.html#cargo-command-config>`_ and run tests
+
+
+Git Hooks
+
+A suggested (opt-in) pre-commit hook lives in ``.githooks``.
+Enable it with ``git config core.hooksPath .githooks``; see ``.githooks/README.md`` for details.
```
mccalluc: If details are needed, suggest a link to the file on github, instead of just the filename.

docs/source/contributing/development-environment.rst
```diff
@@ -417,3 +417,14 @@ A few notes on Intellij IDEA:
 * Be sure to install the Python and Rust plugins for interactivity
 * Be sure to "attach" the Cargo.toml in the red banner the first time you open a Rust source file
 * Use run configurations to `build the Rust library <https://plugins.jetbrains.com/plugin/8182-rust/docs/cargo-command-configuration.html#cargo-command-config>`_ and run tests
+
+
+Git Hooks
```
mccalluc, line 422:
```suggestion
Git Hooks
---------
```
RST formatting

mccalluc (commented):
Formatting comments. I'd wait for Mike's thoughts: If he's fine with it, I don't need to block.

eddiestudies:
> I would rather not add new bash scripts to the repo. Would https://github.com/ry-sun/rust-pre-commit-hooks#code-formatting work? With the other repos that are using githooks, I've been happy with https://pre-commit.com/, and generally reusing tools rather than writing new wrappers.
> 
> (I'll put some comments on these files that can be taken as style preferences, but I think it might be good to start a new smaller PR.)

I would hesitate to link external bash files that run before each commit. If they are ever maliciously updated that can cause issues. We could pin them to a specific hash as a possible approach. 

If the issue is with bash I can convert this to a python file execution where the bash file calls the python file directly like test_subprocess.py

mccalluc:
> If the issue is with bash I can convert this to a python file execution where the bash file calls the python file directly like test_subprocess.py

Bash isn't a show-stopper. I'm mostly concerned with adding more code in any language, if there is a configuration-based alternative. But since you've looped in Mike for review, let's see what he thinks: Might put this on the agenda for the weekly developer call, if he doesn't reply before then. (Sorry if this is bikeshedding.)

eddiestudies:
> > If the issue is with bash I can convert this to a python file execution where the bash file calls the python file directly like test_subprocess.py
> 
> Bash isn't a show-stopper. I'm mostly concerned with adding more code in any language, if there is a configuration-based alternative. But since you've looped in Mike for review, let's see what he thinks: Might put this on the agenda for the weekly developer call, if he doesn't reply before then. (Sorry if this is bikeshedding.)

I was looking to add flake8 and mypy since I have noticed builds tend to fail with mypy at build time instead of commit time. Maybe the checks can be lifted from the test_subprocess.py code so that it has a single source. I can get the repo root  to use .hooks-suggested/pre-commit. And update it to mirror the current subprocess logic + fmt.

Mike was looped in as suggested. I believe his concerns were addressed when the script became optional.

docs/source/contributing/development-environment.rst
```diff
@@ -417,3 +417,17 @@ A few notes on Intellij IDEA:
...
+---------
+
+A suggested (opt-in) pre-commit hook lives in ``.hooks-suggested``.
+Enable it with ``git config core.hooksPath .hooks-suggested``.
+`See <https://github.com/opendp/opendp/tree/main/.hooks-suggested/README.md>` for more details.
+md
```
mccalluc, line 428:
```suggestion
```
Stray line?

.hooks-suggested/pre-commit
```diff
@@ -0,0 +1,33 @@
+#!/bin/sh
+#
+# Suggested pre-commit hook: checks Rust formatting before a commit.
+#
+# This is opt-in. Enable it with:
+#   git config core.hooksPath .hooks-suggested
+# Disable it again with:
+#   git config --unset core.hooksPath
```
mccalluc, line 8:
```suggestion
```

Let's just keep the user documentation in one place. Worry about drift if it's repeated.

.hooks-suggested/README.md
```diff
@@ -0,0 +1,26 @@
+# Suggested git hooks
+
+Optional git hooks for rust fmt.
+
+```
+git config core.hooksPath .hooks-suggested
```
mccalluc, line 6: Just drop this. Covered by the RST.

.hooks-suggested/README.md
```diff
@@ -0,0 +1,26 @@
+# Suggested git hooks
+
+Optional git hooks for rust fmt.
```
mccalluc, line 3:
```suggestion
Optional git hooks.
```

The plan is to add mypy as well?

docs/source/contributing/development-environment.rst
```diff
@@ -417,3 +417,17 @@ A few notes on Intellij IDEA:
 * Be sure to install the Python and Rust plugins for interactivity
 * Be sure to "attach" the Cargo.toml in the red banner the first time you open a Rust source file
 * Use run configurations to `build the Rust library <https://plugins.jetbrains.com/plugin/8182-rust/docs/cargo-command-configuration.html#cargo-command-config>`_ and run tests
+
+
+Git Hooks
+---------
+
+A suggested (opt-in) pre-commit hook lives in ``.hooks-suggested``.
+Enable it with ``git config core.hooksPath .hooks-suggested``.
```
mccalluc, line 426:
```suggestion
Enable it with ``cp .hooks-suggested/pre-commit .git/hooks/``
```
If we are thinking of the pre-commits as something that is left to the preference of individual developers, slight preference for `cp`. But if we want to encourage the same patterns across the team, this is better. 

Decided I don't really care.

mccalluc (commented):
Last round of comments. Only big concern is repeating the developer instructions across the files: I worry about maintenance.

Checking my understanding, there will be a follow-up to add mypy and other format checks, but this is a start?

mccalluc:
Removing Mike as reviewer: His only preference is that it not be required.

PR #2792, opened 2026-07-14
feat: Add make_fully_adaptive_composition_k  to improve DP-SGD performance.

eddiestudies:
The current make_fully_adaptive_composition implementation evaluates each measurement separately. However, for DP-SGD, grouping the identical maps and evaluating once per order instead of per step reaches the same result.

make_fully_adaptive_composition_k adds this optimization with support for multiple privacy maps each with it's own k_i iterations. 

Rather than multiplying out the cost(as is done in opacus) we inf_add them to account for floating point sensitivity.

A further optimization is considered that performs the inf_adds with O(log_2(k)) adds. It leverages Double-and-add, similar to how shift-and-add works in hardware multipliers. It is implemented and tested as inf_add_k_doubling but not adopted yet.

It causes a discrepancy between compose and compose_k since there is less floating point exposure.

### DIFF
--- a/rust/src/combinators/sequential_composition/ffi.rs
+++ b/rust/src/combinators/sequential_composition/ffi.rs
@@ -5,7 +5,7 @@ use crate::{
     measures::{Approximate, MaxDivergence, RenyiDivergence, ZeroConcentratedDivergence},
 };
 
-use super::{Adaptivity, ComposeK, CompositionMeasure};
+use super::{Adaptivity, CompositionMeasure};
 
 impl CompositionMeasure for AnyMeasure {
     fn composability(&self, adaptivity: Adaptivity) -> Fallible<Composability> {
@@ -42,29 +42,5 @@ impl CompositionMeasure for AnyMeasure {
         dispatch!(monomorphize, [
             (self.type_, [MaxDivergence, Approximate<MaxDivergence>, ZeroConcentratedDivergence, Approximate<ZeroConcentratedDivergence>, RenyiDivergence])
         ], (self, d_i))
-    }
-}
-
-impl ComposeK for AnyMeasure {
-    fn compose_k(&self, d_mid: Self::Distance, k: u32) -> Fallible<Self::Distance>
-    where
-        Self::Distance: Clone,
-    {
-        fn monomorphize<M: 'static + ComposeK>(
-            self_: &AnyMeasure,
-            d_mid: AnyObject,
-            k: u32,
-        ) -> Fallible<AnyObject>
-        where
-            M::Distance: Clone,
-        {
-            self_
-                .downcast_ref::<M>()?
-                .compose_k(d_mid.downcast_ref::<M::Distance>()?.clone(), k)
-                .map(AnyObject::new)
-        }
-        dispatch!(monomorphize, [
-            (self.type_, [MaxDivergence, Approximate<MaxDivergence>, ZeroConcentratedDivergence, Approximate<ZeroConcentratedDivergence>, RenyiDivergence])
-        ], (self, d_mid, k))
     }
 }
--- a/rust/src/combinators/sequential_composition/ffi.rs
+++ b/rust/src/combinators/sequential_composition/ffi.rs
@@ -5,7 +5,7 @@ use crate::{
     measures::{Approximate, MaxDivergence, RenyiDivergence, ZeroConcentratedDivergence},
 };
 
-use super::{Adaptivity, CompositionMeasure};
+use super::{Adaptivity, ComposeK, CompositionMeasure};
 
 impl CompositionMeasure for AnyMeasure {
     fn composability(&self, adaptivity: Adaptivity) -> Fallible<Composability> {
@@ -42,5 +42,29 @@ impl CompositionMeasure for AnyMeasure {
         dispatch!(monomorphize, [
             (self.type_, [MaxDivergence, Approximate<MaxDivergence>, ZeroConcentratedDivergence, Approximate<ZeroConcentratedDivergence>, RenyiDivergence])
         ], (self, d_i))
+    }
+}
+
+impl ComposeK for AnyMeasure {
+    fn compose_k(&self, d_mid: Self::Distance, k: u32) -> Fallible<Self::Distance>
+    where
+        Self::Distance: Clone,
+    {
+        fn monomorphize<M: 'static + ComposeK>(
+            self_: &AnyMeasure,
+            d_mid: AnyObject,
+            k: u32,
+        ) -> Fallible<AnyObject>
+        where
+            M::Distance: Clone,
+        {
+            self_
+                .downcast_ref::<M>()?
+                .compose_k(d_mid.downcast_ref::<M::Distance>()?.clone(), k)
+                .map(AnyObject::new)
+        }
+        dispatch!(monomorphize, [
+            (self.type_, [MaxDivergence, Approximate<MaxDivergence>, ZeroConcentratedDivergence, Approximate<ZeroConcentratedDivergence>, RenyiDivergence])
+        ], (self, d_mid, k))
     }
 }
--- a/rust/src/combinators/sequential_composition/ffi.rs
+++ b/rust/src/combinators/sequential_composition/ffi.rs
@@ -22,10 +22,10 @@ impl CompositionMeasure for AnyMeasure {
             (self.type_, [MaxDivergence, Approximate<MaxDivergence>, ZeroConcentratedDivergence, Approximate<ZeroConcentratedDivergence>, RenyiDivergence])
         ], (self, adaptivity))
     }
-    fn compose(&self, d_i: Vec<Self::Distance>) -> Fallible<Self::Distance> {
+    fn compose(&self, d_i: Vec<(Self::Distance, u32)>) -> Fallible<Self::Distance> {
         fn monomorphize<M: 'static + CompositionMeasure>(
             self_: &AnyMeasure,
-            d_i: Vec<AnyObject>,
+            d_i: Vec<(AnyObject, u32)>,
         ) -> Fallible<AnyObject>
         where
             M::Distance: Clone,
@@ -34,8 +34,8 @@ impl CompositionMeasure for AnyMeasure {
                 .downcast_ref::<M>()?
                 .compose(
                     d_i.iter()
-                        .map(|d_i| d_i.downcast_ref::<M::Distance>().map(Clone::clone))
-                        .collect::<Fallible<Vec<M::Distance>>>()?,
+                        .map(|(d_i, k_i)| Ok((d_i.downcast_ref::<M::Distance>()?.clone(), *k_i)))
+                        .collect::<Fallible<Vec<(M::Distance, u32)>>>()?,
                 )
                 .map(AnyObject::new)
         }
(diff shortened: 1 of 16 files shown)


rust/src/combinators/sequential_composition/fully_adaptive/mod.rs
```diff
@@ -174,3 +177,123 @@ where
...
+    input_domain: DI,
+    input_metric: MI,
+    output_measure: MO,
+    data: DI::Carrier,
+) -> Fallible<OdometerQueryable<Measurement<DI, MI, MO, TO>, TO, MI::Distance, MO::Distance>>
+where
+    MO::Distance: Clone,
+    (DI, MI): MetricSpace,
+{
+    let mut inner = new_fully_adaptive_composition_queryable(
```
eddiestudies: To avoid running unproven code this PR leverages new_fully_adaptive_composition_queryable internally

rust/src/combinators/sequential_composition/fully_adaptive/mod.rs
```diff
@@ -174,3 +177,123 @@ where
...
+                        let group = (privacy_maps.iter_mut())
+                            .find(|(map, _)| Arc::ptr_eq(&map.0, &meas.privacy_map.0));
+                        match group {
+                            Some((_, k)) => *k += 1,
+                            None => privacy_maps.push((meas.privacy_map.clone(), 1)),
+                        }
+                    }
+                    answer
+                }
+                Query::Internal(query) => inner.eval_query(Query::Internal(query))?,
```
eddiestudies: Fallback to inner if unsupported

eddiestudies: Given compose_k is an optimization, this seemed like a good fallback

rust/src/combinators/sequential_composition/fully_adaptive/test.rs
```diff
@@ -95,3 +96,177 @@ fn test_fully_adaptive_interactive_postprocessing() -> Fallible<()> {
     assert_eq!(qbl.privacy_loss(1)?, (1.0, 1e-7));
     Ok(())
 }
+
+/// Under a Sequential measure, spawning a new query locks earlier interactive children.
+fn assert_sequentiality_enforced(
```
eddiestudies: These tests failed with the ? from the original implementation. Added to ensure Sequential did not break under compose_k

rust/src/combinators/sequential_composition/ffi.rs
eddiestudies, line 32: Originally wanted to add compose_k here. But it would have a proof implication and  that seemed worth avoiding for now. The below code mostly goes away for an internal call to self.compose if moved here

rust/src/combinators/sequential_composition/fully_adaptive/mod.rs
```diff
@@ -125,7 +126,9 @@ where
                             enforce_sequentiality,
                             move || {
                                 if *enforce_sequentiality.borrow() {
-                                    self_.eval_internal(&AskPermission(child_id))?
+                                    // no ? operator: it would make inference expect a Result answer payload,
+                                    // implementation now matches adaptive/mod.rs
+                                    self_.eval_internal(&AskPermission(child_id))
```
eddiestudies, line 135: adaptive/mod.rs example:

[`if *enforce_sequentiality.borrow() {
                                    self_.eval_internal(&AskPermission(child_id))}`](https://github.com/opendp/opendp/blob/main/rust/src/combinators/sequential_composition/adaptive/mod.rs) line 157

eddiestudies:
Moving into fully_adaptive as per offline discussion with @Shoeboxam

rust/src/combinators/sequential_composition/fully_adaptive/mod.rs
```diff
@@ -147,9 +153,22 @@ where
                 }
                 // evaluate external map query
                 Query::External(OdometerQuery::PrivacyLoss(d_in)) => {
-                    let d_mids = (privacy_maps.iter())
-                        .map(|m| m.eval(d_in))
-                        .collect::<Fallible<_>>()?;
+                    // evaluate each distinct privacy map once, reusing the distance for repeats,
```
eddiestudies: change this to also have the optimization and change the spec

eddiestudies: take state and pass to compositor when it is adjusted

rust/src/combinators/sequential_composition/mod.rs
```diff
@@ -118,10 +119,21 @@ impl CompositionMeasure for RenyiDivergence {
     }
 
     fn compose(&self, d_mids: Vec<Self::Distance>) -> Fallible<Self::Distance> {
+        // curves sharing an allocation are grouped as one (curve, k) pair, in first-occurrence order,
+        // so that each distinct curve is evaluated once, not once per copy
+        let mut groups: Vec<(Self::Distance, u32)> = Vec::new();
+        for d_mid in d_mids {
+            let group = (groups.iter_mut())
+                .find(|(curve, _)| Arc::ptr_eq(&curve.function, &d_mid.function));
```
eddiestudies: implement Eq for function so that pointer does not exist here. Compositor checks for equality

eddiestudies: maybe hash

eddiestudies: hash and eq

rust/src/combinators/sequential_composition/fully_adaptive/mod.rs
Shoeboxam: This is a good change/catch.

rust/src/combinators/sequential_composition/fully_adaptive/mod.rs
```diff
@@ -140,15 +147,21 @@ where
                     *enforce_sequentiality.borrow_mut() = true;
 
                     // we've now increased our privacy spend. This is our only state modification
-                    privacy_maps.push(meas.privacy_map.clone());
+                    num_queries += 1;
```
Shoeboxam, line 150: Super finnicky, but prefer checked add.

rust/src/combinators/sequential_composition/fully_adaptive/mod.rs
```diff
@@ -140,15 +147,21 @@ where
...
+                    // each query joins the group of queries sharing its privacy map
+                    match (privacy_maps.iter_mut()).find(|(map, _)| map == &meas.privacy_map) {
```
Shoeboxam, line 152: Why not use a hashmap here? The hashmap would keep composition O(1) per invoke instead of O(m).

Shoeboxam: Also look into the entry api. Something along the lines of x.entry(meas.privacy_map).or_default()

rust/src/combinators/sequential_composition/mod.rs
```diff
@@ -45,12 +45,13 @@ pub enum Composability {
 }
 
 /// # Proof Definition
-/// `composability` returns `Ok(out)` if the composition of a vector of privacy parameters `d_mids`
+/// `composability` returns `Ok(out)` if the composition of a vector of privacy parameters `d_mids`,
+/// where each parameter `d_mid_i` is charged with multiplicity `k_i`,
```
Shoeboxam, line 49: We need it to more specifically say that the privacy definition's composition is commutative, because we're exploiting that now.

rust/src/combinators/sequential_composition/mod.rs
```diff
@@ -117,11 +126,20 @@ impl CompositionMeasure for RenyiDivergence {
...
+    fn compose(&self, d_mids: Vec<(Self::Distance, u32)>) -> Fallible<Self::Distance> {
+        // equal curves are merged into a single group, in first-occurrence order,
+        // so that each distinct curve is evaluated once, not once per copy
+        let mut groups: Vec<(Self::Distance, u32)> = Vec::new();
+        for (d_mid, k_i) in d_mids {
+            match (groups.iter_mut()).find(|(curve, _)| curve == &d_mid) {
+                Some((_, k)) => *k += k_i,
+                None => groups.push((d_mid, k_i)),
+            }
+        }
```
Shoeboxam, line 138: Using a hashmap as a backer would let this be dropped?

rust/src/core/mod.rs
```diff
@@ -81,6 +81,23 @@ impl<TI, TO> Clone for Function<TI, TO> {
...
+    }
+}
+impl<TI, TO> Eq for Function<TI, TO> {}
+
+impl<TI, TO> std::hash::Hash for Function<TI, TO> {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        (Arc::as_ptr(&self.function) as *const ()).hash(state)
+    }
+}
+
```
Shoeboxam, line 100: Why add this?

rust/src/core/mod.rs
```diff
@@ -143,6 +160,23 @@ impl<MI: Metric, MO: Measure> Clone for PrivacyMap<MI, MO> {
...
+    }
+}
+impl<MI: Metric, MO: Measure> Eq for PrivacyMap<MI, MO> {}
+
+impl<MI: Metric, MO: Measure> std::hash::Hash for PrivacyMap<MI, MO> {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        (Arc::as_ptr(&self.0) as *const ()).hash(state)
+    }
+}
+
```
Shoeboxam, line 179: This is currently unused, but if you switched to a hashmap-based budget store for faster composition, then you'd have to keep it.

rust/src/data/ffi/mod.rs
```diff
@@ -1231,6 +1231,9 @@ impl Clone for AnyObject {
                     } else {
                         dispatch!(clone_hashmap, [(K, @hashable), (V, @primitives)], (self))
                     }
+                } else if *name == "Function" {
+                    // privacy curves, e.g. Rényi divergence distances; a Function clone shares the allocation
+                    dispatch!(clone_plain, [(self.type_, [Function<f64, f64>])], (self))
```
Shoeboxam, line 1236: Vestigial?

Shoeboxam (changes requested):
Noticed some regressions!

rust/src/data/ffi/mod.rs
eddiestudies: I ran into a failure without this. Somewhere in the ffi layer. I can undo it to get the failure. I cannot recall exactly what it was but this resolved it

eddiestudies: I removed the comment since it is useful outside of the compose k context

rust/src/core/mod.rs
eddiestudies: previouse compose_k version used this for dictionary key handling. It was required again now when we added OrderedCounts

rust/src/combinators/sequential_composition/fully_adaptive/mod.rs
eddiestudies: ran into some ordering issues. used OrderedCounts to have both o(1) invoke and no floating point ordering variance. At the cost of o(m) space

rust/src/combinators/sequential_composition/fully_adaptive/mod.rs
eddiestudies: I added them throughout. Then moved it into OrderedCount. The alternative had a weird corner case where we would want to check before incrementing that I did not like

PR #2794, opened 2026-07-16
fix: allow test_subprocess to run from any dir

eddiestudies:
Running uv run pytest python/ from root causes test_subprocess to find documentation and release code that it is not configured to support.

To avoid this, the change made fixes the cwd to python/ using the relative location of that directory from the test file.

python/test/test_subprocess.py/.parent(= test/).parent(= python/)

### DIFF
--- a/python/test/test_subprocesses.py
+++ b/python/test/test_subprocesses.py
@@ -1,5 +1,7 @@
 import subprocess
 import sys
+from pathlib import Path
+
 import pytest
 
 tests = {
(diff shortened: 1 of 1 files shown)


mccalluc (commented):
Great... but R is failing for an unrelated reason. I just restarted that job, just in case the error is something sporadic.

I'll take a look at the R: Hopefully it's a simple fix, and then when main is green again, main can be merged to this branch.

eddiestudies:
> Great... but R is failing for an unrelated reason. I just restarted that job, just in case the error is something sporadic.
> 
> I'll take a look at the R: Hopefully it's a simple fix, and then when main is green again, main can be merged to this branch.

I saw that. I think we should have a 1am scheduled build. It would help us catch errors that come from external changes.

I suspect one of our dependencies released an update

mccalluc (approved):
looks good. Retitling to "fix:" for changelog grouping.

PR #2802, opened 2026-07-16
bug: patch tail bounds exponent rounding

eddiestudies:
Found a small rounding direction bug in tail bounds while scoping out work on intervals.


Shoeboxam (approved):
Thanks for finding this! Good catch.

PR #2806, opened 2026-07-20
docs: Add known 403 URLs

mccalluc:
For liveramp, I can't find anything on their site that mentions this, but we could use the wayback machine, if that would be better.


eddiestudies (approved):
Linkedin link requires login but looks correct

Issue #2776, opened 2026-06-24
Example for `scan_parquet`


mccalluc:
> We don't have any examples for how to use `scan_parquet`. I think this is a part of the analysis setup that we'd like users to know about, rather than just hitting the warning at run time. I'll file a docs issue.

_Originally posted by @mccalluc in https://github.com/opendp/opendp/pull/2775#discussion_r3467900761_

PR #2808, opened 2026-07-21
docs: move data loading discussion to previous page

mccalluc:
- The index page, as it stands, is just loading an empty lazyframe into Context
- Add a link to scan_parquet. (see also: #2776)
- The discussion of `scan_csv` isn't really an "essential statistic". 

(Ignore branch name: I started in the wrong place.)


eddiestudies (approved):
lgtm

Issue #1139, opened 2023-12-18
Scan source code for URLs, and make sure they work


mccalluc:
For example:
```
E         FFI("No match for concrete type AbsoluteDistance<f64>. You've got a debug binary! Debug binaries support fewer types. Consult https://docs.opendp.org/en/stable/contributor/development-environment.html#build-opendp")
```
There are others as well: We want to be sure they all work.

... but I also feel weird about this: I feel like OpenDP has more error messages with URLs that many libraries. Should more info be in the error?

mccalluc:
I think it would make sense to move this information to the docs, and reference it there. (If it's not already there.)

mccalluc:
The 298 remains. Maybe a new trouble-shooting section in the docs?

PR #2810, opened 2026-07-21
docs: use feature listing rather than github discussion

mccalluc:
- Towards #1139.
- It's not obvious to me whether there is already a point in the docs which could substitute for https://github.com/opendp/opendp/discussions/298. Maybe a new trouble-shooting section?


eddiestudies (approved):
lgtm. links worked

Issue #2804, opened 2026-07-17
Document signed counts


mccalluc:
We now have (optional) signed counts: 
- #2761

We didn't make this the default for the kwarg, because it would have been a breaking API change, but we'd like to steer users toward unbiased results, so the tutorial examples should be updated to use it.

PR #2812, opened 2026-07-22
docs: document signed counts

mccalluc:
- followup to #2761 
- towards #2804 

The goal is to prefer signed counts in the docs, but right now it seems like the higher-level functions don't actually use the optional `signed` kwarg. Is this going in the right direction? Filing as draft.


mccalluc:
Other PRs are taking care of this:
- #2840 
- #2842

Closing.

PR #2820, opened 2026-07-23
bug: fix cdp delta and introduce rounding util for testing

eddiestudies:


Shoeboxam (approved):
This is preferable to being anti-conservative, but going forward hopefully intervals will avoid both the bug and the complexity.

Shoeboxam:
Thanks for catching this!

Issue #2787, opened 2026-07-10
Dev tooling: optional pre-commit and ruff


mccalluc:
Decision from [July 8 dev meeting](https://docs.google.com/document/d/1Y1PiUzMqT53yiSdKVmnyv_QlxH78Gsjp7sOMpwCkbU0/edit?tab=t.0#bookmark=id.elki4ilio5gs):
-  [ ] After Bayard's PR (#2761) is merged, apply ruff to entire python codebase
- [ ] Remove formatting check from subprocess tests
- [ ] Configure and document an optional ruff hook with https://pre-commit.com/ (Just report problems: do not auto-format)
- [ ] In CI, run the precommit ruff check even before the rust build so we can fail fast.
- [ ] With this much in place, revisit and see what other candidates are good to remove from subprocess tests and add to precommit

PR #2823, opened 2026-07-23
maintenance: ruff format two top-level files

mccalluc:
- Toward #2787: Ruff format two files.

(Splitting into a couple PRs to minimize conflicts.)


eddiestudies (approved):
lgtm

Issue #2587, opened 2025-12-05
Decide on direction for data-dependent behavior of `scan_csv`, and implement


mccalluc:
We should consistently use `ignore_errors=True`, and a pair of PRs add an explanation why:
- In DP Wizard: https://github.com/opendp/dp-wizard/pull/737 👉  https://github.com/opendp/dp-wizard/pull/737/commits/f49ba5f53ebd720e0394ce732b218d006c7f8234
- In OpenDP docs: #2580 👉   #2581

Requiring the user to use a non-default setting, or risk data dependent errors, is not great. Should this setting either be enforced, or applied by default, by OpenDP itself? 
- #2132

But, even without data-dependent errors when loading the CSV, there could be data dependent errors downstream, if we try to treat a columns as different type than what Polars detected by reading the first few rows. (I've hit this when trying to specify public keys.) 
- Maybe all columns should be strings, until cast to type? #2374
- Or rather than casting, require the [`schema` kwarg on scan_csv](https://docs.pola.rs/api/python/stable/reference/api/polars.scan_csv.html)? (What happens if there are unexpected columns in the CSV? Drop them, or error? If it's characteristic of the dataset, rather than a row, does it still count as a bad data-dependent error?)
- Or something else?
 
See also: 
- #2234

For now, I'll close #2132 and #2374, so we don't jump into this without considering the options.

mccalluc:
Decision from Dec 10 developer meeting:

> Make sure docs are consistent; acknowledge that this is outside the DP scope; Point out that they should know their schema.

Shoeboxam:
One more consistency proposal:

Update https://docs.opendp.org/en/stable/getting-started/tabular-data/index.html to include data load with csv, and give a commentary on data loading best practices.

mccalluc:
While some work is still to be merged:
- #2824

... I think it makes sense to consider this fixed as of:
- #2581

PR #2824, opened 2026-07-23
docs: Remove references to `ignore_errors`

### LINKED ISSUE #2580: Justify `ignore_errors=True`

mccalluc:
- Towards #2587 
- Fix #2580: if we're not using it, we don't need to justify it.

We want to demonstrate best practices, and that means not using `ignore_errors` everywhere.

I don't remember the history of why these warnings were added in the first place. I have not looked into it farther than observing that the tests are still passing.


eddiestudies (approved):
lgtm

PR #2825, opened 2026-07-23
fix: Unbiased One Way Marginals in Synthetic Data Generation

cbaycity:
This adjusts the default length calculation within MBI to utilize signed integer returns types. This will remove bias from the one way marginal calculations.

To enable that work, changes in the rust library are made so that unbiased length calculations pass the `is_len_expr` check that groupby queries have. This is functionality that should be supported in the current implementation of the length measurement.


rust/src/measurements/make_private_lazyframe/group_by/matching.rs
```diff
@@ -221,7 +221,13 @@ fn is_len_expr(expr: &Expr, name: Option<&str>) -> Option<(String, NoisePlugin)>
 
     let (inputs, args) = match_trusted_plugin::<NoisePlugin>(&expr).ok().flatten()?;
 
-    if let Expr::Len = &inputs[0] {
+    // If it is a cast, unwrap to get the len expr.
+    let input = match &inputs[0] {
+        Expr::Cast { expr, .. } => expr.as_ref(),
+        x => x,
+    };
```
Shoeboxam, line 228: This current constructions available in the library make this arbitrary match valid, but it would be safer to reject nonstrict casts/invalid dtype pairings. Can avoid somehow admitting a null into the stability predicate.

Shoeboxam (changes requested):
Technically the synthetic bucket for nulls should no longer have the clamp:

```python
lookup[None] += max(total - count["len"].sum(), 0)
```

becomes

```python
lookup[None] += total - count["len"].sum()
```

Otherwise this PR looks quite close.

cbaycity:
> Technically the synthetic bucket for nulls should no longer have the clamp:
> 
> ```python
> lookup[None] += max(total - count["len"].sum(), 0)
> ```
> 
> becomes
> 
> ```python
> lookup[None] += total - count["len"].sum()
> ```
> 
> Otherwise this PR looks quite close.

Ahh, good catch. I'll update this. Thank you.

Edit: Updated!

rust/src/measurements/make_private_lazyframe/group_by/matching.rs
cbaycity: Are you saying that the problem is users can create a `len().cast(<unsupported_dtype>)` and then this will pass the `is_len_expr` function without the system throwing an error? And that we can help users avoid this by throwing an error in the `is_len_expr` function if the cast expression is invalid?

cbaycity: I updated the `test_find_len_expr` to try and cast to int 128 (which is not supported) and got the following error:

```
failures:

---- measurements::make_private_lazyframe::group_by::test::test_find_len_expr stdout ----
Error: MakeTransformation("make_cast_aggregation cannot cast from u32 to i128.")
disabled backtrace
```

In my opinion this is a great error message that explains things clearly. Does this change your opinion?

mccalluc (commented):
(Review limited to python interface.) Looks good to me. I double checked that there are only three references to `dp_len` in the MBI directory. In other places we will want to expose the `signed` option in the public interface, but I don't think that is necessary here. 

(Leaving this as a comment so Mike can give the final approval.)

cbaycity:
@Shoeboxam I just pushed changes to remove covariance checking for approx-dp versions of MBI. I don't know why specifically approx-dp queries saw a covariance decrease but I cannot solve the unpredictability by increasing the delta budget without getting errors from the underlying method. Should I file a ticket to enable large delta queries? These are helpful when testing to remove noise.

rust/src/measurements/make_private_lazyframe/group_by/matching.rs
Shoeboxam: To be fair, the change I was asking for isn't currently reachable, but it would ensure the stability histogram would remain valid if it did become reachable due to changes elsewhere in the library.

Shoeboxam (dismissed):
I'm actually ok with the removal of that test. I think the reason why it is failing is due to the extra uncertainty in the synthetic bin, which I had been glossing over before. But I can modify MBI's objective to compensate for the synthetic bin's variance and add it back in/re-enable the test. I'll do that in another commit, because it's a bit involved.

rust/src/measurements/make_private_lazyframe/group_by/matching.rs
cbaycity: I'm not opposed to making a change here to ensure that this cannot be abused if changes elsewhere occur. I don't think it is a good idea to have multiple places in the library where valid/invalid cast expressions are set. The logic needs to exist in one location and be used or applied in each location it is needed. If that's the case, then the logic should exist in the `make_stable_expr/expr_cast_aggregation` code, which it currently does. Is there a way to have this code extend easily to a situation like this? Or should we add a method, `check_valid_cast_expr`, to that module which can be imported and used here?

cbaycity:
> I'm actually ok with the removal of that test. I think the reason why it is failing is due to the extra uncertainty in the synthetic bin, which I had been glossing over before. But I can modify MBI's objective to compensate for the synthetic bin's variance and add it back in/re-enable the test. I'll do that in another commit, because it's a bit involved.

Sounds good. Does a ticket exist for this or should I create one?

Shoeboxam (approved):
Please make one! Thanks.

cbaycity:
> Please make one! Thanks.

Here's the [ticket](https://github.com/opendp/opendp/issues/2831).

PR #2828, opened 2026-07-24
chore: Configure and document precommit 

mccalluc:


docs/source/contributing/development-environment.rst
```diff
@@ -54,7 +54,7 @@ Now run ``cargo build`` in the ``rust`` subdirectory of the repo.
     cd rust
```
mccalluc, line 54: Sorry about the whitespace changes here: Right now, it will remove trailing whitespace on commit by default, if you touch a file. (As I did in this case.)

Would you like that disabled?

eddiestudies: all good. I used to run into this all the time. Vim on my machine would clean these up. It is not an issue at all for me

eddiestudies (approved):
lgtm

PR #2829, opened 2026-07-24
chore: pre-commit in CI

mccalluc:
The idea is that the fast checks that we'd run locally on just the changed files, can also be run in CI for all files, and fail fast if there are problems.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -35,6 +35,8 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v6
 
+      ### Quick Rust checks:
+
       - uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
           rustflags: ""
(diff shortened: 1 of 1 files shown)


.github/workflows/smoke-test.yml
```diff
@@ -47,8 +49,24 @@ jobs:
...
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.10'
+
+      - name: Install minimal dependencies
+        run: pip install ruff pre-commit
+
+      - name: Run pre-commit checks
+        run: |
+          cd ../python
```
mccalluc, line 63: From Eddie:
> I am trying to find a way to move the cd into the cli command itself in case it messes with other parts of the ci that require the code to be in the rust/ dir. None of the options are better yet. Maybe adding cd ../rust/ at the end is a happy medium

mccalluc: but is there a problem? [It runs](https://github.com/opendp/opendp/actions/runs/30270980822/job/89993156851#step:9:20). My memory is that successive steps are isolated, and changing cwd in one doesn't bother successive steps.

eddiestudies (approved):
lgtm. It does work.

PR #2835, opened 2026-07-31
feat: Adjusts Minimum Threshold Check from ApproxDP Queries

cbaycity:
This adjusts the minimum value that the threshold can be in an Approx DP query - dropping it from 1 to 0.

In Line with Algorithm 1 in this [paper](https://arxiv.org/pdf/2309.09170).
R. Rogers, “A Unifying Privacy Analysis Framework for Unknown Domain Algorithms in Differential Privacy,” 2024. Accessed: July 31, 2026. [Online]. Available: https://arxiv.org/pdf/2309.09170

### DIFF
--- a/rust/src/measurements/make_private_lazyframe/group_by/mod.rs
+++ b/rust/src/measurements/make_private_lazyframe/group_by/mod.rs
@@ -286,13 +286,6 @@ where
         if margin.invariant.is_some() || is_join {
             ()
         } else if let Some((_, noise, threshold_value, _)) = &threshold_info {
-            if *threshold_value <= 0 {
-                return fallible!(
-                    FailedMap,
-                    "This algorithm requires a threshold greater than 0."
-                );
-            }
-
             let d_instability = threshold_value.neg_inf_sub(&li)?;
             let delta_single =
                 integrate_discrete_noise_tail(noise.distribution, noise.scale, d_instability)?;
--- a/rust/src/measurements/make_private_lazyframe/group_by/mod.rs
+++ b/rust/src/measurements/make_private_lazyframe/group_by/mod.rs
@@ -286,8 +286,8 @@ where
         if margin.invariant.is_some() || is_join {
             ()
         } else if let Some((_, noise, threshold_value, _)) = &threshold_info {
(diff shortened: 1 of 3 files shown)


rust/src/measurements/make_private_lazyframe/group_by/pseudocode/make_private_group_by.py
```diff
@@ -180,8 +180,8 @@ def privacy_map(d_in: Bounds):  # |\label{line:privacy-map}|
             pass
         elif threshold_info is not None:  # |\label{line:privacy-map-threshold}|
             _, noise, threshold_value, _ = threshold_info
-            if li >= threshold_value:
-                raise f"Threshold must be greater than {li}."
+            if 0 > threshold_value:
+                raise "Threshold must be greater than 0."
```
mccalluc, line 184: What should happen if threshold_value == 0?

pseudocode so it doesn't really matter, but I always like to include more information in the error: What was the value, if wasn't right?

rust/src/measurements/make_private_lazyframe/group_by/mod.rs
```diff
@@ -286,10 +286,6 @@ where
         if margin.invariant.is_some() || is_join {
             ()
         } else if let Some((_, noise, threshold_value, _)) = &threshold_info {
-            if li >= *threshold_value {
-                return fallible!(FailedMap, "threshold must be greater than {:?}", li);
-            }
```
mccalluc, line 291: Do you need to add a comparison to match the pseudocode?

mccalluc (commented):
I half-heard the conversation yesterday, but didn't follow any of the details, so these comments are superficial. See what Mike thinks!

Shoeboxam:
As we (briefly!) discussed, the delta parameter is the mass of the noise distribution to the right of a support point at `threshold - l_\infty`. The current function for this only supports positive support points, resulting in delta above 0.5 being rejected. The fix would be to adjust the mass computation to support negative support values, possibly by computing a conservative lower bound to the tail and returning the complement. I can provide more detail to narrow this down to a concrete implementation.

PR #2836, opened 2026-08-01
chore: harden len-expr cast matching in make_private_lazyframe

### LINKED ISSUE #2832: harden stability histogram len selector
Just for future-proofing- conduct some checks on the matched cast in a DP len expr to ensure well-formed-ness.

_Originally posted by @cbaycity in https://github.com/opendp/opendp/pull/2825#discussion_r3683983722_

Shoeboxam:
Closes #2832

### DIFF
--- a/rust/src/measurements/make_private_lazyframe/group_by/matching.rs
+++ b/rust/src/measurements/make_private_lazyframe/group_by/matching.rs
@@ -1,8 +1,11 @@
 use std::{collections::BTreeSet, sync::Arc};
 
-use polars::prelude::{DslPlan, JoinOptions, JoinType, PlSmallStr, Selector};
+use polars::{
+    datatypes::DataType,
+    prelude::{DslPlan, JoinOptions, JoinType, PlSmallStr, Selector},
+};
 use polars_plan::{
-    dsl::{Expr, Operator},
+    dsl::{DataTypeExpr, Expr, Operator},
     prelude::GroupbyOptions,
     utils::expr_output_name,
 };
@@ -221,10 +224,14 @@ pub(super) fn is_len_expr(expr: &Expr, name: Option<&str>) -> Option<(String, No
 
     let (inputs, args) = match_trusted_plugin::<NoisePlugin>(&expr).ok().flatten()?;
 
-    // If it is a cast, unwrap to get the len expr.
+    // Signed frame lengths are represented as a stable widening cast from u32.
     let input = match &inputs[0] {
-        Expr::Cast { expr, .. } => expr.as_ref(),
-        x => x,
+        Expr::Cast {
+            expr,
+            dtype: DataTypeExpr::Literal(DataType::Int64),
+            ..
+        } => expr.as_ref(),
+        input => input,
     };
 
     if let Expr::Len = input {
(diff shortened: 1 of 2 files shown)


Shoeboxam:
* **#2836** <a href="https://app.graphite.com/github/pr/opendp/opendp/2836?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opendp/opendp/2836?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rust/src/measurements/make_private_lazyframe/group_by/matching.rs
```diff
@@ -221,10 +224,14 @@ pub(super) fn is_len_expr(expr: &Expr, name: Option<&str>) -> Option<(String, No
...
     let (inputs, args) = match_trusted_plugin::<NoisePlugin>(&expr).ok().flatten()?;
 
-    // If it is a cast, unwrap to get the len expr.
+    // Signed frame lengths are represented as a stable widening cast from u32.
     let input = match &inputs[0] {
-        Expr::Cast { expr, .. } => expr.as_ref(),
-        x => x,
+        Expr::Cast {
+            expr,
+            dtype: DataTypeExpr::Literal(DataType::Int64),
```
cbaycity, line 231: In [opendp/rust/src/measurements/make_private_expr/expr_dp_frame_len/mod.rs](https://github.com/opendp/opendp/blob/089d59ff02d622c4e483c8fff4619a914c58f1fa/rust/src/measurements/make_private_expr/expr_dp_frame_len/mod.rs#L92) the only transformation that can be built with `signed=true` is a cast to Int64. But in the [opendp/rust/src/transformations/make_stable_expr/expr_cast_aggregation/mod.rs](https://github.com/opendp/opendp/blob/089d59ff02d622c4e483c8fff4619a914c58f1fa/rust/src/transformations/make_stable_expr/expr_cast_aggregation/mod.rs#L21) the allowed cast transformations include `Int8`, `Int16`, and `Int32`. I'm worried that this very strict check of Int64 will prevent someone who is using the framework API from completing a `len` transformation that is cast to a valid type. 

This concern may be too strict though. Polars currently defaults `len` expressions to cast to u32, so this is the only supported use case right now. If they change their default type, or if OpenDP supports Int128 in the future though, this is too tight of a check. I'd consider changing this, and perhaps having an `allowed_cast` constant in the len expression module. BTW, I'm fine with the answer being that the library can accept this level of potential inflexibility. It can't lead to a privacy leak, just a potential poor developer experience.

cbaycity: QQ, is there every a use case for a len expression to be cast to a float so that non-discrete noise can be added?

rust/src/measurements/make_private_lazyframe/group_by/test.rs
```diff
@@ -194,5 +196,28 @@ fn test_find_len_expr() -> Fallible<()> {
             supported_expr,
         );
     }
+
+    // Casts other than the stable u32-to-i64 widening cast are rejected.
```
cbaycity, line 200:
```suggestion
    // Casts other than the stable u32-to-i64 widening case are rejected.
```

cbaycity (commented):
Looks generally very good. Only two preference comments. The test addition is great.

Shoeboxam:
Closing this as I think this isn't worth the complexity for the gain, especially given potential future developments.

PR #2837, opened 2026-08-01
fix: clip counts before casting u32 to i32 in make_stable_marginals

### LINKED ISSUE #2833: harden signed-ness cast of synthetic data counts
The multi-marginal transformation used by synthetic data mechanisms collects u32 into i32. Technically a clip to [0, 2^31) is necessary.

Shoeboxam:
Assigning a numpy u32 into an i32 array causes counts above i32::MAX (2^31) to silently wrap, resulting in sensitivity of 2^32 (although unlikely to be reached, unless you're dealing with a very large dataset!

Clip to i32::MAX before cast so any potentially overflowing count saturates instead.

### DIFF
--- a/python/src/opendp/extras/mbi/_utilities/__init__.py
+++ b/python/src/opendp/extras/mbi/_utilities/__init__.py
@@ -238,7 +238,11 @@ def make_stable_marginals(
 
     def pivot(x, clique: tuple[str, ...]):
         y = np.zeros(shape(clique), dtype=np.int32)
-        y[tuple(x[clique].to_numpy().T)] = x["len"].to_numpy()
+        # u32 counts are clipped before narrowing to i32: numpy treats
+        # uint32->int32 as a same-kind cast and silently wraps instead of erroring.
+        y[tuple(x[clique].to_numpy().T)] = (
+            x["len"].clip(upper_bound=np.iinfo(np.int32).max).cast(pl.Int32).to_numpy()
+        )
         return y
 
     def function(data):
--- a/python/test/mbi/test_utilities.py
+++ b/python/test/mbi/test_utilities.py
@@ -116,6 +116,35 @@ def test_make_stable_marginals():
         )
 
 
+def test_make_stable_marginals_clips_u32_counts(monkeypatch):
+    pytest.importorskip("mbi")
+    import numpy as np  # type: ignore[import-not-found]
+    import polars as pl  # type: ignore[import-not-found]
+
+    domain = dp.lazyframe_domain(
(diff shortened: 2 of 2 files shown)


Shoeboxam:
* **#2837** <a href="https://app.graphite.com/github/pr/opendp/opendp/2837?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opendp/opendp/2837?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

python/src/opendp/extras/mbi/_utilities/__init__.py
```diff
@@ -238,7 +238,11 @@ def shape(by: tuple[str, ...]) -> tuple[int, ...]:
 
     def pivot(x, clique: tuple[str, ...]):
         y = np.zeros(shape(clique), dtype=np.int32)
-        y[tuple(x[clique].to_numpy().T)] = x["len"].to_numpy()
+        # u32 counts are clipped before narrowing to i32: numpy treats
+        # uint32->int32 as a same-kind cast and silently wraps instead of erroring.
+        y[tuple(x[clique].to_numpy().T)] = (
+            x["len"].clip(upper_bound=np.iinfo(np.int32).max).cast(pl.Int32).to_numpy()
```
eddiestudies, line 244: Do we need to use pl.Int32. Can we use np.int32 instead given the function uses it for y?

python/src/opendp/extras/mbi/_utilities/__init__.py
```diff
@@ -238,7 +238,11 @@ def shape(by: tuple[str, ...]) -> tuple[int, ...]:
 
     def pivot(x, clique: tuple[str, ...]):
         y = np.zeros(shape(clique), dtype=np.int32)
-        y[tuple(x[clique].to_numpy().T)] = x["len"].to_numpy()
+        # u32 counts are clipped before narrowing to i32: numpy treats
+        # uint32->int32 as a same-kind cast and silently wraps instead of erroring.
```
mccalluc, line 242:
```suggestion
        # uint32->int32 as a same-kind cast and silently wraps instead of erroring.
        # See https://stackoverflow.com/a/1970697
```
To provide more context. (The more authoritative link to the mailing list archive doesn't work for me.)

mccalluc (approved):
Looks good to me, with or without a tweak to the comment.

python/src/opendp/extras/mbi/_utilities/__init__.py
Shoeboxam: No, the cast is a Polars API, I'd use a Polars dtype.

PR #2838, opened 2026-08-01
feat: model stable oneway marginals as linear constraints (synthetic data)

### LINKED ISSUE #2831: Modify MBI Objective to Fix Synthetic Bin Issue
When adjusting MBI's one way marginals calculation to use signed len calculations and unbiased noise, the covariance fit effectiveness test started to fail unexpectedly. We believe that this is caused by uncertainty in the synthetic bin and MBI's objective. This was discussed in this [thread](https://github.com/opendp/opendp/pull/2825#issuecomment-5132826383) on the PR which disabled the edge cases breaking the covariance expectation tests.

This issue is to resolve the underlying issue causing covariance in these tests to not converge on the true covariance. This test should then be re-enabled by dropping the `if not approximate` condition [here](https://github.com/opendp/opendp/blob/402b24f549b960f6f6ab82767e1c2fd580d2c84f/python/test/mbi/test_table.py#L107).

Note: Dropping this `if not approximate` condition in [opendp/python/test/mbi/test_table.py](https://github.com/opendp/opendp/blob/main/python/test/mbi/test_table.py#L107) will recreate the covariance instability that needs to be resolved.

Shoeboxam:
The synthetic data implementation releases one-way marginals over unknown keysets via stability histograms. The tricky bit is that stability histograms don't sum to total, which breaks the symmetry in MBI. The current implementation inserts a synthetic bin for null counts based on best-known information about the total count. This "works", but the cell is correlated and has much higher variance, which improperly influences MBI's variance weighting.

This PR adjusts the corresponding LinearMeasurements to ignore the synthetic bin, which keeps it out of the variance calculations. In the process, I could no longer roll multiple releases for the same marginal into one via inverse variance weighting because multiple marginals for the same set of grouping columns can have different linear measurement queries.

### DIFF
--- a/python/src/opendp/extras/mbi/_utilities/__init__.py
+++ b/python/src/opendp/extras/mbi/_utilities/__init__.py
@@ -1,6 +1,6 @@
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
-from typing import Callable, Literal, Optional, Any, Iterator, cast, get_args, TYPE_CHECKING
+from typing import Callable, Literal, Optional, Any, Iterator, get_args, TYPE_CHECKING
 from functools import reduce
 from math import sqrt
 from opendp._internal import (
@@ -35,8 +35,10 @@ from opendp.mod import (
     Metric,
     Transformation,
 )
+
 if TYPE_CHECKING:  # pragma: no cover
     from opendp.extras.polars import Bound
+
 
 @dataclass
 class Count:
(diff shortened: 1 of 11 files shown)


Shoeboxam:
* **#2838** <a href="https://app.graphite.com/github/pr/opendp/opendp/2838?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#2839](https://github.com/opendp/opendp/pull/2839) <a href="https://app.graphite.com/github/pr/opendp/opendp/2839?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2905](https://github.com/opendp/opendp/pull/2905) <a href="https://app.graphite.com/github/pr/opendp/opendp/2905?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/opendp/opendp/2838?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

python/src/opendp/extras/mbi/_utilities/__init__.py
```diff
@@ -342,35 +353,92 @@ def reducer(a, b):
...
+    """
 
-    marginals = marginals.copy()
+    length: int
 
-    for new_marginal in new_marginals:
-        if not isinstance(new_marginal, LinearMeasurement):
-            raise ValueError("each new marginal must be of type LinearMeasurement")
+    def __call__(self, factor):
+        return factor.datavector()[: self.length]
```
eddiestudies, line 371: Can we add a test to verify this is always dropping the "other" bucket as intended even with Nones

eddiestudies (approved):
lgtm.

Might be worth adding a test to make sure we are dropping the "other" bin even with nulls. Although it does look like the current PR drops them and creates a new vector with [*observed_keys, None] which means the .length  is correct since it is the length of *observed_keys.

I really liked the Marginals refactor. Much easier to follow then the previous approach

PR #2840, opened 2026-08-01
feat: add signed in all count APIs

### LINKED ISSUE #2803: Extend Cast Aggregation Functionality
Aggregations within OpenDP return a data type based on what the non-private Polars expression will return, and then noise is added before the user receives a value. This creates usability concerns, for example: negative noise added to an unsigned int will be truncated to zero. This problem was discussed in a prior [issue](https://github.com/opendp/opendp/issues/2749) related to the DP len function, which returns a u32 by default resulting in biased noise that prevents some post-processing steps from being run. This request is to extend the `expr_cast_aggregation` functionality to allow for more data type transformations (right now it just supports casting to integers). There could be use cases for this method to return float values or other data types.

The prior issue and PR address dp-len queries, but other methods, like max/min, which return unsigned integers and some users could hit the same biased noise issue that was encountered in the dp-len use case. Ideally every aggregation within the library should be castable to another valid type before noise is added. 

There are two requests described here:
1. Enable additional data type casts in the `expr_cast_aggregation`
2. Extend casting after aggregation but before measurement to additional methods. Potentially any aggregation but definitely ones that can result in unsigned integer evaluations - which can result in biased noise.

## Links to related work as reference.
Cast-Aggregation PR: https://github.com/opendp/opendp/issues/2749
DP-len issue: https://github.com/opendp/opendp/issues/2749

Shoeboxam:
Before reviewing, humor another API: What if we don't add a signed argument to `.dp.x(scale, signed)`, default them to cast/nonclip, and instead require users write the explicit `.x().dp.noise(scale)` if they want clipped counts? Then existing call sites also don't have a compile-change or gain an extra bool. They would, though, admittedly, silently change behavior. If we'd rather that API, the next PR adds warnings that could be folded in here.


Shoeboxam:
* **#2840** <a href="https://app.graphite.com/github/pr/opendp/opendp/2840?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#2842](https://github.com/opendp/opendp/pull/2842) <a href="https://app.graphite.com/github/pr/opendp/opendp/2842?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2872](https://github.com/opendp/opendp/pull/2872) <a href="https://app.graphite.com/github/pr/opendp/opendp/2872?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/opendp/opendp/2840?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

python/src/opendp/extras/polars/__init__.py
```diff
@@ -230,25 +238,29 @@ def count(self, scale: float | None = None):
         │ ...    │
         └────────┘
 
-        Output is noise added to three.
+        Output is noise added to two.
```
Shoeboxam, line 241: Yes, this was off by one.

docs/source/api/user-guide/polars/expressions/aggregation.rst
```diff
@@ -169,6 +169,46 @@ This is more likely to happen when the true value is small.
 This release tells us that the number of null values is relatively
 small.
 
+Signed Counts
+-------------
+
+Pass ``signed=True`` to any counting query (``dp.len``, expression ``len``,
+``count``, ``null_count`` or ``n_unique``) to cast the exact count to a
+signed ``Int64`` before noise is added, so negative noisy outputs are
+preserved instead of being clamped to zero:
```
mccalluc, line 178: Before going farther, is this right? My sense was that we wanted to move towards signed counts being the default, though I may be misremembering the discussion. Suggest clarifying direction at a Wednesday afternoon meeting.

Putting that aside, even if the current state in this PR is what we want, I don't think we want to explain it this way.

> cast the exact count to a signed ``Int64`` before noise is added, so negative noisy outputs are preserved instead of being clamped to zero

That is really centered on the implementation. I think we just should focus on the trade-off the user has:
- unsigned: will seem less weird to a non-dp audience, but biased upwards.
- signed: will seem more weird to a non-dp audience, but unbiased.

> Before reviewing, humor another API: What if we don't add a signed argument to `.dp.x(scale, signed)`, default them to cast/nonclip, and instead require users write the explicit `.x().dp.noise(scale)` if they want clipped counts?

I do think opt-in for clipped counts would be ok, but this syntax seems confusing to me.

Will save a longer review until the direction is clear.

mccalluc (commented):
Not sure this is the right direction, but probably shouldn't try to sort it out in a PR.

eddiestudies:
> Before reviewing, humor another API: What if we don't add a signed argument to `.dp.x(scale, signed)`, default them to cast/nonclip, and instead require users write the explicit `.x().dp.noise(scale)` if they want clipped counts? Then existing call sites also don't have a compile-change or gain an extra bool. They would, though, admittedly, silently change behavior. If we'd rather that API, the next PR adds warnings that could be folded in here.

I think having signed default to False but have a warning, will default to true in the future, set if you require false.

then you end up with .dp.x(scale) as the final api and it would be default be signed

docs/source/api/user-guide/polars/expressions/aggregation.rst
eddiestudies: my comment was highlighting the default decision Chuck mentioned here.

rust/src/polars/mod.rs
```diff
@@ -481,11 +487,11 @@ impl DPExpr {
 ///
 /// # Arguments
 /// * `scale` - parameter for the noise distribution
-/// * `signed` - parameter to flip len output from unsigned int to signed int.
+/// * `signed` - if true, the exact count is cast to `Int64` before noise is added,
```
eddiestudies, line 490: always int64?

rust/src/measurements/make_private_expr/mod.rs
```diff
@@ -158,16 +158,7 @@ where
...
-            return expr_dp_counting_query::make_expr_dp_len(
-                input_domain,
-                input_metric,
-                output_measure,
-                self,
-                global_scale,
-            );
-        }
-
+        counting_query!(DPLenShim, make_expr_dp_len);
```
eddiestudies, line 161: this was nice

rust/src/polars/mod.rs
Shoeboxam: yep

Shoeboxam:
@eddiestudies @mccalluc Yes about default behavior. This PR adds the signed arguments to all counts. The next PR makes the backwards incompatible default change. https://github.com/opendp/opendp/pull/2842

eddiestudies (approved):
lgtm. There is no warning in this PR. We can add the warning in the next one and then move it to True in the follow up

PR #2842, opened 2026-08-01
feat: deprecate unsigned defaults for counting queries

### LINKED ISSUE #2813: Deprecate dp.len() with default `signed`
I think the decision we reached was that the default value would change to `True` in a release or two, and we need to give users a heads up.

Tricky part: I don't think it's possible to distinguish `False` provided explicitly by the user, and `False` as a default.

Proposal:
- Make the default `None`
- For now, if `None` is received, use `False` internally, and emit a warning.
- (Explicit `False` is not a warning.)
- Add a note in the docstring to point out upcoming change in behavior.
- Either:
  - Examples in docs shouldn't emit warnings, so add `signed=True` in all the docs.
  - Or (not my favorite), tweak pytest config to allow the deprecation warning, which would otherwise be a test failure.
- In a future release: 
  - Make the default `True`
  - Remove the deprecation warning
  - Remove the `signed=True` from all the docs

Shoeboxam:
Closes #2813 using the described approach.


Shoeboxam:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/opendp/opendp/2842?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2842** <a href="https://app.graphite.com/github/pr/opendp/opendp/2842?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opendp/opendp/2842?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2840** <a href="https://app.graphite.com/github/pr/opendp/opendp/2840?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#2872](https://github.com/opendp/opendp/pull/2872) <a href="https://app.graphite.com/github/pr/opendp/opendp/2872?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

docs/source/api/user-guide/polars/expressions/aggregation.rst
```diff
@@ -66,9 +66,9 @@ rows.
             ...     .agg(
             ...         [
             ...             # total number of rows in the frame, including nulls
-            ...             dp.len(),
+            ...             dp.len(signed=True),
```
mccalluc: From #2813:
> In a future release:
>- Make the default True
>- Remove the deprecation warning
>- Remove the signed=True from all the docs

Maybe as you read it, that "future release" work is out of scope for #2813? If that's the case, can we go ahead and pull out that future work into a separate issue? I don't want to auto-close this and lose track of the final goal.

(Or I might be confused in some other way! This is just my best guess of where we're not lining up.)

docs/source/api/user-guide/polars/expressions/aggregation.rst
```diff
@@ -156,26 +156,33 @@ result in one more, or one less, unique value.
...
             ...
             └─────────┘
 
 
-Noise added to a count can make the count go negative, but since the
-output data type is an unsigned integer, the library may return zero.
-This is more likely to happen when the true value is small.
+These examples use ``signed=True`` to preserve negative noisy counts as
+``Int64`` values. With ``signed=False``, negative noisy counts may be
+returned as zero.
```
mccalluc, line 167: It's not "may":

```suggestion
``Int64`` values. With ``signed=False``, negative noisy counts will be
clamped to zero.
```

I also worry about focusing too much on Rust types in the Python docs.

docs/source/api/user-guide/polars/expressions/aggregation.rst
```diff
@@ -156,26 +156,33 @@ result in one more, or one less, unique value.
...
-``count``, ``null_count`` or ``n_unique``) to cast the exact count to a
-signed ``Int64`` before noise is added, so negative noisy outputs are
-preserved instead of being clamped to zero:
+Pass ``signed=True`` to cast the exact count to a signed ``Int64`` before
+noise is added, so negative noisy outputs are preserved instead of being
+clamped to zero. This applies to all counting queries (``dp.len``, expression
+``len``, ``count``, ``null_count`` and ``n_unique``).
+
+For now, omitting ``signed`` (or passing ``None``) retains unsigned output
+and emits a warning. The default will change to ``True`` in a future release.
```
mccalluc, line 181: This should be explained in terms of the results the user sees, and not in terms of the implementation. I wouldn't mention "cast the exact count" (That describes the process, not the result.) or "retains unsigned output" (Just confusing to say "retains" when we're losing information by clamping.).

docs/source/api/user-guide/polars/expressions/aggregation.rst
```diff
@@ -156,26 +156,33 @@ result in one more, or one less, unique value.
...
+Pass ``signed=True`` to opt in now, or ``signed=False`` to explicitly retain
+unsigned output. Neither explicit boolean emits a warning.
```
mccalluc, line 183: > explicitly retain unsigned output

This is good for migration notes, but I think the docs themselves should focus on the way it works in this version.

docs/source/api/user-guide/polars/data-types.ipynb
```diff
@@ -136,14 +136,14 @@
...
       ],
       "text/plain": [
        "shape: (3, 2)\n",
        "┌───────┬──────┐\n",
        "│ alpha ┆ len  │\n",
        "│ ---   ┆ ---  │\n",
-       "│ cat   ┆ u32  │\n",
+       "│ cat   ┆ i64  │\n",
        "╞═══════╪══════╡\n",
        "│ A     ┆ 1000 │\n",
```
mccalluc, line 148: The notebooks also encode results... which would probably be different with signed counts... but rerunning all the notebooks to get new results introduces more diffs.

So, not recommending any changes here, but pointing out that things could look unusual!

python/src/opendp/extras/polars/__init__.py
```diff
@@ -58,6 +58,21 @@
 def _get_opendp_polars_lib_path():
     return os.environ.get("OPENDP_POLARS_LIB_PATH", lib_path)
 
+def _resolve_signed(signed: bool | None) -> bool:
+    if signed is None:
+        warn(
+            "The default for counting queries will change to signed=True in a future "
+            "release. Pass signed=True to preserve negative noisy counts as Int64, "
+            "or signed=False to retain the current unsigned behavior.",
```
mccalluc, line 66:
```suggestion
            "The default value is currently signed=False, and counts are clamped at zero, "
            "consistent with past releases, but the default will be changing to signed=True. "
```
I think the use of both "preserve" and "retain" could be confusing, and I'd like to minimize mentions of rust types like Int64 in python.

python/src/opendp/extras/polars/__init__.py
```diff
@@ -58,6 +58,21 @@
...
+            FutureWarning,
+            stacklevel=3,
```
mccalluc, line 68: I wasn't familiar with this kwarg. Interesting.

Feels like higher numbers could be a little fragile, but since we're planning not to keep this around forever, I think this is fine.

mccalluc (approved):
Great! There are places where I think it could be described with less attention to the implementation and more focus on the user experience: If that resonates with you, I'm happy to re-review, or if you want to merge now, I won't block. 

I've filed a new issue, so we don't lose track of the next, and last, step:
- #2947

python/src/opendp/extras/polars/__init__.py
Shoeboxam: I use it in a number of places, just search for stacklevel. It keeps the stack traces trim.

PR #2807, opened 2026-07-21
feat: spherical kmeans

Shoeboxam:

PR #2843, opened 2026-08-01
feat: sklearn API base class

Shoeboxam:
Adds a shared base class for OpenDP’s sklearn-style estimators and integrates it with `Context.query()`.

Estimators now keep only algorithm parameters, while the query supplies the input space and privacy budget. Callers are meant to use `estimator.fit(context.query())` or `context.query().sklearn(estimator).release()`.

Documentation and examples are in the stack.


Shoeboxam:
* **#2918** <a href="https://app.graphite.com/github/pr/opendp/opendp/2918?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#2807](https://github.com/opendp/opendp/pull/2807) <a href="https://app.graphite.com/github/pr/opendp/opendp/2807?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2844](https://github.com/opendp/opendp/pull/2844) <a href="https://app.graphite.com/github/pr/opendp/opendp/2844?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#2843** <a href="https://app.graphite.com/github/pr/opendp/opendp/2843?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opendp/opendp/2843?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

asaha1203 (dismissed):
The PR mostly looks good, but I would consider renaming `SklearnEstimator` to `DPEstimator` or something like this, since it inherits from the BaseEstimator class anyways and only adds fields for use with the context API. 

We've also chatted about this offline, but it would be good to consider the API design for the final version of the context-query interplay (e.g. do we want to add a column selection mechanism when the context object contains tabular data). I have no specific thoughts on how this should be done, and I believe that the existing API in this PR appears to suffice for this goal, but we should think about it carefully. 

No other major comments.

python/src/opendp/extras/sklearn/_estimator.py
```diff
@@ -0,0 +1,158 @@
...
+    """Base used when the optional scikit-learn dependency is unavailable."""
+
+
+_BaseEstimator = (
+    _sklearn_base.BaseEstimator if _sklearn_base is not None else _FallbackBaseEstimator
+)
+
+
+class SklearnEstimator(_BaseEstimator, ABC):  # type: ignore
+    """Base class for OpenDP scikit-learn-style DP estimators.
```
eddiestudies, line 50: for sklearn the estimator API also includes a .predict on the model object for predictions.

Although the API change for fit is not typical I understand that the extra tracking needed from query is needed. 

For predict would we match the Estimator spec? 

This would mean the OpenDP estimator cannot be in a pipeline for training but it can be in a pipeline for predictions 

Won't support:
from sklearn.pipeline import Pipeline
from sklearn.preprocessing import StandardScaler
from sklearn.svm import SVC

# Define the pipeline
pipe = Pipeline([
    ('scaler', StandardScaler()),
    ('classifier', SVC())
])

# Fit and predict in one step
pipe.fit(X_train, y_train)
predictions = pipe.predict(X_test)

But if we mark is trained on the estimator we may be able to have it follow the chanined predict calls of a pipeline

eddiestudies: We may be able to have a OpenDPPipeline that takes Query throughout to mirror one of these examples with each component following the sklearn API.

That would be closer to matching what I have seen in production scenarios. There are almost always preprocessor hurdles for operationalizing a model, having a clean even if custom story for this would help

Shoeboxam: > For predict would we match the Estimator spec?

Yes. I'm walking back the property magic. I think it will be able to work in a pipeline- it could desugar into framework API calls and chaining...

python/src/opendp/extras/sklearn/_estimator.py
```diff
@@ -0,0 +1,158 @@
+"""Base class shared by OpenDP scikit-learn-style differentially private estimators.
+
+An estimator instance carries *only algorithm hyperparameters* (e.g. ``n_clusters``,
+``n_components``).  Everything privacy- or data-related -- the input domain, input
```
eddiestudies, line 4:
```suggestion
``n_components``).  Everything privacy or data-related -- the input domain, input
```

eddiestudies (dismissed):
lgtm.

The query as input helps avoid privacy bugs which is nice. I hope we can make the seams around this ergonomic to sklearn users over time. 

The extra error handling already helps on that front. A auto translation of sklearn steps to opendp sklearn steps on a dppipeline might be an ergonomic way to do this but I am sure more design iterations will be needed before then.

The no y check on the base _prepare_fit_query function was unexpected but made sense after thinking about unfitted preprocessors. There might be scenarios where y is set but ignored if the pipeline needs it for some but not all fits.

mccalluc:
Moved this to the next release milestone. We'll look forward to getting this in, and the PRs that build on it, but that's not going to happen immediately.

PR #2846, opened 2026-08-01
feat: support one-sided binary-search bounds

Shoeboxam:
Adds support for one-sided binary-search bounds, allowing callers to provide only a known lower or upper endpoint while the other is inferred. Used by other `PrivacyCurve` PRs to invert monotone δ(ε) and log-δ profiles where the lower bound is known (ε = 0).

### DIFF
--- a/R/opendp/R/mod.R
+++ b/R/opendp/R/mod.R
@@ -599,8 +599,10 @@ unbox2 <- function(x) {
 #' @param make_chain a function that takes a number and returns a Transformation or Measurement
 #' @param d_in how far apart input datasets can be
 #' @param d_out how far apart output datasets or distributions can be
-#' @param bounds a 2-tuple of the lower and upper bounds on the input of `make_chain`
-#' @param .T type of argument to `make_chain`, either "float" or "int"
+#' @param bounds a two-element list of optional lower and upper bounds on the input
+#'   of `make_chain`; use `NULL` for either bound to infer it
+#' @param .T type of argument to `make_chain`, either "float" or "int"; when set,
+#'   takes precedence over the type inferred from `bounds`
 #' @return a Transformation or Measurement (chain) that is (`d_in`, `d_out`)-close.
 #' @export
 #' @examples
@@ -625,8 +627,10 @@ binary_search_chain <- function(make_chain, d_in, d_out, bounds = NULL, .T = NUL
 #' @param make_chain a function that takes a number and returns a Transformation or Measurement
 #' @param d_in how far apart input datasets can be
 #' @param d_out how far apart output datasets or distributions can be
-#' @param bounds a 2-tuple of the lower and upper bounds on the input of `make_chain`
-#' @param .T type of argument to `make_chain`, either "float" or "int"
+#' @param bounds a two-element list of optional lower and upper bounds on the input
+#'   of `make_chain`; use `NULL` for either bound to infer it
+#' @param .T type of argument to `make_chain`, either "float" or "int"; when set,
+#'   takes precedence over the type inferred from `bounds`
 #' @return the parameter to `make_chain` that results in a (`d_in`, `d_out`)-close Transformation or Measurement
 #' @export
 binary_search_param <- function(make_chain, d_in, d_out, bounds = NULL, .T = NULL) {
@@ -684,18 +688,21 @@ binary_search_param <- function(make_chain, d_in, d_out, bounds = NULL, .T = NUL
 }
 
 .infer_search_type <- function(predicate, .T = NULL, bounds = NULL) {
-  if (!is.null(bounds)) {
-    if (inherits(bounds, "integer")) {
-      return(rt_parse("int"))
-    }
-    if (inherits(bounds, "numeric")) {
-      return(rt_parse("float"))
-    }
-    stop("bounds must be either float or int", call. = FALSE)
-  }
-
   if (!is.null(.T)) {
     return(rt_parse(.T))
+  }
+
+  if (!is.null(bounds)) {
+    present_bounds <- Filter(Negate(is.null), as.list(bounds))
+    if (length(present_bounds) > 0) {
+      if (all(vapply(present_bounds, inherits, logical(1), "integer"))) {
+        return(rt_parse("int"))
+      } else if (all(vapply(present_bounds, inherits, logical(1), "numeric"))) {
+        return(rt_parse("float"))
+      } else {
+        stop("bounds must be either float or int", call. = FALSE)
+      }
+    }
   }
 
   check_type <- function(v) {
@@ -714,16 +721,34 @@ binary_search_param <- function(make_chain, d_in, d_out, bounds = NULL, .T = NUL
 
 #' Find the closest passing value to the decision boundary of `predicate`
 #'
-#' If bounds are not passed, conducts an exponential search.
+#' Missing bounds are inferred. Use `NULL` for either element of `bounds` to
+#' conduct a one-sided search, or omit `bounds` to conduct an exponential search.
 #'
 #' @concept mod
 #' @param predicate a monotonic unary function from a number to a boolean
-#' @param bounds a 2-tuple of the lower and upper bounds on the input of `make_chain`
-#' @param .T type of argument to `predicate`, one of float or int
+#' @param bounds a two-element list of optional lower and upper bounds on the input
+#'   of `predicate`; use `NULL` for either bound to infer it
+#' @param .T type of argument to `predicate`, one of float or int; when set,
+#'   takes precedence over the type inferred from `bounds`
 #' @param return_sign if True, also return the direction away from the decision boundary
 #' @return the discovered parameter within the bounds
 #' @export
+#' @examples
+#' binary_search(\(x) x <= -5L, bounds = list(-10L, NULL))
 binary_search <- function(predicate, bounds = NULL, .T = NULL, return_sign = FALSE) {
+  if (is.null(bounds)) {
+    lower <- NULL
+    upper <- NULL
+  } else {
+    if (length(bounds) != 2) {
+      stop(
+        "bounds must contain exactly two elements; use list(lower, NULL) or list(NULL, upper) for a one-sided search",
+        call. = FALSE
+      )
+    }
+    lower <- bounds[[1]]
+    upper <- bounds[[2]]
+  }
   .T <- .infer_search_type(predicate, .T, bounds)
 
   result <- .call_search_callback(
@@ -731,7 +756,7 @@ binary_search <- function(predicate, bounds = NULL, .T = NULL, return_sign = FAL
     .T = .T,
     bounds = bounds,
     call_expr = function(wrapped) {
-      `_binary_search`(wrapped, bounds = bounds, .T = .T, return_sign = return_sign)
+      `_binary_search`(wrapped, lower = lower, upper = upper, .T = .T, return_sign = return_sign)
     }
   )
 
--- a/python/src/opendp/mod.py
+++ b/python/src/opendp/mod.py
@@ -1314,7 +1314,7 @@ M = TypeVar("M", Transformation, Measurement)
 def binary_search_chain(
         make_chain: Callable[[float], M],
         d_in: Any, d_out: Any,
-        bounds: tuple[float, float] | None = None,
+        bounds: tuple[float | None, float | None] | None = (None, None),
         T=None) -> M:
     """Find the highest-utility (`d_in`, `d_out`)-close Transformation or Measurement.
     
@@ -1327,7 +1327,7 @@ def binary_search_chain(
     :param make_chain: a function that takes a number and returns a Transformation or Measurement
     :param d_in: how far apart input datasets can be
     :param d_out: how far apart output datasets or distributions can be
-    :param bounds: a 2-tuple of the lower and upper bounds on the input of `make_chain`
+    :param bounds: a 2-tuple of optional lower and upper bounds on the input of `make_chain`; use `None` for either bound to infer it
     :param T: type of argument to `make_chain`, one of {float, int}
     :return: a chain parameterized at the nearest passing value to the decision point of the relation
     :rtype: Union[Transformation, Measurement]
@@ -1379,7 +1379,7 @@ def binary_search_chain(
 def binary_search_param(
         make_chain: Callable[[float], Union[Transformation, Measurement]],
         d_in: Any, d_out: Any,
-        bounds: tuple[float, float] | None = None,
+        bounds: tuple[float | None, float | None] | None = (None, None),
         T=None) -> float:
     """Solve for the ideal constructor argument to `make_chain`.
     
@@ -1389,7 +1389,7 @@ def binary_search_param(
     :param make_chain: a function that takes a number and returns a Transformation or Measurement
     :param d_in: how far apart input datasets can be
     :param d_out: how far apart output datasets or distributions can be
-    :param bounds: a 2-tuple of the lower and upper bounds on the input of `make_chain`
+    :param bounds: a 2-tuple of optional lower and upper bounds on the input of `make_chain`; use `None` for either bound to infer it
     :param T: type of argument to `make_chain`, one of {float, int}
     :return: the nearest passing value to the decision point of the relation
     :raises TypeError: if the type is not inferrable (pass T) or the type is invalid
@@ -1448,7 +1448,7 @@ def _call_rust_search(
     predicate: Callable[[float], bool],
     T: "RuntimeType | str",
     search: Callable[[Callable[[float], bool]], Any],
-    bounds: tuple[float, float] | None = None,
+    bounds: tuple[float | None, float | None] | None = (None, None),
 ) -> Any:
     first_exception = None
     had_success = False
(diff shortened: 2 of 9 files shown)


Shoeboxam:
* **#2848** <a href="https://app.graphite.com/github/pr/opendp/opendp/2848?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#2849](https://github.com/opendp/opendp/pull/2849) <a href="https://app.graphite.com/github/pr/opendp/opendp/2849?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2860](https://github.com/opendp/opendp/pull/2860) <a href="https://app.graphite.com/github/pr/opendp/opendp/2860?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#2847** <a href="https://app.graphite.com/github/pr/opendp/opendp/2847?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2846** <a href="https://app.graphite.com/github/pr/opendp/opendp/2846?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opendp/opendp/2846?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

python/src/opendp/mod.py
```diff
@@ -1314,7 +1314,7 @@ def assert_features(*features: str) -> None:
 def binary_search_chain(
         make_chain: Callable[[float], M],
         d_in: Any, d_out: Any,
-        bounds: tuple[float, float] | None = None,
+        bounds: tuple[float | None, float | None] | None = None,
```
mccalluc: Here and below, would `(None, None)` be different from `None`? If not, I'd lean towards:

```suggestion
        bounds: tuple[float | None, float | None] = (None, None),
```

python/src/opendp/mod.py
```diff
@@ -1542,6 +1543,8 @@ def binary_search(
     6
     >>> dp.binary_search(lambda x: x < 5, T=int)
     4
+    >>> dp.binary_search(lambda x: x <= -5, bounds=(-10, None))
+    -5
```
mccalluc: Sorry if I'm missing something here, but is there a reason to introduce negative numbers for this example, or could it work just as well with positive?

python/src/opendp/mod.py
```diff
@@ -1572,15 +1575,26 @@ def binary_search(
...
-    if bounds is not None and len(set(map(type, bounds))) != 1:
-        raise TypeError("bounds must share the same type")  # pragma: no cover
-
-    runtime_T = RuntimeType.infer(bounds[0]) if bounds is not None else RuntimeType.parse(T or _infer_type(predicate))
+    if bounds is not None:
+        lower, upper = bounds
+        if lower is not None and upper is not None and type(lower) is not type(upper):
+            raise TypeError("bounds must share the same type")
+    else:
+        lower, upper = None, None
```
mccalluc, line 1588: Answering my own question, it looks like `None` will be treated the same as `(None, None)`.

python/src/opendp/typing.py
```diff
@@ -470,6 +470,12 @@ def get_first(value):
         return None
     return next(iter(value))
 
+
+def get_first_non_null(*values):
+    """Get the first non-null value."""
+    return next((value for value in values if value is not None), None)
```
mccalluc:
```suggestion
    for value in values:
        if value is not None:
            return value
```
If the comprehension gives you something the simple loop doesn't, please explain.

python/test/test_binary_search.py
```diff
@@ -55,6 +55,19 @@ def test_binary_search():
...
+
+    with pytest.raises(ValueError, match="decision boundary is below the lower bound"):
+        dp.binary_search(lambda x: x <= -5, bounds=(0, None))
+    with pytest.raises(ValueError, match="decision boundary is above the upper bound"):
+        dp.binary_search(lambda x: x <= -5, bounds=(None, -10))
+
+def test_mixed_type_bounds():
+    with pytest.raises(TypeError, match="bounds must share the same type"):
+        dp.binary_search(lambda x: x <= -5, bounds=(-10, 20.))
+
```
mccalluc: Slight preference to parameterize this, so the focus is clearly on what changes: the bounds. See https://docs.pytest.org/en/7.1.x/example/parametrize.html#parametrizing-conditional-raising

python/src/opendp/typing.py
mccalluc: Also, this doesn't seem to be used elsewhere in the python code. Maybe the next PR in the stack will use it?

mccalluc (commented):
Some superficial comments. It looks like this is under construction, so here are some comments on the python, and I can look at it again next week.

rust/src/utilities/search/test.rs
```diff
@@ -3,39 +3,75 @@ use crate::error::ErrorVariant;
...
+    assert_eq!(binary_search(|x: &i32| *x < 5, (10, 0))?, 4);
     Ok(())
 }
 
 #[test]
-fn test_signed_binary_search_reports_direction() -> Fallible<()> {
+fn test_bound_specs_resolve() {
+    assert_eq!(BoundSpec::<i32>::resolve(()), (None, None));
+    assert_eq!(BoundSpec::resolve((0, 10)), (Some(0), Some(10)));
+    assert_eq!(BoundSpec::resolve(Above(0)), (Some(0), None));
```
eddiestudies, line 24: the None::<i32> (turbofish ??) is possible for Above and Below as well:

assert_eq!(BoundSpec::resolve(Above(0)), (Some(0), None::int<32>));
assert_eq!(BoundSpec::resolve(Below(10)), (None::int<32>, Some(10)));

R/opendp/R/mod.R
```diff
@@ -714,24 +721,42 @@ binary_search_param <- function(make_chain, d_in, d_out, bounds = NULL, .T = NUL
...
-#' @param .T type of argument to `predicate`, one of float or int
+#' @param bounds a two-element list of optional lower and upper bounds on the input
+#'   of `predicate`; use `NULL` for either bound to infer it
+#' @param .T type of argument to `predicate`, one of float or int; when set,
+#'   takes precedence over the type inferred from `bounds`
 #' @param return_sign if True, also return the direction away from the decision boundary
 #' @return the discovered parameter within the bounds
 #' @export
+#' @examples
+#' binary_search(\(x) x <= -5L, bounds = list(-10L, NULL))
```
eddiestudies, line 737: why escape the first open parens but not list( ? I am assuming this is an R thing but that seems off

rust/src/utilities/search/test.rs
```diff
@@ -60,20 +96,34 @@ fn test_fallible_binary_search_recovers_from_exception_boundary() -> Fallible<()
...
+    assert_eq!(binary_search(|x: &i8| *x <= 0, (i8::MIN, i8::MAX))?, 0);
     Ok(())
 }
 
+#[test]
+fn test_binary_search_one_sided_float_ranges() -> Fallible<()> {
+    assert_eq!(binary_search(|x: &f32| *x <= 5.0, Above(0.0))?, 5.0);
+    assert_eq!(binary_search(|x: &f64| *x <= 5.0, Above(1.0))?, 5.0);
+    assert_eq!(binary_search(|x: &f64| *x >= -5.0, Below(-1.0))?, -5.0);
+    assert_eq!(binary_search(|x: &f64| *x >= 5.0, Above(0.0))?, 5.0);
```
eddiestudies, line 109: can you explain this test case? 

bound is above 0
the check is >= 5.

So this assumes directionally going down to 5 and then returns 5.0.

I would assume the sign
assert_eq!(binary_search(|x: &f64| *x >= -5.0, Below(-1.0))?, -5.0);
would be
assert_eq!(binary_search(|x: &f64| *x <= -5.0, Below(-1.0))?, -5.0);
To mirror it? Originally I thought it was the same as the second case but that one had the sign flip.

rust/src/utilities/search/test.rs
```diff
@@ -3,39 +3,75 @@ use crate::error::ErrorVariant;
...
     Ok(())
 }
 
 #[test]
 fn test_exponential_bounds_search_bands() {
-    assert_eq!(
-        exponential_bounds_search::<i32>(&|x| *x > 5),
-        Some((1, 65_536))
-    );
+    assert_eq!(exponential_bounds_search::<i32>(&|x| *x > 5), Some((1, 16)));
```
eddiestudies, line 74: I see this is making sure we drop to the tightest bound with the new base 16 approach

rust/src/utilities/search/mod.rs
```diff
@@ -21,7 +25,11 @@ macro_rules! impl_binary_searchable_float {
     ($($ty:ty),+ $(,)?) => {
         $(impl BinarySearchable for $ty {
             fn midpoint(lower: &Self, upper: &Self) -> Self {
-                lower + (upper - lower).halve()
+                if lower.is_sign_negative() != upper.is_sign_negative() {
+                    lower / 2.0 + upper / 2.0
```
Shoeboxam, line 29: @eddiestudies upper - lower can overflow. Makes it (pedantically) ever so slightly more robust. It's admittedly kind of extra.

rust/src/utilities/search/test.rs
Shoeboxam: The bound doesn't determine directionality, it determines valid search range. The predicate then returns the closest true value to the crossing that still passes within the search range. -5.0 satisfies both of your predicates and is below -1.

R/opendp/R/mod.R
Shoeboxam: That's the lambda syntax R chose. `\(params) body`

python/src/opendp/mod.py
Shoeboxam: It was maybe overkill.

python/src/opendp/typing.py
Shoeboxam: Yep! Updated PR description.

python/src/opendp/mod.py
```diff
@@ -1314,7 +1314,7 @@ def assert_features(*features: str) -> None:
 def binary_search_chain(
         make_chain: Callable[[float], M],
         d_in: Any, d_out: Any,
-        bounds: tuple[float, float] | None = None,
+        bounds: tuple[float | None, float | None] | None = (None, None),
```
mccalluc, line 1317:
```suggestion
        bounds: tuple[float | None, float | None] = (None, None),
```
Maybe?

python/src/opendp/mod.py
```diff
@@ -1379,7 +1379,7 @@ def binary_search_chain(
 def binary_search_param(
         make_chain: Callable[[float], Union[Transformation, Measurement]],
         d_in: Any, d_out: Any,
-        bounds: tuple[float, float] | None = None,
+        bounds: tuple[float | None, float | None] | None = (None, None),
```
mccalluc, line 1382:
```suggestion
        bounds: tuple[float | None, float | None] = (None, None),
```

python/src/opendp/mod.py
```diff
@@ -1448,7 +1448,7 @@ def _call_rust_search(
     predicate: Callable[[float], bool],
     T: "RuntimeType | str",
     search: Callable[[Callable[[float], bool]], Any],
-    bounds: tuple[float, float] | None = None,
+    bounds: tuple[float | None, float | None] | None = (None, None),
```
mccalluc, line 1451:
```suggestion
    bounds: tuple[float | None, float | None] = (None, None),
```

python/src/opendp/mod.py
```diff
@@ -1470,7 +1470,7 @@ def wrapped(arg):
     except OpenDPException as err:
         if first_exception is not None:
             exc = first_exception
-            if bounds is None and not had_success and hasattr(exc, "add_note"):
+            if (bounds is None or bounds == (None, None)) and not had_success and hasattr(exc, "add_note"):
```
mccalluc, line 1473:
```suggestion
            if bounds == (None, None) and not had_success and hasattr(exc, "add_note"):
```
Only if the simplification of the signatures above makes sense. If someone explicitly passed `None`, there would be an error downstream, but I think that's ok: It's not python style to aggressively validate inputs.

(An argument on the other side would be to keep this, even if the signatures change, for backward compatibility.)

python/src/opendp/mod.py
```diff
@@ -1488,7 +1488,7 @@ def wrapped(arg):
 @overload
 def binary_search(
         predicate: Callable[[float], bool],
-        bounds: tuple[float, float] | None = ...,
+        bounds: tuple[float | None, float | None] | None = ...,
```
mccalluc, line 1491:
```suggestion
        bounds: tuple[float | None, float | None] = ...,
```

python/src/opendp/mod.py
```diff
@@ -1498,7 +1498,7 @@ def binary_search(
...
```
mccalluc, line 1501:
```suggestion
        bounds: tuple[float | None, float | None] = ...,
```

python/src/opendp/mod.py
```diff
@@ -1508,23 +1508,24 @@ def binary_search(
 @overload
 def binary_search(
         predicate: Callable[[float], bool],
-        bounds: tuple[float, float] | None,
+        bounds: tuple[float | None, float | None] | None,
```
mccalluc, line 1511:
```suggestion
        bounds: tuple[float | None, float | None],
```

python/src/opendp/mod.py
```diff
@@ -1508,23 +1508,24 @@ def binary_search(
...
         T: Type[float] | None,
         return_sign: Literal[True]) -> tuple[float, int]:
     ...
 
 
 def binary_search(
         predicate: Callable[[float], bool],
-        bounds: tuple[float, float] | None = None,
+        bounds: tuple[float | None, float | None] | None = (None, None),
```
mccalluc, line 1519:
```suggestion
        bounds: tuple[float | None, float | None] = (None, None),
```

python/src/opendp/mod.py
```diff
@@ -1572,15 +1575,31 @@ def binary_search(
     from opendp._internal import _binary_search
     from opendp.typing import RuntimeType
 
-    if bounds is not None and len(set(map(type, bounds))) != 1:
-        raise TypeError("bounds must share the same type")  # pragma: no cover
-
-    runtime_T = RuntimeType.infer(bounds[0]) if bounds is not None else RuntimeType.parse(T or _infer_type(predicate))
+    if bounds is not None:
```
mccalluc, line 1578: Possibly drop this wrapping `if`, depending on how you feel about the signature.

R/opendp/R/mod.R
```diff
@@ -684,20 +688,23 @@ binary_search_param <- function(make_chain, d_in, d_out, bounds = NULL, .T = NUL
...
-      return(rt_parse("float"))
-    }
-    stop("bounds must be either float or int", call. = FALSE)
-  }
-
   if (!is.null(.T)) {
     return(rt_parse(.T))
   }
 
+  if (!is.null(bounds)) {
```
mccalluc, line 695: This doesn't exactly parallel the changes in python. It doesn't need to be an exact parallel, but unless there is some difference between idioms or data structures in r and Python, I think it might be good to try to keep them in sync? Or comment on why they diverge?

R/opendp/R/mod.R
```diff
@@ -714,24 +721,42 @@ binary_search_param <- function(make_chain, d_in, d_out, bounds = NULL, .T = NUL
...
+#'   of `predicate`; use `NULL` for either bound to infer it
+#' @param .T type of argument to `predicate`, one of float or int; when set,
+#'   takes precedence over the type inferred from `bounds`
 #' @param return_sign if True, also return the direction away from the decision boundary
 #' @return the discovered parameter within the bounds
 #' @export
+#' @examples
+#' binary_search(\(x) x <= -5L, bounds = list(-10L, NULL))
 binary_search <- function(predicate, bounds = NULL, .T = NULL, return_sign = FALSE) {
+  if (is.null(bounds)) {
```
mccalluc, line 739: Similar to python, confirm that we want to continue to allow null bounds, either for backward compatibility, or because it's good in its own right?

rust/src/utilities/search/mod.rs
mccalluc: Suggest moving this into a comment in the code!

mccalluc (commented):
You could tweak it more, but I think it's good enough. Thanks for tweaks.

- I only skimmed the Rust: it looks like Eddie is reading it carefully.
- Unless we really want people to be able to supply `None` instead of a tuple, I'd simplify the typing, and drop the final `| None` in all of these functions.
  - If you still want to check for `None` explicitly, sure: the backward compatibility is good, and it doesn't add much complexity. (I've gone back and forth on this.)
- Last point: Do you expect this to work with non-monotonic functions? If that's not a usecase, maybe add a caveat in the docstrings?

mccalluc:
(I would have approved, but then one more commit came in, so I could only comment.)

eddiestudies (approved):
rust and python side lgtm

Nice reduction in the macro code, the BoundSpec trait was really interesting to read through

python/src/opendp/mod.py
Shoeboxam: I think keeping None is a better/preferred route. It's the standard for an optional arg and remains backcompat.

PR #2865, opened 2026-08-03
chore: remove codeowners

mccalluc:
https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners

> Code owners are automatically requested for review when someone opens a pull request that modifies code that they own. Code owners are not automatically requested to review draft pull requests

I just don't think we need this functionality right now:
- Increasingly, Eddie should be able to take the load from Mike on Rust code.
- Where Python is barely touched, Chuck's review might not add much.

If we're concerned about PRs falling through the cracks, I could imagine having a single codeowner for the whole repo, just with the understanding that that person would find the right person? But I think it would also work for us to assign reviewers manually, and just keep an eye on the queue, in case a PR comes from the outside.

(We have a statement of responsibility with the OpenDP Commons note in the README, which also partially covers this.)


eddiestudies (approved):
lgtm. Code writer knows who should review

Shoeboxam (approved):
Looked into codeowners notifications settings. There are none. The feature is not aligned with our needs, therefore we should cut it.

PR #2879, opened 2026-08-07
docs: Improve readability of getting-started

### LINKED ISSUE #2869: Revise docs with feedback from Joshua
After Joshua's talk, Mike and I talked and came up with these items:

- Remove the framework (and R) tabs from getting-started.
- Add a pointer to a new section on next-steps that introduces the Framework API.
  - and from that section, point to Measurments in the user-guide.
- Make the typical workflow polars.
- Move scan_csv back one page, and point to the polars IO docs. (Don't create an empty dataframe in the docs.)
- Move the discussion of the structure of eurostat into a folddown.

mccalluc:
- Fix #2869
- Tweak RST style: `:doc:` links will be checked at build-time, so this is a small improvement.

The part I'm the least confident about is the new API comparison page. I just copied it over, but dropped most of the interleaved text, because here we're just comparing the APIs, rather than discussing the actual analysis... but this might not be the right call.


docs/source/getting-started/tabular-data/essential-statistics.rst
```diff
@@ -133,8 +133,8 @@ release the query:
             ... )  # doctest: +ELLIPSIS
             len: ...
 
-Other variations of counting queries are discussed in the `Aggregation
-section <../../api/user-guide/polars/expressions/aggregation.ipynb>`__.
+Other variations of counting queries are discussed in the
+`Aggregation section <../../api/user-guide/polars/expressions/aggregation>`.
```
eddiestudies: no :doc: ?

docs/source/api/user-guide/api-comparison/index.rst
```diff
@@ -0,0 +1,176 @@
+API Comparison
+==============
+
+OpenDP has two APIs:
+
+* The **Context API** is simpler and helps to enforce best practices. Currently available only for Python, it is used in the :doc:`../../../getting-started/index` documentation.
```
eddiestudies: helps enforce best practices? Both likely work but I think the to is not needed

eddiestudies (commented):
looks good. 
looked up the orphan marker.

 one minor doc: omission and an optional nit. the doc may not be needed but it seemed to follow a different rule so I wanted you to check it

mccalluc:
@eddiestudies : Good catch on the link syntax. We have some tests for RST syntax, since it's easy to mess up, and this would now be caught.

(I think I did a little search for an RST linter, but didn't find a great solution.  Many things are valid, but aren't what we intend.)

docs/source/getting-started/tabular-data/essential-statistics.rst
mccalluc: (Added another test so this would be caught by CI.)

python/test/test_docs.py
```diff
@@ -87,7 +87,7 @@ def test_single_backticks(rst_path: Path):
             r"""
             ([^`:]|^)   # Non-backtick or start of line
             `           # backtick
-            ([^`<>_:]+) # content, excluding RST links and tags
+            ([^`<>_:]+?) # content, excluding RST links and tags
```
eddiestudies, line 90: nit: comments get misaligned. might be worth lining them up

docs/source/api/user-guide/api-comparison/index.rst
```diff
@@ -0,0 +1,176 @@
...
+            :language: r
+            :start-after: privacy-loss
+            :end-before: /privacy-loss
+
+3. Collect Public Information
+-----------------------------
+
+.. tab-set::
+
+    .. tab-item:: Python
```
mccalluc:
```suggestion
    .. tab-item:: Context API (Python)
```

mccalluc:
Getting a build warning locally:
```
opendp/python/src/opendp/extras/mbi/_utilities/__init__.py:
docstring of opendp.extras.mbi._utilities.Marginals:1:
WARNING: py:class reference target not found: mbi.LinearMeasurement
```
but unless CI fails, I'll assume that's just a problem on my end.

eddiestudies (approved):
Looks good. Only concern is if the Theory notebooks that were moved interact with the references for theory/framework. I assume not but it was not easy to verify within the review window

mccalluc:
> Only concern is if the Theory notebooks that were moved interact with the references for theory/framework

We should be good. One of the strengths, or aggravations, or sphinx is that it points out broken links, as long as you're using its reference syntax.

PR #2892, opened 2026-08-12
feat: add `ibis` extra to bring in `polars-to-ibis`

### LINKED ISSUE #2889: Add polars-to-ibis
Doesn't necessarily need to be merged, but there's enough here that we could make a PR to surface remaining questions.

mccalluc:
- fix #2889
- Waiting to move out of draft until we can support means, and tests pass.
- Unlike the other extras, ibis and mbi are independent of each other: neither, one, or both can be installed.
- Documentation: Down the road, maybe we have a whole section about how to load data, but I don't want to invest in that until this is really solid. Are these few changes ok as a stopgap?


mccalluc:
Two CI errors: I need to add another typing ignore. The other is weird:
```
FAILED api/user-guide/measurements/randomized-response.rst::randomized-response.rst
- pytest.PytestUnraisableExceptionWarning: Exception ignored in:
  <sqlite3.Connection object at 0x7f74b386f970>
```
Why is there any connection between the randomized-response docs and sqlite? Will rerun.

mccalluc:
TODO:
- Change name to `ibis`
- Will be independent of `mbi`: You can have one or the other or both. Test matrix should reflect this.
- upstream, polars-to-ibis should at least support mean across databases.
- tests should pass!

.github/workflows/smoke-test.yml
```diff
@@ -180,6 +180,7 @@ jobs:
           - 'numpy'
           - 'scikit-learn'
           - 'polars'
+          - 'sql'
```
eddiestudies: ibis as discussed

Issue #2876, opened 2026-08-06
Check branch protections


mccalluc:
Besides approval, shouldn't passing CI be required? We're surprised that #2875 was able to merge.

mccalluc:
Oh, we didn't actually have any checks turned on. Before:

<img width="448" height="369" alt="Image" src="https://github.com/user-attachments/assets/a7fcf57b-0572-40d8-97cc-b1f74dda0b02" />

After (We have not been actually been requiring branches to be up to date with main. That would be a little safer, but I think it would be really painful.)

<img width="380" height="646" alt="Image" src="https://github.com/user-attachments/assets/b1f0615d-90b1-43b2-af5a-a758aebea147" />

mccalluc:
Must not have saved changes? Reopen.

mccalluc:
Ok: Saved now!

PR #2894, opened 2026-08-13
chore: Prefix all smoke-test jobs with "ci-"

### LINKED ISSUE #2893: Add "ci-" prefix to smoke-test jobs
For branch protection, we need to list the required jobs. It will be easiest to pick these out if they all have a common prefix. (Making `rust-build` required isn't strictly necessary, but it's simplest just to add everywhere.)

After that, update the branch protections: https://github.com/opendp/opendp/settings/branch_protection_rules/19589231

mccalluc:
- Fix #2893
- Towards #2876: We want to be able to easily pick out just these jobs for branch protection.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -12,7 +12,7 @@ on:
   pull_request:
 
 jobs:
-  rust-build:
+  ci-rust-build:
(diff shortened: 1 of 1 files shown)


eddiestudies (approved):
lgtm

PR #2895, opened 2026-08-13
docs: Consistent usage of "Framework API"

### LINKED ISSUE #2886: Replace "Programming Framework" with "Framework API"
Follow-up to:
- #2869 

While I recognize that there is a more abstract "framework" idea separate from our particular implementation, I don't think that distinction matters, even for the users using the framework API. I suggest replacing references to  "Programming Framework" with "Framework API", particularly the `programming-framework` page.

mccalluc:
- Fix #2886

We have two distinct APIs for using OpenDP. This PR tries to make the distinction more clear by:
- Reviewing instances of (lowercase) "framework", and usually replacing them with (capital) "Framework API", but in a minority of cases, what is meant is the more abstract idea expressed in the paper. Rephrase as necessary.
- Similarly, where "OpenDP Framework" or "Programming Framework" is already used, clarify whether the API or the abstract idea is intended.

A follow-up item could be to add either a page or a section for "Context API" in the User Guide, but that will intersect with another PR which I'm waiting to land:

- #2879


eddiestudies (approved):
lgtm. I couldn't test the links since some of them change.

PR #2897, opened 2026-08-13
docs: Revise limitations page

### LINKED ISSUE #2896: Less self-deprecation
`api/user-guide/limitations.html`: This doesn't really describe where we are now:

> OpenDP is early in its development history. We aim to work with the community to make it the trustworthy implementation of differential privacy algorithms, but that will take some time. OpenDP already can be used to build some applications, but please be aware of the following limitations when deploying it.

We're still 0.x.y, but the language can be updated.

mccalluc:
- Fix #2896
- Less wordy
- Describe the feature flags from the user POV: "they will drop the `contrib` feature flag" describes a change in the source code, not a change in the user's code.
- Less about our plans for the future, when the roadmap really isn't that clear. "OpenDP has not ~~yet~~ been hardened"
- Mention semantic versioning in the text

### DIFF
--- a/docs/source/api/user-guide/limitations.rst
+++ b/docs/source/api/user-guide/limitations.rst
@@ -3,25 +3,23 @@
 Limitations
 ===========
 
-OpenDP is early in its development history.
-We aim to work with the community to make it *the* trustworthy implementation of differential privacy algorithms, but that will take some time.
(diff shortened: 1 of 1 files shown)


eddiestudies (approved):
lgtm

PR #2899, opened 2026-08-14
docs: add a link to cookbook

mccalluc:
This PR is really a placeholder. What do we think the cookbook needs before it can be official?
- Better web design?
  - or maybe, instead of having a separate site, the examples should be pulled into the docs here?
  - or maybe we have a separate sphinx site there?
- If it is a separate site, a CNAME?
- More examples?
- Better examples?
- More tooling? (More description on the homepage, and not just the filename?)


mccalluc:
Replaced by 
- #2914

PR #2901, opened 2026-08-19
bug: use R_alloc on the vector to avoid gc cleanup of epsilons

### LINKED ISSUE #2900: R garbage collector can prematurely reduce epsilon to 0
## Describe the bug

R garbage collection can zero out epsilons between calls. The repro here iterates over 100 elements with garbage collection set to run each call. 3/100 iterations free and then override epsilon. We create a pad object to zero out freed up memory if there is any, otherwise, the memory might stay as the original epsilon value even if it was freed.

## Your environment

```r
library(opendp)
enable_features("contrib", "honest-but-curious")

space <- c(atom_domain(.T = "f64", nan = FALSE), absolute_distance(.T = "f64"))
meas <- make_fix_delta(make_zCDP_to_approxDP(space |> then_gaussian(1.)), 1e-7)

budget <- meas(d_in = 1.)                     # eps = 5.670859
claimed <- c(budget[[1]] * 1.01, 1e-7)  # strictly above: check should be true

wrong <- 0
gctorture(TRUE)
for (i in 1:100) {
# setting anything cleaned by gc to 0 with numeric
pad <- numeric(i %% 37 + 1). # zero vector of length i %% 37 + 1 to wipe the epsilon after gc
if (!isTRUE(measurement_check(meas, distance_in = 1., distance_out = claimed)))
   wrong <- wrong + 1 # eps is wiped so now our check is no longer correct
}
gctorture(FALSE)
cat(wrong, "/100 sufficient-budget checks saw a zeroed epsilon\n")
```

Observed: `3 /100 sufficient-budget checks saw a zeroed epsilon`
Expected: `0 /100`

If using OpenDP with Python:
- `python --version`
- The version of OpenDP installed, and any extras
- `pip freeze` might be overkill, but is sometimes useful

## A reproducer

- Code snippet to reproduce the behavior
- And if possible, a reference to a gist or notebook or repo where we can see the whole situation.

eddiestudies:


R/opendp/tests/testthat/test-convert.R
```diff
@@ -0,0 +1,15 @@
+opendp::enable_features("contrib", "honest-but-curious")
+
+test_that("tuple gc", {
+  input_space <- c(vector_domain(atom_domain(.T = f64, nan = FALSE)), symmetric_distance())
+
+  gctorture(TRUE)
+  # node reuse depends on heap layout
+  for (i in 1:30) {
```
mccalluc, line 8: Add a little math to explain why 30 is a good number of tries to surface the problem? Like, I could imagine trying 1000 times, but you can probably explain why that would be excessive.

mccalluc (approved):
LGTM, but I'm a little out of my depth, so feel free to loop in Mike if you want a double check.

Also, mark this PR as fixing #2900, I think?

R/opendp/tests/testthat/test-convert.R
eddiestudies: no math here. I found a number where the test passed 10 times for the new code and failed 10 times on the previous code. We can keep as is and if it gets flaky we can expand it further with some statistical check to increase passing percentage. It does not fail now so it should not be necessary, the buggy approach did.

PR #2903, opened 2026-08-19
chore: add CLA exceptions notes

### LINKED ISSUE #2902: Add note where we do not have CLA on file
We have a few contributions without a CLA on file.
- For details, see https://github.com/opendp/clabot-config/pull/36 

One partial fix would be to note in the codebase that an individual retains their copyright on individual files.

mccalluc:
- Fix: #2902

The contributions where we do not have the CLA are small tweaks across multiple files, and I don't think an in-file notice is necessary.

I would suggest leaving this as draft, and then when we are farther in sponsorship discussions, we can ask if this is useful, or not.

### DIFF
--- a/CLA_AND_COPYRIGHT.md
+++ b/CLA_AND_COPYRIGHT.md
@@ -1,0 +1,61 @@
+OpenDP contributors are asked to sign our [Contributor License Agreement](https://docs.opendp.org/en/stable/contributing/cla.html) via a Githb workflow. Contributors retain copyright, but have granted to Harvard a broad license going beyond the terms of the MIT License under which the software is released.
+
+Currently, our [clabot](https://github.com/opendp/clabot-config/) confirms that each contribution has a CLA on file, but this was not in place in the early years of the project and there are some discrepancies.
+
+## Post-hoc signature
+
+After noticing this gap, we asked past contributors to sign the CLA, noting however that the CLA does not cover past contributions:
+
+> Contributor accepts and agrees to the terms and conditions of this CLA, which govern any present or future original work of authorship
+
+| CLA signature |
+|---|
+| matchaginseng |
+| orespo |
+| abigail-gentle |
+| silviacasac |
+| PaulineMauryL | 
+| aramasethu |
+| shlomihod |
+| maxine-park |
+| TedTed |
+| ClaireMcKayBowen |
+| rionides |
+| GKD-stack |
+| pdurbin |
+
+
+## Name mismatch
+
+These users have signed the CLA, but under a different username than that shown by `git blame`:
+
+| real name | CLA signature | git blame |
+|---|---|---|
+| Phil Durbin | pdurbin | philip_durbin | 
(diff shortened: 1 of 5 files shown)


mccalluc:
After running for nearly 6 hours, installing R dependencies failed. It might have been a server outage up stream, but I'm surprised it didn't time out earlier. Perhaps there's an option we can add to force timeout, if it's not the default?

Re-running

mccalluc:
The CLA does not actually assign copyright, so I'm less sure that this does anything useful for us. Back to draft, and can revisit next week.

CLA_AND_COPYRIGHT.md
```diff
@@ -0,0 +1,61 @@
...
+| rionides |
+| GKD-stack |
+| pdurbin |
+
+
+## Name mismatch
+
+These users have signed the CLA, but under a different username than that shown by `git blame`:
+
+| real name | CLA signature | git blame |
```
mccalluc, line 32: Drop `real name`

CLA_AND_COPYRIGHT.md
```diff
@@ -0,0 +1,61 @@
...
+| Maxine Park | maxine-park | maxinempark |
+
+## Institution rather than contributor
+
+For Christian Janos Lebeda's work on sparse histogram ALP ([commit](https://github.com/opendp/opendp/commit/f7095caadbe3fe57a6f3cebea0ccaf8da7c902ff)/[PR](https://github.com/opendp/opendp/pull/388)), the CLA was signed by an instutional representative.
+
+> I believe that I already filled out a CLA.
+> I specifically remember doing so because I could not sign the CLA myself due to Danish university rules for intellectual property.
+> 
+> The document was either signed by the head of department (Peter Sestoft) or the director of the university (Martin Zachariasen/Jens Christian Godskesen depending on the date).
```
mccalluc, line 47: Drop personal names

CLA_AND_COPYRIGHT.md
```diff
@@ -0,0 +1,61 @@
...
+|---|---|---|
+| Phil Durbin | pdurbin | philip_durbin | 
+| Raman Prasad | raprasad | raman_prasad |
+| Ethan Cowan | ecowan | ethan24 |
+| Damien Desfontaines | TedTed | damienbfs |
+| Maxine Park | maxine-park | maxinempark |
+
+## Institution rather than contributor
+
+For Christian Janos Lebeda's work on sparse histogram ALP ([commit](https://github.com/opendp/opendp/commit/f7095caadbe3fe57a6f3cebea0ccaf8da7c902ff)/[PR](https://github.com/opendp/opendp/pull/388)), the CLA was signed by an instutional representative.
```
mccalluc, line 42: replace with username

mccalluc: drop the quotation. "We don't have access to that message"

mccalluc:
Addresses comments from conversation with Salil.

PR #2906, opened 2026-08-21
bug: fix aim penalty penalize candidates by the expected error of the upcoming measure step

eddiestudies:


python/src/opendp/extras/mbi/_aim/__init__.py
```diff
@@ -360,9 +363,14 @@ def is_small(clique: tuple[str, ...]) -> bool:
     if not candidates:
         return None
 
+    # penalize candidates by the expected error of the upcoming MEASURE step,
+    # not of this SELECT step
+    sensitivity = max(d_in[q.by] for q in candidates)
+    expectation = get_scale(output_measure, d_measure, sensitivity) * to_mu
```
Shoeboxam, line 369: It would be better to weight the scores by each marginal's sensitivity:

```python
expectations = {
    q.by: get_scale(output_measure, d_measure, d_in[q.by]) * to_mu
    for q in candidates
}
```

Shoeboxam (changes requested):
Really small nit. It can be calibrated a bit tighter. Good to go after.

Shoeboxam (approved):
Thanks!

Issue #2907, opened 2026-08-25
Nightly (really weekly) failing since August 2


mccalluc:
Several warnings and a fatal error in the [log](https://github.com/opendp/opendp/actions/runs/32617624517/job/97140893996):
```
fatal: ambiguous argument 'cdp_delta.rs': unknown revision or path not in the working tree.
...
LaTeX Warning: Citation `CKS20' on page 1 undefined on input line 12.
LaTeX Warning: Citation `CKS20' on page 1 undefined on input line 26.
LaTeX Warning: Reference `plrv' on page 1 undefined on input line 37.
LaTeX Warning: Citation `CKS20' on page 1 undefined on input line 89.
LaTeX Warning: Reference `delta-bound' on page 2 undefined on input line 102.
LaTeX Warning: Citation `CKS20' on page 2 undefined on input line 140.
LaTeX Warning: Reference `renyidp-approxdp-delta' on page 2 undefined on input 
line 148.
LaTeX Warning: Reference `renyidp-approxdp-epsilon' on page 2 undefined on inpu
t line 160.
...
./pseudocode/cdp_delta.py:39: LaTeX Error: Invalid UTF-8 byte sequence (�\lst@F
illFixed@).
```

eddiestudies:
Looking into this. Seems to be a utf-8 handling change. I have the failures passing locally, working on the warnings now.

eddiestudies:
Looks like the patch worked https://github.com/opendp/opendp/actions/runs/32893592084/job/97950982543?pr=2908

Waiting for the rest of the builds on https://github.com/opendp/opendp/pull/2908

mccalluc:
Closed!

PR #2908, opened 2026-08-25
build: Patch latex build

eddiestudies:
Resolves https://github.com/opendp/opendp/actions/runs/32617624517/job/97140893996

### DIFF
--- a/rust/src/combinators/measure_cast/zCDP_to_approxDP/cdp_delta/cdp_delta.tex
+++ b/rust/src/combinators/measure_cast/zCDP_to_approxDP/cdp_delta/cdp_delta.tex
@@ -8,7 +8,7 @@
 \begin{document}
 \maketitle
 
-Proves soundness of \texttt{fn cdp\_delta} in \asOfCommit{cdp_delta.rs}{0b8f4222}.
+Proves soundness of \texttt{fn cdp\_delta} in \asOfCommit{mod.rs}{0b8f4222}.
 This proof is an adaptation of \href{https://arxiv.org/pdf/2004.00010.pdf#subsection.2.3}{subsection 2.3} of \cite{CKS20}.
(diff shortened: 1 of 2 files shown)


rust/src/combinators/measure_cast/zCDP_to_approxDP/cdp_delta/cdp_delta.tex
```diff
@@ -8,7 +8,7 @@
 \begin{document}
 \maketitle
 
-Proves soundness of \texttt{fn cdp\_delta} in \asOfCommit{cdp_delta.rs}{0b8f4222}.
+Proves soundness of \texttt{fn cdp\_delta} in \asOfCommit{mod.rs}{0b8f4222}.
```
mccalluc, line 11: I'm guessing the rename happened after the hash, so will the generated link work? Perhaps confirm no changes to the code, and update the hash as well?

mccalluc (commented):
Wondering if the hash should also be updated?

Is there anything we should change in the build to catch this earlier? Perhaps not: If the build is expensive and only rarely has problems, then no. (If you have ideas, a follow-up issue rather than holding up this PR would be best.)

rust/src/combinators/measure_cast/zCDP_to_approxDP/cdp_delta/cdp_delta.tex
eddiestudies: we can keep the incorrect name. I couldn't find the hash in the history 0b8f4222. I can revert the file rename change, but we definitely should update these hashes. We made a bug fix recently for cdp delta

mccalluc (approved):
- Filed #2910

So there is some more work to do, but getting the build fixed is good!

I think this fixes the issue?

- fix #2907

PR #2914, opened 2026-08-28
docs: remove several theory notebooks that are now in the cookbook

mccalluc:
I've left two pages under Theory... but if you think it all should go, that wouldn't be impossible.

I'll put this in the release milestone: Might be nice to have the publication of the cookbook along with the next release, though I don't think it's essential. (Might sense is that you are still the project architect, and you should make the call on this, but if you don't think you need to be looped in, that's ok, too.)

- Towards #2869


docs/source/theory/mechanisms/exponential-mechanism-quantiles.ipynb
Shoeboxam, line 1: This file feels like it should stay with OpenDP. It's not a cookbook, it's explanatory. Perhaps move it into user guide?

docs/source/theory/accuracy-pitfalls.ipynb
Shoeboxam, line 1: Do you think this notebook should similarly remain in the library docs?

Shoeboxam (changes requested):
I think Theory should be completely removed in favor of Cookbooks. What's left is:
[A Framework to Understand DP](https://docs.opendp.org/en/stable/theory/a-framework-to-understand-dp.html)
[Differential Privacy with OpenDP](https://docs.opendp.org/en/stable/theory/dp-with-opendp.html)

DP with OpenDP could go to cookbooks. The framework notebook could go under user guide/ programming framework?

mccalluc:
> The framework notebook could go under user guide/ programming framework?

My sense is that the framework notebook is a primary example of a theory-centered notebook. It is not describing the software, and in this form, I don't want it to intrude on other parts of the docs.

If it were to go to the user guide, is it because it is information the user needs, or because we just need to find a place for it? 

> re exponential-mechanism-quantiles.ipynb
> This file feels like it should stay with OpenDP. It's not a cookbook, it's explanatory. Perhaps move it into user guide?

Maybe I'm missing it, but why should a user of the software need to know these details of the algorithm? Is there a gap in the user guide that this would fill, or are we just looking for a place to put it?

> Do you think this notebook should similarly remain in the library docs?

Similarly, no.

These are all just my opinions. Can you take a look at the Cookbook, and see if it makes sense as a place for more than just "recipes"? I'm more concerned about keeping the scope of the library docs clear, than about keeping miscellanea out of the cookbook.

PR #2915, opened 2026-09-02
Add DP logistic regression estimator extending DPEstimator

rtonpe2:
Add DP logistic regression estimator (DPEstimator subclass):

Adds LogisticRegression extending the DPEstimator base class from #2843.

Clips per-example gradients to an L2 bound, sums under Gaussian noise, steps, with the budget split across iterations under zCDP via make_adaptive_composition. Fits through the Context (est.fit(context.query())), exposes coef_/intercept_/classes_/predict/predict_proba, and converges to sklearn's unregularized solution at high budget (max abs weight diff ~0.01).

Structure: thin subclass in linear_model/__init__.py delegating to a calibrated builder in _make_private_logistic_regression/, mirroring Theil-Sen.

Friction worth discussing (as a base-class test case):

Sized-data d_in: odd d_in floors to zero sensitivity, giving ρ=0 at all scales and an opaque calibration failure — a clearer error would help.
Budget-split rounding: d_out / n_iters overshoots in floating point and gets rejected. Was resolved via binary_search_param, but a helper might be worth it.
predict/predict_proba not scaffolded: classifiers reimplement these + classes_ from the regression-shaped base.
Supervised y: fit(X, y) implies a target, but the array2_domain model led me to a last-column convention with y ignored in _prepare_fit_query. Intended, or should it accept y as a column selector and reorder?

Tests: basic fit-through- Minimal Context test

### DIFF
--- a/python/src/opendp/extras/sklearn/linear_model/__init__.py
+++ b/python/src/opendp/extras/sklearn/linear_model/__init__.py
@@ -58,7 +58,7 @@ class LinearRegression:
         if len(x_bounds) != 1:
             msg = f"For now, the x_bounds array must consist of a single tuple, not {x_bounds}"
             raise Exception(msg)
-        
+
         self.measurement = _make_private_theil_sen(
             output_measure=output_measure,
             x_bounds=x_bounds[0],
@@ -139,24 +139,24 @@ class LinearRegression:
 
 class LogisticRegression(DPEstimator):
     """
-        DP Logistic Regression
-    
-        The interface is parallel to that offered by sklearn's
-        `LogisticRegression <https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html>`_.
-        Fitting is performed with differentially private stochastic gradient descent (DP-SGD):
-        each iteration clips per-example gradients to an L2 norm bound, sums them under Gaussian
-        noise, and takes a gradient step, with the privacy budget split across iterations under
-        zero-concentrated differential privacy. After ``fit``, the model exposes ``coef_``,
-        ``intercept_``, and ``classes_``, and supports ``predict`` and ``predict_proba``.
+    DP Logistic Regression
 
-        Data is expected as a single 2-dimensional array whose last column is the binary target
-        and whose remaining columns are the features.
+    The interface is parallel to that offered by sklearn's
+    `LogisticRegression <https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html>`_.
+    Fitting is performed with differentially private stochastic gradient descent (DP-SGD):
+    each iteration clips per-example gradients to an L2 norm bound, sums them under Gaussian
+    noise, and takes a gradient step, with the privacy budget split across iterations under
+    zero-concentrated differential privacy. After ``fit``, the model exposes ``coef_``,
+    ``intercept_``, and ``classes_``, and supports ``predict`` and ``predict_proba``.
 
-        :param n_iters: Number of DP-SGD iterations. Also controls the budget split: the total privacy budget is divided across this many gradient steps, so more iterations means more noise per step.
-        :param learning_rate: Step size for the gradient update applied to the weights each iteration.
-        :param clip_norm: Per-example gradient clipping bound. Each example's gradient is scaled to have L2 norm at most this value, bounding one record's influence on the update. Larger values distort gradients less but require more noise.
-        :param l2_penalty: Strength of optional L2 regularization added to the gradient update. Defaults to 0.0 (no regularization).
-        """
+    Data is expected as a single 2-dimensional array whose last column is the binary target
+    and whose remaining columns are the features.
+
+    :param n_iters: Number of DP-SGD iterations. Also controls the budget split: the total privacy budget is divided across this many gradient steps, so more iterations means more noise per step.
+    :param learning_rate: Step size for the gradient update applied to the weights each iteration.
+    :param clip_norm: Per-example gradient clipping bound. Each example's gradient is scaled to have L2 norm at most this value, bounding one record's influence on the update. Larger values distort gradients less but require more noise.
+    :param l2_penalty: Strength of optional L2 regularization added to the gradient update. Defaults to 0.0 (no regularization).
+    """
 
     def __init__(self, n_iters=100, learning_rate=0.1, clip_norm=1.0, l2_penalty=0.0):
         self.n_iters = n_iters
@@ -181,33 +181,34 @@ class LogisticRegression(DPEstimator):
         """
 
         self._reject_fit_params(fit_params)
-        return X #we are assuming that the last column is the target (open design question)
+        return X  # we are assuming that the last column is the target (open design question)
 
     def make(self, input_domain, input_metric, output_measure, d_in, d_out):
         return _make_private_logistic_regression(
-            input_domain, input_metric, output_measure, d_in, d_out,
-            n_iters=self.n_iters, learning_rate=self.learning_rate,
-            clip_norm=self.clip_norm, l2_penalty=self.l2_penalty,
+            input_domain,
+            input_metric,
+            output_measure,
+            d_in,
+            d_out,
+            n_iters=self.n_iters,
+            learning_rate=self.learning_rate,
+            clip_norm=self.clip_norm,
+            l2_penalty=self.l2_penalty,
         )
 
     def _ingest_release(self, release):
-        import numpy as np
+        np = import_optional_dependency("numpy")
         theta = np.asarray(release)
-        self.coef_ = theta[:-1]        
-        self.intercept_ = theta[-1]    
+        self.coef_ = theta[:-1]
+        self.intercept_ = theta[-1]
         self.classes_ = np.array([0, 1])
 
     def predict_proba(self, X):
-        from scipy.special import expit
-        import numpy as np
+        np = import_optional_dependency("numpy")
+        special = import_optional_dependency("scipy.special")
         Xb = np.hstack([np.asarray(X), np.ones((len(X), 1))])
-        p = expit(Xb @ np.concatenate([self.coef_, [self.intercept_]]))
+        p = special.expit(Xb @ np.concatenate([self.coef_, [self.intercept_]]))
         return np.column_stack([1 - p, p])
 
     def predict(self, X):
         return (self.predict_proba(X)[:, 1] >= 0.5).astype(int)
-
-        
-
-        
-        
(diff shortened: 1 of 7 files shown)


python/src/opendp/extras/sklearn/linear_model/_make_private_logistic_regression/__init__.py
```diff
@@ -0,0 +1,102 @@
...
+from opendp.extras.numpy import make_np_clamp
+from scipy.special import expit
+
+
+def make_per_example_gradients(data, theta, np):
+    y = data[:, -1]                                         
+    Xf = data[:, :-1]                                       
+    X = np.hstack([Xf, np.ones((data.shape[0], 1))])        
+    z = X @ theta
+    p = expit(z)
```
eddiestudies: we usually move the scipy calls to use a delayed import to avoid breaking unrelated scenarios. 

similar to
from opendp._lib import import_optional_dependency

special = import_optional_dependency("scipy.special")

then special.expit.

python/src/opendp/extras/sklearn/linear_model/__init__.py
```diff
@@ -131,3 +135,79 @@ def predict(X):
...
+        from scipy.special import expit
+        import numpy as np
+        Xb = np.hstack([np.asarray(X), np.ones((len(X), 1))])
+        p = expit(Xb @ np.concatenate([self.coef_, [self.intercept_]]))
+        return np.column_stack([1 - p, p])
+
+    def predict(self, X):
+        return (self.predict_proba(X)[:, 1] >= 0.5).astype(int)
+
+
```
eddiestudies: nit: the formatter is likely going to want these lines cleaned up

python/src/opendp/extras/sklearn/linear_model/__init__.py
```diff
@@ -131,3 +135,79 @@ def predict(X):
...
+    def _ingest_release(self, release):
+        import numpy as np
+        theta = np.asarray(release)
+        self.coef_ = theta[:-1]        
+        self.intercept_ = theta[-1]    
+        self.classes_ = np.array([0, 1])
+
+    def predict_proba(self, X):
+        from scipy.special import expit
```
eddiestudies: same import convention comment. Also for the numpy import below

rtonpe2: fixed.

python/src/opendp/extras/sklearn/linear_model/__init__.py
rtonpe2: fixed.

python/src/opendp/extras/sklearn/linear_model/_make_private_logistic_regression/__init__.py
rtonpe2: fixed.

asaha1203 (approved):
Thanks for the great PR! I've taken a look at and played with your implementation a bit. From my end, it looks like all non-FP-safe calculations are wrapped with the correct OpenDP-side NumPy functions (e.g. `make_private_np_sum`, `make_np_clamp`). 

I also believe that the privacy accounting is done as-expected within the library, via `make_adaptive_composition`, so I don't anticipate new issues there. 

The gradient computation logic appears to be correct for logistic regression, so I don't foresee any significant correctness issues.

Two points you may wish to address: (1) you may want to add additional losses or penalties (e.g. $L_1$ regularization) to match the `sklearn` implementation more closely. (2) you may also want to stress in your naming conventions that this is one of several possible algorithms for logistic regression (e.g. creating a `GDClassifier` runner that wraps the provided code). This second point is mostly external, and there are probably larger design decisions that affect it. 

Altogether, this looks great to me! I feel comfortable merging this PR in its current form.

python/test/extras/test_sklearn_estimator.py
```diff
@@ -143,3 +140,34 @@ def test_direct_measurement_and_context_fit_share_estimator_path():
     )
     assert estimator.fit(context.query()) is estimator  # type: ignore[arg-type]
     assert isinstance(estimator.count_, int)
+
+
+def test_logistic_regression_fit_through_context():
+    import opendp.prelude as dp
+    import numpy as np
```
eddiestudies: np = pytest.importorskip("numpy")
pytest.importorskip("scipy")

eddiestudies: Another import nuance. For code we use the import wrapper tool. For tests we skip if the dependency is not available since we use the same test directory command across our test matrix.

python/test/extras/test_sklearn_estimator.py
```diff
@@ -143,3 +140,34 @@ def test_direct_measurement_and_context_fit_share_estimator_path():
...
+    assert measurement.map(2) <= 1.0
+
+    context = dp.Context.compositor(
+        data=data,
+        domain=domain,
+        privacy_unit=dp.unit_of(contributions=2),
+        privacy_loss=dp.loss_of(rho=1.0),
+        split_evenly_over=1,
+    )
+    assert estimator.fit(context.query()) is estimator
```
eddiestudies: we should add a call to predict and predict prob to cover the full model call pattern

eddiestudies: If the output is inconsistent, we can make the check loose. Either the verifying the output type or a large range that keeps the test passing consistently

eddiestudies: assert estimator.fit(context.query()) is estimator  # type: ignore[arg-type]

for linter checks

python/test/extras/test_sklearn_estimator.py
```diff
@@ -46,7 +42,12 @@ def test_sklearn_estimator_is_abstract():
 
     with pytest.raises(NotImplementedError):
         DPEstimator.make(
-            _DummyEstimator(), None, None, None, 1, 1  # type: ignore[arg-type]
+            _DummyEstimator(),
+            None,
```
eddiestudies: None,  # type: ignore[arg-type]
None,  # type: ignore[arg-type]
None,  # type: ignore[arg-type]

for linter checks

python/src/opendp/extras/sklearn/linear_model/__init__.py
```diff
@@ -131,3 +135,80 @@ def predict(X):
...
+        query's data, with the preceding columns as features. The target is not passed
+        separately: ``y`` is accepted for scikit-learn signature compatibility but is not
+        used to relocate or attach a target column, and supplying it has no effect on the
+        released model. Fit metadata (``fit_params``) is not supported and is rejected.
+
+        :param X: a Context query whose data has the binary target as its final column
+        :param y: accepted for sklearn compatibility; ignored (the target must already be the last column of ``X``)
+        :param fit_params: not supported; any fit metadata raises
+        :raises TypeError: if any ``fit_params`` are supplied
+        :return: the input query ``X`` unchanged
```
eddiestudies, line 179: :return: the input query ``X`` unchanged
:raises TypeError: if any ``fit_params`` are supplied

For some reason, I do not know why, the docs do not like raises before return.

python/src/opendp/extras/sklearn/linear_model/_make_private_logistic_regression/__init__.py
```diff
@@ -0,0 +1,131 @@
...
+    output_measure,
+    d_in,
+    d_mids,
+    n,
+    d,
+    clip_norm,
+    learning_rate,
+    l2_penalty,
+):
+
```
eddiestudies, line 35: We should assert_features earlier on in the function similar to: 

https://github.com/opendp/opendp/blob/94bce371b0d3715ae3143b6f139272377c62ed55/python/src/opendp/extras/numpy/_make_np_sscp/__init__.py#L19

    dp.assert_features("contrib", "honest-but-curious")

python/src/opendp/extras/sklearn/linear_model/_make_private_logistic_regression/__init__.py
```diff
@@ -0,0 +1,131 @@
...
+    *,
+    n_iters: int,
+    learning_rate: float,
+    clip_norm: float,
+    l2_penalty: float = 0.0,
+) -> dp.Measurement:
+
+    import opendp.prelude as dp
+
+    dp.assert_features("contrib", "idealized-numerics")
```
eddiestudies, line 82: assert_features for honest-but-curious similar to here but for log_regression_train

eddiestudies:
Changes look close, just a few small changes for the builds to pass and checks to happen earlier. I believe there are a few input parameter based invariants that can be checked to warn/ fail earlier if the configuration is wrong. 

We do not need to solve all of the input invariants but if you know of a few, the test used 2 instead of 1 for example, it would be worth adding guardrails for values that are knowingly not supported.

python/test/extras/test_sklearn_estimator.py
rtonpe2: fixed. tests use pytest.importorskip for numpy and scipy

python/test/extras/test_sklearn_estimator.py
rtonpe2: fixed.

python/src/opendp/extras/sklearn/linear_model/_make_private_logistic_regression/__init__.py
rtonpe2: done.

python/test/extras/test_sklearn_estimator.py
rtonpe2: fixed.

python/src/opendp/extras/sklearn/linear_model/__init__.py
rtonpe2: fixed.

python/src/opendp/extras/sklearn/linear_model/_make_private_logistic_regression/__init__.py
rtonpe2: added.

python/test/extras/test_sklearn_estimator.py
rtonpe2: have added with loose checks (shape + valid labels).

rtonpe2:
Thanks Amit, really appreciate the careful review. L1/elastic-net would bring it closer to sklearn parity. I'd propose adding that as a follow-up PR. Also, a name like GDClassifier would make that explicit. You're right - this would be a broader design discussion but would be good to explore.

python/test/extras/test_sklearn_estimator.py
eddiestudies: thanks

